PR #3, opened 2011-03-16
Here I go

lzap:
The only issue is that Ruby 1.8 contains a bug - it always returns 0 as return code here. V 1.9 is fine.

There are some workarounds but I dont think we need it now since we do not wait until the process finishes. I properly docummented this behaviour. You will see. Need to work now on my roles... please finish this for me.


lzap:
Youll face the issue only when you call join. I call join only in the unit tests.

Also it COUULD be good idea to call join_all to wait until all wget processes finishes when stopping the smart proxy. Consider it. The method is there. Maybe I could write a test for it. Should be easy (two echos + join all).

lzap:
Is there anything wrong with my request at all? I am not sure what you asked on the IRC.

PR #4, opened 2011-03-17
Couple of quick fixes

mbainter:
I was having some issues with the smart-proxy app.  One was just a typo I believe, and the other is switching up the logging to look up the constant, as the current setup was causing crashes for me.

Files changed:
lib/proxy/log.rb
lib/proxy/puppetca.rb


ohadlevy:
great - thanks!

I've rebased a bit, and merged... hopefully you are ok with that :)

Ohad

PR #7, opened 2011-11-04
Fixes #1295 to verify puppet certificate before attempting to clean

treydock:


ohadlevy:
@treydock I cant figure out why this is required....

PR #11, opened 2012-02-20
Change default log_level for debians log_level to INFO

mrud:
Starting with commit 5821a2694b1d62ad6c250fdced123bfe8d17cf4e the
log level is looked up using the get_const method which requires
unqualified strings. This patch fixes the default log level for the
debian package and updates the comments.


ohadlevy:
merged!

PR #12, opened 2012-02-20
Fix log_level comment for settings.yml.example

mrud:
Logger.constants does not contain Fatal but only FATAL. This change
adjusts the comment for log_level.


ohadlevy:
merged!

PR #13, opened 2012-02-20
Add to_json method to lease and reservation module

mrud:
The /dhcp/:network returned strings instead of dicts for
reservation and leases. This patch adds to_json methods
for reservation and lease to return dicts instead of strings.


mrud:
Seems to be merged -> closing

PR #14, opened 2012-03-27
Possible fix for Feature 1552

jantman:
this little hack expands $confdir in modulepath if it's present. Only within the dynamic environments block. Not really tested other than in my environment, please review, but should provide a starting point for work on Feature 1552.


ohadlevy:
merged - thanks!

PR #17, opened 2012-05-12
Fix proxy for 1.9

GregSutcliffe:
Two parts to this pull req...

1) The files needed to rebuild the Debian package. I felt we should store these in Git, so I've added them.

2) Changes for ruby 1.9. I run my proxy on an Archlinux box, and this is what I needed to change to make the proxy run there (Arch defaults to ruby1.9). I think this commit would actually break 1.8 compatibility, so we probably need to do more work, but it needed writing down somewhere :)

### DIFF
--- a/bin/smart-proxy
+++ b/bin/smart-proxy
@@ -18,7 +18,7 @@ class SmartProxy < Sinatra::Base
 
   set :root, APP_ROOT
   set :views, APP_ROOT + '/views'
-  set :public, APP_ROOT + '/public'
+  set :public_folder, APP_ROOT + '/public'
   set :logging, true
   set :env,     :production
   set :run,     true
--- a/extra/debianbuild/dot-nightly-smart-proxy
+++ b/extra/debianbuild/dot-nightly-smart-proxy
@@ -1,0 +1,15 @@
+PACKAGE_NAME='foreman-proxy'
+VERSION='0.5'
+MAINTAINER='Greg Sutcliffe <gsutcliffe@gmail.com>'
+
+BUILD_DIR="/home/build/nightly-${PACKAGE_NAME}"
+TARGET="${BUILD_DIR}/${PACKAGE_NAME}"
+REPO='git://github.com/theforeman/smart-proxy.git'
+BRANCH='develop'
+
+REPO_DIR='/home/build/foreman-repo'
+DEB_REPO='nightly'
+
+function prepare_build() {
+  mv ./extra/debian .
+}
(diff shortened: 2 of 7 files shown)


bin/smart-proxy
```diff
@@ -18,7 +18,7 @@ class SmartProxy < Sinatra::Base
 
   set :root, APP_ROOT
   set :views, APP_ROOT + '/views'
-  set :public, APP_ROOT + '/public'
+  set :public_folder, APP_ROOT + '/public'
```
ohadlevy, line 21: i think this depends on the version of sinatra vs ruby 1.9?

extra/debianbuild/dot-nightly-smart-proxy
```diff
@@ -0,0 +1,19 @@
...
+REPO='git://github.com/theforeman/smart-proxy.git'
+BRANCH='develop'
+
+REPO_DIR='/home/greg/build-area/foreman-repo'
+DEB_REPO='nightly'
+
+function prepare_build() {
+  cd /home/greg/github/smart-proxy && git pull && cd -
+  cp -r /home/greg/github/smart-proxy/extra/debian .
+}
```
ohadlevy, line 19: i have no problems merging the above two scripts (just in case they were in a different pull request ;))

lib/proxy/dhcp/subnet.rb
```diff
@@ -193,7 +192,7 @@ def icmp_pingable? ip
     end
 
     def privileged_user
-      (PLATFORM =~ /linux/i and Process.uid == 0) or PLATFORM =~ /mingw/
+      (RUBY_PLATFORM =~ /linux/i and Process.uid == 0) or RUBY_PLATFORM =~ /mingw/
     end
```
ohadlevy, line 196: switch it to a method and figure out if its ruby19 or 18?

extra/debianbuild/dot-nightly-smart-proxy
GregSutcliffe, line 19: Yeah, this should be a separate request - it was late :)

I'll strip back the rest of this pullreq and rework it so you can merge these two - they're only here so we can store it somewhere anyway

bin/smart-proxy
GregSutcliffe, line 21: OK, I think it was only a warning anyway. Gem list says I'm on sinatra 1.3.1, for reference. I'll try and retest.

lib/proxy/dhcp/subnet.rb
GregSutcliffe, line 196: Sensible. I'll see what I can do.

PR #19, opened 2012-05-29
Environment listing fixes

Orphis:
This fixes bug #1654 by making a copy of the hash we're iterating one.

It also optimizes a little environment discovery by filtering invalid folder names.
In my setup, we uses git replication to map branches to an environment and for each branch we have a folder called "foobar.somegithash" and a symlink just called "foobar" at the same level.
Having both folders makes it very slow as foreman tries to get the classes from these two and in the end fails to add these environments because they have some invalid name.


ohadlevy:
merged - thanks!

PR #20, opened 2012-05-31
Was not returning true on command success, caused puppet_api.rb to print...

skihero:
... false messages.


ohadlevy:
merged - thanks!

PR #23, opened 2012-07-05
Fix for 1733 - proxy doesn't handle the default puppet modulepath

GregSutcliffe:
Fixes http://theforeman.org/issues/1733


ohadlevy:
merged

PR #25, opened 2012-07-10
Export arguments of parameterized classes

ofavre:
Export a `params` sub-object whose keys are the parameter names, along with their values, or `null` if the parameter is mandatory.

### DIFF
--- a/lib/proxy/puppet/puppet_class.rb
+++ b/lib/proxy/puppet/puppet_class.rb
@@ -1,3 +1,5 @@
+require 'puppet/parser'
+
 module Proxy::Puppet
 
   class PuppetClass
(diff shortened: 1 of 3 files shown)


GregSutcliffe:
testing by request from @ohadlevy: seems to work for me!

Obviously foreman doesn't show anything new about the detected classes, but it still correctly finds the new classes and deletes the old ones. Good stuff.

ohadlevy:
do we have any idea which versions of puppet support the parser? which version is the oldest one that works?

GregSutcliffe:
testing with wget seems good, returns things like (formatted for clarity):

{"dhcp":
  {"params":
    {"dnsupdatekey":null,
     "dhcp_monitor":true,
     "pxeserver":null,
     "interface":"NOTSET",
     "logfacility":"local7",
     "dnsdomain":null,
     "ntpservers":null,
     "interfaces":null,
     "pxefilename":null,
     "nameservers":null},
   "module":null,"name":"dhcp
  }
}

GregSutcliffe:
I can probably try a few older versions tonight, but now I should get some work done :)

ofavre:
Tests updated to work against Puppet 2.6

ohadlevy:
merged! thanks @ofavre!!

PR #26, opened 2012-07-31
Added a REST interface for the Facts known about the proxy

Telmo:
It has 2 calls

```
/facts
```

Returns all Facts known of the smart-proxy in a Hash (json)

```
/fact/:fact
```

Returns a has with the requested fact value

```
curl -X GET http://localhost:8443/fact/uptime
{"uptime":"32 days"}
```

It adds a dependency to facter which I added to the smart-proxy binary


ohadlevy:
merged - thanks!

PR #28, opened 2012-08-09
fixes #1685 - Windows DNS: Secure connection using GSS-TSIG

ohadlevy:
Signed-off-by: theforemanuser123 oliver_weinmann@gmx.de

### DIFF
--- a/lib/proxy/dns/bind.rb
+++ b/lib/proxy/dns/bind.rb
@@ -1,5 +1,6 @@
 require "proxy/dns"
 require 'resolv'
+require 'date'
 
 module Proxy::DNS
   class Bind < Record
@@ -52,6 +53,25 @@ module Proxy::DNS
 
     private
 
+    def nsupdate_args
+      args = ""
+      args = "-k #{SETTINGS.dns_key} " if SETTINGS.dns_key
+      if SETTINGS.dns_tsig_keytab
+        args += "-g "
+        logger.debug "DNS TSIG authentication enabled."
+      end
+      args
+    end
+
+    def kinit_args
+      args = ""
+      if SETTINGS.dns_tsig_keytab
+        args += "-F -k -t #{SETTINGS.dns_tsig_keytab} #{SETTINGS.dns_tsig_principal}"
+        logger.debug "kinit #{args}."
+      end
+      args
+    end
+
     def find_nsupdate
       @nsupdate = which("nsupdate", "/usr/bin")
       unless File.exists?("#{@nsupdate}")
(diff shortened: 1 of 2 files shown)


ohadlevy:
it looks good, however, I would say that the kerb stuff probably need to exists in a class by itself, more minor comments bellow

lib/proxy/dns/bind.rb
```diff
@@ -60,11 +80,60 @@ def find_nsupdate
...
+    def find_kinit
+      @kinit = which("kinit", "/usr/bin")
+      unless File.exists?("#{@kinit}")
+        logger.warn "unable to find kinit binary, maybe missing krb5-user package?"
+        raise "unable to find kinit binary"
+      end
+    end
+
+    def find_krb5tgt
+      krb5tgt = system("klist | grep -i #{SETTINGS.dns_tsig_principal} 2>&1")
```
ohadlevy, line 92: what if klist is not in path? I guess we need to do which(klist..).

also, if we move this code into its own line, we could parse it without reverting to grep :)

lib/proxy/dns/bind.rb
```diff
@@ -60,11 +80,60 @@ def find_nsupdate
...
+      if krb5tgt == false
+        exp = 1
+        now = 0
+        @loggermsg = "unable to find kerberos ticket. Trying to aquire a valid TGT..."
+        @raisemsg = "unable to find kerberos ticket. Trying to aquire a valid TGT..."
```
ohadlevy, line 97: these error messages are never used i think

lib/proxy/dns/bind.rb
```diff
@@ -60,11 +80,60 @@ def find_nsupdate
...
+        kinit
+      else
+        format = "%d/%m/%y %H:%M"
+        krbexp = `klist | grep A.SPACE.CORP | grep '/' | awk -F ' ' '{print $3,$4}'`
```
ohadlevy, line 101: I guess the kerb domain needs to be extracted into a setting value?

lib/proxy/dns/bind.rb
r0k5t4r, line 92: what if klist is not in path? I guess we need to do which(klist..).

--> I added which klist

also, if we move this code into its own line, we could parse it without reverting to grep :)

hmm can you give me a small hint, how to do this. Basically all I did was copy & paste for most of the code. :)

lib/proxy/dns/bind.rb
r0k5t4r, line 97: fixed. I forgot the "_" in the var names.

lib/proxy/dns/bind.rb
r0k5t4r, line 101: Forgot to remove it. :( I have added a new setting value "dns_tsig_realm"

r0k5t4r:
t looks good, however, I would say that the kerb stuff probably need to exists in a class by itself, more minor comments bellow

--> Would this be as easy as creating a new file e.g. kerberos.rb, put the code in there and put a require kerberos.rb in bind.rb?

r0k5t4r:
closing this one as it's a duplicate of #29.

domcleal:
Please ignore the test failure comment, the job was misconfigured, I believe it passes.  Deleting the bot's comments should cause it to get retested.

domcleal:
#77 has been merged for Foreman 1.2 to do the same via rkerberos instead of klist/kinit.  Thanks for the contribution!

PR #30, opened 2012-08-17
Ipmi

logicminds:
This will add support to smartproxy that allows bmc management via http(s).  

authentication is handled via http basic auth.


ohadlevy:
merged - thanks!

PR #31, opened 2012-08-22
Allow a Squid proxy to be managed by the Smart Proxy

doherty:
When you are provisioning hosts that cannot reach the
Foreman server due to the network's design, you may
want to have a Squid proxy running on the subnet to
provide access. This permits such a proxy to be
managed with the Smart Proxy.

Currently, "managed" means adding and removing files
which control access to the proxy for a single host:

```
acl foreman_clients src host_that_cant_reach_foreman
```

This permits the host to use the Squid proxy to reach
Foreman, for example when notifying the server that the
build has completed.

Set "squid: true" and "squid_conf_dir" to the configuration
directory for your proxy (typically, /etc/squid or
/etc/squid3) to enable Squid management. Inside this
directory, you must create a directory called "foreman.d"
which will store the files that the Smart Proxy manages.
Include the directory with "include /etc/squid/foreman.d/*.conf"
and provide a url_regex ACL called "foreman_url" - this is the
resource that the Squid proxy will permit access to for the
ACL entries that the Smart Proxy manages.

An example configuration follows:

```
#           hostname                type    proxy port  icp port    options
cache_peer  foreman.localdomain     parent  80          0           no-query

http_port   80

acl Safe_ports port 80 21 443 563 70 210 1025-65535
http_access deny !Safe_ports

acl SSL_ports port 443
acl CONNECT method CONNECT
http_access deny CONNECT !SSL_ports

acl foreman_url url_regex ^http://foreman.localdomain/unattended/
include /etc/squid3/foreman.d/*.conf

# Permit access only to known safe ports
http_access deny !safe_ports
http_access allow foreman_clients foreman_url

# Must be the last rule
http_access deny all
```

The user the Smart Proxy runs as must be able to use sudo
to run the squid binary as root without a password.

Caveats:
- There is no test suite for this feature yet

Resolves http://theforeman.org/issues/969

### DIFF
--- a/lib/proxy/squid.rb
+++ b/lib/proxy/squid.rb
@@ -5,7 +5,27 @@ module Proxy::Squid
   require "proxy/settings"
 
   class << self
-    def reconfigure *host
+    def add *hosts
+      hosts.each do |ip_addr|
+        filename = File.join(SETTINGS.squid_conf_dir, 'foreman.d', "#{ip_addr}.conf")
+        File.open( filename, 'w' ) do |f|
+          f.puts "acl foreman_clients src #{ip_addr}"
+        end
+      end
+      reconfigure
+    end
+
+    def rm *hosts
+      hosts.each do |ip_addr|
+        filename = File.join(SETTINGS.squid_conf_dir, 'foreman.d', "#{ip_addr}.conf")
+        File.unlink( filename )
+      end
+      reconfigure
+    end
+  end
+
+  private
+  def reconfigure *host
       path  = ['/usr/sbin', '/usr/bin', '/opt/squid3/bin', '/opt/squid/bin']
       squid = which('squid3', path) || which('squid', path)
       sudo  = which('sudo', '/usr/bin')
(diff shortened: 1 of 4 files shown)


doherty:
This goes with theforeman/foreman#142

doherty:
One caveat I forgot to mention is the lack of access control. I didn't add anything to restrict access to the Squid functions - only the Foreman server should be able to access those.

lib/proxy/squid.rb
```diff
@@ -0,0 +1,51 @@
+module Proxy::Squid
+  extend Proxy::Log
+  extend Proxy::Util
+
+  require "proxy/settings"
+
+  class << self
+    def reconfigure *host
```
ohadlevy: I guess reconfigure can be a private method?

lib/squid_api.rb
```diff
@@ -0,0 +1,19 @@
+# Should only be accessible by Foreman
+class SmartProxy
+  # Having two handlers isn't DRY
```
ohadlevy: I actually think it does make sense to split them, as one should be a POST and the other should be DELETE (to conform with REST)

lib/squid_api.rb
```diff
@@ -0,0 +1,19 @@
...
+  post %r{/squid/(add|rm)$} do
+    require 'pp'
+    pp params
```
ohadlevy: debug leftover?

ohadlevy:
this is pretty awesome work!

one concern I have is if we can use the new UUID token based deployment with this approach too (e.g. can squid decide on ACL based on a request param - as anaconda etc cant send additional headers afaik), or worst case this would be limited to only ip based deployments?

secondly, any chance you would be willing to add the puppet installer bits? it would not be trivial for everyone to configure sudo etc - thanks!

doherty:
Sorry, I don't work with Karl any more, so someone else will need to take this and theforeman/foreman#142 on.

dLobatog:
Closing this as the author mentioned someone else should take it on the Foreman/foreman PR

PR #33, opened 2012-09-11
fixes #1856 - adds a config.ru and allow the SP to run as a rack app.

ohadlevy:


skottler:
Do you think we should also add the rackup configuration values to settings.yml.example?

```
    environment: production
    address: 0.0.0.0
    user: root
    group: root
    port: 4567
    pid: extra/proxy.pid
    rackup: ../config.ru
    max_conns: 1024
    timeout: 30
    max_persistent_conns: 512
    daemonize: true
```

If they don't get included then you'll have to set them manually. The lack of these values isn't an issue if you're running the proxy via Passenger, but it is one if you use unicorn or Thin.

ohadlevy:
@skottler yep, makes sense, does that file has to be static?

skottler:
Yup, I believe the YAML does have to be static. Unfortunately that also make's it very hard to configure the proxy in unknown configurations (like those many Foreman users have) so relative paths seem to be the answer to this problem. There was a pull request merged last week that creates the parent directory for the log from the settings; we could add that same thing fr the pid and any other directories that need to exist. Not completely sure if that answers the question - let me know if you need for info.

ohadlevy:
@skottler so I guess it means we cant extract values from settings into rackup, it does make sense (e.g. when using passenger) that the port values you specify in a vhost are the one that going to apply anyway.

so was that a +1 to merge it as is?

GregSutcliffe:
Tested on Passenger, works perfectly. I think most users are on Passenger for Foreman, so we can get to Unicorn/Thin/etc as we need to. +1 to merge from me :)

skottler:
Merge on! I will submit a pull request with the additional settings.

PR #35, opened 2012-10-31
#Fixes 1761 - default to /etc/puppet if we can't find a value for $confdir

GregSutcliffe:


lib/proxy/puppet/environment.rb
```diff
@@ -49,8 +49,13 @@ def puppet_environments
...
           if modulepath and modulepath.include?("$environment")
             # expand $confdir if defined and used in modulepath
-            if conf[:main][:confdir] and modulepath.include?("$confdir")
-              modulepath.sub!("$confdir", conf[:main][:confdir])
+            if modulepath.include?("$confdir")
+              if conf[:main][:confdir]
+                modulepath.sub!("$confdir", conf[:main][:confdir])
+              else
+                # /etc/puppet is the default if $confdir is not defined
+                modulepath.sub!("$confdir", "/etc/puppet")
```
GregSutcliffe, line 57: I'd love to use "puppet config --mode master print confdir" instead of assuming "/etc/puppet" but unfortunately the proxy is running as non-root, so you get "/usr/share/foreman-proxy/.puppet" back....

ohadlevy:
thanks @GregSutcliffe

PR #36, opened 2012-11-07
fixes #1929 - set umask sensibly to prevent world writable files

domcleal:
Previous umask wasn't very sensible at all, this locks it down.

It does restrict group writes, but I can't think of an instance where this would be needed for files created by the proxy.


domcleal:
Updated the branch as CVE-2012-5477 has been assigned by the Red Hat SRT.

ohadlevy:
merged - thanks!

PR #42, opened 2012-11-21
fixes #1915 - load environments from Puppet 3

domcleal:
Two changes to support loading environment settings from Puppet 3:
- explicitly specify the run_mode as being 'master' to ensure the config is
  parsed in a puppetmaster context
- call private method to fully clear the state in Puppet::Settings allowing for
  a full reparsing of the config to pick up changes in environments (Puppet bug
  #17783)

This supersedes GH-40 as it explicitly loads the master run_mode and will allow for environments to be reloaded without restarting the proxy.  I've filed #17783 with Puppet to give us  back a better public interface for doing this, it should included in the Puppet 3.1 refactor of the settings stuff.

### DIFF
--- a/lib/proxy/puppet/environment.rb
+++ b/lib/proxy/puppet/environment.rb
@@ -21,10 +21,25 @@ module Proxy::Puppet
 
       def puppet_environments
         Puppet.clear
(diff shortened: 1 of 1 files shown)


ohadlevy:
merged - thanks @domcleal

PR #44, opened 2012-11-26
Store and re-use the last IP we pinged

GregSutcliffe:
Store the last successful IP returned to the Foreman UI in a file in /tmp. Use a file lock to make sure concurrent process don't try and offer the same IP to two users.

### DIFF
--- a/lib/proxy/dhcp/subnet.rb
+++ b/lib/proxy/dhcp/subnet.rb
@@ -1,9 +1,12 @@
 require 'checks'
 require 'ipaddr'
 require 'proxy/dhcp/monkey_patches' unless IPAddr.new.respond_to?('to_range')
+require 'proxy/dhcp/monkey_patch_subnet' unless Array.new.respond_to?('rotate')
 require 'ping' unless RUBY_1_9
 require 'proxy/validations'
 require 'net/ping'
+require 'timeout'
+require 'tmpdir'
 
 module Proxy::DHCP
   # Represents a DHCP Subnet
@@ -32,6 +35,10 @@ module Proxy::DHCP
 
     def to_s
       "#{network}/#{netmask}"
+    end
+
+    def cidr
+      IPAddr.new(netmask).to_i.to_s(2).count("1")
     end
 
     def range
@@ -104,6 +111,30 @@ module Proxy::DHCP
       return false
     end
 
+    def get_index_and_lock filename
+      # Store for use in the unlock method
+      @filename = "#{Dir::tmpdir}/#{filename}"
+      @lockfile = "#{@filename}.lock"
+
+      # Loop if the file is locked
+      Timeout::timeout(30) { sleep 0.1 while File.exists? @lockfile }
+
+      # Touch the lock the file
+      File.open(@lockfile, "w") {}
+
+      @file = File.new(@filename,'r+') rescue File.new(@filename,'w+')
+
+      # this returns the index in the file
+      return @file.readlines.first.to_i rescue 0
+    end
+
+    def set_index_and_unlock index
+      @file.reopen(@filename,'w')
+      @file.write index
+      @file.close
+      File.delete @lockfile
+    end
+
     # returns the next unused IP Address in a subnet
     # Pings the IP address as well (just in case its not in Proxy::DHCP)
     def unused_ip args = {}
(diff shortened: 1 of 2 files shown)


GregSutcliffe:
@ohadlevy first pass at this. Two questions:

Should we make the filename a Setting?
Is a lock over the whole IP ping process acceptable?

lib/proxy/dhcp/subnet.rb
```diff
@@ -104,6 +104,27 @@ def add_record record
       return false
     end
 
+    def get_ip_and_lock
+      # Use Settings here
+      filename = "/tmp/foreman-proxy_ip-store.tmp"
+
+      # Lock the file
+      (@file = File.new(filename,'r+') rescue File.new(filename,'w+')).flock(File::LOCK_EX)
```
ohadlevy: I dont think that would work, assuming one proxy can handle multiple subnets

lib/proxy/dhcp/subnet.rb
```diff
@@ -119,16 +140,34 @@ def unused_ip args = {}
...
-            return ip
+        ip = false
+        begin
+          # Lock the storage file
+          if stored_ip = get_ip_and_lock
+            logger.debug "Got stored last_ip from file: #{stored_ip}"
+            # rescue true to start from the beginning if the stored IP has been taken since
+            index = (free_ips.index(stored_ip)+1) rescue 0
+            # Rotate the array to where we last got to, plus one
+            free_ips.rotate!(index)  # rotate! is 1.9 specific
```
ohadlevy: then why do we use it if it doesnt work on 1.87?

lib/proxy/dhcp/subnet.rb
```diff
@@ -119,16 +140,34 @@ def unused_ip args = {}
...
+              logger.debug "Found free ip #{ip} out of a total of #{free_ips.size} free ips"
+              set_ip_and_unlock ip
+              return ip
+            end
           end
+          logger.warn "No free IPs at #{to_s}"
+          set_ip_and_unlock
+        rescue Exception => e
+          logger.debug e.message
+          # ensure we unlock the storage file
```
ohadlevy: i would just put the unlock in  an ensure call?

lib/proxy/dhcp/subnet.rb
GregSutcliffe: Good point. I'll add the subnet to the filename tomorrow.

lib/proxy/dhcp/subnet.rb
GregSutcliffe: Because I forgot to change it before submitting :) Will fix tomorrow.

lib/proxy/dhcp/subnet.rb
GregSutcliffe: I tried that, but I had issues getting the value of "ip" (when successful) into the ensure block - it was coming up nil. I thought I'd handled the scope correctly, so I took it out. I'll try again with an instance variable tomorrow

GregSutcliffe:
@ohadlevy done. The only thing left is maybe to cleanup old lockfiles when the proxy starts up?

GregSutcliffe:
@ohadlevy also I can't find a ticket for this on the smartproxy issue tracker. Shall we create one, or did I just miss it?

lib/proxy/dhcp/subnet.rb
```diff
@@ -34,6 +36,11 @@ def to_s
       "#{network}/#{netmask}"
     end
 
+    def directory_safe_name_with_cidr
```
ohadlevy: add a cidr method instead?

lib/proxy/dhcp/subnet.rb
```diff
@@ -104,6 +111,30 @@ def add_record record
       return false
     end
 
+    def get_index_and_lock filename
+      # Store for use in the unlock method
+      @filename = filename
+      @lockfile = "#{filename}.lock"
+
+      # Loop if the file is locked
+      Timeout::timeout(30) { sleep 1 while File.exists? @lockfile }
```
ohadlevy: 1s sounds pretty long? can we drop it to 0.1 or so ?

lib/proxy/dhcp/subnet.rb
```diff
@@ -119,16 +150,28 @@ def unused_ip args = {}
...
-          logger.debug "Searching for free ip - pinging #{ip}"
-          if tcp_pingable?(ip) or icmp_pingable?(ip)
-            logger.info "Found a pingable IP(#{ip}) address which does not have a Proxy::DHCP record"
-          else
-            logger.debug "Found free ip #{ip} out of a total of #{free_ips.size} free ips"
-            return ip
+        @index = 0
+        begin
+          # Read and lock the storage file
+          stored_index = get_index_and_lock("/tmp/foreman-proxy_#{directory_safe_name_with_cidr}.tmp")
```
ohadlevy: /tmp wont work on windows? I would suggest to figure out the tmp dir in a different way?

ohadlevy: also, not sure if we want to put it /tmp, maybe /var/run or any other better place for it?

maybe we should consider adding a proxy tmp dir (just like rails does)

GregSutcliffe: Maybe use Tempfile? http://www.ruby-doc.org/stdlib-1.9.3/libdoc/tempfile/rdoc/Tempfile.html

That should be cross platform.

lib/proxy/dhcp/subnet.rb
GregSutcliffe: Sure, if you prefer.

GregSutcliffe:
Updated with discussed changes. I went with Dir::tmpdir which should return the correct tmp path for each operating system.

ohadlevy:
merged - thanks!

PR #45, opened 2012-11-28
fixes #1983 - use /etc/puppet/puppet.conf by default

domcleal:
When initialising Puppet 3 settings directly, a config dir or file is required,
so default to /etc/puppet/puppet.conf (as the puppetca code also does).

(As mentioned in the commit message, the puppetca code already has a fallback for /etc/puppet, so the environments code should do the same.)

### DIFF
--- a/lib/proxy/puppet/environment.rb
+++ b/lib/proxy/puppet/environment.rb
@@ -28,8 +28,12 @@ module Proxy::Puppet
           Puppet.settings.send(:clear_everything_for_tests)
         end
 
-        Puppet[:config] = SETTINGS.puppet_conf if SETTINGS.puppet_conf
-        raise("Cannot read #{Puppet[:config]}") unless File.exist?(Puppet[:config])
+        if SETTINGS.puppet_conf
+          Puppet[:config] = SETTINGS.puppet_conf
+          raise("Cannot read #{Puppet[:config]}") unless File.exist?(Puppet[:config])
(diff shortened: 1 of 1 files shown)


lib/proxy/puppet/environment.rb
```diff
@@ -28,8 +28,12 @@ def puppet_environments
           Puppet.settings.send(:clear_everything_for_tests)
         end
 
-        Puppet[:config] = SETTINGS.puppet_conf if SETTINGS.puppet_conf
-        raise("Cannot read #{Puppet[:config]}") unless File.exist?(Puppet[:config])
+        if SETTINGS.puppet_conf
+          Puppet[:config] = SETTINGS.puppet_conf
+          raise("Cannot read #{Puppet[:config]}") unless File.exist?(Puppet[:config])
+        elsif Puppet::PUPPETVERSION.to_i >= 3
+          Puppet[:config] = "/etc/puppet/puppet.conf"
```
ohadlevy, line 35: Why only for puppet v3? would it be safe to assume we can set the default regardless?

domcleal, line 35: I made it v3 only on the off-chance that it wasn't safe and since v2 is known to always read the system-wide puppet.conf.  It probably is safe to override, but not sure it's of any benefit.

ohadlevy:
merged - thanks @domcleal

PR #47, opened 2012-11-28
Fixes #1985 - Replace ping with nmap

GregSutcliffe:
Replaces ping with nmap. Checks if the binary is on the system and raises a sensible error if not.

As for removing tcp_pingable, from the nmap man page:

The default host discovery done with -sn consists of an ICMP echo request, TCP SYN to port 443, TCP ACK to port 80, and an ICMP timestamp request by default. When executed by an unprivileged user, only SYN packets are sent (using a connect call) to ports 80 and 443 on the target.

### DIFF
--- a/lib/proxy/dhcp/subnet.rb
+++ b/lib/proxy/dhcp/subnet.rb
@@ -2,9 +2,7 @@ require 'checks'
 require 'ipaddr'
 require 'proxy/dhcp/monkey_patches' unless IPAddr.new.respond_to?('to_range')
 require 'proxy/dhcp/monkey_patch_subnet' unless Array.new.respond_to?('rotate')
-require 'ping' unless RUBY_1_9
 require 'proxy/validations'
-require 'net/ping'
 require 'timeout'
 require 'tmpdir'
 
@@ -16,6 +14,7 @@ module Proxy::DHCP
 
     include Proxy::DHCP
     include Proxy::Log
+    include Proxy::Util
     include Proxy::Validations
 
     def initialize server, network, netmask
@@ -217,23 +216,21 @@ module Proxy::DHCP
     end
 
     private
-    def tcp_pingable? ip
-      Net::Ping::TCP.service_check=true
-      Net::Ping::TCP.new(nil,nil,1).ping(ip)
+    def pingable? ip
+      unless which("nmap", ENV["PATH"])
+        logger.debug "No nmap executable, falling back to ping"
+        return icmp_pingable?
+      end
+      # nmap covers tcp and icmp in a single check, so no need for two methods
+      logger.debug "running nmap on #{ip}"
+      (`nmap -n -sn -PR -PE -PP -PS22,80,443 #{ip}` =~ /Host seems down/).nil?
     rescue
       # We failed to check this address so we should not use it
       true
     end
 
     def icmp_pingable? ip
-      if privileged_user
-        Net::Ping::ICMP.new(nil,nil,1).ping(ip)
-      else
-        system("ping -c 1 -W 1 #{ip} > /dev/null")
-      end
-    rescue
-      # We failed to check this address so we should not use it
-      true
+      system("ping -c 1 -W 1 #{ip} > /dev/null")
     end
 
     def privileged_user
--- a/Rakefile
+++ b/Rakefile
@@ -47,7 +47,6 @@ spec = Gem::Specification.new do |s|
   s.extra_rdoc_files = ["README"]
   s.add_dependency 'json'
   s.add_dependency 'sinatra'
-  s.add_dependency 'net/ping'
   s.rubyforge_project = 'rake'
   s.description = <<EOF
 Foreman Proxy is used via The Foreman Project, it allows Foreman to manage


lib/proxy/dhcp/subnet.rb
```diff
@@ -121,7 +119,8 @@ def unused_ip args = {}
       else
         free_ips.each do |ip|
           logger.debug "Searching for free ip - pinging #{ip}"
-          if tcp_pingable?(ip) or icmp_pingable?(ip)
+          `which nmap` ; raise "No nmap executable" unless $?.success?
```
ohadlevy: you should use our internal version of which (much faster then shelling out to which)

lib/proxy/dhcp/subnet.rb
```diff
@@ -174,20 +173,10 @@ def broadcast
...
-    def icmp_pingable? ip
-      if privileged_user
-        Net::Ping::ICMP.new(nil,nil,1).ping(ip)
-      else
-        system("ping -c 1 -W 1 #{ip} > /dev/null")
-      end
+    def pingable? ip
+      # nmap covers tcp and icmp in a single check, so no need for two methods
+      logger.debug "running nmap on #{ip}"
+      (`nmap -sn #{ip}` =~ /Host seems down/).nil?
```
ohadlevy: what about other options (e.g. timeout?)

GregSutcliffe: In my tests, the timeout was fast (~1s) by default. There are options for it if you think it should be set.

GregSutcliffe:
By the way, this paves the way for doing something like "nmap -sn 192.168.1.10-200" which is much faster than scanning individual IPs and returns a list of all 'up' IP addresses

lib/proxy/dhcp/subnet.rb
GregSutcliffe: oh, yeah. fixing now

GregSutcliffe: Done

bgupta:
Greg, any chance we can add port 22 to the list of nmap ports we scan for? -Brian

ohadlevy:
@GregSutcliffe can we

# fallback to ping ? (if nmap is not installed)

# ensure that nmap does arp based detection too (not sure if that requires root permissions, if it does, we maybe want to fall back to execute arp).

GregSutcliffe:
@bgupta I'll see if that's easily doable. Probably is.

@ohadlevy no need to shout :) A ping fallback is easy enough. Arp does require root I think, and is done by default if nmap is run as root:

"-sn: ... When a privileged user tries to scan targets on a local ethernet network, ARP requests are used unless --send-ip was specified."

GregSutcliffe:
@bgupta @ohadlevy I've changed it to: nmap -sn -PR -PP -PS22,80,443 $ip

so that's -PR for an Arp scan (does nothing if we're not root, so safe to include), -PE for an ICMP echo-request, -PP for an ICMP echo-timestamp (because echo-request is always blocked :P), and finally 3 TCP SYN connections to 22, 80 and 443.

If all of that fails, I think we can probably use the IP - or at least accuse the user of being overly paranoid :)

This produces the same list of in-use IPs on my network as both root and non-root. So, I think it's good to go....

lib/proxy/dhcp/subnet.rb
```diff
@@ -174,23 +173,21 @@ def broadcast
     end
 
     private
-    def tcp_pingable? ip
-      Net::Ping::TCP.service_check=true
-      Net::Ping::TCP.new(nil,nil,1).ping(ip)
+    def pingable? ip
+      if which("nmap", ENV["PATH"]) == false
```
ohadlevy: unless ?

lib/proxy/dhcp/subnet.rb
```diff
@@ -174,23 +173,21 @@ def broadcast
...
+        logger.debug "No nmap executable, falling back to ping"
+        return icmp_pingable?
+      end
+      # nmap covers tcp and icmp in a single check, so no need for two methods
+      logger.debug "running nmap on #{ip}"
+      (`nmap -sn -PR -PE -PP -PS22,80,443 #{ip}` =~ /Host seems down/).nil?
```
ohadlevy: my tests took 11 second to run this command for a machine which is up
for example:

nmap -sn -PR -PE -PP -PS22,80,443 host

```
Warning:  You are not root -- using TCP pingscan rather than ICMP

Starting Nmap 6.01 ( http://nmap.org ) at 2012-12-02 10:55 IST
Stats: 0:00:07 elapsed; 0 hosts completed (0 up), 1 undergoing Ping Scan
Parallel DNS resolution of 1 host. Timing: About 0.00% done
Nmap scan report for file (10.35.255.8)
Host is up (0.00034s latency).
rDNS record for 10.35.255.8:xx.yy.zz
Nmap done: 1 IP address (1 host up) scanned in 11.23 seconds
```

GregSutcliffe: Interesting. I can't get it to take longer that 3s on my local network. Try adding '--max-rtt-timeout 100ms' and see if it helps.

Also, try the full command on a subnet - for my network, 'nmap -sn -PR -PE -PP -PS22,80,443 --max-rtt-timeout 100ms 172.20.10.*' returned all 10 hosts in 2.7s. That's compered to 1.2s for a single host. So once we merge this I can look at how to make use of this ;)

lib/proxy/dhcp/subnet.rb
GregSutcliffe: You mean something like "unless which("nmap", ENV["PATH"]) " with no further tests? I can do that :)

GregSutcliffe:
Updated with "-n" to avoid the DNS lookup time factor

GregSutcliffe:
Suspending this until https://github.com/theforeman/smart-proxy/pull/49 is merged, at which point I'll rebase for further discussion.

GregSutcliffe:
Going to close this, i don't think we're going to do it any time soon ;)

PR #49, opened 2012-12-07
Remove dependency on net-ping

GregSutcliffe:
Tested on ruby 1.8 and ruby 1.9.


lib/proxy/dhcp/subnet.rb
```diff
@@ -218,19 +217,47 @@ def broadcast
 
     private
     def tcp_pingable? ip
-      Net::Ping::TCP.service_check=true
-      Net::Ping::TCP.new(nil,nil,1).ping(ip)
+      # This code is from net-ping, and stripped down for use here
+      # We don't need all the ldap dependencies net-ping brings in
+
```
ohadlevy, line 222: any tests we can borrow? :)

ohadlevy:
rake test does not work on 1.9... merged regardless

PR #51, opened 2012-12-31
fixes #2077 - proxy should not return nil for parameters with a function...

ohadlevy:
... call

in cases where puppet has class declaration such as:

  class xyz($param = myfunction('p'))

the proxy would return a that param value is nil.
this patch changes that, so it return it as a string, allowing users to know
that a function call was there, and allow them to override it if they want.

### DIFF
--- a/lib/proxy/puppet/puppet_class.rb
+++ b/lib/proxy/puppet/puppet_class.rb
@@ -96,6 +96,8 @@ module Proxy::Puppet
             when Puppet::Parser::AST::ASTHash
               Hash[value.value.each.inject([]) { |arr, (k,v)| (arr << [ast_to_value(k), ast_to_value(v)]) rescue arr }]
             # Let's see if a raw evaluation works with no scope for any other type
+            when Puppet::Parser::AST::Function
+              "${#{value}}"
             else
               if value.respond_to? :evaluate
                 # Can probably work for: (depending on the actual content)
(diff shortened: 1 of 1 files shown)


lib/proxy/puppet/puppet_class.rb
```diff
@@ -96,6 +96,8 @@ def ast_to_value value
             when Puppet::Parser::AST::ASTHash
               Hash[value.value.each.inject([]) { |arr, (k,v)| (arr << [ast_to_value(k), ast_to_value(v)]) rescue arr }]
             # Let's see if a raw evaluation works with no scope for any other type
```
domcleal, line 98: This comment should be next to the "else" clause

lib/proxy/puppet/puppet_class.rb
```diff
@@ -96,6 +96,8 @@ def ast_to_value value
...
+            when Puppet::Parser::AST::Function
+              "${#{value}}"
```
domcleal, line 100: This is just cosmetic, isn't it?  As `${function(..)}` isn't valid puppet syntax, maybe just return `value`.

ohadlevy, line 100: it would return:

```
${ffunction_name(parameters)}
```

just like we do for parameters, would it be better to return plain:

```
function_name(parameters)
```

?

domcleal, line 100: I think plain `function_name(params)` since that's valid puppet DSL syntax.  `${params}` is valid, so it makes sense to see it in the UI, but `${function(params)}` isn't valid.

ohadlevy:
merged - thanks!

PR #52, opened 2013-01-04
#2101 gemfile tests

domcleal:


ohadlevy:
when i run bundle exec rake, not all tests are passing, there are more errors on 1.87 than on 1.9 but sitll:

e.g

```
 5) Failure:
test_should_find_class_in_a_manifest(PuppetClassTest)
    [/home/olevy/git/proxy/test/puppet_class_test.rb:44:in `test_should_find_class_in_a_manifest'
     /home/olevy/.gem/gems/mocha-0.13.1/lib/mocha/integration/test_unit/ruby_version_186_and_above.rb:29:in `__send__'
     /home/olevy/.gem/gems/mocha-0.13.1/lib/mocha/integration/test_unit/ruby_version_186_and_above.rb:29:in `run']:
<1> expected but was
<0>.
```

Gemfile
```diff
@@ -0,0 +1,8 @@
+source 'http://rubygems.org'
+
+gem 'sinatra'
```
ohadlevy: i would add

``` ruby
gem 'json'
```

too

domcleal:
I can reproduce that particular failure with Puppet 3 - it appears to be [#2085](http://theforeman.org/issues/show/2085).  There are other failures though which are due to the tests expecting a certain settings.yml setup, so I'd like to fix that - will update the PR once I've got it working.

domcleal:
@ohadlevy branch updated.  This should now work from an entirely clean checkout after copying settings.yml.example in place.  The Puppet errors you saw were #2085, so they should now be gone (plus I've added a dummy puppet.conf so you don't need it installed system-wide).

Gemfile
domcleal: Fixed, thanks.

ohadlevy:
merged - thanks @domcleal

PR #53, opened 2013-01-07
fix handling of multiple module paths

luxflux:
Replace every occurrence of $confdir, not only the first one


ohadlevy:
@luxflux looks good to me, @domcleal ?

domcleal:
+1 to merge.

ohadlevy:
@domcleal do you know if we have an open ticket for this one?

domcleal:
Not that I'm aware of.  We've got a very similar issue in [#2099](http://theforeman.org/issues/2099) where $confdir's only interpolated if $environment is present (due to how the conditionals there are structured).  We could bundle them together.

dcleal:
Don't think you meant to send this to me....
- Dave Cleal

PR #54, opened 2013-01-07
Add warning if no environments found 

DoubleMalt:
... and smart-proxy falls back to defaults

### DIFF
--- a/lib/proxy/puppet/environment.rb
+++ b/lib/proxy/puppet/environment.rb
@@ -61,6 +61,7 @@ module Proxy::Puppet
         if env.values.compact.size == 0
           # fall back to defaults - we probably don't use environments
           env[:production] = conf[:main][:modulepath] || conf[:master][:modulepath] || '/etc/puppet/modules'
+          logger.warn "No environments found - falling back to defaults (production - #{env[:production]})"
         end
 
         new_env = env.clone


lib/proxy/puppet/environment.rb
```diff
@@ -60,6 +60,7 @@ def puppet_environments
         end
         if env.values.compact.size == 0
           # fall back to defaults - we probably don't use environments
+          logger.warn "No environments found - falling back to defaults (production - /etc/puppet/modules)"
```
domcleal: It'd be best to move it down a line and use `#{env[:production]}` just in case it picks up the modulepath from the config.

domcleal:
+1, thanks @DoubleMalt.  Could you prefix the commit with `fixes #2114 -` too?

DoubleMalt:
Done

domcleal:
Thanks, though looks like you've got two commits on the branch and two log lines now!  You might want to fix that, squash the commits (`git rebase -i HEAD~2`) and force-update this branch (`git push -f`).

DoubleMalt:
Done.

domcleal:
+1 to merge.  Thanks for the contribution!

GregSutcliffe:
This doesn't look quite right to me. That logger line will be printed regardless of which of the 3 logical clauses are matched on the line above - sure this log line only applies to the last of them?

DoubleMalt:
The line actually applies to all of them. The line above only concerns itself with the module path. For all cases it falls back to a default environment called "production".

GregSutcliffe:
Ah, I take your point. Whilst [:main][:modulepath] could point anywhere, it's still a default 'environment' called production.

+1 to merge, @ohadlevy

skottler:
Looks good to me, too. Thanks @DoubleMalt!

ohadlevy:
thanks @DoubleMalt !

PR #57, opened 2013-01-17
fixes #2085 - load Puppet 3 app defaults for master mode too

domcleal:
Allows $confdir and other run mode specific settings to be initialised from
the defaults present in Puppet::Util::RunMode.  When requesting detail on
classes, the Puppet parser indirectly requests app level settings so both
global and app settings are needed.  Refactored into shared code for both
parser and environments to use.

No new tests, since the old ones were already failing!


ohadlevy:
merged - thanks!

PR #59, opened 2013-01-21
Fixes some autosign.conf issues (comments, EOL on last line)

vStone:
Ignore commented lines when reading all certificates in autosign and make sure the autosign file contains a EOL on the last line

### DIFF
--- a/lib/proxy/puppetca.rb
+++ b/lib/proxy/puppetca.rb
@@ -35,6 +35,7 @@ module Proxy::PuppetCA
       if found
         autosign = open(autosign_file, File::TRUNC|File::RDWR)
         autosign.write entries.join("\n")
+        autosign.write "\n"
         autosign.close
         logger.info "Removed #{certname} from autosign"
       else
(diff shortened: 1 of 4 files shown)


lib/proxy/puppetca.rb
```diff
@@ -59,7 +59,7 @@ def autosign certname
 
     # list of hosts which are now allowed to be installed via autosign
     def autosign_list
-      File.exist?(autosign_file) ? File.read(autosign_file).split : []
+      File.exist?(autosign_file) ? File.read(autosign_file).split("\n").reject { |v| v =~ /^\s*#.*|^$/ } : []
```
domcleal: The change to split means trailing spaces aren't stripped any more, perhaps add a `.map {|v| v.chomp}` or similar?

domcleal:
+1 tested OK, with one comment above.

vStone:
Also, there seem to be no tests for parsing the autosign file? :/

domcleal:
@vStone thanks.  Indeed, contributions welcome!

vStone:
Added some tests and also fix some other tests that were failing for no good (trivial) reason.

ohadlevy:
@domcleal mind giving this a go if you have time before puppetcamp? I know I wont :(

domcleal:
:+1: to  4ddfc7f and  4fa205c, thanks for doing that!  For 4884d4a I'd prefer to merge GH-52 which modifies settings rather than explicitly loading subsystems.

ohadlevy:
merged without 4884d4a  - thanks @vStone!

PR #65, opened 2013-02-27
fixes #2261 - fixes for CI testing under Ruby 1.9

domcleal:
Support minitest in Jenkins rake tasks, add rdoc and minitest deps due to
old 1.9.2 versions.  Fix PLATFORM under 1.9, fix 1.9.3 specific test.


domcleal:
I've fixed the same issue as GH-61 here, but by adding `require 'checks'` as per my last comment.

ohadlevy:
merged - thanks

PR #70, opened 2013-03-15
fixes #2067 - set puppet parser filename so 'include' directives work

domcleal:


domcleal:
Please don't merge this yet, I've just realised the test doesn't actually fail without the patch.  Fixing.

domcleal:
fixed

ohadlevy:
merged - thanks!

PR #71, opened 2013-03-18
The path to puppetca is /opt/puppet/sbin for PE.

apenney:


skottler:
:+1:

domcleal:
We should keep both sbin and bin here I think, in anticipation of PE shipping Puppet 3 and going back to the "puppet" binary only.  Otherwise :+1:, @ohadlevy can probably fix that on merge.

[test]

logicminds:
This should contain both since pe version 1 uses bin.  It would be better to let the user define in config along with these default parameters.

ohadlevy:
fixed with @domcleal suggestion at 32d9749

PR #72, opened 2013-03-27
Add shell provider to the BMC API

GregSutcliffe:
- Only Accepts on/off/status/cycle for power commands
- Needs sudo for the proxy to call /sbin/shutdown
- Adds a simple test for calling the shutdown api

### DIFF
--- a/config/settings.yml.example
+++ b/config/settings.yml.example
@@ -62,10 +62,10 @@
 :puppet_conf: /etc/puppet/puppet.conf
 
 # enable BMC management  (Bare metal power and bios controls)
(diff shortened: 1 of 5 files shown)


logicminds:
I don't think a shell provider should be mixed in with BMC providers.  Its confusing to the user.  Maybe a simpler more generalized version of this that can run "power" events to all types of compute resources?

GregSutcliffe:
@logicminds I guess you could see it that way - I was looking at it from the "keep all the power management together" viewpoint. @ohadlevy do you have a preference?

ohadlevy:
[test]

ohadlevy:
merged - thanks

PR #73, opened 2013-03-29
Making proxy to be compliant with Sinatra 1.4.2 release

lzap:
This change in Sinatra 1.4.2

https://github.com/sinatra/sinatra/pull/626/files

introduces new class AcceptEntry which is used for comparison of the accept
flag. Since we use request.accept.include? it does not work anymore on my
system:

```
(rdb:3) pp request.accept
[application/json]

(rdb:3) pp request.accept.include?("application/json")
false

(rdb:3) pp request.accept[0].class
Sinatra::Request::AcceptEntry
```

This patch changes all places where we use this comparison and changes it to
the standard way. This should be compatible across old Sinatra versions, but
please confirm this explicitly.

### DIFF
--- a/lib/sinatra-patch.rb
+++ b/lib/sinatra-patch.rb
@@ -49,4 +49,13 @@ module Sinatra
       puts "== Someone is already performing on port #{port}!"
     end
   end
+
+  module MonkeyRequest
+    # We need request.accept? method also in pre-1.3.0 versions. This is simplified
+    # version of the method that only accept one parameter (mime type string).
+    def accept? type
+      accept.include? type
+    end
+  end
+  Request.send :include, MonkeyRequest if not Request.method_defined? :accept?
 end
--- a/lib/dhcp_api.rb
+++ b/lib/dhcp_api.rb
@@ -37,7 +37,7 @@ class SmartProxy < Sinatra::Base
 
   get "/dhcp" do
     begin
-      if request.accept.include?("application/json")
+      if request.accept? 'application/json'
         content_type :json
 
         log_halt 404, "No subnets found on server @{name}" unless @subnets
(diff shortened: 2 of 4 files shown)


lzap:
Ok I have tracked down what is our minimum version for my patch (search for "accept?"):

```
= 1.3.0 / 2011-09-30
```

https://github.com/sinatra/sinatra/blob/master/CHANGES

Is this acceptable?

domcleal:
+1, fixes it for me on sinatra 1.4.3 and 1.3.2.

domcleal:
1.3.0's not going to be old enough for our RHEL 6 install, which has rubygem-sinatra 1.0.2 [from EPEL](http://dl.fedoraproject.org/pub/epel/6/x86_64/repoview/rubygem-sinatra.html).

lzap:
Let me do a monkey patch for older Sinatra versions.

lzap:
Here it comes, tested with Sinatra 1.2.9 and 1.4.2.

domcleal:
[test]

ohadlevy:
@lzap does it mean that 1.2.9 also represents 1.02?

lib/sinatra-patch.rb
```diff
@@ -49,4 +49,13 @@ def self.run!(options={})
...
   end
+
+  module MonkeyRequest
+    # We need request.accept? method also in pre-1.3.0 versions. This is simplified
+    # version of the method that only accept one parameter (mime type string).
+    def accept? type
+      accept.include? type
+    end
+  end
+  Request.send :include, MonkeyRequest
```
ohadlevy: shouldnt this be restricted based on sinatra version?

lzap:
Right that is better. Rebased. [test]

ohadlevy:
merged - thanks

PR #74, opened 2013-04-19
Override the initialize method for the shell provider, as no connection ...

GregSutcliffe:
...setup is needed


ohadlevy:
merged at 9316a46

PR #75, opened 2013-04-19
REVIEW ONLY: fixes #1685 - GSS-TSIG support for DNS updates

domcleal:
Based on patch by theforemanuser123 oliver_weinmann@gmx.de.

---

Posting for review as an alternative to GH-28.  I found the kinit/klist method rather fragile when testing against FreeIPA (which also uses GSS-TSIG), so this is an alternative using the krb5_auth library.  It uses the djberg96-krb5-auth fork (the more actively maintained one) to request Kerberos credentials using the keytab and load them into the credential cache.

I've got a spec for EL6 for the rubygem here: https://github.com/domcleal/foreman-rpms/commit/76f424a8 (@skottler), though we'll use a proper release instead of patching 0.9.0 or using :git in the Gemfile.

### DIFF
--- a/lib/proxy/dns/nsupdate.rb
+++ b/lib/proxy/dns/nsupdate.rb
@@ -2,13 +2,18 @@ require "proxy/dns"
 require 'resolv'
 
 module Proxy::DNS
-  class Bind < Record
+  class Nsupdate < Record
 
     include Proxy::Util
     attr_reader :resolver
 
     def initialize options = {}
       raise "Unable to find Key file - check your dns_key settings" unless SETTINGS.dns_key == false or File.exists?(SETTINGS.dns_key)
+      if SETTINGS.dns_tsig_keytab
+        raise "Kerberos principal required - check dns_tsig_principal setting" unless SETTINGS.dns_tsig_principal
+        logger.debug "DNS GSS-TSIG authentication enabled"
+        require 'krb5_auth'
+      end
       super(options)
     end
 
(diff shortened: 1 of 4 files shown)


lib/proxy/dns/nsupdate.rb
```diff
@@ -52,6 +57,13 @@ def remove
 
     private
 
+    def nsupdate_args
+      args = ""
```
skottler, line 61: This line is superfluous.

skottler, line 61: Never mind - didn't see the SETTING.dns_tsig_keytab.

lib/proxy/dns/nsupdate.rb
```diff
@@ -2,13 +2,18 @@
...
 module Proxy::DNS
-  class Bind < Record
+  class Nsupdate < Record
 
     include Proxy::Util
     attr_reader :resolver
 
     def initialize options = {}
       raise "Unable to find Key file - check your dns_key settings" unless SETTINGS.dns_key == false or File.exists?(SETTINGS.dns_key)
+      if SETTINGS.dns_tsig_keytab
```
ohadlevy, line 12: why not have a tsig class < nsupdate ?

domcleal, line 12: Can do, would you prefer it?  I'm not fussed either way, I don't think it adds much.

domcleal:
Need to change this to use rkerberos instead (relaunch of the same library).

PR #76, opened 2013-04-29
Develop

pikelly:
Hi Ohad, I had some issues with performance using the native_ms handler so I added some caching and code optimization. I also fixed the issue where leases were not being captured. I hope that pull requests are a supported mechanism for submitting changes as I cannot remember how to use the git patch mechanism. :-( It has been a long time. . . .

### DIFF
--- a/lib/proxy/dhcp/server/native_ms.rb
+++ b/lib/proxy/dhcp/server/native_ms.rb
@@ -86,27 +86,26 @@ module Proxy::DHCP
 
     def loadSubnetData subnet
       super
-      cmd = "scope #{subnet.network} show reservedip"
+      cmd = "scope #{subnet.network} show client 1"
       msg = "Enumerated hosts on #{subnet.network}"
 
       # Extract the data
       execute(cmd, msg).each do |line|
-        #     172.29.216.6      -    00-a0-e7-21-41-00-
-        if line =~ /^\s+([\w\.]+)\s+-\s+([-a-f\d]+)/
-          ip  = $1
+        #192.29.205.4    - 255.255.255.128- 00-1e-68-65-55-f8   -4/23/2013 6:41:21 PM   -D-
+        #192.29.205.5    - 255.255.255.128-00-1b-24-93-35-09   - NEVER EXPIRES        -U-  host.brs.company.com
+        if (match = line.match(/^([\d\.]+)\s*-\s*[\d\.]+\s*- ?([-a-f\d]+)\s*-\s*([^-]+?)\s*-(\w)-\s*(.*)/))
+          ip, mac, expire, kind, name  = match[1,5]
           next unless subnet.include?(ip)
-          mac = $2.gsub(/-/,":").match(/^(.*?).$/)[1]
+          # Some mac addresses appear to be more than 6 bytes!
+          mac = mac[0,17].gsub!(/-/,":")
           begin
-            opts = {:subnet => subnet, :ip => ip, :mac => mac}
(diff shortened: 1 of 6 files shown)


ohadlevy:
[test]

lib/proxy/dhcp/server/native_ms.rb
```diff
@@ -221,28 +221,30 @@ def report msg, response, error_only
...
-        when /For vendor class \[([^\]]+)\]:/
-          options[:vendor] = "<#{$1}>"
-        when /OptionId : (\d+)/
-          optionId = $1.to_i
-        when /Option Element Value = (\S+)/
-          #TODO move options to a class or something
-          opts = SUNW.update(Standard)
+    def parse_options ip
+      dump_file = "#{name}.dump"
+      if !File.exist?(dump_file) or Time.now > File.mtime(dump_file) + 5 * 60
```
ohadlevy: where does this gets stored? while its on windows platform only, I still assume there are some potential file security limitations? :)

ohadlevy: also, should that be configurable (duration)? how about be able to force flush it? I assume that after updating a record you want to flush the cache for the next request?

GregSutcliffe: This looks like it will go in the cwd of the proxy execution. I think that should be set to the OS Tempdir (via `Dir::tmpdir`) like we do in https://github.com/theforeman/smart-proxy/blob/develop/lib/proxy/dhcp/subnet.rb#L115

lib/proxy/dhcp/server/native_ms.rb
```diff
@@ -221,28 +221,30 @@ def report msg, response, error_only
...
+        dump = execute("dump", "dummy message", false, true)
+        File.new("#{name}.dump", "w").puts(dump)
```
GregSutcliffe: Is it possible to get a collision here? If there's no options cache and two clients both make a request at the same time, will two calls to this code result in errors, or maybe a mangled file? Might be safest to use a lock file on it....

lib/proxy/dhcp/server/native_ms.rb
```diff
@@ -221,28 +221,30 @@ def report msg, response, error_only
...
+        @options_cache = {}
+      end
+      if @options_cache.empty?
+        dump = File.open("#{name}.dump").readlines
```
GregSutcliffe: Another file lock here I think - what happens if clientA is reading this file and clientB causes it to update (because the cache timeout has _just_ been hit)?

ohadlevy:
@pikelly ping ?

pikelly:
Wow. I see that I have missed a few points! OK. I will take these valid criticisms into account.

ohadlevy:
@pikelly why did you close it?

pikelly:
Oops. I think that I closed the request by mistake. This github thing is cool but takes a bit of betting used to

ohadlevy:
@pikelly thanks for updating, sadly GH wont send a notification when pushing new code, so this was missed a bit.

@GregSutcliffe can you please review - thanks!

GregSutcliffe:
Will review this week. Meantime, let's have Jenkins do some work :) [test]

GregSutcliffe:
@ohadlevy looks good to me. do we need any new tests?
@pikelly can you squash the commits and reword them, so that they're all "Fixes #NNNN - [comment]" - we require a Redmine reference for each commit now.

ohadlevy:
@pikelly ping ?

PR #77, opened 2013-05-08
fixes #1685 - GSS-TSIG support for DNS updates

domcleal:
Replaces GH-75 and GH-28.

Changes since GH-75:
- replaced djberg96-krb5-auth with [rkerberos](https://github.com/djberg96/rkerberos) - a relaunch of the same lib, though I'm waiting on a new release with a patch
- moved GSS-TSIG support into a separate class, added dns_provider setting

### DIFF
--- a/bundler.d/krb5.rb
+++ b/bundler.d/krb5.rb
@@ -1,0 +1,3 @@
+group :krb5 do
+  gem 'rkerberos', '>= 0.1.1'
+end
--- a/lib/dns_api.rb
+++ b/lib/dns_api.rb
@@ -1,8 +1,22 @@
-require "proxy/dns/bind"
-
 class SmartProxy
   def setup(opts)
-    @server = Proxy::DNS::Bind.new(opts.merge(:server => SETTINGS.dns_server))
+    raise "Smart Proxy is not configured to support DNS" unless SETTINGS.dns
+    case SETTINGS.dns_provider
+    when "nsupdate"
+      require 'proxy/dns/nsupdate'
+      @server = Proxy::DNS::Nsupdate.new(opts.merge(:server => SETTINGS.dns_server))
+    when "nsupdate_gss"
+      require 'proxy/dns/nsupdate_gss'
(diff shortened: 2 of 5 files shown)


bundler.d/krb5.rb
```diff
@@ -0,0 +1,3 @@
+group :krb5 do
+  gem 'rkerberos', '>= 0.2.0'
```
domcleal: Waiting on this version to get pushed.

GregSutcliffe: I'm not currently using bundler to do the proxy packaging (the proxy does actually run on system gems :P). That shouldn't stop this being merged, I'm just noting that it will break the current nightlies.

lib/dns_api.rb
```diff
@@ -1,8 +1,24 @@
-require "proxy/dns/bind"
+require "proxy/dns/nsupdate"
 
 class SmartProxy
   def setup(opts)
-    @server = Proxy::DNS::Bind.new(opts.merge(:server => SETTINGS.dns_server))
+    raise "Smart Proxy is not configured to support DNS" unless SETTINGS.dns
+    case SETTINGS.dns_provider
+    when "nsupdate"
+      require 'proxy/dns/nsupdate'
```
GregSutcliffe, line 6: same as line 2?

domcleal, line 6: oops, removed line 1, ta!

bundler.d/krb5.rb
domcleal: I've pushed this new version (0.1.1).

GregSutcliffe:
I don't have a GSS server to test the new code against, but I don't think you'll break the existing DNS inetgration with this, so good to merge from my view. Let's see what Jenkins thinks of it :)

ohadlevy:
[test]

domcleal:
Test failure turned out to be a problem with not finding some headers.. I'll fix rkerberos later to do that, set the appropriate environment variable for now.

Pushed as 7e72434.

PR #78, opened 2013-05-10
Fixes #2116: mcollective support for the proxy to initialize puppetrun

domcleal:
Reworked #58 a bit to copy the provider pattern used in DHCP/DNS.


domcleal:
(tested both puppetrun and mco paths)

GregSutcliffe:
Reviewed/tested. Looks great. :+1:

PR #79, opened 2013-05-12
fixes #2482 Global search in DHCP reservations/leases using MAC

ohadlevy:


lib/dhcp_api.rb
```diff
@@ -50,6 +50,20 @@ def load_subnet
     end
   end
 
+  get %r{\/dhcp\/(([a-f0-9]{1,2}:){5}[a-f0-9]{1,2})} do |mac, unused_match|
+    begin
+      log_halt 404, "No subnets found on this server" unless @subnets
+      records = @subnets.map{|s| s.has_mac?(mac)}.compact.delete_if{|r| r == false }
```
domcleal, line 56: Is this not the same as `@subnets.find_all{|s| s.has_mac?(mac)}`?

lib/dhcp_api.rb
```diff
@@ -50,6 +50,20 @@ def load_subnet
...
+      log_halt 404, "not found" unless records[0]
+    rescue => e
```
dmitri-d, line 58: Why a rescue here - doesn't look like we can get an exception?

domcleal:
Closing this so we're not duplicating work with @korekhov's #97.

PR #80, opened 2013-05-13
Fixes #2119 - Properly support colon-separated modulepath

GregSutcliffe:
We were not properly handling module paths with colons in. Due to a bad regex (/\$environment.*/,'/') the secondary paths are dropped, and then lost when Dir can't find the first parent.

Best viewed with ?w=1 as indentation has changed ;)


ohadlevy:
looks ok, but would need a test imho :)

GregSutcliffe:
@ohadlevy done - thanks for reminding me. In writing the test I found that my fix didn't work in all cases - this should be much better ;)

GregSutcliffe:
Hang on, it's still not quite right. More tests incoming.

GregSutcliffe:
@ohadlevy ok added _a lot_ more tests; there are so many environment corner-cases ;)

I had to move/add some fixtures so that some of the `$environment` code could be properly tested.

GregSutcliffe:
Pushed fix for array-sorting test errors.

ohadlevy:
[test]

domcleal:
Thanks @GregSutcliffe!  Hairy code indeed.  Merged as 722bba.

PR #81, opened 2013-05-16
Added support for building with tito

mbacovsky:
I used slightly modified spec file that is being used for building  foreman-proxy in katello,
added rel-eng dir with tito related stuff and added rhel/fedora specific stuff to extra dir.


mbacovsky:
Added support for SCL Test scratch build is here http://koji.katello.org/koji/taskinfo?taskID=35739

skottler:
Merged, thanks!

PR #85, opened 2013-05-21
fixes #2523 - add missing require

domcleal:


ohadlevy:
merged at 7776616  - thanks @domcleal

PR #86, opened 2013-05-21
Rkerberos dep in spec + SCL turned off

mbacovsky:


domcleal:
Thanks @mbacovsky for the patch, squashed and merged as 7e6f5650 and 2e10e585.

PR #87, opened 2013-05-23
DHCP global search by record + pxegrub option

korekhov:
DHCP global search by record (MAC, IP or hostname). Example:
- <proxy-host>:8443/dhcp/find/<MAC|IP|hostname>

DHCP option "pxegrub" added to allow to specify a path to pxegrub file
to override default menu.lst-<MAC>. Requires dhcpd.conf to have the
following 2 lines in it:

option grubmenu code 150 = text;
option dhcp-parameter-request-list = concat(option
dhcp-parameter-request-list,bootmenu);

Example:
curl -X POST http://dhcp:8443/dhcp/<network> -d 'hostname=blah' -d
'ip=<ip>' -d 'mac=<mac>' -d
'grubmenu=pxelinux.cfg/01-<mac-with-dashes>' -d 'filename=pxegrub.0'

### DIFF
--- a/lib/proxy/dhcp/server.rb
+++ b/lib/proxy/dhcp/server.rb
@@ -76,7 +76,7 @@ module Proxy::DHCP
     def find_record record
       subnets.each do |s|
         s.records.each do |v|
-          return v if record.is_a?(String) and (v.ip == record or v.mac == record)
+          return v.options if record.is_a?(String) and (v.ip == record or v.mac == record or v.options[:hostname] == record)
           return v if record.is_a?(Proxy::DHCP::Record) and v == record
           return v if record.is_a?(IPAddr) and v.ip == record.to_s
         end
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ bundler.d/Gemfile.local.rb
 .rvmrc*
 .ruby-version
 .bundle
+tmp
--- a/lib/dhcp_api.rb
+++ b/lib/dhcp_api.rb
@@ -50,6 +50,17 @@ class SmartProxy < Sinatra::Base
     end
   end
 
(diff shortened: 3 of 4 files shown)


lib/proxy/dhcp/server.rb
```diff
@@ -76,7 +76,7 @@ def find_subnet value
     def find_record record
       subnets.each do |s|
         s.records.each do |v|
-          return v if record.is_a?(String) and (v.ip == record or v.mac == record)
+          return v.options if record.is_a?(String) and (v.ip == record or v.mac == record or v.options[:hostname] == record)
```
domcleal, line 79: Why the change to the return value?

domcleal:
Would you mind adding a test under test/server_test.rb for the options[:hostname] search please?

domcleal:
@ohadlevy any opinion over this search method versus your own in GH-79?  This is more flexible, but the API is less strongly typed, due to the use of find_record.

lib/proxy/dhcp/server.rb
korekhov, line 79: This is to allow a user to use a hostname as a search string in addition to MAC and IP.
I did a search of uses of find_record method and only found it was used to find an IP and MAC in the same server.rb so thought this change would not break things.
Was I wrong and this breaks something?

## Thanks!

Konstantin Orekhov

domcleal, line 79: Agreed on the hostname search, that's good - it's the beginning of the line I'm looking at where it used to return `v` (the record) and now returns `v.options`.

korekhov, line 79: Yes, this is wrong. I'll update that. Should be just "v".

## 

Konstantin Orekhov

PR #88, opened 2013-05-23
fixes #2547 - fix ruby-abi on non-SCL EL6

domcleal:
@mbacovsky could you review please?

Scratch build here: https://koji.katello.org/koji/taskinfo?taskID=36520

Tested with the latest non-SCL rkerberos build and it installs on EL6 using sinatra from EPEL.


mbacovsky:
ACK

PR #90, opened 2013-06-03
fixes #2520 - defines default settings for dns and puppet

dmitri-d:


lib/proxy/settings.rb
```diff
@@ -2,18 +2,40 @@
...
-raw_config = File.read(Pathname.new(__FILE__).join("..", "..","..","config","settings.yml"))
+
+require 'proxy/dns/default_dns_settings'
+require 'proxy/puppet/default_puppet_settings'
 
 class Settings < OpenStruct
-  def method_missing args
+  DEFAULTS = {}
+
+  include ::DefaultDnsSettings, ::DefaultPuppetSettings
```
domcleal: Using mixins for default settings feels really complex to me, can't we just put them in the hash above?

lib/proxy/settings.rb
```diff
@@ -2,18 +2,40 @@
...
+      settings.delete :puppetca if settings.has_key? :puppetca
+      settings.delete :puppet   if settings.has_key? :puppet
+      settings[:x86_64] = File.exist?('c:\windows\sysnative\cmd.exe')
+    end
+    load(settings)
+  end
+
+  def self.load(ahash)
+    s = Settings.new(ahash)
+    s.initialize_default_values!
```
domcleal: Could we just merge ahash with our defaults hash instead of needing initialize_default_values?

lib/smart_proxy.rb
```diff
@@ -27,12 +27,6 @@ class SmartProxy < Sinatra::Base
     set :public, APP_ROOT + '/public'
   end
 
-  require "tftp_api"     if SETTINGS.tftp
```
domcleal: These removals break the proxy, it no longer responds to anything apart from the features API.

test/test_helper.rb
```diff
@@ -1,15 +1,11 @@
...
-require "proxy/settings"
-
-# Override settings to enable subsystems we intend to test
-[ "tftp", "puppet", "puppetca", "bmc" ].each { |s| SETTINGS.send("#{s}=", true) }
 logdir = File.join(File.dirname(__FILE__), '..', 'logs')
 FileUtils.mkdir_p(logdir) unless File.exists?(logdir)
-SETTINGS.log_file = File.join(logdir, 'test.log')
-SETTINGS.puppet_conf = File.join(File.dirname(__FILE__), 'fixtures', 'puppet.conf')
 
+#require File.expand_path '../testing_proxy_settings.rb', __FILE__
```
domcleal: leftover?

test/test_helper.rb
```diff
@@ -1,15 +1,11 @@
 require "test/unit"
 $: << File.join(File.dirname(__FILE__), '..', 'lib')
 
-require "proxy/settings"
-
-# Override settings to enable subsystems we intend to test
-[ "tftp", "puppet", "puppetca", "bmc" ].each { |s| SETTINGS.send("#{s}=", true) }
```
domcleal, line 7: nice!

test/test_helper.rb
```diff
@@ -1,15 +1,11 @@
...
 logdir = File.join(File.dirname(__FILE__), '..', 'logs')
 FileUtils.mkdir_p(logdir) unless File.exists?(logdir)
-SETTINGS.log_file = File.join(logdir, 'test.log')
-SETTINGS.puppet_conf = File.join(File.dirname(__FILE__), 'fixtures', 'puppet.conf')
 
+#require File.expand_path '../testing_proxy_settings.rb', __FILE__
+require 'testing_proxy_settings.rb'
```
domcleal: We normally leave out .rb

lib/proxy/settings.rb
dmitri-d: I wanted to keep defaults together with individual modules, hence the implementation.

lib/proxy/settings.rb
dmitri-d: Since we only want to set default values for keys not provided in the ahash, we'd have to use Hash#update(.., &blk), as it allows for control over hash collisions. I think the end result would be similar...

lib/smart_proxy.rb
dmitri-d: Ugh. I'll track this down (and will add tests).

test/test_helper.rb
dmitri-d: indeed.

test/test_helper.rb
dmitri-d: Normally we do the same. Which this should be. But for some reason isn't.

lib/proxy/settings.rb
domcleal: I was thinking the other way around, start with DEFAULTS and merge in ahash, so it overwrites any defaults that are set.  The bit in initialize_default_values I particularly don't like is calling ostruct internals, when a simple hash merge would suffice.

dmitri-d: That is better.

lib/proxy/settings.rb
dmitri-d: I could build DEFAULTS here by merging individual modules' default values hashes. Would that be more clear?

domcleal: Yeah, that'd be much clearer (rather than included magic), but still looks over-engineered to me, having two extra files just for two settings.  @ohadlevy?

dmitri-d: Wouldn't there be more default settings?

domcleal: Perhaps..

dmitri-d:
@domcleal: how about now?

lib/proxy/dns/default_dns_settings.rb
```diff
@@ -0,0 +1,3 @@
+module DefaultDnsSettings
```
domcleal: Do you think these should be namespaced under Proxy::DNS?

dmitri-d: That would make sense.

ohadlevy:
what was the reasoning to add settings modules?

dmitri-d:
@ohadlevy I wanted to keep defaults together with their modules.

lib/proxy/settings.rb
ohadlevy: :+1: for keeping it simple... how about merge each defaults hash to a SETTINGS constant?

dmitri-d: @ohadlevy: you are commenting on old changes ;).

dmitri-d:
@domcleal what do you think?

domcleal:
:+1: will wait for CI

domcleal:
Merged as c4a245d, thanks @witlessbird.

PR #93, opened 2013-06-11
fixes #2652 - systemd service now working

lzap:
SSIA

There were two options - to use "-" flag to ignore missing file, or to
remove the statement. Since there is no big point of keeping that file for
modern Fedora distros, I am removing it out.


domcleal:
:+1: could you add a new %changelog entry and bump the release please?

domcleal:
@lzap ping?

lzap:
Eeeeh, I need more sleep! Done.

domcleal:
Thanks @lzap, merged as 9923a7d.

PR #94, opened 2013-06-12
2654 2659 env fixes

GregSutcliffe:
Also a very minor fix around using to_sym in the dynamic environments to be consistent with the static ones.

### DIFF
--- a/lib/proxy/puppet/environment.rb
+++ b/lib/proxy/puppet/environment.rb
@@ -40,6 +40,12 @@ module Proxy::Puppet
           env[:production] = conf[:main][:modulepath] || conf[:master][:modulepath] || '/etc/puppet/modules'
           logger.warn "No environments found - falling back to defaults (production - #{env[:production]})"
         end
+        if env.size == 1 and env.keys.first == :master and !env.values.first.include?('$environment')
+          # If we only have an entry in [master] it should really be called production
+          logger.warn "Re-writing single 'master' environment as 'production'"
+          env[:production] = env[:master]
+          env.delete :master
+        end
 
         new_env = env.clone
         # are we using dynamic puppet environments?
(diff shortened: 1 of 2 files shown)


domcleal:
Tests?

GregSutcliffe:
Ah yes, some of those failures are expected as we're renaming master to production when there's only one environment (which then permits you to have 'master' as well as production if you're using git or something). I'll update.

GregSutcliffe:
@domcleal test updated. the failing PuppetApi test isn't me, it was already failing; i reverted to develop and it fails there too. I'm looking into it, but it's not part of this PR.

lib/proxy/puppet/environment.rb
```diff
@@ -40,6 +40,12 @@ def puppet_environments
           env[:production] = conf[:main][:modulepath] || conf[:master][:modulepath] || '/etc/puppet/modules'
           logger.warn "No environments found - falling back to defaults (production - #{env[:production]})"
         end
+        if env.size == 1 and env.keys.first == :master and !env.values.first.match /\$environment/
```
domcleal: Nitpicky thing here, but `.include?` might be slightly nicer than match.

domcleal:
@GregSutcliffe looks good, thanks.  I won't hold the PuppetApi test against you :)

Any chance of adding an $environment test too for the second commit?

GregSutcliffe:
If I can avoid a headache long enough to structure one, sure ;) - oddly, I'm not sure how the previous tests were passing, since it was clearly broken...

lib/proxy/puppet/environment.rb
GregSutcliffe: True, I'll fix that.

GregSutcliffe:
@domcleal tests updated - we already had "test_single_modulepath_in_single_env_with_dynamic_path" but it was only checking that the output of the _environment_ names was correct. I've extended it to check it actually finds some classes, which correctly fails without the fix.

GregSutcliffe:
[test]

domcleal:
Ah yes, that was unfortunate.  Thanks for the tests, merged as 1c60be63 and 92a8bdeff.

PR #95, opened 2013-06-14
fixes #2662 - Proxy::Util.escape_for_shell is being used in PuppetRun.run and MCollective.run methods

dmitri-d:


domcleal:
What about using arrays with `IO#popen`?  This way it bypasses the shell entirely, which is safer.

I suspect escaping _and_ using arrays in 216728d09495fe5a2940760df702910642853bca was belt and braces, and popen3 isn't necessary unless we're using stdin.

dmitri-d:
@domcleal: IO#popen using an array won't work - there are issues around reassembling a command line when more than a single command is used. But this is a bit cleaner still.

domcleal:
@witlessbird I don't follow - it's a single command, but with multiple arguments.  Supply this as an array prevents it going via the shell and being subject to shell expansion.

dmitri-d:
@domcleal how about now?

domcleal:
@witlessbird excellent, and with tests!  Merged as 4a8411d, thanks!

PR #96, opened 2013-06-18
fixes #2673 - none/wildcard http 'accept' headers are being handled corr...

dmitri-d:
...ectly now

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,7 +1,7 @@
 source 'http://rubygems.org'
 
 gem 'json'
-gem 'sinatra'
+gem 'sinatra', '< 1.4.3'
 
 Dir["#{File.dirname(__FILE__)}/bundler.d/*.rb"].each do |bundle|
  # puts "adding custom gem file #{bundle}"
--- a/lib/sinatra-patch.rb
+++ b/lib/sinatra-patch.rb
@@ -2,6 +2,7 @@ require "sinatra/base"
 require "openssl"
 require "webrick/https"
 require "daemon" unless PLATFORM =~/mingw/
+
 module Sinatra
   class Base
     # Run the Sinatra app as a self-hosted server using
@@ -53,9 +54,9 @@ module Sinatra
   module MonkeyRequest
     # We need request.accept? method also in pre-1.3.0 versions. This is simplified
     # version of the method that only accept one parameter (mime type string).
-    def accept? type
+    def accept?(type)
       accept.include? type
     end
   end
-  Request.send :include, MonkeyRequest if not Request.method_defined? :accept?
+  Request.send :include, MonkeyRequest unless Request.method_defined?(:accept?)
 end
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -4,8 +4,9 @@ $: << File.join(File.dirname(__FILE__), '..', 'lib')
 logdir = File.join(File.dirname(__FILE__), '..', 'logs')
 FileUtils.mkdir_p(logdir) unless File.exists?(logdir)
 
-require 'testing_proxy_settings'
+require './testing_proxy_settings'
 require "proxy"
 require "mocha/setup"
 require "rack/test"
 require 'sinatra'
+require 'sinatra-patch'


lib/helpers.rb
```diff
@@ -14,7 +14,7 @@ def log_halt code=nil, exception=nil
         message += e.message
         code     = code || 400
       end
-      content_type :json if request.accept?("application/json")
+      content_type :json if request.preferred_type("application/json").include?("application/json")
```
ohadlevy: does this require a specific version of rack ?

domcleal: So we have an unfortunate (somewhat crippling) constraint with sinatra in that EPEL6 ships sinatra 1.0, which doesn't contain the preferred_type method :)
(we have a monkey patch in lib/sinatra-patch.rb that adds `accept?` to Request for pre-1.3.0)

domcleal:
Is the test not supplying an accept header, or is it a wildcard?  Maybe we can fix it that way for now, since Foreman itself does supply application/json and this code path works fine I believe (@GregSutcliffe tested it).

lib/helpers.rb
dmitri-d: @ohadlevy: Sinatra::Request#preferred_type was introduced in Sinatra 1.2. I guess that should be the minimal version then?

dmitri-d:
The smaller issue is that we don't set 'accept' headers in tests. The larger issue is that in Sinatra 1.4.3 handling of wildcards in 'accept' header (as in 'application/*') and of missing 'accept' headers is broken. Simply fixing the tests would work, but my concern is that since this is concerns public API, it should be fixed in the proxy itself.

lib/helpers.rb
domcleal: I'd rather not move from EPEL6's version unless we can help it, this seems a trivial reason.

dmitri-d:
@domcleal, @ohadlevy: thoughts?

test/puppet_api_test.rb
```diff
@@ -1,4 +1,4 @@
-require 'test_helper'
+require File.join(File.dirname(__FILE__), 'test_helper')
```
domcleal: why these changes?  Shouldn't be needed.

dmitri-d: I needed those to run the tests under RubyMine, if it's not too big of a deal, I'd prefer to keep them...

dmitri-d:
I can add ">= 1.4.4" requirement for sinatra to the Gemfile. I think it's worth keeping the check in the code though. Alternatively, I could remove the version check and add a test instead

domcleal:
We can't bump the version to 1.4.4 min as we need to support 1.0 for EL6.  I'd prefer to see a <= 1.4.3 dependency added if anything, until the fixed version is released and then remove the dependency (as people will just get 1.4.4 then).

I'm not really convinced that adding a monkey patch to core is worth it to fix one broken version, but not entirely opposed to it if you want to do this?

test/puppet_api_test.rb
domcleal: We'd need to add them to every file for it to be properly useful though, can't you configure test in the search path somehow (sorry, not familiar with it)?

dmitri-d:
Apologies, I actually meant "< 1.4.4" requirement for now, and the opposite after the fix is in. And yeah, I wasn't sure about the monkey patch myself. I'll remove it.

test/puppet_api_test.rb
dmitri-d: I spent a minute on it, but didn't find anything. I'll look more.

domcleal: Check with @isratrade, he's a rubymine user too I think.

dmitri-d:
Added '< 1.4.3' constraint to sinatra gem and removed the rest of the changes. (btw - the reason for not needing explicit paths in tests is https://github.com/theforeman/smart-proxy/blob/develop/Rakefile#L21). Adding test/ dir to load path for tests under rubymine fixes the issue.

domcleal:
Glad you sorted it, thanks.  Merged as c164bdd!

Could you file a redmine bug to remove the version dependency once fixed upstream?  ta

PR #97, opened 2013-06-20
Feature #2482 - Global search by MAC, IP or hostname

korekhov:


ohadlevy:
[test]

ohadlevy:
Hmm... are you aware that the second commit deletes the changes by the first one?

korekhov:
Yes, the only difference is the title, really.

Thanks!

korekhov:
Oh, I see what you mean. I thought you were talking about my previous PR.

Let me fix this really quick.

lib/dhcp_api.rb
```diff
@@ -50,6 +50,17 @@ def load_subnet
...
+    begin
+      content_type :json
+      records = @server.find_record(params[:record])
+      log_halt 404, "Record #{params[:record]} not found" unless records
+      records.options.to_json
+    rescue => e
+      log_halt 400, e
+    end
+  end
+
```
dmitri-d, line 63: This is not RESTful. A way to implement this could be to introduce scoping to the index operation, ie: GET /dhcp would return all records, while GET /dhcp?ip=192.168.42.1,192.168.42.3 would only return records with the ip addresses included in the list. filtering by mac addresses could be done similarly: GET /dhcp?mac=aa:aa:aa:aa:aa,bb:bb:bb:bb:bb.

Also, in this particular approach we'd be doing search based on the explicit parameters passed in by the user, instead of trying to guess the meaning of supplied data, which is always better.

dmitri-d, line 63: @isratrade: thoughts on adding of search semantics to the index operation above?

korekhov, line 63: I agree, this is not ideal. I was just trying to minimize the changes and used existing @server.find_record.

Also, if you look at the list of issues I posted here http://projects.theforeman.org/issues/2687, it becomes very clear (at least to me), that there are a lot changes have to happen in DHCP smart-proxy and better (and faster) API for record search is just one of them.

Going back to API itself. In ISC DHCP there are 3 main objects - lease, host and group. With "host" object being the only one that is fully RW, I thought the following URL structure should be truly RESTful:

curl -X <GET | POST | DELETE> http://<proxy>:<port>/dhcp/host -d 'ip=x.x.x.x' -d 'mac=xx.xx.xx.xx.xx.xx', etc.

so basically exactly the same thing you mentioned, but with object "reservation". For example, if somebody wanted to implement the lease manipulation API (like expiration or something), they would just use "PUT /dhcp/lease".

Not sure if other DHCP implementations (MS, etc) have other objects available, but having this structure will allow for easier API modifications, IMHO.

dmitri-d:
I think API for lease, host, and group can be dealt with separately. Looks like the most pressing need is hosts :).
Not sure how useful it is to combine multiple filters, unless we want fancy filters (subnet matching?). If so, I think the fancy stuff can be added later, but it would be safer to introduce multi-filter semantic right away.

lib/dhcp_api.rb
isratrade, line 63: @witlessbird, passing individual parameters in the URL like you did as fine.  If there are multiple parameters that you are searching on, maybe consider using the scoped_search gem like we do for hosts.  In this case, you assign a string to the :string parameter in the URL like this:
http://0.0.0.0:3000/hosts?search=+environment+%3D+production+and++hostgroup+%3D++Base

ohadlevy, line 63: @isratrade sadly this is using sinatra and has no db backend, so scoped search would not work here.

isratrade, line 63: @ohadlevy, ok, in that case, passing the parameters in the URL seems fine with me like you did
GET /dhcp?ip=192.168.42.1,192.168.42.3, @witlessbird

lib/dhcp_api.rb
```diff
@@ -50,6 +50,17 @@ def load_subnet
     end
   end
 
+  get "/dhcp/find/:record" do
```
ohadlevy, line 53: @korekhov I agree with @witlessbird that its better to update the GET /dhcp action to accept search parameters.

you could still use the find_record method, but we would need to introduce a better helper to parse the parameters, e.g:
- if network parameter is passed, we search within that network
- if mac parameter and network parameter are passed, we search only within that network
- if mac is passed, but no network, we search on all subnets etc etc.

ohadlevy, line 53: hmm.. actually, get  on /DHCP returns a list of subnets, not records, let me think about this again :)

ohadlevy, line 53: I would consider adding the following API routes:
- /dhcp/records  - return all records (not ideal, but maybe useful)
- /dhcp/records?ip=<ip address or range>
- /dhcp/records?mac=<mac or prefix, e.g. all systems that start with 00:xxxxxxxxx>
- /dhcp/records?name=...
- /dhcp/records?network= (equiv to /dhcp/<network>)
- /dhcp/records?option_67=<..> 
- /dhcp/records?mac=00:11:22:*&option_67=<my_value>
  etc?

we could also consider adding a more reach search language that includes OR and NOT, but maybe thats outside the scope of this feature.

any comments @witlessbird @domcleal @isratrade and @korekhov of course?

korekhov, line 53: Sounds reasonable to me. Thanks for looking into this.

Hint, hint - when you do the testing of this feature, can you please use the large dataset I uploaded some time ago to help with http://projects.theforeman.org/issues/2687? That should give you an idea of the performance issues that preventing us from using DHCP proxy at the moment and it would much appreciated if you look into those too. This particular feature is really useless for anyone running a more or less large DHCP set up as the lookup/search API query just cannot even complete in reasonable time.

Thanks again!

dmitri-d:
Closing this PR as it's been inactive for awhile now.

PR #100, opened 2013-06-26
Refs #969 - Proxy-side changes for serving templates from the proxy

GregSutcliffe:
Compliment to https://github.com/theforeman/foreman/pull/751

Probably also refs #1969 as `:template_url` is the client-facing proxy name, and so could be used for `puppetserver` too. `:foreman_url` could be used for implementing an auto-update-on-feature-change, if we want to.

There's an open issue about serving http on an https proxy, which will probably require a 2-threaded proxy listening on two ports. For now, this functionality is for http proxies only.

@domcleal, do we have somewhere to start putting documentation for 1.3 features?

### DIFF
--- a/lib/proxy/template.rb
+++ b/lib/proxy/template.rb
@@ -1,0 +1,27 @@
+require "proxy/util"
+require 'uri'
+
+module Proxy::Template
+
+  class Handler
+    extend Proxy::Log
+
+    # Gets a template from Foreman
+    def self.get_template kind, token
+      # Parse the Foreman URI for a connection
+      uri              = URI.parse("#{SETTINGS.foreman_url}/unattended/#{kind}?token=#{token}")
+      req              = Net::HTTP::Get.new(uri.request_uri)
+      http             = Net::HTTP.new(uri.host, uri.port)
+      http.use_ssl     = uri.scheme == 'https'
+      # TODO: handle CA properly; for now we'll accept self-signed certs
+      http.verify_mode = OpenSSL::SSL::VERIFY_NONE if http.use_ssl?
+      res              = http.start { |http| http.request(req) }
+
+      # You get a 201 from the 'built' URL
+      raise "Error retrieving #{kind} for #{token} from #{uri.host}: #{res.class}" unless ["200", "201"].include?(res.code)
+      logger.info "Template: request for #{kind} using #{token} at #{uri.host}"
+      res.body
+    end
+  end
+
+end
--- a/config/settings.yml.example
+++ b/config/settings.yml.example
@@ -25,6 +25,17 @@
 :tftproot: /var/lib/tftpboot
 # Defines the TFTP Servername to use, overrides the name in the subnet declaration
 #:tftp_servername: tftp.domain.com
+#
+# Set this to true if the Proxy should handle template requests on behalf of Foreman
+:templates: false
+# Where is Foreman?
+# no need to specify http(s) here as templates are always retrieved on http
+# Examples:
+# https://myforemanserver.mydomain.com
+# http://1.2.3.4:3000
+#:foreman_url: http://127.0.0.1:3000
+# Where should provisioning hosts request their template from?
+#:template_url: http://192.168.122.1:8443
 
 # Enable DNS management
 :dns: false
--- a/lib/proxy.rb
+++ b/lib/proxy.rb
@@ -1,5 +1,5 @@
 module Proxy
-  MODULES = %w{dns dhcp tftp puppetca puppet bmc}
+  MODULES = %w{dns dhcp tftp puppetca puppet bmc templates}
   VERSION = "1.2-develop"
 
   require "checks"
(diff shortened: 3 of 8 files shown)


domcleal:
On 26/06/13 18:48, Greg Sutcliffe wrote:

> @domcleal https://github.com/domcleal, do we have somewhere to start
> putting documentation for 1.3 features?

No, but I'd hold off copying the 1.2 manual until we've released it.

The trouble too with copying it is every PR then has to be duplicated to
it and we missed some at first last time.  Solutions to this problem are
welcome.

## 

Dominic Cleal
Red Hat Engineering

GregSutcliffe:
I'm fine with holding off on docs for another week, I expect the review process for this one might take a bit anyway ;)

lib/proxy/template.rb
```diff
@@ -0,0 +1,24 @@
...
+
+module Proxy::Template
+
+  class Handler
+    extend Proxy::Log
+
+    # Gets a template from Foreman
+    def self.get_template kind, token
+      # Parse the Foreman URI for a connection
+      uri         = URI.parse("http://#{SETTINGS.foreman_url}/unattended/#{kind}?token=#{token}")
```
ohadlevy: should us use https for proxy -> foreman?

ekohl: I'd expect SETTINGS.foreman_url to be a URL, including the protocol.

ohadlevy: yeah, lets not get do the same issue we had with foreman.ip=URL in the
discovery ;)

GregSutcliffe: Agreed, that was stupid of me. I'll fix it.

GregSutcliffe: Worth noting though, that we accept 'http' for all template requests, because anaconda/deb-install cannot handle SSL for the kickstart/preseed retrieval. Since we accept http for such requests, using http here means I don't have to handle certificates :)

GregSutcliffe:
@ekohl @ohadlevy updated

lib/proxy/template.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    # Gets a template from Foreman
+    def self.get_template kind, token
+      # Parse the Foreman URI for a connection
+      uri              = URI.parse("#{SETTINGS.foreman_url}/unattended/#{kind}?token=#{token}")
+      req              = Net::HTTP::Get.new(uri.request_uri)
+      http             = Net::HTTP.new(uri.host, uri.port)
+      http.use_ssl     = uri.scheme == 'https'
+      # TODO: handle CA properly; for now we'll accept self-signed certs
+      http.verify_mode = OpenSSL::SSL::VERIFY_NONE if http.use_ssl?
```
ekohl, line 17: I have no clue about ruby, but couldn't you just use false instead of looking like you do something with verification?

dLobatog:
This works well on my setup, although I'm running this from your branch directly instead of rebasing first. Is there anything blocking this from being merged, other than verifying SSL certificates support?

> I'm fine with holding off on docs for another week, I expect the review process for this one might take a bit anyway ;)

:laughing: 

cross-reference to this: https://github.com/theforeman/foreman/pull/751

domcleal:
One issue was that we may get loops between Foreman and the proxy when making /unattended/provision type calls, because it might make Puppet CA requests back to the proxy.  Unsure if it was the direction we wanted to take at the time.

GregSutcliffe:
@eLobato i'll be revisiting this once the pluggable proxy PR is merged as it then supports running both http and https simultaneously.

dLobatog:
@GregSutcliffe you were waiting for this? https://github.com/theforeman/smart-proxy/commit/038fa6cf7079880b16d998346d05112acefec81e

lib/proxy/template.rb
dLobatog, line 17: It could be 0 (Fixnum), but I think `OpenSSL::SSL::VERIFY_NONE` is more readable.

dLobatog:
@GregSutcliffe [PR in core](https://github.com/theforeman/foreman/pull/751) will likely need a revisit and a rebase too so we can test it.

GregSutcliffe:
@eLobato indeed it will, but this is not a priority for me at the moment. I'll try to get back to it sometime around 2017, I think there's space in my calendar then :P

sodabrew:
We used nginx on an alternate port to proxy the provision script and wrote a replacement snippet for the foreman_url variable. The private side is plain HTTP, but nginx proxies to the Foreman HTTPS endpoint.

My 2p, 2¢, €0.02, ₪0.02, 0.02Kč (picking cities for your team meetings must be really fun!) - Building provision script proxy into the system would be awesome. The PR at foreman#751 is marked "Reached an impasse" - is there anything I can do to help?

dLobatog:
@sodabrew You can pick up any of the two PR (or both), rebase them against the latest develop branch and finish them off if you're interested :smile:

GregSutcliffe:
For the record, the impasse was waiting for concurrent http and https support in the proxy - now that we have that, this can probably move forward again. If you have time to look at it @sodabrew that would be awesome :)

dLobatog:
Closing in favor of https://github.com/theforeman/smart-proxy/pull/224

PR #102, opened 2013-07-15
Fixes #2790 - Optionally set DNS TTL value from settings.yml config file

pdaukintis:
This allows overriding default DNS TTL value (86400) per smart-proxy instance. It is possible via setting newly introduced dns_ttl variable in settings.yml configuration file.

### DIFF
--- a/config/settings.yml.example
+++ b/config/settings.yml.example
@@ -35,6 +35,8 @@
 #:dns_key: /etc/rndc.key
 # use this setting if you are managing a dns server which is not localhost though this proxy
 #:dns_server: dns.domain.com
+# use this setting if you want to override default TTL setting (86400)
+#:dns_ttl: 60
 # use dns_tsig_* for GSS-TSIG updates using Kerberos.  Required for Windows MS DNS with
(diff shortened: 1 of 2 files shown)


ares:
Why do you add SETTINGS inside of `Proxy::DNS::Record#initialize` ? You can use existing options mechanism I think. See https://github.com/theforeman/smart-proxy/blob/develop/lib/dns_api.rb.

pdaukintis:
Yes, indeed. I've reverted changes in lib/proxy/dns.rb and made them in lib/dns_api.rb.

ares:
Seems good from my side. Squashing into one commit would be nice but I believe we can handle that :-) Thank you!

lzap:
Nice patch, thank you for contribution. Would you mind squashing all changes into one commit and re-pushing changes into the same branch?

pdaukintis:
Pushed changes, hope this would be ok.

domcleal:
Thanks for your contribution @pdaukintis, merged as 1242cab for Foreman 1.3.

PR #103, opened 2013-07-29
Update mcollective.rb

Slydder:
add missing spaces and missing ID to cmd which calls mco. Fails without these changes.

### DIFF
--- a/lib/proxy/puppet/mcollective.rb
+++ b/lib/proxy/puppet/mcollective.rb
@@ -4,15 +4,20 @@ module Proxy::Puppet
   class MCollective < Runner
     def run
       cmd = []
-      cmd.push(which("sudo", "/usr/bin"))
-      cmd.push(which("mco", ["/usr/bin", "/opt/puppet/bin"]))
+      cmd.push(which("sudo", "/usr/bin") + " ")
+      cmd.push(which("mco", ["/usr/bin", "/opt/puppet/bin"]) + " ")
 
       if cmd.include?(false)
         logger.warn "sudo or the mco binary is missing."
         return false
       end
 
-      shell_command(cmd + ["puppet", "runonce", "-I"] + shell_escaped_nodes)
+      nodenames = []
+      shell_escaped_nodes.each { |x|
+        y = x.split('.')
+        nodenames.push(y.first)
+      }
+      shell_command(cmd + ["puppet ", "runonce ", "-I ", nodenames, " "] + shell_escaped_nodes)
     end
   end
 end


lib/proxy/puppet/mcollective.rb
```diff
@@ -4,15 +4,20 @@ module Proxy::Puppet
   class MCollective < Runner
     def run
       cmd = []
-      cmd.push(which("sudo", "/usr/bin"))
-      cmd.push(which("mco", ["/usr/bin", "/opt/puppet/bin"]))
+      cmd.push(which("sudo", "/usr/bin") + " ")
```
domcleal, line 7: This is going to break on Ruby 1.9 since we call popen with an array.. and on Ruby 1.8 I can't see that it does anything except change the log output.  In lib/proxy/puppet.rb, we use `.join(" ")` so they are all space-separated.

lib/proxy/puppet/mcollective.rb
```diff
@@ -4,15 +4,20 @@ module Proxy::Puppet
...
       if cmd.include?(false)
         logger.warn "sudo or the mco binary is missing."
         return false
       end
 
-      shell_command(cmd + ["puppet", "runonce", "-I"] + shell_escaped_nodes)
+      nodenames = []
+      shell_escaped_nodes.each { |x|
+        y = x.split('.')
+        nodenames.push(y.first)
```
domcleal, line 18: If you run an `mco ping`, are your hostnames fully qualified or short names?  Here they're fully qualified, matching Foreman (which only deals with FQDNs), so I'm suspecting yours aren't showing up that way.

Slydder:
on my 1.8 Debian box I can push the "run puppet" button and it actually works (no 500 error) and no error in the logs. and seeing as how I split on the "." in the FQDN I would have to say that I do get the FQDN. Added to that shell_command executes the command as it is in mcollective.rb. I'm not seeing how it's getting hit with a "join" anywhere in my latest stable version of puppet.rb in smart-proxy. are you maybe referring to the dev version?

domcleal:
Yes, you're getting in the FQDN in, but it seems like your MCollective infra isn't using FQDNs.  I don't think we can accept that, as Foreman is designed to work in FQDN and multi-domain environments.

The join code is here for 1.8 in the stable proxy code:
https://github.com/theforeman/smart-proxy/blob/1.2-stable/lib/proxy/puppet.rb#L39

Slydder:
you are correct. I was in the wrong directory. I did find the join now. also updated the server.cfg for mcollective on the remote node and it works. too bad the change just couldn't be made on the server. it would ease the deployment of mcollective a bit. but hey, it works.

domcleal:
Glad it's sorted, I'll close the PR.  You might want to manage your MCollective agents via the Puppet modules, they work quite well in my experience.

Thanks @Slydder!

PR #104, opened 2013-07-31
adds Proxy::Logger to Proxy::DNS::Nsupdate class

kitchen:
without this, 'logger.debug' and such does nothing at all

### DIFF
--- a/lib/proxy/dns/nsupdate.rb
+++ b/lib/proxy/dns/nsupdate.rb
@@ -4,6 +4,7 @@ require 'resolv'
 module Proxy::DNS
   class Nsupdate < Record
 
+    include Proxy::Logger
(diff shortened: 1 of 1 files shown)


lib/proxy/dns/nsupdate.rb
```diff
@@ -4,6 +4,7 @@
 module Proxy::DNS
   class Nsupdate < Record
 
+    include Proxy::Logger
```
domcleal: Just `Log`, not `Logger`?

kitchen:
hrm. That is correct. I did this through the github web editor, my bad :(

domcleal:
Thanks @kitchen, merged as 0dd2f70.

PR #105, opened 2013-08-01
Always empty PXEClient on native_ms.rb server

tomas-edwardsson:
Often DHCP scopes have PXEClient set by default. This interferes with
the way Foreman does pxe/tftp. With this patch the default is to set
PXEClient to empty string when using native_ms.

### DIFF
--- a/lib/proxy/dhcp.rb
+++ b/lib/proxy/dhcp.rb
@@ -5,6 +5,7 @@ module Proxy::DHCP
   require "proxy/dhcp/server"
   Standard = {
               :hostname              => {:code => 12, :kind => "String"    }, # The host's name
+              :PXEClient             => {:code => 60, :kind => "String"    }, # Needs to be empty for foreman to function
               :nextServer            => {:code => 66, :kind => "String"    }, # From where we download the pxeboot image via TFTP
               :filename              => {:code => 67, :kind => "String"    }  # The pxeboot image
              }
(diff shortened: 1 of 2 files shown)


tomas-edwardsson:
When I say PXEClient by default, I mean at least in windows type setups I've seen.

skottler:
[test]

skottler:
Thanks for the PR @tomas-edwardsson! Can you please create a ticket explaining the issue at http://projects.theforeman.org and then add the ticket number to the commit?

tomas-edwardsson:
I've created Issue 2870 in redmine as instructed by @skottler 

http://projects.theforeman.org/issues/2870

skottler:
@tomas-edwardsson looks good and tests just fine for me. Can you add the ticket number into the commit?

tomas-edwardsson:
@skottler not sure what you mean? Rewriting the commit comment to include "#2870" ?

skottler:
@tomas-edwardsson yep, that way redmine will automatically close the issue which makes writing changelogs much easier.

tomas-edwardsson:
@skottler commit updated in a82b393, hope this is sufficient.

skottler:
@tomas-edwardsson almost, but not quite. Anyhow, no worries, we can take care of that during the merge process.

@ohadlevy @domcleal are you comfortable with clearing the PXEClient for scopes that have it set already?

domcleal:
@skottler yes, I think it makes sense.

The only couple of things I'd suggest to improve this would be:
1. set PXEClient as a default value in the options hash so if a user wants to override it via the API, they can:
   
   def addRecord options={}
     record = super({'PXEClient' => '""'}.merge(options))
2. instead of specifying a string containing double quotes, the loop to set the option could quote every value in the command or replace empty strings with double quotes

tomas-edwardsson:
I have done as @domcleal suggested and merged PXEClient into the options and quoted every value set. It seems to work fine. @skottler anything else I should mind?

ohadlevy:
@domcleal looks good to you?, aside rebasing, :+1: (without testing)

ohadlevy:
merged as 296617a for 1.3 thanks @tomas-edwardsson!

PR #106, opened 2013-08-02
added puppetrun over ssh support

cyberkov:
Hello!

I wrote a puppetrun over ssh extension. 
Unfortuantely `( Process.detach(c.pid); return 0 ) unless wait` doesn't seem to be very ruby-like...
Maybe someone can help me out with that one :-)

Cheers
Hannes

### DIFF
--- a/lib/proxy/puppet/puppetssh.rb
+++ b/lib/proxy/puppet/puppetssh.rb
@@ -1,0 +1,23 @@
+require 'proxy/puppet'
+
+module Proxy::Puppet
+  class PuppetSsh < Runner
+    def run
+      cmd = []
+      cmd.push(which("sudo", "/usr/bin")) if SETTINGS.puppetssh_sudo
+      cmd.push(which("ssh", "/usr/bin"))
+      cmd.push("-l #{SETTINGS.puppetssh_user}") if SETTINGS.puppetssh_user
+      cmd.push("-i #{SETTINGS.puppetssh_keyfile}") if SETTINGS.puppetssh_keyfile
+
+      if cmd.include?(false)
+        logger.warn "sudo or the ssh binary is missing."
+        return false
+      end
+
+      ssh_command = escape_for_shell(SETTINGS.puppetssh_command || '/usr/bin/puppet agent --onetime --no-usecacheonfailure')
+      nodes.each do |node|
+        shell_command(cmd + [ escape_for_shell(node), ssh_command ], false )
+      end
+    end
+  end
+end
--- a/test/puppetssh_test.rb
+++ b/test/puppetssh_test.rb
@@ -1,0 +1,83 @@
+require 'test_helper'
+require 'proxy/puppet/puppetssh'
+
+class PuppetSshTest < Test::Unit::TestCase
+  def setup
+    @puppetssh = Proxy::Puppet::PuppetSsh.new(:nodes => ["host1", "host2"])
+  end
+
+  def test_command_line_with_default_command
+    @puppetssh.stubs(:which).with("sudo", anything).returns("/usr/bin/sudo")
+    @puppetssh.stubs(:which).with("ssh", anything).returns("/usr/bin/ssh")
+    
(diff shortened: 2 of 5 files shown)


lzap:
You know, maybe it's fine. I am not great rubyist :-)

Thanks for the patch!

## 

Later,

 Lukas "lzap" Zapletal
 irc: lzap #theforeman

ohadlevy:
[test]

ares:
How about

``` ruby
return 0 if Process.detach(c.pid) && !wait
```

or just simply (since I'm not sure about return value of detach)

``` ruby
unless wait
  Process.detach(c.pid)
  return 0
end
```

cyberkov:
Thank you :) 
@ares I think go with the "unless wait" option just to be on the safe side :)

I also managed to put in some tests for the module but i keep getting ssh errors at the end. I'm not sure how I could suppress those.

```
Finished tests in 0.954535s, 143.5254 tests/s, 239.9075 assertions/s.

137 tests, 229 assertions, 0 failures, 0 errors, 0 skips
ssh: Could not resolve hostname host1: Name or service not known
ssh: Could not resolve hostname host2: Name or service not known
```

lzap:
I can confirm, investigating.

test/puppetssh_test.rb
```diff
@@ -0,0 +1,61 @@
...
+
+    assert !@puppetssh.run
+  end
+
+  def test_missing_sudo_and_not_needed
+    SETTINGS.stubs(:puppetssh_sudo).returns(false)
+    @puppetssh.stubs(:which).with("sudo", anything).returns(false)
+    @puppetssh.stubs(:which).with("ssh", anything).returns("/usr/bin/ssh")
+
+    assert @puppetssh.run
```
lzap: Okay you are missing this

```
@puppetssh.stubs(:shell_command).returns(true)
```

before the assert in this test. Thats it.

Looks good to me.

skottler:
Thanks for your PR @cyberkov. So I think there are a few things this needs before it can get merged:
1. The ability to specify a key. `ssh-agent` likely won't always have the right key available, which will break in the current design.
2. Input needs to be properly escaped.

I'm not particularly keen on having the proxy be able to SSH to all the hosts it manages. What's the use case for this beyond just simplicity?

lib/proxy/puppet/puppetssh.rb
```diff
@@ -0,0 +1,21 @@
...
+      cmd.push(which("ssh", "/usr/bin"))
+
+      if cmd.include?(false)
+        logger.warn "sudo or the ssh binary is missing."
+        return false
+      end
+
+      ssh_command = (SETTINGS.puppetssh_command || '/usr/bin/puppet agent --onetime --no-usecacheonfailure')
+      nodes.each do |node|
+        shell_command(cmd + [ node, ssh_command ], false )
```
skottler: `Proxy::Util.escape_for_shell` should get used here. It's a 1.8 + 1.9 compatible version of the escape methods that are available in 1.9.

ohadlevy:
note that we also have something similar in code (thanks to @lzap who wrote it a long time ago) 
https://github.com/theforeman/smart-proxy/blob/develop/lib/proxy/util.rb#L15

maybe we can reuse?

lzap:
Hmmm this had to be the other lzap, the smarter one :-)

Indeed, we might want to refactor shell_command to use this utility spawner, which improves logging a bit.

This might be out of scope of this effort tho. Would be a bonus.

lzap:
@skottler I think both your points are resolved now! LGTM [test]

on my system:

```
140 tests, 238 assertions, 0 failures, 0 errors, 0 skips
```

ohadlevy:
I think it make sense to use the CommandTask class here, to have a consistent usage across the app, @cyberkov if you can't refactor it, let me know and I'll fix it myself.

ohadlevy:
thanks! merged with minor changes as 13ed471 for 1.3!

PR #110, opened 2013-10-03
fixed feature #3150 - adds ability to specify the user to invoke sudo as...

gpetras:
Here's a minor enhancement to allow a user to specify a user to invoke sudo as.

### DIFF
--- a/config/settings.yml.example
+++ b/config/settings.yml.example
@@ -82,6 +82,9 @@
 #:puppetssh_user: root
 #:puppetssh_keyfile: /etc/foreman-proxy/id_rsa
 
+# Which user to invoke sudo as to run mco commands
+#:puppet_mco_user: root
+
 # enable BMC management  (Bare metal power and bios controls)
 # Available providers:
 # - freeipmi / ipmitool - requires the appropriate package installed, and the rubyipmi gem
--- a/config/settings.yml.example
+++ b/config/settings.yml.example
@@ -82,8 +82,8 @@
 #:puppetssh_user: root
 #:puppetssh_keyfile: /etc/foreman-proxy/id_rsa
 
-# Which user to invoke sudo as to run mco commands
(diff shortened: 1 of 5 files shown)


domcleal:
Sorry for the delay @gpetras, I somehow missed this in my inbox.

This looks good, just a couple of points:
1. Could you add a test for this behaviour?  There's an existing test/mcollective_test.rb file (run them all with `rake`) that checks the command is correct (test_run_command).  You could copy this test and add this to it to test the behaviour when configured this way:
   
   ```
   SETTINGS.stubs(:puppet_mco_user).returns("example")
   ```
2. I think it'd be useful to have this in puppetrun.rb too for running "puppet kick".  Do you think the setting could be more generic?

Thanks for the PR!

ohadlevy:
[test]

gpetras:
Dominic -

Thanks for the reply, and sorry for my delay in responding too.

I've not done a lot of ruby programming yet, so this simple example is a
good exercise for me.

Also, as far as making this more generic, I suppose I could simply call the
setting puppet_user. Sound good?

We aren't using "puppet kick" here, do you have a place where this can be
tested after I write this part of the enhancement?

Thanks!

Greg

domcleal:
puppet_user sounds good, let's do that.

I can test the puppet kick, as it's still set up on my environment, no problem.  We could duplicate the test I suggested for mcollective_test.rb in puppetrun_test.rb which will help to verify the result.

gpetras:
Great. I just submitted a couple more patches to the feature branch.

Let me know what you think.

Thanks,

Greg

domcleal:
Looks good, let's [test].

domcleal:
Works perfectly.  Merged as 81a3445 for Foreman 1.4.0.  Thanks for your contribution @gpetras!

PR #112, opened 2013-10-26
First support of chefproxy

rvrignaud:
First support of chefproxy on smart-proxy.

Related to : theforeman/chef-handler-foreman#4 and  theforeman/foreman#983

@ohadlevy, @ares, @GregSutcliffe : thoughts, remarks ?


Gemfile
```diff
@@ -1,6 +1,7 @@
 source 'http://rubygems.org'
 
-gem 'json'
+#Chef have a strong requirement on json version
+gem 'json', '= 1.7.7'
```
GregSutcliffe: That's going to be interesting from a packaging standpoint, I'm not sure what versions are available for which OSes. For example, Wheezy has only got 1.7.3

lib/chefproxy_api.rb
```diff
@@ -0,0 +1,52 @@
+require 'proxy/chefproxy'
+require 'chef'
+require 'digest/sha2'
+require 'base64'
+
+
```
GregSutcliffe, line 3: nitpick: remove one of these two blank lines

Gemfile
rvrignaud: In fact this is not accurate but I didn't know how to manage it.
Official chef requirements on JSON are:
json <= 1.7.7, >= 1.4.4

domcleal: If their gem specifies it, no need to here.

lib/chefproxy_api.rb
```diff
@@ -0,0 +1,51 @@
...
+      auth = true
+      auth = verify_signature_request(client_name,signature,facts) if SETTINGS.authenticate_nodes
+      if auth
+        Proxy::Chefproxy::Facts.post_facts(facts)
+      else
+       log_halt 401, "Failed to authenticate node"
+      end
+  end
+
+  post "/reports" do
```
ares, line 19: this seems as a big duplication of /facts, could you extract it to some method which would accept a block with code to execute when authenticated?

result would look similar to this

``` ruby
post "/facts/" do
  authenticated do |request|
    Proxy::Chefproxy::Facts.post_facts(request.env["rack.input"].read)
  end
end
```

lib/proxy/chefproxy.rb
```diff
@@ -0,0 +1,46 @@
+#require "proxy/util"
+require 'net/http'
+require 'net/https'
+require 'uri'
+
+def foreman_request(path, body)
```
ares: this should be part of Proxy::Chefproxy::Facts I think, so we don't mess global space, also I suggest not using class methods. How about this model

``` ruby
module Proxy::Chefproxy
  class ForemanRequest
    def foreman_request(path, body)
      #heavylifting here
    end
  end

  class Reports
    def post_report(report)
      ForemanRequest.new('/api/reports', report)
    end
  end

  class Facts
    def post_facts(facts)
      ForemanRequest.new('/api/hosts/facts', facts)
    end
  end
end
```

and then in chefproxy.api something like

``` ruby
post "/facts/" do
  Proxy::Chefproxy::Facts.new.post_facts(request.env["rack.input"].read)
end
```

ares:
Last comment, I'd prefer `Chefproxy` class name called `ChefProxy` :-)

lib/chefproxy_api.rb
```diff
@@ -0,0 +1,48 @@
+require 'proxy/chefproxy'
+require 'chef'
+require 'digest/sha2'
+require 'base64'
+
+class SmartProxy
+
+  def verify_signature_request(client_name,signature,body)
```
ares: Well, I'm not a Sinatra experienced user but these two methods would be shared across all APIs because they all open SmartProxy class. Not sure how to handle this best, maybe @domcleal or @GregSutcliffe will know. I think the problem is to share SmartProxy class for all *_api.rb. So if there's no better way I'd prefer to move those methods to separate class. Especially `authenticated` name can result in conflict.

ares:
And one another thing, could you please cover this by tests? You can see some examples here https://github.com/rvrignaud/smart-proxy/tree/develop/test

lib/chefproxy_api.rb
```diff
@@ -0,0 +1,48 @@
+require 'proxy/chefproxy'
+require 'chef'
```
ohadlevy: only if chef is enabled? otherwise it would break

lib/chefproxy_api.rb
ohadlevy: i would assume the fastest way would be to have a dedicated class for authentication? and maybe a sub class for a chef auth?

lib/chefproxy_api.rb
ohadlevy: scratch that, i didnt look below

lib/proxy.rb
```diff
@@ -1,5 +1,6 @@
 module Proxy
   MODULES = %w{dns dhcp tftp puppetca puppet bmc}
+  MODULES = %w{dns dhcp tftp puppetca puppet bmc chefproxy}
```
guilhem, line 2: I may be wrong, but it looks like you forget to remove previous MODULES affectation.

lib/proxy/authentication.rb
```diff
@@ -0,0 +1,47 @@
...
+          signature   = request.env['HTTP_X_FOREMAN_SIGNATURE']
+          raise "401 Failed to authenticate node. Missing some headers" if client_name.nil? or signature.nil?
+          auth = verify_signature_request(client_name,signature,content)
+        end
+
+        if auth
+          raise "400 Body is empty" if content.nil?
+          block.call(content)
+        else
+          raise "401 Failed to authenticate node"
```
ares: I think you should create your own exception class for this (one for 401 second for 400) and raise those. You can find some info here http://www.sinatrarb.com/intro.html#Not%20Found

(applies to all raises in this method)

lib/proxy/authentication.rb
```diff
@@ -0,0 +1,47 @@
...
+        #We need to retrieve node public key
+        #to verify signature
+        chefurl = SETTINGS.chefserver_url
+        client_name = SETTINGS.smartproxy_clientname
+        key = SETTINGS.smartproxy_privatekey
+
+        rest = ::Chef::REST.new(chefurl,client_name,key)
+        begin
+          public_key = OpenSSL::PKey::RSA.new(rest.get_rest("/clients/#{client_name}").public_key)
+        rescue
```
ares: Here you should at least log the trace of exception. Also you should specify the exception class you are rescuing, these are hard to debug later since it swallow every exception. And I'm not sure what kind of error do you expect here.

lib/chefproxy_api.rb
```diff
@@ -0,0 +1,24 @@
+require 'proxy/chefproxy'
+require 'proxy/authentication'
+
+class SmartProxy
+  post "/facts" do
+    begin
+      Proxy::Authentication::Chef.new.authenticated(request) do |content|
+          Proxy::ChefProxy::Facts.new.post_facts(content)
+      end
+    rescue => e
```
ares: I think that this error handling should happen outside of post blocks. Again, I'm not sinatra user but I think that `error MyException do...end` should be used. Also you should specify the exception class.

lib/chefproxy_api.rb
```diff
@@ -0,0 +1,24 @@
...
+      log_halt(e.message[/^\d+/].to_i , e.message[/\D+/])
+    end
+  end
+
+  post "/reports" do
+    begin
+      Proxy::Authentication::Chef.new.authenticated(request) do |content|
+        Proxy::ChefProxy::Reports.new.post_report(content)
+      end
+    rescue => e
```
ares: same as above, it should share the error handler

test/authentication_chef_test.rb
```diff
@@ -0,0 +1,71 @@
...
+
+  def test_auth_enable_without_headers_should_raise_an_error
+    SETTINGS.stubs(:authenticate_nodes).returns(true)
+    s = StringIO.new('Hello')
+    request = Sinatra::Request.new(env={'rack.input' => s})
+    begin
+      result = @chefauth.authenticated(request) do |content|
+        true
+      end
+    rescue => e
```
ares: you should rescue only specific class of expcetions (ot test their type, but the first seems better to me)

lib/proxy/authentication.rb
rvrignaud: "rest.get_rest" is make an HTTP call to chef-server URI and that call can fail

test/chefproxy_test.rb
```diff
@@ -0,0 +1,28 @@
...
+class ChefProxyTest < Test::Unit::TestCase
+  def setup
+    @foreman_url = 'https://foreman.example.com'
+    SETTINGS.stubs(:foreman_url).returns(@foreman_url)
+    SETTINGS.stubs(:authenticate_nodes).returns(false)
+  end
+
+  def test_post_facts
+    facts = {'fact' => "sample"}
+    stub_request(:post,@foreman_url+'//api/hosts/facts')
```
ares: nitpick - double slash

lib/proxy/authentication.rb
ares: so maybe something like this should be rescued

``` ruby
rescue Timeout::Error, Errno::EINVAL, Errno::ECONNRESET, EOFError, Net::HTTPBadResponse, 
    Net::HTTPHeaderSyntaxError, Net::ProtocolError, Errno::ECONNREFUSED, 
    OpenSSL::SSL::SSLError => e
```

test/authentication_chef_test.rb
```diff
@@ -0,0 +1,71 @@
...
+
+  def test_signing_and_checking_with_2_different_keys_sould_not_work
+    # We mock chef-server response but with a wrong publick key to make signature check fail
+    response = '{"public_key":"'+@testnode2_pubkey+'","name":"testnode1","admin":false,"validator":false,"json_class":"Chef::ApiClient","chef_type":"client"}'
+    stub_request(:get, 'https://chef.example.com//clients/testnode1').to_return(:body => response.to_s, :headers => {'content-type' => 'application/json'} )
+    
+    assert_equal(false,@chefauth.verify_signature_request('testnode1',@signature,@mybody), "Signing and checking with different keys should not pass")
+  end
+
+  def test_auth_disabled_should_always_sucess
```
ares: double c in "success" I think

ares:
:+1: from my side (if tests remain green after rebase)

lib/proxy/chefproxy.rb
```diff
@@ -0,0 +1,47 @@
...
+      req.add_field('Accept', 'application/json,version=2')
+      req.content_type = 'application/json'
+      req.body         = body
+      response         = http.request(req)
+    end
+  end
+
+  class Facts
+    def post_facts(facts)
+      ForemanRequest.new.send_request('/api/hosts/facts',facts)
```
domcleal: Perhaps Facts and Reports should inherit (i.e. `class Facts < ForemanRequest`) instead of instantiating ForemanRequest?

lib/proxy/chefproxy.rb
```diff
@@ -0,0 +1,47 @@
...
+          http.verify_mode = OpenSSL::SSL::VERIFY_PEER
+        end
+
+        if SETTINGS.foreman_ssl_cert && !SETTINGS.foreman_ssl_cert.to_s.empty? && SETTINGS.foreman_ssl_key && !SETTINGS.foreman_ssl_key.to_s.empty?
+          http.cert = OpenSSL::X509::Certificate.new(File.read(SETTINGS.foreman_ssl_cert))
+          http.key  = OpenSSL::PKey::RSA.new(File.read(SETTINGS.foreman_ssl_key), nil)
+        end
+      end
+
+      req = Net::HTTP::Post.new("#{uri.path}#{path}")
```
domcleal: `URI.join(uri, path).path`?

config/settings.yml.example
```diff
@@ -85,6 +86,15 @@
 # Which user to invoke sudo as to run puppet commands
 #:puppet_user: root
 
+# enable Chef management
+:chefproxy: false
+# :authenticate_nodes: true
```
domcleal: Can we prefix all of the Chef related settings with "chef"?

lib/chefproxy_api.rb
```diff
@@ -0,0 +1,24 @@
+require 'proxy/chefproxy'
+require 'proxy/authentication'
+
+class SmartProxy
+  error Proxy::Error::BadRequest do
+    log_halt(400, "Bad request : " + env['sinatra.error'].message )
```
domcleal, line 6: We should try and pass the exception itself into log_halt as it will log the backtrace at debug level if the Exception is passed in.  Maybe we should extend log_halt to take a message and exception separately?

lib/proxy/authentication.rb
```diff
@@ -0,0 +1,52 @@
...
+
+          raise Proxy::Error::Unauthorized, "Failed to authenticate node. Missing some headers" if client_name.nil? or signature.nil?
+          auth = verify_signature_request(client_name,signature,content)
+        end
+
+        if auth
+          raise Proxy::Error::BadRequest, "Body is empty" if content.nil?
+          block.call(content)
+        else
+          raise Proxy::Error::Unauthorized, "Failed to authenticate node"
```
domcleal: Should we log a bit more here, like the client_name?

lib/proxy/authentication.rb
```diff
@@ -0,0 +1,52 @@
...
+    require 'proxy/error'
+
+    def verify_signature_request(client_name,signature,body)
+        #We need to retrieve node public key
+        #to verify signature
+        chefurl = SETTINGS.chefserver_url
+        client_name = SETTINGS.smartproxy_clientname
+        key = SETTINGS.smartproxy_privatekey
+
+        rest = ::Chef::REST.new(chefurl,client_name,key)
```
domcleal: Some debug logging could be really valuable here later I think.

domcleal:
Code-wise this looks pretty good, subject to minor comments above.  Could you check over the patch too for indentation please?  Some of the classes and methods are 4-space indented rather than 2-space.

Maybe something for @ares or a subsequent ticket, but we might need to update the RPM spec file and create a -chef subpackage?

Can you file a redmine ticket too please?

ares:
Subpackage sounds good but I'm not sure how to handle modules constant

``` ruby
module Proxy
  MODULES = %w{dns dhcp tftp puppetca puppet bmc chefproxy}
```

so maybe we should first push on http://projects.theforeman.org/issues/3351 and split when it's done?

domcleal:
I was thinking to still ship the code, but have a package that pulls in chef and ships bundler.d/chef.rb, but I just realised we don't run the proxy via bundler, so never mind!

domcleal:
Merged as 4c2e7d8 for Foreman 1.4.0.  Thanks @rvrignaud, exciting stuff!

PR #114, opened 2013-11-21
fixes #3712 - change develop versioning scheme to indicate next version

domcleal:


lzap:
ACK

But how are we going to put the datetags in it? By that I mean things like 20131122...

domcleal:
Jenkins does that at the moment, just by inserting it before %{?dist} in the spec file when building.

lzap:
OK

domcleal:
Was merged as 133bac9fbafbea13121bf07262050f20377bd91e.

PR #118, opened 2013-12-08
Ship config.ru in RPMs

ekohl:
Shipping config.ru makes running under passenger trivial.

Note that I haven't tested it and just copied what foreman.spec does.


domcleal:
Thanks @ekohl, merged as 42bab91 for Foreman 1.4.

PR #119, opened 2013-12-12
fixes #3874 - extend paths for FreeBSD support

mmoll:
FreeBSD is putting packaged software into /usr/local, so paths have to be extended for it to work.

### DIFF
--- a/lib/proxy/puppet/puppet_ssh.rb
+++ b/lib/proxy/puppet/puppet_ssh.rb
@@ -4,8 +4,8 @@ module Proxy::Puppet
   class PuppetSSH < Runner
     def run
       cmd = []
-      cmd.push(which('sudo', '/usr/bin')) if SETTINGS.puppetssh_sudo
-      cmd.push(which('ssh', '/usr/bin'))
+      cmd.push(which('sudo')) if SETTINGS.puppetssh_sudo
+      cmd.push(which('ssh'))
       cmd.push("-l #{SETTINGS.puppetssh_user}") if SETTINGS.puppetssh_user
       if (file = SETTINGS.puppetssh_keyfile)
         if File.exists?(file)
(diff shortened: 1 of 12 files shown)


lib/proxy/puppet/puppet_ssh.rb
```diff
@@ -4,7 +4,7 @@ module Proxy::Puppet
   class PuppetSSH < Runner
     def run
       cmd = []
-      cmd.push(which('sudo', '/usr/bin')) if SETTINGS.puppetssh_sudo
+      cmd.push(which('sudo', ['/usr/bin', '/usr/local/bin'])) if SETTINGS.puppetssh_sudo
       cmd.push(which('ssh', '/usr/bin'))
```
mmoll: ssh is part of the "base system", so it's in /usr

lib/proxy/puppet/puppet_ssh.rb
```diff
@@ -20,10 +20,10 @@ def run
         return false
       end
 
-      ssh_command = escape_for_shell(SETTINGS.puppetssh_command || '/usr/bin/puppet agent --onetime --no-usecacheonfailure')
+      ssh_command = escape_for_shell(SETTINGS.puppetssh_command || 'puppet agent --onetime --no-usecacheonfailure')
```
mmoll, line 23: not sure about that one, but PE is in /opt/puppet/bin, so it might be better that way.

lib/proxy/puppet/puppet_ssh.rb
lzap: Hey,

thanks for the patch. Would you mind changing this to setting it as the
default 2nd param of the which method and wiping out all the other
parameters for all of its calls?

I was planning to do this, that'd do it!

## 

Later,

 Lukas "lzap" Zapletal
 irc: lzap #theforeman

lib/proxy/puppet/puppet_ssh.rb
lzap, line 23: From my experiences, incorrect path in $PATH usually do not hurt
(nothing is printed on Linux, no error or warning). We could add that as
the default parameter (in the array).

Anyone?

## 

Later,

 Lukas "lzap" Zapletal
 irc: lzap #theforeman

lib/proxy/puppet/puppet_ssh.rb
mmoll: all done

lzap:
ACK thanks. @domcleal I think this can be merged.

domcleal:
Thanks @mmoll!  Merged as 7e3ca5c2 for Foreman 1.4.

PR #121, opened 2013-12-18
added a method of running a custom tool, to trigger puppet.

nelg:
I've found this useful, as a way to run a script to trigger puppet. 
Please consider adding this.     Happy to discuss this if I've got something wrong.

### DIFF
--- a/config/settings.yml.example
+++ b/config/settings.yml.example
@@ -78,10 +78,11 @@
 :puppet_provider: puppetrun
 
 # customrun command details
-# The example below run for each host: /usr/local/bin/myscript.py -ay -f -s hostname
-:customrun:
-  :cmd: /usr/local/bin/myscript.py
-  :args: -ay -f -s
+# Set :customrun_cmd to the full path of the script you want to run, instead of /bin/false
+:customrun_cmd: /bin/false
+# Set :customrun_args to any args you want to pass to your custom script. The hostname of the
+# system to run against will be appended after the custom commands.
+:customrun_args: -ay -f -s
 
 :puppet_conf: /etc/puppet/puppet.conf
 # whether to use sudo before the ssh command
--- a/config/settings.yml.example
+++ b/config/settings.yml.example
@@ -74,7 +74,15 @@
 #   mcollective (uses mco puppet)
 #   puppetssh   (run puppet over ssh)
 #   salt        (uses salt puppet.run)
+#   customrun   (calls a custom command with args)
 :puppet_provider: puppetrun
+
+# customrun command details
+# The example below run for each host: /usr/local/bin/myscript.py -ay -f -s hostname
+:customrun:
+  :cmd: /usr/local/bin/myscript.py
+  :args: -ay -f -s
+
 :puppet_conf: /etc/puppet/puppet.conf
 # whether to use sudo before the ssh command
 :puppetssh_sudo: false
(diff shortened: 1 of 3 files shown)


lib/proxy/puppet/customrun.rb
```diff
@@ -0,0 +1,26 @@
...
+        logger.error "failed to load configuration #{e}"
+        return false
+      end
+
+      cmd = settings[:customrun][:cmd]
+
+      unless File.exists?( cmd )
+        logger.warn "#{cmd} not found."
+        return false
+      end
```
lzap, line 12: Hello, thanks for the contribution [test]

We have a method that is able to tell you if command exists:

```
unless which(cmd)
```

You do not need to provide any optional parameters as there is already pending pull request (https://github.com/theforeman/smart-proxy/pull/119) that removes all the optional params and adds sane default

lib/proxy/puppet/customrun.rb
```diff
@@ -0,0 +1,26 @@
...
+
+      shell_command( [ cmd, settings[:customrun][:args], shell_escaped_nodes ] )
```
lzap: Again, we have method called `escape_for_shell` which should be used to encode the cmd and all parameters. Please use it here, grep the codebase for examples.

Why do you separate the command and args? I think this could be in one single string, what you think?

config/settings.yml.example
```diff
@@ -74,7 +74,15 @@
...
 #   salt        (uses salt puppet.run)
+#   customrun   (calls a custom command with args)
 :puppet_provider: puppetrun
+
+# customrun command details
+# The example below run for each host: /usr/local/bin/myscript.py -ay -f -s hostname
+:customrun:
+  :cmd: /usr/local/bin/myscript.py
+  :args: -ay -f -s
+
```
lzap, line 86: As I've said, this could be only one option (command and parameters).

Please make the default something that does not hurt on most systems:

```
:cmd: /usr/bin/true
```

lzap:
Other than that, I think this is good to go. Thank you!

lib/proxy/puppet/customrun.rb
nelg, line 12: I don't think I can use which, because I want the user to be able to specify the full location of their custom script, which may well not be in the default path.

lib/proxy/puppet/customrun.rb
nelg: I separated command and args, so that we can test if the command exists and fail nicely if it does not. 
Also, escape_for_shell works OK for my cmd, but not for args.  If I use it for args, it puts all the args into $1 within the script that's being called.   IE, it passes "-a -ya" rather than -a -ya.

nelg:
Anything else you would like me to do?  I've commented on the other stuff.  Glen

ohadlevy:
[test]

ohadlevy:
@nelg mind having a look at why the tests are failing? thanks!

nelg:
@ohadlevy I am trying to look at the tests, as requested, but I am not sure where to look.  I've looked at http://ci.theforeman.org/job/test_proxy_develop_pull_request/189/, however, these failures don't seem to be anything to do with the code I've worked on, as far as I can tell, so I guess I am in need of a little help, or at least pointers as to what I should look at.

ohadlevy:
@nelg thanks! I've double checked, and those issues are not cause because of your patch, rather then a breaking change in puppet 3.4.

ohadlevy:
do you know if we have a redmine ticket for this patch?

ohadlevy:
:+1: otherwise

ohadlevy:
added a ticket and merged as e478c9e for 1.4, thanks @nelg!

PR #122, opened 2013-12-19
FEATURE: added puppetdb support to smart proxy

logicminds:
This adds support to use the same query language for puppetdb but against the smart proxy which proxies the request to the puppetdb.  This isn't meant to be merged just yet, but I wanted to put it up for code review.  I still need to fix some comments and write up some docs first before merging this.

Also, I added the puppet gem to test.rb because its required but wasn't included.  I guess most people already have puppet gem installed so it must have slipped through the cracks.


domcleal:
What's this for, what's the goal?

I don't see any value in just proxying PuppetDB's API - if you need proxy capabilities, use a reverse proxy?  This doesn't add any high level constructs, nor does it abstract away the implementation.

If it's something specific to your infrastructure, I'd prefer to wait for plugin support in the proxy and let you maintain a plugin locally.

logicminds:
A common problem with big infrastructures is network operations is completely manual and ridden with mistakes so creating ACLs and VIPs to those services is a huge pain in the ass.  By having smart proxy relay those requests you no longer need an ACL for each individual service.  Instead you just need one ACL/VIP that combines all the services via a single endpoint.  So when you need a new service you just add the code to smart proxy.  When a new datacenter is created only one ACL/VIP is required to consume all the data/services that the proxy responds to.  While this is most important to my own infrastructure I think other people could make use of it as well.  I could imagine future support for puppetdb with foreman if foreman ever decides to not use custom reports, but that is a big design decision.

Additionally, there is no need to abstract anything away from the puppetdb api to make it simpler.  One just needs to follow the same puppetdb api documentation.

The smart proxy should be completely decoupled from The Foreman so that enhancements can be made to smart proxy without concern of how downstream projects like the foreman would use it.  We are using smart proxy without the use of foreman.

I would be happy to rewrite as a plugin when the plugin api is available, but for now, if anybody wants puppetdb support its available to merge now.

ohadlevy:
@logicminds I wonder if the better long term support is something like ovs-bridge, where you could create your own networks across multiple networks for your management devices

it sounds like solving a single problem until the next one shows up?

what do you mean by foreman as a downstream project? i thought its all upstream ;)

domcleal:
Ok, I think a plugin once we've got the support available is going to be best for this, as I disagree on a couple of points raised.

a) an abstraction to some extent is useful, e.g. nsupdate support is abstracted into a DNS API, ISC dhcpd is abstracted into a DHCP API, all of which add some value in hiding implementation details.  Just proxying requests, while it solves a network ACL issue, looks different to me from the proxy generally does.

b) the proxy should be decoupled from Foreman, but that doesn't mean we should make changes without concern for downstream projects - especially since frankly, Foreman is the primary user.

Mostly on the basis of (a), I'm going to close this for now.

PR #123, opened 2013-12-22
fixes #2878 - Import Classes fails when using parser = future

rickard-von-essen:
This enables _smart-proxy_ to read puppet classes which contains new _lambda_ constructs when `parser = future` option is used.

For more information see:
http://docs.puppetlabs.com/puppet/3/reference/experiments_future.html
http://projects.theforeman.org/issues/2878


ohadlevy:
[Test]

lib/proxy/puppet/environment.rb
```diff
@@ -122,8 +122,11 @@ def to_s
...
-      paths.map {|path| PuppetClass.scan_directory path}.flatten
+      conf = Puppet.settings.instance_variable_get(:@values)
+      eparser = conf[:master] && conf[:master][:parser] == 'future'
+
+      paths.map {|path| PuppetClass.scan_directory path, eparser}.flatten
     end
 
   end
-end
+end
```
ekohl: Nitpick, but I'd keep the line ending here.

rickard-von-essen: Fixed!

ohadlevy:
[test]

domcleal:
This is awesome, thanks for the contribution @rickard-von-essen.  I fixed a couple of weird issues under Ruby 1.8.7 and Puppet 3.2+, but all merged into af58b9f for Foreman 1.4.

PR #124, opened 2014-01-07
fixes #3965 Change API endpoint for chef_proxy to be the same as Foreman's

rvrignaud:
This is needed to be transparent from client side.

@ares : you're ok ?


ares:
Would it make sense to keep both variants instead of moving?

domcleal:
ping @rvrignaud

ares:
discussed on IRC, I'm ok with moving

domcleal:
Merged as a7e5964 for Foreman 1.4, thanks @rvrignaud.

PR #125, opened 2014-01-07
fixes #3966 Puppet is not listed as bundler dependency

rvrignaud:


lzap:
Hello Romain,

this is on purpose, we do not have it since our users mainly use puppet version from operating system.

If you need to live on the edge and use puppet from rubygems, you can put your bit into:

```
bundler.d/Gemfile.local.rb
```

This file is in our gitignore. This is ideal for development purposes.

logicminds:
Puppet is actually needed to run all the tests successfully.  This should be put in the development bundle at a minimum. 

Corey

domcleal:
@lzap I think it's fine, since we don't use bundler when installing from packages.  We did the same recently for facter in Foreman core.

Users running rvm etc will find this useful, but if you've installed from source and are using bundler under the system Ruby and have then installed puppet or facter packages, you can exclude this group.

lzap:
Then this file sould be excluded in the SPEC file because this is what we do:

```
cp -p -r app bundler.d config config.ru extras lib locale Rakefile script %{buildroot}%{_datadir}/%{name}
```

RPM build will likely fail if we merge this.

domcleal:
Don't worry, we don't package Gemfile or bundler.d today.

domcleal:
In RPMs that is, I think it might be the same for debs.. @GregSutcliffe?

GregSutcliffe:
Correct: Depends: ruby1.9.1|ruby-interpreter, rake, ruby-sinatra, ruby-json, ruby-rkerberos (>= 0.1.1)

in all honesty, even ruby-rkerberos shouldn;t be there, it's an optional component (as is puppet - it;s not required on a dns-only proxy, for example)

domcleal:
Merged as 6da3d66, thanks @rvrignaud!

PR #126, opened 2014-01-09
Update settings.yml.example

rvrignaud:
sorry about that


lzap:
Thanks

@domcleal can I merge just like that, or do we need an issue for that?

domcleal:
@lzap I just referenced the ticket that put it there.

Merged as 860a288, thanks @rvrignaud.

PR #127, opened 2014-01-10
fixes #3991 - dnscmd provider for smart-proxy (Windows)

mmatuska:
nsupdate against a Microsoft DNS server does not create static records. This causes DNS entries to be scavenged if aging is turned on. A dnscmd provider running on a Windows host (similar to MS DHCP) can use the dnscmd command to create/remove static entries.

### DIFF
--- a/lib/proxy/dns/dnscmd.rb
+++ b/lib/proxy/dns/dnscmd.rb
@@ -1,0 +1,133 @@
+require "proxy/dns"
+require 'resolv'
+require 'checks'
+require 'rubygems' if USE_GEMS
+require 'win32/open3'
+
+module Proxy::DNS
+  class Dnscmd < Record
+
+    include Proxy::Log
+    include Proxy::Util
+    attr_reader :resolver
+
+    def initialize options = {}
+      super(options)
+    end
+
+    # create({ :fqdn => "node01.lab", :value => "192.168.100.2"}
+    # create({ :fqdn => "node01.lab", :value => "3.100.168.192.in-addr.arpa",
+    #          :type => "PTR"}
+    def create
+      @resolver = Resolv::DNS.new(:nameserver => @server)
+      case @type
+      when "A"
+        if ip = dns_find(@fqdn)
+          raise(Proxy::DNS::Collision, "#{@fqdn} is already used by #{ip}") unless ip == @value
+        else
+          zone = @fqdn.sub(/[^.]+./,'')
+          msg = "Added DNS entry #{@fqdn} => #{@value}"
+          cmd = "/RecordAdd #{zone} #{@fqdn}. A #{@value}"
+          execute(cmd, msg)
+        end
+      when "PTR"
+        if name = dns_find(@value)
+          raise(Proxy::DNS::Collision, "#{@value} is already used by #{name}") unless name == @fqdn
+        else
+          # TODO: determine reverse zone names
+          #
+          # zone = nil
+          # msg = "Added reverse DNS entry #{@value} => #{@fqdn}"
+          # data = @value.split(".")
+          # if data[3].eql?("10")
+          #   zone = "10.in-addr.arpa"
+          # end
+          # return nil if zone.nil?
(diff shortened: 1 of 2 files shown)


lib/proxy/dns/dnscmd.rb
```diff
@@ -0,0 +1,133 @@
...
+          zone = @fqdn.sub(/[^.]+./,'')
+          msg = "Added DNS entry #{@fqdn} => #{@value}"
+          cmd = "/RecordAdd #{zone} #{@fqdn}. A #{@value}"
+          execute(cmd, msg)
+        end
+      when "PTR"
+        if name = dns_find(@value)
+          raise(Proxy::DNS::Collision, "#{@value} is already used by #{name}") unless name == @fqdn
+        else
+          # TODO: determine reverse zone names
```
lzap, line 37: Thanks, this might be a new proxy issue instead of TODO line.

lib/proxy/dns/dnscmd.rb
```diff
@@ -0,0 +1,133 @@
...
+        return true
+      end
+    end
+
+    private
+
+    def execute cmd, msg=nil, error_only=false
+      tsecs = 5
+      response = nil
+      interpreter = SETTINGS.x86_64 ? 'c:\windows\sysnative\cmd.exe' : 'c:\windows\system32\cmd.exe'
```
lzap, line 84: Can't we detect this automatically?

Also, I am missing settings.yaml.example changes. Please add sane examples then.

lib/proxy/dns/dnscmd.rb
```diff
@@ -0,0 +1,133 @@
...
+      command  = interpreter + ' /c c:\Windows\System32\dnscmd.exe ' + "#{@server} #{cmd}"
```
lzap, line 85: No escaping for shell?

lib/proxy/dns/dnscmd.rb
mmatuska, line 84: I have reused code from:
https://github.com/theforeman/smart-proxy/blob/develop/lib/proxy/dhcp/server/native_ms.rb
If you want things different (better) I suggest it gets unified and/or updated in both locations

lib/proxy/dns/dnscmd.rb
mmatuska, line 85: Same as my previous comment, code reused from here:
https://github.com/theforeman/smart-proxy/blob/develop/lib/proxy/dhcp/server/native_ms.rb

lib/proxy/dns/dnscmd.rb
mmatuska, line 37: I think this will be expensive - you have to scan all reverse zones and search for the (closest) match. This will require a lot of command calls and parsing. You simply do not now how the zones is named - the zones may be created using any subnet rules.

A possibility would be to include the zones in the configuration file a similar way you can restrict subnets in native_ms dhcp.

lib/proxy/dns/dnscmd.rb
lzap, line 85: If I read the code correctly, the proxy server is vulnerable to Windows shell code injection at the moment if native_ms implementation is configured. The same for the new implementation. I am fine with all the other minor remarks I had, but I vote for fixing this using the same patter (escape_for_shell methods) as we do for UNIX here and in native_ms. One small method to be used in both implementation. That would be nice.

lib/proxy/dns/dnscmd.rb
domcleal, line 37: Filed as http://projects.theforeman.org/issues/4025

lib/proxy/dns/dnscmd.rb
domcleal, line 85: Filed as http://projects.theforeman.org/issues/4026

domcleal:
Thanks @mmatuska, merged as 168ee4e for Foreman 1.4.0.  I've filed the issues discussed above as new redmines so we can work on them at a later date - this change is relatively independent from any other code, so I'd prefer to get it into the upcoming release all things considered.

PR #129, opened 2014-01-15
Combine paths

ares:
If we run foreman under some prefix we have to join both paths.


domcleal:
Thanks @ares, merged as c1c0dee.

PR #131, opened 2014-01-28
refs #3874 - add which() to wget call

mmoll:
another missing which() I found while testing...


domcleal:
[test]

domcleal:
Thanks @mmoll, merged as 835cf3a for Foreman 1.4.0.

PR #132, opened 2014-01-29
Fixes #4215 smart-proxy needlessly creates tmp/pids directory

mmatuska:
tmp/pids does not need to be created if not on mingw


domcleal:
Thanks @mmatuska, merged as a25e7b9 for Foreman 1.4.1.

PR #133, opened 2014-02-12
Initial ABRT proxy support

mmilata:
This change makes it possible for the proxy to pretend it is ABRT
(Automatic Bug Reporting Tool[1]) server. When it receives a problem
report, it forwards notification of the error to Foreman, and optionally
also forwards the original report to an actual ABRT server.

Non strictly ABRT-related changes:
- Move module Proxy::ChefProxy to Proxy::Request
- Add :foreman_ssl_ options to settings example

[1] https://github.com/abrt/abrt/wiki/ABRT-Project

For some notes regarding usage, see https://groups.google.com/forum/#!topic/foreman-dev/Yk2GrW9ne20

I am quite new to Ruby so please let me know if anything should be done in a different way:)

### DIFF
--- a/config/settings.yml.example
+++ b/config/settings.yml.example
@@ -12,7 +12,13 @@
 #:trusted_hosts:
 #- foreman.prod.domain
 #- foreman.dev.domain
+# Following is needed for communication with foreman initiated by the proxy.
 #:foreman_url: http://127.0.0.1:3000
+# SSL settings for client authentication against foreman. In most cases, these
+# will be the same as the ssl_* settings above.
+#:foreman_ssl_ca: ssl/certs/ca.pem
+#:foreman_ssl_cert: ssl/certs/fqdn.pem
+#:foreman_ssl_key: ssl/private_keys/fqdn.pem
 
 # enable the daemon to run in the background
 :daemon: true
@@ -112,6 +118,16 @@
 # in order to retrive all nodes public keys
 # :chef_smartproxy_clientname: 'host.example.net'
 # :chef_smartproxy_privatekey: '/etc/chef/client.pem'
+
+# enable ABRT uReport forwarding
+:abrtproxy: false
+# FAF server instance the reports will be forwarded to (optional)
+#:abrt_server_url: https://retrace.fedoraproject.org/faf
+# Set to true if FAF server uses self-signed certificate
+#:abrt_server_ssl_noverify: true
+# Following two options enable client authentication to FAF server
+#:abrt_server_ssl_cert: /etc/foreman-proxy/faf_client_cert.pem
+#:abrt_server_ssl_key: /etc/foreman-proxy/faf_cleint_key.pem
 
 # enable BMC management  (Bare metal power and bios controls)
 # Available providers:
--- a/lib/abrtproxy_api.rb
+++ b/lib/abrtproxy_api.rb
@@ -1,0 +1,64 @@
+require 'openssl'
+require 'json'
+
+require 'proxy/request'
+require 'proxy/abrtproxy'
+
+STATUS_ACCEPTED = 202
+
+class SmartProxy
+  post "/abrt/reports/new/" do
+    begin
+      cn = Proxy::AbrtProxy::common_name request
+    rescue Proxy::Error::Unauthorized => e
+      log_halt 403, "Client authentication failed: #{e.message}"
+    end
+
+    ureport_json = request['file'][:tempfile].read
+    ureport = JSON.parse(ureport_json)
+
+    #forward to FAF
+    response = nil
+    if SETTINGS.abrt_server_url
+      begin
+        result = Proxy::AbrtProxy::faf_request "/reports/new/", ureport_json
+        response = result.body if result.code.to_s == STATUS_ACCEPTED.to_s
+      rescue RuntimeError => e
+        logger.error "Unable to forward to ABRT server: #{e}"
+      end
+    end
+    if not response
+      # forwarding is not configured or failed
+      # FAF source that generates replies is in src/webfaf/reports/views.py
+      response = { "result" => false,
+                   "message" => "Report forwarded to Foreman server" }.to_json
+    end
+
+    #send report to Foreman
+    begin
+      foreman_report = Proxy::AbrtProxy::create_report cn, ureport
+      Proxy::Request::Reports.new.post_report(foreman_report.to_json)
+    rescue RuntimeError => e
+      log_halt 503, "Unable to forward to Foreman server: #{e}"
+    end
+
+    status STATUS_ACCEPTED
+    response
+  end
+
+  post "/abrt/reports/:action/" do
+    # pass through to real FAF if configured
+    if SETTINGS.abrt_server_url
+      body = request['file'][:tempfile].read
+      result = Proxy::AbrtProxy::faf_request "/reports/#{params[:action]}/", body
+      if result
+        status result.code
+        result.body
+      else
+        log_halt 503, "ABRT server unavailable"
+      end
+    else
+      log_halt 501, "foreman-proxy does not implement /reports/#{params[:action]}/"
+    end
+  end
+end
--- a/lib/proxy.rb
+++ b/lib/proxy.rb
@@ -1,5 +1,5 @@
 module Proxy
-  MODULES = %w{dns dhcp tftp puppetca puppet bmc chefproxy}
+  MODULES = %w{dns dhcp tftp puppetca puppet bmc chefproxy abrtproxy}
   VERSION = "1.5-develop"
 
   require "checks"
@@ -18,7 +18,7 @@ module Proxy
   require "proxy/dns"        if SETTINGS.dns
   require "proxy/dhcp"       if SETTINGS.dhcp
   require "proxy/bmc"        if SETTINGS.bmc
-  require "proxy/chefproxy"  if SETTINGS.chefproxy
+  require "proxy/abrtproxy"  if SETTINGS.abrtproxy
 
   def self.features
     MODULES.collect{|mod| mod if SETTINGS.send mod}.compact
--- a/lib/proxy/abrtproxy.rb
+++ b/lib/proxy/abrtproxy.rb
@@ -1,0 +1,111 @@
+require 'net/http'
+require 'net/https'
+require 'uri'
+
+module Proxy::AbrtProxy
+  extend Proxy::Log
+
+  # Generate multipart boundary separator
+  def self.suggest_separator
+      separator = "-"*28
+      base = ('a'..'z').to_a
+      16.times { separator << base[rand(base.size)] }
+      separator
+  end
+
+  # It seems that Net::HTTP does not support multipart/form-data - this function
+  # is adapted from http://stackoverflow.com/a/213276 and lib/proxy/request.rb
+  def self.form_data_file(content, file_content_type)
+    # Assemble the request body using the special multipart format
+    thepart =  "Content-Disposition: form-data; name=\"file\"; filename=\"*buffer*\"\r\n" +
+               "Content-Type: #{ file_content_type }\r\n\r\n#{ content }\r\n"
+
+    boundary = self.suggest_separator
+    while thepart.include? boundary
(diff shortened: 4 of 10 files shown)


lzap:
Thanks for your contribution, the formal review will follow shortly. [test]

lib/proxy.rb
```diff
@@ -18,7 +18,8 @@ module Proxy
   require "proxy/dns"        if SETTINGS.dns
   require "proxy/dhcp"       if SETTINGS.dhcp
   require "proxy/bmc"        if SETTINGS.bmc
-  require "proxy/chefproxy"  if SETTINGS.chefproxy
+  #require "proxy/chefproxy"  if SETTINGS.chefproxy
```
lzap: Leftover?

lzap:
Initial code reading looks fine, can you paste some info how to configure the ABRT client. I will be happy to test this tomorrow.

mmilata:
The client configuration is described on https://groups.google.com/forum/#!topic/foreman-dev/Yk2GrW9ne20 (please note that `# abrt-auto-reporting enabled` and other lines below starting with # are commands to be run as root; lines starting with # above are config file comments).

If anything is unclear, you can also contact me via IRC (either internal or as _b42_ on #abrt@freenode).

lib/proxy.rb
mmilata: Oops. Should I reopen the PR or wait after you finish the review?

config/settings.yml.example
```diff
@@ -12,7 +12,11 @@
 #:trusted_hosts:
 #- foreman.prod.domain
 #- foreman.dev.domain
+# Following is needed for communication with foreman initiated by the proxy.
 #:foreman_url: http://127.0.0.1:3000
+#:foreman_ssl_ca: ssl/certs/ca.pem
+#:foreman_ssl_cert: ssl/certs/fqdn.pem
```
lzap: Hmmm I am trying to find use of `foreman_ssl_*` variables in this patch and I am not succesful. I only see validations for the `abrt_server_ssl*` variables bellow. Am I searching correctly, or is this expected. I wonder what is unused settings for then.

config/settings.yml.example
```diff
@@ -12,7 +12,11 @@
...
+#:foreman_ssl_key: ssl/private_keys/fqdn.pem
```
lzap: One idea - the `foreman_ssl_*` variables could default to `ssl_*` variables when not set. In case when foreman and smart-proxy are on the same host (typical small deployment setups) this could work "out-of-box" without uncommenting those lines. Just an idea, opinions?

config/settings.yml.example
mmilata: The variables are used in `lib/proxy/request.rb` (which was renamed from `lib/proxy/chefproxy.rb` in the commit). They should probably have been added in the patch that introduced the chef proxy.

lib/proxy.rb
lzap: Let us finish the review, then fix it, amend the changes and force-push into the same branch. Github will notice the changes, no need to close/open/reopen the PR.

When amending, please reword your commit message to:

```
fixes #4332 - initial abrt proxy support
```

I created feature for this: http://projects.theforeman.org/issues/4332

config/settings.yml.example
mmilata: Yes, I think it might make sense. In fact, you would usually want to set the `foreman_ssl_*` variables same as the `ssl_*` even if you are running the proxy on a different host.

I can't think of a reason why would you want those two to be different - perhaps it's useful for the Chef proxy?

lzap:
Sidenote: In the instructions for this feature, we must make sure that abrt-addon-ccpp package is installed.

lzap:
Ok this is _working fine_. The only issue I had was that abrt autoupload feature was not working properly on Fedora 20 minimal install. Workaround:

```
yum -y install abrt abrt-addon-ccpp abrt-python
```

I will look on the ruby code now.

config/settings.yml.example
lzap: I see, thanks.

config/settings.yml.example
lzap: Ah I see, out of scope for this patch. Ok.

lib/abrtproxy_api.rb
```diff
@@ -0,0 +1,63 @@
...
+    begin
+      cn = Proxy::AbrtProxy::common_name request
+    rescue Proxy::Error::Unauthorized => e
+      log_halt 403, "Client authentication failed: #{e.message}"
+    end
+
+    ureport_json = request['file'][:tempfile].read
+    ureport = JSON.parse(ureport_json)
+
+    #forward to FAF
```
lzap, line 20: Can you also enclose the whole FAF sending to a begin/rescue block catching all the Exceptions with a message "Unable to send to FAF"? Use log_halt to generate the error and return 503.

lib/abrtproxy_api.rb
```diff
@@ -0,0 +1,63 @@
...
+    end
+
+    #send report to Foreman
+    begin
+      foreman_report = Proxy::AbrtProxy::create_report cn, ureport
+      Proxy::Request::Reports.new.post_report(foreman_report.to_json)
+    rescue SystemCallError => e
+      logger.warn e
+      status 503
+      return "Unable to forward to Foreman server"
```
lzap: Cange these three lines into:

```
log_halt 503, "Unable to forward to Foreman server: #{e}"
```

lib/abrtproxy_api.rb
```diff
@@ -0,0 +1,63 @@
...
+    response = nil
+    if SETTINGS.abrt_server_url
+      result = Proxy::AbrtProxy::faf_request "/reports/new/", ureport_json
+      response = result.body if result && result.code.to_s == STATUS_ACCEPTED.to_s
+    end
+    if not response
+      # FAF source that generates replies is in src/webfaf/reports/views.py
+      response = { "result" => false,
+                   "message" => "Report forwarded to Foreman server" }.to_json
+    end
```
lzap: This is maybe too complicated, why not something like:

```
    if SETTINGS.abrt_server_url
       result = Proxy::AbrtProxy::faf_request "/reports/new/", ureport_json
       response = result.body if result && result.code.to_s == STATUS_ACCEPTED.to_s
    else
       # FAF source that generates replies is in src/webfaf/reports/views.py
       response = { "result" => false,
                    "message" => "Report forwarded to Foreman server" }.to_json
     end
```

lzap: Assuming you do not need to check the result.body tho.

lib/abrtproxy_api.rb
```diff
@@ -0,0 +1,63 @@
...
+      if result
+        status result.code
+        result.body
+      else
+        status 503
+        "ABRT server unavailable"
+      end
+    else
+      status 501
+      "foreman-proxy does not implement /reports/#{params[:action]}/"
```
lzap: `log_halt` again maybe? same for the guy above.

lib/proxy/abrtproxy.rb
```diff
@@ -0,0 +1,111 @@
...
+  # Generate multipart boundary separator
+  def self.suggest_separator
+      separator = "-"*28
+      base = ('a'..'z').to_a
+      16.times { separator << base[rand(base.size)] }
+      separator
+  end
+
+  # It seems that Net::HTTP does not support multipart/form-data - this function
+  # is adapted from http://stackoverflow.com/a/213276 and lib/proxy/request.rb
```
lzap, line 17: IANAL, is this fine?

lib/proxy/abrtproxy.rb
```diff
@@ -0,0 +1,111 @@
...
+    end
+
+    headers, body = self.form_data_file content, content_type
+
+    path = [uri.path, path].join unless uri.path.empty?
+    begin
+      response = http.start { |con| con.post(path, body, headers) }
+    rescue SystemCallError => e
+      # FAF unreachable
+      logger.warn e
```
lzap: Better would be logger.error or something like that.

lib/proxy/abrtproxy.rb
mmilata, line 17: The commenter said _The code in this post is licensed under the WTFPL_ so I guess there shouldn't be problems. IANAL either though.

lib/abrtproxy_api.rb
mmilata: In the next version, the request is wrapped in do..rescue block and we need to set the response when the request fails. So I'll leave the structure as it is now.

lib/abrtproxy_api.rb
lzap, line 20: Thanks. [test]

Ok this needs some additional opinions from the team.

domcleal:
I don't think this should be in the proxy core, especially as it masquerades via configuration management reports.  This should be written as a plugin once the proxy has plugin support of its own.

lzap:
Well my assumption was to generalize reports to some degree. I am fine
with the plugin.

## 

Later,

 Lukas "lzap" Zapletal
 irc: lzap #theforeman

mmilata:
@domcleal Afaik it is planned for foreman to support multiple report types: ~~http://projects.theforeman.org/issues/4332~~ http://projects.theforeman.org/issues/4151. Then the ABRT proxy can be rewritten to use its own report format. Would that be acceptable? Do you have any estimate for the smart-proxy plugin support?

domcleal:
Yes, Foreman should support different report types (http://projects.theforeman.org/issues/4151), but I don't think that would be appropriate still.  Reports in Foreman's context are configuration management reports, so it would additionally require different treatment - e.g. we show a host as out of sync when it doesn't send in reports.  If an ABRT "report" arrived, that would incorrectly imply that configuration management is working.  I think it would require even more changes to how "reports" are handled, or simply to define a new object type.. which can already be done today in a plugin.

I don't know when proxy plugin support (http://projects.theforeman.org/issues/3351) will be done, as it's not on our current sprint and not prioritised in the backlog.

mmilata:
Oops, #4151 was what I meant to link.

I'll take a closer look at foreman plugins. From what you say it certainly makes sense to treat them separately from the configuration management reports.

iNecas:
@domcleal what about having reports more generic, and a subtype of them would be ConfigManagement related: this subtypes would update the sync status, while ABRT type would not cause it. Since the reports are all that similar, I don't see why we would need something completely different for the ABRT type of reports?

domcleal:
@iNecas indeed.  Tthough I don't know what ABRT sends, that'd be sensible.

mmilata:
The ABRT reports look like this: https://github.com/mmilata/smart-proxy/blob/abrtproxy/test/fixtures/abrt/ureport1.json

ares:
I think report is still a report. We may set host status based on config mgmt reports only but I think we could reuse a lot of existing report functionality. 
1) if ABRT report was imported, it would be displayed on dashboard
2) ABRT report will later aggregate crashes, so counters (e.g. for failures) seems appropriate and each failure should be separate log in report

I think it would be better to make reports more general. If we make STI which separates "config reports" and "crash reports" it will be easy to keep hosts statuses based only on config reports.

Also maybe later we could have more host statuses based on reports type. E.g. crash report could make host in some failing state. I can imagine other reports from let's say monitoring tools.

EDIT: I should have reloaded the page before submitting, seems like @iNecas already said it.

lzap:
I agree with Marek.

Having that said, I think this deserves some investigation of the field.
We should look on technologies we would like to see in Foreman and then
refactor this properly for once.

I suppose we aim for:
- puppet reports
- chef reports
- ansible reports
- abrt reports

In addition to that, it would be good to consider support for syslog to
some degree (syslog server integration). But this should be perhaps
better candidate for a plugin as this is something else than reporting.
It's a constant stream of logs.

## 

Later,

 Lukas "lzap" Zapletal
 irc: lzap #theforeman

iNecas:
Maybe SCAP reports as well?

mmilata:
Thanks for the feedback! I've opened new pull request for this, since the code changed significantly: https://github.com/theforeman/smart-proxy/pull/144

PR #135, opened 2014-02-20
Don't call sudo if the calling user is the same as :puppet_user

m4ce:
Hi,

the idea behind this change is that we don't really need to invoke sudo if the calling user (ENV['USER']) is the same as the one defined in :puppet_user.

Cheers,
Matteo

### DIFF
--- a/lib/proxy/puppet/mcollective.rb
+++ b/lib/proxy/puppet/mcollective.rb
@@ -4,10 +4,12 @@ module Proxy::Puppet
   class MCollective < Runner
     def run
       cmd = []
-      cmd.push(which("sudo"))
+      unless ENV['USER'] == SETTINGS.puppet_user
+        cmd.push(which("sudo"))
 
-      if SETTINGS.puppet_user
-        cmd.push("-u", SETTINGS.puppet_user)
+        if SETTINGS.puppet_user
+          cmd.push("-u", SETTINGS.puppet_user)
+        end
       end
 
       cmd.push(which("mco", "/opt/puppet/bin"))


domcleal:
Seems reasonable to me [test]

lib/proxy/puppet/mcollective.rb
```diff
@@ -4,10 +4,12 @@ module Proxy::Puppet
   class MCollective < Runner
     def run
       cmd = []
-      cmd.push(which("sudo"))
+      unless ENV['USER'] == SETTINGS.puppet_user
```
domcleal, line 7: We should check that puppet_user is actually set, as it's optional.  If `ENV['USER']` wasn't present and the user hadn't specified this setting, we might skip using sudo.  Maybe...

```
unless ENV['USER'] == (SETTINGS.puppet_user || 'root')
```

It might be good to use `Etc.getlogin` instead of ENV and add `require 'etc'`, but that's a nitpick.

lib/proxy/puppet/mcollective.rb
```diff
@@ -4,10 +4,12 @@ module Proxy::Puppet
   class MCollective < Runner
```
domcleal, line 4: We've got this code also in puppetrun.rb and a bit of it in puppet_ssh.rb so they would need updating too.  Maybe just move the sudo section into runner.rb and use it in all three.

lzap:
@m4ce I am closing this one, feel free to reopen and discuss with us what @domcleal propose. Thanks.

PR #139, opened 2014-03-14
fixes #4658 - parse puppet.conf with augeas instead of puppet internals

domcleal:
@ares?

Tested on EL6 with Puppet 3.5 successfully.

### DIFF
--- a/lib/proxy/puppet/config_reader.rb
+++ b/lib/proxy/puppet/config_reader.rb
@@ -1,0 +1,35 @@
+require 'augeas'
+
+module Proxy
+  module Puppet
+    class ConfigReader
+      attr_reader :config
+
+      def initialize(config)
+        raise "Puppet config at #{config} was not found" unless File.exist?(config)
+        @config = config
+      end
+
+      def get
+        return @config_hash if @config_hash
+
+        aug = nil
+        begin
+          aug = Augeas::open(nil, nil, Augeas::NO_MODL_AUTOLOAD)
+          aug.set('/augeas/load/Puppet/lens', 'Puppet.lns')
+          aug.set('/augeas/load/Puppet/incl', config)
+          aug.load
+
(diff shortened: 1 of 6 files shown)


domcleal:
[test]

lib/proxy/puppet/config_reader.rb
```diff
@@ -0,0 +1,35 @@
...
+        begin
+          aug = Augeas::open(nil, nil, Augeas::NO_MODL_AUTOLOAD)
+          aug.set('/augeas/load/Puppet/lens', 'Puppet.lns')
+          aug.set('/augeas/load/Puppet/incl', config)
+          aug.load
+
+          @config_hash = {}
+          aug.match("/files#{config}/*/*[label() != '#comment']").each do |path|
+            (section, key) = path.split('/')[-2..-1]
+            (@config_hash[section.to_sym] ||= {})[key.to_sym] = aug.get(path)
```
ares: I'd consider defining config_hash like this

``` ruby
@config_hash = Hash.new { |h, k| h[k] = {} }
```

so I could do this

``` ruby
@config_hash[section.to_sym][key.to_sym] = aug.get(path)
```

also I'd

``` ruby
(section, key) = path.split('/')[-2..-1].map(&:to_sym)
```

so result would look like

``` ruby
@config_hash[section][key] = aug.get(path)
```

domcleal: Nice suggestions, updated.

ares:
Code :+1:, untested though, should I try to install puppet nightlies or did you tested carefully?

domcleal:
I've tested reasonably careful I think, patching a package installation and then doing a Puppet module import into Foreman successfully.

Just fixed the tests, I hope.

ares:
Merged as dd37400544ff6f50bca77a1c13a82a3073cb8ebd, thanks @domcleal

PR #140, opened 2014-03-14
fixes #4661 - store hostname with virsh DHCP reservations

domcleal:
@lzap?

### DIFF
--- a/lib/proxy/virsh.rb
+++ b/lib/proxy/virsh.rb
@@ -54,14 +54,21 @@ module Proxy::Virsh
     raise Proxy::DNS::Error, "Failed to update DNS: #{e}"
   end
 
-  def virsh_update_dhcp command, mac, ip
+  def virsh_update_dhcp command, mac, ip, name
     mac = escape_for_shell(mac)
     ip = escape_for_shell(ip)
     net = escape_for_shell(network)
+
+    if name
+      name = escape_for_shell(name)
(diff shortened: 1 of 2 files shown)


lib/proxy/virsh.rb
```diff
@@ -54,14 +54,21 @@ def virsh_update_dns command, hostname, ip
...
+      xml = "'<host mac=\"#{mac}\" ip=\"#{ip}\"/>'"
+    end
+
     virsh "net-update", net, command, "ip-dhcp-host",
-      "--xml", "'<host mac=\"#{mac}\" ip=\"#{ip}\"/>'",
-      "--live", "--config"
+      "--xml", xml, "--live", "--config"
   rescue Exception => e
-    raise Proxy::DHCP::Error, "Failed to update DNS: #{e}"
+    raise Proxy::DHCP::Error, "Failed to update DHCP: #{e}"
```
lzap, line 72: Yeah the problem with this code is it is used both by DNS and DHCP. I guess we can live with this.

Going to test now, thanks.

domcleal, line 72: Not in the DHCP method it isn't.. the DNS one is above.

lzap:
Aah okay.

Works great, A+ Thanks

:+1:

lzap:
Merged as 2f75941 for Foreman 1.5. Thanks @domcleal

PR #141, opened 2014-03-14
fixes #4660 - stub omshell during DHCP tests

domcleal:
It looks like omcmd/shell wasn't actually stubbed out, so we were probably making real calls to a DNS server at 192.168.122.1.  Extended tests for the SPARC attributes while I was here, so we're checking more than the response code.

Three green runs:
http://ci.theforeman.org/job/test_proxy_develop_pull_request/236/
http://ci.theforeman.org/job/test_proxy_develop_pull_request/237/
http://ci.theforeman.org/job/test_proxy_develop_pull_request/238/

### DIFF
--- a/test/server_isc_test.rb
+++ b/test/server_isc_test.rb
@@ -15,8 +15,8 @@ class ServerIscTest < Test::Unit::TestCase
     SmartProxy.new
   end
 
-  def test_sparc_host
-    data = {
+  def sparc_attrs
+    {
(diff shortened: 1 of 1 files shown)


lzap:
Jeeez that was hard. https://www.youtube.com/watch?v=Nr6rWjT-zcA

I think I can merge now?

lzap:
Merged as f4d9d0b for Foreman 1.5. Thanks @domcleal

PR #136, opened 2014-02-25
Added the ability to use rvmsudo when calling `puppet cert` commands.

Ginja:
Setup:

CentOS 6.4
RVM
Ruby 2.0.0-p451 via RVM
Puppet, Facter, Hiera, etc... provided by Gem installs
Foreman (git repo, just an ENC, served out through Passenger)
smart-proxy (git repo)
Puppet Master (served out through Passenger)

With my stack, I noticed that smart-proxy was throwing an error when attempting to use sudo with the puppet cert command:

$ sudo -S /usr/share/foreman/foreman-proxy/vendor/ruby/2.0.0/bin/puppet cert --ssldir /var/lib/puppet/ssl --list --all
/usr/lib/ruby/site_ruby/1.8/rubygems.rb:779:in `report_activate_error': Could not find RubyGem puppet (>= 0) (Gem::LoadError)
    from /usr/lib/ruby/site_ruby/1.8/rubygems.rb:214:in`activate'
    from /usr/lib/ruby/site_ruby/1.8/rubygems.rb:1082:in `gem'
    from /usr/share/foreman/foreman-proxy/vendor/ruby/2.0.0/bin/puppet:22

From the looks of it, it's trying to use the system's Ruby installation. In order to resolve this, I've added another config option that specifies whether or not to use rvmsudo, which passes the appropriate RVM environment variables to sudo. If :use_rvmsudo: is commented out (default), or set to anything other than true (no quotes), smart-proxy will use sudo. So it shouldn't interfere with existing installations.

Here is a snippet from proxy.log with :use_rvmsudo: set to true:

D, [2014-02-25T15:11:39.825499 #4441] DEBUG -- : Found puppetca at /usr/share/foreman/foreman-proxy/vendor/ruby/2.0.0/bin/puppet
D, [2014-02-25T15:11:39.825753 #4441] DEBUG -- : Found sudo at /usr/local/rvm/bin/rvmsudo
D, [2014-02-25T15:11:39.825815 #4441] DEBUG -- : Executing /usr/local/rvm/bin/rvmsudo -S /usr/share/foreman/foreman-proxy/vendor/ruby/2.0.0/bin/puppet cert --ssldir /var/lib/puppet/ssl --list --all

And with :use_rvmsudo: set to false:

D, [2014-02-25T15:14:33.972073 #5612] DEBUG -- : Found puppetca at /usr/share/foreman/foreman-proxy/vendor/ruby/2.0.0/bin/puppet
[...]

PR #142, opened 2014-03-17
puppet cert sudo now optional

lzap:
REVIEW ONLY

How about this? Please someone test.

### DIFF
--- a/config/settings.yml.example
+++ b/config/settings.yml.example
@@ -74,6 +74,8 @@
 :puppetca: false
 #:ssldir: /var/lib/puppet/ssl
 #:puppetdir: /etc/puppet
+#:puppetca_use_sudo: true
+#:sudo_command: /usr/bin/sudo
 
 # enable Puppet management
 :puppet: false
--- a/lib/proxy/puppetca.rb
+++ b/lib/proxy/puppetca.rb
@@ -71,7 +71,7 @@ module Proxy::PuppetCA
     # list of all certificates and their state/fingerprint
     def list
       find_puppetca
-      command = "#{@sudo} -S #{@puppetca} --list --all"
+      command = "#{@sudo} #{@puppetca} --list --all"
       logger.debug "Executing #{command}"
       response = `#{command}`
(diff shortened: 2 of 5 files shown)


config/settings.yml.example
```diff
@@ -74,6 +74,8 @@
 :puppetca: false
 #:ssldir: /var/lib/puppet/ssl
 #:puppetdir: /etc/puppet
+#:use_sudo: true
```
domcleal: What's the intended scope of this setting?

lib/proxy/puppetca.rb
```diff
@@ -71,7 +71,7 @@ def autosign_list
     # list of all certificates and their state/fingerprint
     def list
       find_puppetca
-      command = "#{@sudo} -S #{@puppetca} --list --all"
```
domcleal, line 74: Why removing -S?

lib/proxy/util.rb
```diff
@@ -45,6 +45,12 @@ def self.join_all
     end
   end
 
+  # convert setting to boolean (with a default value)
+  def to_bool(value, default = false)
```
domcleal, line 49: Can you add tests?

config/settings.yml.example
lzap: puppetca

I will rename to puppetca_use|_sudo.

lib/proxy/puppetca.rb
lzap, line 74: Because it's added elsewhere. Line 135.

lzap:
Test added, rebased.

Ginja:
@lzap, I've tested this and it resolves the issue I raised in #136. :+1: 

@domcleal, @lzap, any thoughts on which PR should be merged?

domcleal:
@Ginja I'm going to take this one as it gives us more flexibility.  Thanks for testing and your earlier PR, much appreciated!

Merged as 3824d18 for Foreman 1.5.0, thanks @lzap!

PR #144, opened 2014-03-26
Initial abrt proxy support (2nd version)

mmilata:
This change makes it possible for the proxy to pretend it is ABRT
(Automatic Bug Reporting Tool[1]) server. When it receives a problem
report it stores it to the disk, and optionally also forwards the
original report to an actual ABRT server.

Forwarding the stored reports to the Foreman server is done by
smart-proxy-abrt-send script which is invoked from cron. It reads the
reports, groups them by the host they came from, optionally merges those
with the same error (the satyr gem and library are needed), and sends
them to the Foreman, one report per host, masquerading as a report from
Puppet.

Non strictly ABRT-related changes:
- Move module Proxy::ChefProxy to Proxy::Request
- Add :foreman_ssl_ options to settings example

[1] https://github.com/abrt/abrt/wiki/ABRT-Project

NOTES:
- the aggregation-by-same-crash needs satyr gem (available on rubygems.org, I can provide specfile if needed) and satyr package (available in F19)
- instead of appending the reports to a file, I decided to store each report in individual file in a spool directory - this way we avoid locking completely
- even if you decide not to merge it (for the reasons given in https://github.com/theforeman/smart-proxy/pull/133#issuecomment-34998283), I would appreciate any feedback as it will mean less changes in the future when smart-proxy supports plugins and foreman accepts multiple report types


lzap:
Thanks, if there are no other volunteers I am happy to take for review. [test]

domcleal:
My position remains unchanged, I think this should be implemented as a plugin and not merged into core, especially given the complexity.

lzap:
Does this mean we shell wait intil proxy is pluginized?

domcleal:
That's what I'd prefer, yes.  @witlessbird is working on the first stories for that this sprint.

bin/smart-proxy-abrt-send
```diff
@@ -0,0 +1,73 @@
+#!/usr/bin/env ruby
+
+$LOAD_PATH.unshift *Dir["#{File.dirname(__FILE__)}/../lib"]
```
iNecas: this could be simplified with:

``` ruby
$LOAD_PATH.unshift File.expand_path('../../lib', __FILE__)
```

bin/smart-proxy-abrt-send
```diff
@@ -0,0 +1,73 @@
...
+      hr.unlink
+    rescue StandardError => e
+      logger.error "Cannot delete #{hr.files}: #{e}"
+    end
+  end
+end
+
+if ARGV[0] == "--daemon"
+  sleep_interval = SETTINGS.abrt_send_period || 600
+  Process.daemon(true) unless PLATFORM =~ /mingw/
```
iNecas: If running in mingw with --daemon switch, shouldn't it fail?

lib/abrtproxy_api.rb
```diff
@@ -0,0 +1,63 @@
...
+    response = nil
+    if SETTINGS.abrt_server_url
+      begin
+        result = Proxy::AbrtProxy::faf_request "/reports/new/", ureport_json
+        response = result.body if result.code.to_s == STATUS_ACCEPTED.to_s
+      rescue StandardError => e
+        logger.error "Unable to forward to ABRT server: #{e}"
+      end
+    end
+    if not response
```
iNecas: `unless` can be used, I would prefer that,  but not insisting on that

lib/proxy.rb
```diff
@@ -18,7 +18,7 @@ module Proxy
   require "proxy/dns"        if SETTINGS.dns
   require "proxy/dhcp"       if SETTINGS.dhcp
   require "proxy/bmc"        if SETTINGS.bmc
-  require "proxy/chefproxy"  if SETTINGS.chefproxy
```
iNecas, line 21: is removing the `chefproxy` intentional here?

lib/proxy/abrtproxy.rb
```diff
@@ -0,0 +1,246 @@
...
+            "metrics"     => { "resources" => { "total" => 0 },
+                               "time"      => { "total" => 0 }
+                             },
+            "logs"        => []
+            }
+      }
+      @reports.each do |ar|
+        message = ar.report["reason"]
+        message << " (repeated #{ar.count} times)" if ar.count > 1
+        foreman_report["report"]["status"]["failed"] += ar.count
```
iNecas: this could probably be extracted into mehotds `failed_reports_count` and `report_logs`, used directly when constructing the `foreman_report _hash`, so that this method would look like this:

``` ruby
    def create_foreman_report
        { "report" => {
            "host" => @host,
            "reported_at" => Time.now.utc.to_s,
            "status" => { "applied" => 0,
                               "restarted" => 0,
                               "failed" => failed_reports_count,
                               "failed_restarts" => 0,
                               "skipped" => 0,
                               "pending" => 0
                             },
            "metrics" => { "resources" => { "total" => 0 },
                               "time" => { "total" => 0 }
                             },
            "logs" => report_logs
            }
      }
    end
```

lib/proxy.rb
iNecas, line 21: Ok, I got it now: it was renamed to `request` as it's not chef specific anymore

bin/smart-proxy-abrt-send
mmilata: Currently it will run in a loop in the foreground. Is it preferable that it fails?

iNecas: Well personally,  I would be confused: if I wanted daemonized mode, I would expect that to go into background or fail when not possible

lzap: > Well personally,  I would be confused: if I wanted daemonized mode, I
> would expect that to go into background or fail when not possible

There are no daemons on Windows platform AFAIK. You have those services
you need to register first. Not a big deal I think :-)

Throwing an error is fine indeed.

## 

Later,

 Lukas "lzap" Zapletal
 irc: lzap #theforeman

bin/smart-proxy-abrt-send
mmilata: The behavior is different when the path that is computed does not exist ($LOAD_PATH unchanged in the first case, prepended with the nonexistend directory in the second), don't know if it matters. I took the line from `bin/smart-proxy`.

mmilata:
Thanks for the feedback, I've force-pushed updated branch. The old one can be found in the `abrtproxy-old` branch in my repo if needed.

mmilata:
Closing, abrt support is being implemented as a plugin: https://github.com/abrt/foreman_proxy_abrt

PR #148, opened 2014-04-09
Bug Fix: Incorrect IP returned as next available on subnet

seanhandley:
If there is an existing DHCP record for a host, the address is re-used even if it's from a different subnet.

This fix introduces a check to ensure that the address which is selected only comes from the existing record if the address also belongs to the given subnet.

### DIFF
--- a/lib/proxy/dhcp/subnet.rb
+++ b/lib/proxy/dhcp/subnet.rb
@@ -139,7 +139,7 @@ module Proxy::DHCP
     def unused_ip args = {}
       # first check if we already have a record for this host
       # if we do, we can simply reuse the same ip address.
-      if args[:mac] and r=has_mac?(args[:mac])
+      if args[:mac] and r=has_mac?(args[:mac]) and valid_range(args).include?(r.ip)
         logger.debug "Found an existing dhcp record #{r}, reusing..."
         return r.ip
       end
--- a/test/subnet_test.rb
+++ b/test/subnet_test.rb
@@ -115,4 +115,26 @@ class Proxy::DHCPSubnetTest < Test::Unit::TestCase
     assert_equal @subnet.size, counter-1
   end
 
+  def test_should_reuse_ip_if_from_same_subnet
+    @subnet.stubs(:has_mac?).returns(stub(:ip => '192.168.0.10'))
+    @subnet.stubs(:icmp_pingable?)
+    @subnet.stubs(:tcp_pingable?)
+    assert_equal '192.168.0.10', @subnet.unused_ip({:mac => '0'})
+  end
+
+  def test_should_not_reuse_ip_if_from_other_subnet
+    @subnet.stubs(:has_mac?).returns(stub(:ip => '10.0.0.10'))
+    @subnet.stubs(:icmp_pingable?)
+    @subnet.stubs(:tcp_pingable?)
+    assert_equal false, @subnet.unused_ip({:mac => '0'}).include?("10.0.0")
+  end
+
+  def test_should_not_reuse_ip_if_from_outside_range
+    @subnet.stubs(:has_mac?).returns(stub(:ip => '192.168.0.250'))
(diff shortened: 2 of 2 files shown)


mattj-io:
Should probably qualify that comment a bit - the address is re-used if it's outside of the autosuggestion range not the subnet, and this checks if the address is in the autosuggestion range before re-using. See Bug #5123 and this thread https://groups.google.com/forum/#!topic/foreman-users/DWcHKPYJRJo

domcleal:
[test]

Could you review please @lzap?

seanhandley:
@domcleal Could you re-[test] please?

domcleal:
sure, [test]

seanhandley:
Hmmm, I get this locally also:

```
1) Failure:
RunnerTest#test_popen [/Users/sean/src/foreman/smart-proxy/test/runner_test.rb:11]:
<0> expected but was
<2>.
```

Can someone tell me what this means and how we fix it?

domcleal:
This seems to fix it: https://github.com/domcleal/smart-proxy/commit/0f88f09ec909c43eda66e1773f2cde5aa1ada5b7

I'm not sure why to be honest.. but I noticed while running `rake` locally that it was pausing on the subnet unused IP tests, which suggested to me it was probably pinging my LAN to find a free IP.  Somehow, stubbing that out fixes the popen test!

seanhandley:
Great stuff @domcleal - thanks. I've fixed, squashed and pushed.

Could you re-[test]?

domcleal:
Here we go.. [test]

lzap:
Sorry for the delay, I will test this on my instance. Can you also add test for scenario when we have MAC reservation but it is actually OUTSIDE of the specified range? We do not cover this one with tests I believe.

lzap:
Ok disregard my previous comment about tests, I think the latter test covers that fine.

I was able to test both scenarios, works fine. One note: when doing provisioning of discovered host, the leased IP is suggested no matter if it is in the specified range or not. This is expected behavior - discovery puts the IP explicitly to the form. +1

mattj-io:
Was about to update the foreman-users thread to that effect. The API call doesn't happen in the edit view in Foreman unless you switch to a different subnet and then back again, at which point with this change to the proxy you will get the auto-suggested IP. In our case, for discovery hosts we'd prefer for the form to be auto-populated with the auto-suggested IP first without having to do that, would a change be acceptable to do that just for discovery hosts ? This is possibly not the right place to ask that question so feel free to point us to the right place for that discussion :)

seanhandley:
I've added another test anyway for completeness @lzap 

I think it's worth having since it makes it completely clear what the behaviour is. Would you re-[test] ?

lzap:
Oh damn I am in the process of squashing it. :-) [test]

lzap:
Hmmm that should go perhaps into the discovery plugin, we create the host record once it is discovered and then we just keep the IP there.

But what could be interesting for you might be a button next to the IP field that would request another IP on click. I was already thinking this could be great addition because during testing I very often need "another" IP address (and dnsmasq which I use seems to return random IPs). That could help with discovery, what you think?

mattj-io:
We haven't dug around in the code to generate that view yet, if the host record exists it presumably just loads whatever is in the record ? It's really just for discovered hosts where we'd want to change the IP, unless we're changing subnets in which case we get a new IP suggested anyway. Having played with the view a bit more, at the point when the form initially loads you don't have a subnet set,  so populating with an auto-suggested IP is possibly not the right behaviour - we were thinking that maybe there should be no IP there for discovered hosts until you select a subnet, at which point the API call should happen. Having a button which iterates the suggested IP's could also be useful in other contexts though, we'll have a poke around in the code.

lzap:
Pushed as
- 85098a3 fixes #5123 - corrected string typos <Matt Jarvis>
- 54ee3a8 fixes #5123 - IP only from the same subnet and range <Sean Handley>

Thanks guys!

PR #150, opened 2014-04-29
Fixes #4866: Refactoring of SmartProxy functionality into isolated modules.

dmitri-d:


dmitri-d:
[test]

dmitri-d:
[test]

lzap:
Is this ready for review?

dmitri-d:
@lzap: yeah, please do (pls. ignore test failures for the time being)!

lzap:
Sure, but we need more reviewers, huge change.

bin/smart-proxy.rb
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env ruby
+
+$LOAD_PATH.unshift *Dir[File.expand_path("../../lib", __FILE__), File.expand_path("../../modules", __FILE__)]
+
+require 'bundler'
```
ohadlevy: do we force bundle now? until now it was not 100% required.

lib/proxy/plugin.rb
```diff
@@ -0,0 +1,145 @@
...
+class ::Proxy::Dependency
+  attr_reader :name, :version
+
+  def initialize(aname, aversion)
+    @name = name.to_sym
+    @version = version
+  end
+end
+
+class ::Proxy::Plugins
```
ohadlevy, line 13: is it possible to keep one class per file?

bin/smart-proxy.rb
lzap: and if we choose to go with bundler, we need to make sure this works with bundler_ext too.

dmitri-d: We need a way to manage/load plugins that are not in the core; Bundler is probably the most straightforward and common way to accomplish that?

config/settings.yml.example
```diff
@@ -18,8 +18,9 @@
 :daemon: true
 :daemon_pid: /var/run/foreman-proxy/foreman-proxy.pid
 
-# Port used by the proxy
-:port: 8443
+# ports used by the proxy
+:http_port: 8000
+:https_port: 8443
```
lzap: :notebook_with_decorative_cover: we need to change the installer defaults for proxy registration

lzap: oh no, scratch that, https port stays the same. okay.

lib/proxy/version.rb
```diff
@@ -0,0 +1,3 @@
+module Proxy
+  VERSION = "1.5.develop"
```
ohadlevy: 1.6.develop?

lib/proxy/plugin.rb
dmitri-d, line 13: Sure, I usually try to; This is an implementation detail however, and is irrelevant to rest of the code base, hence my decision to keep it here.

lib/smart_proxy.rb
```diff
@@ -1,85 +1,128 @@
...
-    set :public_folder, APP_ROOT + '/public'
-  else
-    set :public, APP_ROOT + '/public'
+require 'sinatra'
+require 'sinatra-patch'
+require 'fileutils'
+require 'pathname'
+require 'checks'
+require 'webrick/https'
+require 'daemon' # FIXME: Do we still need this?
```
ohadlevy, line 13: i think we do? at least for windows?

lib/smart_proxy.rb
```diff
@@ -1,85 +1,128 @@
...
+  SETTINGS_PATH = Pathname.new(__FILE__).join("..","..","config","settings.yml")
+  SETTINGS = Settings.load_from_file
+
+  ::Sinatra::Base.set :run, false
+  ::Sinatra::Base.set :root, APP_ROOT
+  ::Sinatra::Base.set :views, APP_ROOT + '/views'
+  ::Sinatra::Base.set :public_folder, APP_ROOT + '/public'
+  ::Sinatra::Base.set :logging, true
+  ::Sinatra::Base.set :env, :production
+#  ::Sinatra::Base.use ::Rack::CommonLogger, logger
```
ohadlevy: leftover?

lib/proxy/plugin.rb
```diff
@@ -0,0 +1,145 @@
...
+  end
+
+  def self.registered_plugins
+    @@registered.values
+  end
+end
+
+#
+# example of plugin DSL
+#
```
lzap: :facepalm:

lib/smart_proxy.rb
```diff
@@ -1,85 +1,128 @@
...
-  require "chefproxy_api" if SETTINGS.chefproxy
-  require "resolv"        if SETTINGS.trusted_hosts
-  require "realm_api"     if SETTINGS.realm
-
-  begin
-    require "facter"
-    require "facts_api"
-  rescue LoadError
-    warn "Facter was not found, Facts API disabled"
+  MODULES = %w{dns dhcp tftp puppetca puppet bmc chefproxy}
```
ohadlevy: missing realm?

lib/smart_proxy.rb
dmitri-d, line 13: We are relying on Rack for daemonization now, there is a chance it handles windows-specific stuff already.

lib/proxy/plugin.rb
```diff
@@ -0,0 +1,145 @@
...
+# class ExamplePlugin < ::Proxy::Plugin
+#  http_rackup_path File.expand_path("http_config.ru", File.expand_path("../", __FILE__)) # note no https rackup path, module will not be available over https
+#  requires :foreman_proxy, ">= 1.5.develop"
+#  requires :another_plugin, "~> 1.3.0"
+#  default_settings :first => 'first', :second => 'second'
+#  plugin :example, "1.2.3"
+#  before_registration { call_this }
+#  after_registration { call_that }
+# end
```
lzap: Nitpick. I'd maybe move this line to the top:

```
 #  plugin :example, "1.2.3"
```

lib/smart_proxy.rb
dmitri-d: No, more of a todo -- Sinatra can log request details if logger is set, which can be useful for debugging.

lib/smart_proxy.rb
```diff
@@ -1,85 +1,128 @@
...
+#  ::Sinatra::Base.use ::Rack::CommonLogger, logger
+
+  require 'root/root'
+  require 'facts/facts'
+  require 'dns/dns'
+  require 'tftp/tftp'
+  require 'dhcp/dhcp'
+  require 'puppetca/puppetca'
+  require 'puppet/puppet'
+  require 'bmc/bmc'
```
ohadlevy, line 46: wont it load regardless of configuration? before we had if statements on each api part.

lib/proxy/plugin.rb
dmitri-d: huh?

lib/proxy/plugin.rb
dmitri-d: Will do. The order is irrelevant, however.

lib/smart_proxy.rb
dmitri-d, line 46: It won't, please see respective plugin classes.

lib/proxy/plugin.rb
```diff
@@ -0,0 +1,145 @@
...
+
+  def validate_dependencies!(dependencies)
+    dependencies.each do |dep|
+      plugin = ::Proxy::Plugins.find_registered_plugin(dep.name)
+      raise ::Proxy::PluginNotFound "Plugin '#{dep.name}' required by plugin '#{plugin_name}' could not be found." unless plugin
+      unless ::Gem::Dependency.new('', dep.version).match?('', version)
+        raise ::Proxy::PluginVersionMismatch "Available version '#{version}' of plugin '#{dep.name}' doesn't match version '#{dep.version}' required by plugin '#{plugin_name}'"
+      end
+    end
+  end
```
lzap, line 173: This method explains why we do need Bundler. Now the question is - will it work with bundler_ext? :-) I think so, if not we should be able to turn the validation off when in bundler_ext mode, because RPM/Yum will take care of this automatically for us.

lib/proxy/proxy.rb
```diff
@@ -0,0 +1 @@
+module Proxy; end
```
lzap: Missing newline, vim will fix this automatically when you open this one.

lib/proxy/plugin.rb
dmitri-d, line 173: Nods, I don't see any reason why it wouldn't (but obviously need to test that).

lib/smart_proxy.rb
lzap, line 13: We need to test on Windows. Now, who has a windows license, raise your :hand: !

lib/smart_proxy.rb
```diff
@@ -1,85 +1,128 @@
...
+      ::Proxy::Plugins.register_loaded_plugins
+
+      launcher = Launcher.new
+      
+      launcher.create_pid_dir
+      http_app = launcher.http_app
+      https_app = launcher.https_app
+
+      t1 = Thread.new { http_app.start }  
+      t2 = Thread.new { https_app.start } unless https_app.nil?
```
lzap: Do I read it right that we always start http port and https only when it is configured? I don't think we want that.

dmitri-d: It's possible that https is either not present (deliberately) or misconfigured (accidentally). I think not starting https then (and providing a warning) is fine, as long as we don't automatically redirect to http?

modules/facts/facts.rb
```diff
@@ -0,0 +1,8 @@
+include ::Proxy::Log
+begin
+  require "facter"
+  require 'facts_plugin'
+rescue LoadError
+  logger.info "Facter was not found, Facts API disabled"
+end
+
```
lzap, line 8: Hmm bundler_ext should make sure we are using system puppet/facter.

lib/proxy/plugin.rb
lzap, line 13: I am fine with this, more than that one class per file sometimes bother me.

lib/proxy/plugin.rb
lzap: ignore, I usually put my face in my palms everytime I learn new DSL. this one was quick. :D

lib/smart_proxy.rb
lzap: I have a problem with the other way around, actually. We should not always expose smart proxy via HTTP. For development this is good, but I read it it is hardcoded here.

lzap:
First look is good, I will test that once you are done with tests and others are good. I had only minor comments. Good job.

lib/proxy/plugin.rb
dmitri-d: I can call this plugin API? :P

lib/smart_proxy.rb
dmitri-d: Hmm, good point. Anyone else? Do we even need to bother with http?

modules/facts/facts.rb
dmitri-d, line 8: yeah, this is a direct port from the old code. I tried not to make too many changes in this round.

lib/smart_proxy.rb
domcleal: HTTP is useful for dev setups at least, please keep it.

lzap: > Hmm, good point. Anyone else? Do we even need to bother with http?

God please yes! I do not want to setup https for development setup :-)

## 

Later,

 Lukas "lzap" Zapletal
 irc: lzap #theforeman

config/settings.yml.example
domcleal: We will need to update the installer template, though my bigger concern is for upgrades - can we make this backwards compatible?

bin/smart-proxy.rb
domcleal: Arguably, while the plugins users want are still in core, we could consider it optional too.

Edit: but yes, for loading other plugins, it's certainly the most straightforward way.  +1

bin/smart-proxy.rb
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env ruby
```
domcleal: Could we keep this without the .rb extension please?

domcleal:
Looks really good, thanks for keeping the changes minimal!  There are three main things to work on IMO:
1. bundler_ext compatibility, it'd be good to ensure this works the same way as Foreman itself with Gemfile.in.
2. HTTP and/or HTTPS - not exposing HTTPS-secured proxies via HTTP automatically when they upgrade to this version.
3. configuration file, I'd suggest we support a .d directory for settings as well as a single file, so we can immediately start shipping plugins with configuration.

dmitri-d:
@domcleal:
1. Indeed. I was planning on testing this bit. 
2. I wonder if we could simply disable http in production? @GregSutcliffe ?
3. On one hand I like the idea of breaking up configuration into per-module config files. How would enabling of a module work? Should we have 'enabled' directory?

domcleal:
1. What do you mean by production?  That's the usual installer configuration, so generally yes, we would.  But some people run without and for development use I think it's necessary.
2. I'm not sure if separate directories are needed, either merge all the configs together (we tend to use namespaced keys today) or specifically load individual config files into each proxy module (though that's less compatible with today's setup).

dmitri-d:
1. Nods; I'm not suggesting (anymore) we get rid of http mode. Just disabling it by default.
2. Wouldn't it be annoying needing to go into separate config files to enable/disable modules? I guess it's not a common scenario when several modules need to enabled/disabled at the same time...

domcleal:
Yeah, I don't mind too much about the default, since for most users that's largely dictated by foreman-installer.

It might be more laborious to configure, yes, though it's the only way we'll be able to package modules separately, i.e.. to supply default config files without having to get the core proxy one changed to show examples of an external module.

dmitri-d:
[test]

domcleal:
@witlessbird it won't test with a merge conflict, rebase time

GregSutcliffe:
+1 for defaulting to https, that's effectively what we do today anyway
+1 for a .d-style dir for config files.

lib/proxy/plugin.rb
```diff
@@ -0,0 +1,145 @@
...
+  attr_reader :name, :version
+
+  def initialize(aname, aversion)
+    @name = name.to_sym
+    @version = version
+  end
+end
+
+class ::Proxy::Plugins
+  @@loaded = [] # {:name, :version, :class}
```
iNecas, line 14: Could we get rid of the `@@class_variables` and just use the `@instance_variables` on the singleton class level? Class variables are not recommended to be used, as they behavior depends on the order of classes being loaded

lib/proxy/plugin.rb
iNecas: :)

dmitri-d:
@domcleal, @GregSutcliffe: per-module config is in!

domcleal:
@witlessbird looks good!

bin/smart-proxy
```diff
@@ -1,45 +1,12 @@
 #!/usr/bin/env ruby
 
-$LOAD_PATH.unshift *Dir["#{File.dirname(__FILE__)}/../lib"]
+$LOAD_PATH.unshift *Dir[File.expand_path("../../lib", __FILE__), File.expand_path("../../modules", __FILE__)]
 
+require 'bundler'
```
GregSutcliffe: Do we _have_ to use bundler? I've managed to avoid it in the proxy thus far, which means also managing to avoid vendoring gems in the Debian papcakges...

config/settings.yml.example
```diff
@@ -15,131 +14,19 @@
...
-# Redhat 5
-#:dhcp_config: /etc/dhcpd.conf
-# Redhat 6
-#:dhcp_config: /etc/dhcp/dhcpd.conf
-#:dhcp_leases: /var/lib/dhcpd/dhcpd.leases
-#:dhcp_key_name: secret_key_name
-#:dhcp_key_secret: secret_key
+# ports used by the proxy
+:http_port: 8000 # disabled by default
+#:https_port: 8443
```
GregSutcliffe, line 30: This seems confusing... the disabled-by-default http service is _enabled_ in the config file, but the enabled-by-default https service is commented out?

bin/smart-proxy
dmitri-d: I think so. There are no external plugins atm, but we need bundler to support them...

config/settings.yml.example
dmitri-d, line 30: I'm not sure what would be a good way to document default settings. Perhaps this is redundant -- default values used during intialization are being logged...

bin/smart-proxy
GregSutcliffe: Bleah, that's not good - the proxy was the _one_ package that could be accepted upstream in Debian. Use of bundler/rubygems denies that. Oh well, it's not a lot of use if core is never going to get there :p

Could we optionally support using bundler --standalone? IIRC that requires a small amount of code in the initializer/setup stage, but should leave the rest untouched.

dmitri-d: Possibly? What would be the usecase(s) for this?

domcleal: bundler --standalone is pretty equivalent to bundler_ext I think, which we already use (would be preferable I think..)

dmitri-d: Don't we have to use bundler_ext on RHEL-derived os', as bundling of deps in packages is frowned upon?

GregSutcliffe: I've not even tried to figure out if bundler_ext will work on Debian, whereas standalone I know works (Ohad just wouldn;t let me merge the changes to core :P). I could look into bundler_ext I guess - I'm not keen on vendoring the proxy as it talks to far more systems than core does (look what happened when we tried to SCL the proxy, we had to SCL puppet too...)

GregSutcliffe: At a first pass, bundler_ext seems workable on debian. Perhaps thats the way to go...

config/settings.yml.example
GregSutcliffe, line 30: i think we need to substantially extend the comments here, something like:

```
# port config
# by default the https service is enabled on 8443
#:https_port: 8443
# by default, the http service is disable, you can enable by <how?>
#:http_port: 8000
```

modules/puppetca/puppetca.rb
```diff
@@ -139,11 +139,11 @@ def find_puppetca
     end
```
GregSutcliffe, line 139: You have two more instances of SETTINGS that need to be converted to `Proxy::PuppetCa::Plugin.settings` on line 128/129

lzap:
@mmilata this thread you are interested in. Once this is merged, please rebase on top of it, thanks!

config/settings.yml.example
dmitri-d, line 30: I wonder if :ssl_certificate, :ssl_ca_file, and :ssl_private_key should have default values...

GregSutcliffe:
With https://github.com/witlessbird/smart-proxy/pull/4 applied, I can use tftp/dns/dhcp/puppetca/puppet modules in a full production environment to correctly provision, tftp, and puppetize a discovered host. Looks good!

GregSutcliffe:
Testing logging, i can see stacktraces in the log again. Looks good.

domcleal:
For the record, we've been trying to package this today and hit some issues that need resolving:
- currently it's requiring all bundler groups on startup, but should only require groups for modules that are enabled, as bundler.d/ files may exist for disabled modules (wb)
- `require 'rubygems'` is needed on Ruby 1.8 installs, but must rescue load errors for packaging on Debian (wb)
- rack 1.1.0 on EL6 failed as it didn't support passing a custom app into the rack server, I'll build a newer rack and sinatra stack (dc)
- bundler_ext, bundler and thor need adding to EL6 comps (dc)
- puppet-foreman_proxy needs a major overhaul to handle separate config files (wb)
- settings migration script required so it can be called from package upgrades etc.

GregSutcliffe:
Update from the Debian side:
- Bundler_ext packaging done for all 4 Debian OSs
- Foreman-proxy packaging updated for all 4
- I also have hit issues with Rack 1.1 on Squeeze. I'm not sure on the feasibility of building a new version of rack though...

I can take a look at a migration script if no-one else is picking that up.

domcleal:
There was a possible workaround for the Rack 1.1 issue - instead of generating the app ourselves (https://github.com/witlessbird/smart-proxy/blob/0f078186/lib/smart_proxy.rb#L69-L74) a new config.ru could be written to a temp file and passed into Rack::Server, but understandably wb's not too keen to go this far backwards.  Or drop squeeze support?

domcleal:
For the record, the Rack 1.1 issue is fixed in the current commit via rest_patch.rb, used on Rack < 1.3.

I'm hitting an interesting problem on an RPM upgrade of a clean package installation - the proxy installs its default config file, which works fine and runs the proxy in HTTP-only mode.  If you then upgrade it, the new example config file no longer starts the proxy by default until you configure https?_port.  Should we keep `:http_port: 8443` enabled by default?

domcleal:
I managed to fix it in packaging only, over in https://github.com/theforeman/foreman-packaging/pull/255.

Could you please squash the second commit straight into the main one, since we've already committed it to master, and also update the commit message with a relevant redmine ticket.  I think we're good to go.

dmitri-d:
@domcleal: done \o/

domcleal:
Merged for Foreman 1.6.0, thanks @witlessbird!

PR #153, opened 2014-05-01
Implements #4630: SmartProxy can query the status of Pulp server now.

dmitri-d:


domcleal:
I remain of the opinion that this should not be in core (it serves no real purpose), best build on top of the plugin framework.

dmitri-d:
Agreed, but I don't think modularized SmartProxy will be ready before May 7th (which is the deadline I was given).

ohadlevy:
Then why not simply patch katello to test for pulp when a proxy is added?

dmitri-d:
I thought about that. That would mean treating pulp proxy as a special case (probably as a Foreman plugin). Probably would involve some sort of orchestration too.
In the end I felt that keeping things in SmartProxy would be pretty simple and I didn't see huge benefits to the other approach. Extracting pulp support into a dedicated module is going to be pretty easy too.

ohadlevy:
Isn't it a simple call back extension in Katello plugin to test if pulp is
reachable e.g. in a before save? (Just like how we detect the features
today)?

And of course a seed for the new feature.

dmitri-d:
I think it's more complicated than that. What if pulp is down when we are attempting to query pulp status? There isn't a good way to distinguish a failure from legitimate absence of service in this case. We could add an additional setting for that, but that would mean extending SmartProxy model.

Again, I feel this is a more straightforward approach.

lib/pulp_api.rb
```diff
@@ -0,0 +1,49 @@
+require 'net/http'
+require 'net/https'
+require 'uri'
+
+class SmartProxy
```
ohadlevy, line 5: can you extract this to another file ? does it make any sense to use an inner class?

ohadlevy:
I cant see any difference between having a feature enabled on the proxy (e.g. in conf file) and a api endpoint that return the value vs. a checkbox in a ui saying that pulp is also installed.

long term you are right that the proxy installation should / could install pulp at the same time and configure the settings file, but I agree with @domcleal here that this should not be in core.
once the plugin arch is avail, we could use that instead, but as it stands now, I think this should be re implemented in Katello. any other opinions?

iNecas:
@ohadlevy what do you mean by reimplementing this in Katello? Does that mean we should not rely on the PR at all nowhere?

iNecas:
Also, it would be nice to bring more people from Katello in discussion regarding this kind of changes/feature //cc @jlsherrill @mccun934, I know it would be nice for all the people watching all the PRs but that's not possible
lack-of-transparency

iNecas:
For now, I will base my work on the katello to add pulp node feature to the smart-proxy on this PR, expecting that by the time of opening the PR to Katello, the plugin-style smart proxy will be merged, with this feature being available.

When there will be major issue with that, we can look into the workaround of having the feature detection logic in Katello as temporary solution.

dmitri-d:
[test]

domcleal:
This was implemented as a plugin (success!) over here: https://github.com/theforeman/smart-proxy-pulp, published as smart_proxy_pulp to rubygems.org and RPMs built to our plugins repo via https://github.com/theforeman/foreman-packaging/pull/284.

PR #155, opened 2014-05-09
Match the DHCP specification of last-wins leases

GregSutcliffe:
Just sending this PR mainly to get the tests to run on Jenkins - I'm not entirely happy with it yet...

Discussion of the issue can be found here: http://projects.theforeman.org/issues/5648

This attempts to preserve the current proxy behaviour of indexing DHCP records by IP, and only re-indexes by mac (thus updating/removing duplicates) when specifically searching for a mac record.

### DIFF
--- a/lib/proxy/dhcp/subnet.rb
+++ b/lib/proxy/dhcp/subnet.rb
@@ -77,26 +77,30 @@ module Proxy::DHCP
     end
 
     def [] record
-      self.load if not loaded?
-      begin
-        return has_mac?(record) if validate_mac(record)
-      rescue
-        nil
-      end
-      begin
-        return has_ip?(record) if validate_ip(record)
-      rescue
-        nil
-      end
+      records_for record, :all
     end
 
-    def has_mac? mac
-      r = records.reverse_each {|r| return r if r.mac == mac.downcase}
+    def has_mac? mac, type
+      r = case type
+          when :reservation
+            reservations
+          when :lease
+            leases
+          else
+            records
+          end.reverse_each {|r| return r if r.mac == mac.downcase}
       return false
     end
 
-    def has_ip? ip
-      r = records.reverse_each {|r| return r if r.ip == ip}
+    def has_ip? ip, type
+      r = case type
+          when :reservation
+            reservations
+          when :lease
+            leases
+          else
+            records
+          end.reverse_each {|r| return r if r.ip == ip}
       return false
     end
 
@@ -205,6 +209,20 @@ module Proxy::DHCP
 
     def leases
       records.collect{|r| r if r.kind == "lease"}.compact
+    end
+
+    def records_for record, type
+      self.load if not loaded?
+      return has_mac?(record, type) if (validate_mac(record) rescue nil)
+      return has_ip?(record, type)  if (validate_ip(record) rescue nil)
+    end
+
+    def reservation_for record
+      records_for record, :reservation
+    end
+
+    def lease_for record
+      records_for record, :lease
     end
 
     def <=> other
--- a/lib/proxy/dhcp/subnet.rb
+++ b/lib/proxy/dhcp/subnet.rb
@@ -22,7 +22,7 @@ module Proxy::DHCP
       @network   = validate_ip network
       @netmask   = validate_ip netmask
       @options   = {}
-      @records   = {}
+      @records   = []
       @timestamp = Time.now
       @loaded    = false
       raise Proxy::DHCP::Error, "Unable to Add Subnet" unless @server.add_subnet(self)
@@ -46,7 +46,7 @@ module Proxy::DHCP
     end
 
     def clear
-      @records = {}
+      @records = []
       @loaded  = false
     end
 
@@ -55,7 +55,7 @@ module Proxy::DHCP
     end
 
     def size
-      records.size
+      @records.size
     end
 
     def load
@@ -73,7 +73,7 @@ module Proxy::DHCP
 
     def records
       self.load if not loaded?
-      @records.values
+      @records
     end
 
     def [] record
@@ -91,23 +91,23 @@ module Proxy::DHCP
     end
 
     def has_mac? mac
-      records.each {|r| return r if r.mac == mac.downcase }
+      r = records.reverse_each {|r| return r if r.mac == mac.downcase}
       return false
     end
 
     def has_ip? ip
-      @records[ip] ? @records[ip] : false
+      r = records.reverse_each {|r| return r if r.ip == ip}
+      return false
     end
 
     # adds a record to a subnet
     def add_record record
-      unless has_mac?(record.mac) or has_ip?(record.ip)
-        @records[record.ip] = record
-        logger.debug "Added #{record} to #{to_s}"
-        return true
-      end
-      logger.warn "Record #{record} already exists in #{to_s} - can't add again"
-      return false
+      # Record all leases, since the definition of a duplicate depends on whether we
+      # are searching by ip or mac. Arrays have fixed sort order so we can rely on this
+      # being the order they were read from the file
+      @records.push record
+      logger.debug "Added #{record} to #{to_s}"
+      return true
     end
 
     def get_index_and_lock filename
@@ -176,7 +176,7 @@ module Proxy::DHCP
     end
 
     def delete record
-      if @records.delete_if{|k,v| v == record}.nil?
+      if @records.delete(record).nil?
         raise Proxy::DHCP::Error, "Removing a Proxy::DHCP Record which doesn't exist"
       end
     end
--- a/lib/dhcp_api.rb
+++ b/lib/dhcp_api.rb
@@ -104,7 +104,7 @@ class SmartProxy < Sinatra::Base
   # delete a record from a network
   delete "/dhcp/:network/:record" do
     begin
-      record = load_subnet[params[:record]]
+      record = load_subnet.reservation_for(params[:record])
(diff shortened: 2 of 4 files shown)


GregSutcliffe:
Alright, tests pass - I think this _might_ be mergable - but input on the direction we should be going is welcome

lib/proxy/dhcp/subnet.rb
```diff
@@ -91,7 +100,7 @@ def [] record
     end
 
     def has_mac? mac
-      records.each {|r| return r if r.mac == mac.downcase }
+      records(:by_mac).each {|r| return r if r.mac == mac.downcase }
```
lzap: Hmmm what's the point of building the whole hash while we use it on this line only to iterate it. That's zero time saved and extra memory consumed. I must be missing something.

lzap: I think you want to search it in O(1) rather than O(n) here :-)

lzap: In other words:

```
records(:by_mac)[mac.downcase]
```

lzap: Hmmm disregard the above, I see the point.

GregSutcliffe: For clarity to other reviewers: yeah, the issue is that the records hash is keyed on IP by default. That causes issues when you GET /dhcp/:network/:mac as the same IP might belong to two macs, and you'll only get one of them.

I considered altering the code to key on mac, but then you hit the same problem when you GET /dhcp/:network/:ip. So instead I re-index on the fly for mac-centric queries. It seems to work from my tests, but I'd appreciate confirmation from others.

lzap:
Ok taking for testing together with https://github.com/theforeman/foreman/pull/1430 in once.

GregSutcliffe:
@lzap new version just been pushed - it turns out I was relying on implicit Hash sorting in ruby 1.9 which breaks on 1.8, so I've added explicit sorting of the records, to make sure we keep them in the order they were listed in the file.

mburns72h:
I've verified that this works in my environment in combination with the above mentioned foreman patch ( theforeman/foreman#1430 )

GregSutcliffe:
@lzap @mburns72h New extra commit added to handle the failure when deleting hosts - see the referenced Redmine issue for an explanation of why this is necessary. Please re-test and confirm

lzap:
Finally had time to finish my new discovery PR testing setup and reproduced:

ERF12-4395 [ProxyAPI::ProxyException]: Unable to retrieve DHCP entry for 52:54:00:bf:fe:77 ([Net::LeaseConflict]: 52:54:00:bf:fe:77/192.168.100.14) for proxy https://foreman.virtual.lan:8443/dhcp

Applying patches. Sorry for the delay.

mburns72h:
tested the update and it works as expected, hosts delete correctly

lzap:
Got this when testing:

```
Started POST "/discovers" for 192.168.100.17 at 2014-05-15 21:27:54 +0200
Processing by DiscoversController#create as */*
  Parameters: {"ip"=>"192.168.100.17"}
Failed to import facts for Host::Discovered: super: no superclass method `importFacts' for #<Host::Discovered:0x00000008c75940>
  Rendered text template (0.0ms)
Completed 400 Bad Request in 2937ms (Views: 0.7ms | ActiveRecord: 1.8ms)
```

And when I click on Refresh Facts, I get:

```
Started GET "/discovers/52540028bbfa/refresh_facts" for 192.168.100.1 at 2014-05-15 21:29:26 +0200
Processing by DiscoversController#refresh_facts as HTML
  Parameters: {"id"=>"52540028bbfa"}
Operation FAILED: Could not get facts from proxy: the scheme http does not accept registry part: :8443 (or bad hostname?)
  Rendered common/500.html.erb within layouts/application (2.8ms)
  Rendered layouts/base.html.erb (1.0ms)
Completed 500 Internal Server Error in 16ms (Views: 4.8ms | ActiveRecord: 2.0ms)
```

lzap:
Hmmm so when I click on Refresh Facts, I get two requests:

```
==> /var/log/foreman/production.log <==


Started GET "/discovers/52540028bbfa/refresh_facts" for 192.168.100.1 at 2014-05-15 21:48:08 +0200
Processing by DiscoversController#refresh_facts as HTML
  Parameters: {"id"=>"52540028bbfa"}
retrieving facts from proxy on ip: 
Operation FAILED: Could not get facts from proxy: the scheme http does not accept registry part: :8443 (or bad hostname?)
  Rendered common/500.html.erb within layouts/application (2.7ms)
  Rendered layouts/base.html.erb (0.9ms)
Completed 500 Internal Server Error in 17ms (Views: 8.4ms | ActiveRecord: 1.5ms)

==> /var/log/httpd/foreman.virtual.lan_access_ssl.log <==
192.168.100.1 - - [15/May/2014:21:48:08 +0200] "GET /discovers/52540028bbfa/refresh_facts HTTP/1.1" 500 1417 "https://foreman.virtual.lan/discovers/52540028bbfa" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/34.0.1847.132 Safari/537.36"
192.168.100.1 - - [15/May/2014:21:48:09 +0200] "GET /assets/application-a9b9547202305b784ffdecf30aa59b20.js HTTP/1.1" 200 389485 "https://foreman.virtual.lan/discovers/52540028bbfa/refresh_facts" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/34.0.1847.132 Safari/537.36"

==> /var/log/foreman/production.log <==


Started POST "/discovers" for 192.168.100.17 at 2014-05-15 21:48:10 +0200
Processing by DiscoversController#create as */*
  Parameters: {"ip"=>"192.168.100.17"}
retrieving facts from proxy on ip: 192.168.100.17
Failed to import facts for Host::Discovered: super: no superclass method `importFacts' for #<Host::Discovered:0x000000088d0bf0>
  Rendered text template (0.0ms)
Completed 400 Bad Request in 2879ms (Views: 0.6ms | ActiveRecord: 1.1ms)

==> /var/log/httpd/foreman.virtual.lan_access_ssl.log <==
192.168.100.17 - - [15/May/2014:21:48:10 +0200] "POST /discovers HTTP/1.1" 400 127 "-" "-"
```

I noticed the first one is missing the IP address. The discovered host has no facts with it. Not sure if this is different bug. Will investigate tomorrow.

lzap:
Ok got it, I was testing with 1.2 (old) discovery.

I was able to reproduce the bug and then with both patches (this and https://github.com/theforeman/foreman/pull/1430) it was working. +1 Good catch, work, report. Thanks guys.

domcleal:
Looks sensible, but I'm concerned about the use of data structures inside the subnet and lack of tests for the first commit.

GregSutcliffe:
@domcleal updated for your comments, thanks for the style cleanup

GregSutcliffe:
@lzap can you retest?

lzap:
Ok testing now, please do not change the code anymore, only add more unit tests if needed :-)

lib/proxy/dhcp/subnet.rb
```diff
@@ -207,6 +211,28 @@ def leases
...
+        return has_mac?(record, type) if validate_mac(record)
+      rescue
+        nil
+      end
+      begin
+        return has_ip?(record, type) if validate_ip(record)
+      rescue
+        nil
+      end
+    end
```
lzap: Hmmm this could be in one block maybe? Or do I miss why is this in two?

lib/proxy/dhcp/subnet.rb
```diff
@@ -176,7 +180,7 @@ def unused_ip args = {}
     end
 
     def delete record
-      if @records.delete_if{|k,v| v == record}.nil?
+      if @records.delete_if{|r| r == record}.nil?
```
domcleal: `@records.delete(record)`

lib/proxy/dhcp/subnet.rb
GregSutcliffe: validate_mac and validate_ip return errors, so you need to rescue them individually, I think. Otherwise the raise in validate_mac will prevent the validate_ip check from running at all.

domcleal: <pre>
self.load unless loaded?
return has_mac?(record, type) if (validate_mac(record) rescue nil)
return has_ip?(record, type)  if (validate_ip(record) rescue nil)
</pre>

domcleal:
untested, looks fine

lib/proxy/dhcp/subnet.rb
GregSutcliffe: @records is an array, so @records.delete() will expct an integer, surely?

domcleal: http://www.ruby-doc.org/core-1.8.7/Array.html#method-i-delete

GregSutcliffe: note to self delete != delete_at

lib/proxy/dhcp/subnet.rb
GregSutcliffe: thats a lot neater, done.

lzap:
I've just tested this. Provisioning do work, no conflicts at all.

domcleal:
Merged as 2080b2e, ad2b465 for Foreman 1.5.1.  Thanks @GregSutcliffe.

PR #156, opened 2014-05-12
fixes #5675 - expect dns_key setting to be nil if not set

domcleal:
Regression from 3824d18, as the default for a setting changed from false to nil, so the check for SETTING.dns_key == false is no longer correct.

Untested....

### DIFF
--- a/lib/proxy/dns/nsupdate.rb
+++ b/lib/proxy/dns/nsupdate.rb
@@ -9,7 +9,7 @@ module Proxy::DNS
     attr_reader :resolver
 
     def initialize options = {}
-      raise "Unable to find Key file - check your dns_key settings" unless SETTINGS.dns_key == false or File.exists?(SETTINGS.dns_key)
+      raise "Unable to find Key file - check your dns_key settings" unless SETTINGS.dns_key.nil? or File.exists?(SETTINGS.dns_key)
       super(options)
     end
 
--- a/test/nsupdate_test.rb
+++ b/test/nsupdate_test.rb
@@ -1,0 +1,17 @@
+require 'test_helper'
+require 'proxy/dns'
+require 'proxy/dns/nsupdate'
+
+class NsupdateTest < Test::Unit::TestCase
+  def test_initialize_nsupdate_returns_no_error_with_missing_key_setting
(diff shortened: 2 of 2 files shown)


dmitri-d:
lgtm.

GregSutcliffe:
:+1: - @lzap?

lzap:
Who tested this, then? :-)

GregSutcliffe:
Tested, confirmed the fix. @domcleal you probably want a couple of tests to ensure we don't regress again - https://gist.github.com/GregSutcliffe/766c8ed3c401474aa72b worked for me, but on top of wb's plugin work, so it may need a slight alteration.

GregSutcliffe:
@lzap ping?

lzap:
I was waiting for the fix. @domcleal do you want me to merge instead?

domcleal:
No, it's waiting on me to integrate those tests.

GregSutcliffe:
Oops, my mistake, I thought I was on a different PR. Unping :)

domcleal:
I've added the test, though @witlessbird may prefer to take Greg's patch for his modularisation branch over what I've added here.

dmitri-d:
@domcleal, @GeregSutcliffe: ported the fix and tests here: https://github.com/witlessbird/smart-proxy/commit/034f4c715bbf80659841e142b23e76809f5e6fb5

GregSutcliffe:
I'm happy to keep this in the plugins branch until merge, it's not critical. You may want to fix your commit message though @witlessbird as it has "thanks Dominic & Greg" in it, and it shouldn't be squashed along with the rest of the branch when merged.

domcleal:
I would like this merged so I can include it in 1.5.1, not as part of the plugins work (it's separate).

@lzap?

lzap:
Yeah, this looks good.

GregSutcliffe:
@domcleal fine. @lzap can you merge? I don't have rights...

lzap:
Pushed as 62c7320 for Foreman 1.6 and 1.5.1. Thanks @domcleal

## 

Later,

 Lukas "lzap" Zapletal
 irc: lzap #theforeman

PR #158, opened 2014-05-14
Fixes #5712: Limit DHCP subnets for ISC if necessary

jimmidyson:
Implementation copied from 700e96b

### DIFF
--- a/config/settings.yml.example
+++ b/config/settings.yml.example
@@ -52,7 +52,7 @@
 #   - native_ms (Microsoft native implementation)
 #   - virsh (simple implementation for libvirt)
 :dhcp_vendor: isc
-# dhcp_subnets is a Native MS implementation setting. It restricts the subnets queried to a
+# dhcp_subnets is an ISC &  Native MS implementation setting. It restricts the subnets queried to a
 # subset, so as to reduce the query time.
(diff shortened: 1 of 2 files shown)


domcleal:
[test]

@lzap could you review please?

jimmidyson:
Not quite sure why this is failing as the errors are unrelated to the change made?

domcleal:
Sorry about that, we found out today that a new mocha release broke on older patch versions of Ruby 2.0.0.  Puppet's just finished updating our build servers (https://github.com/theforeman/foreman-infra/commit/69d5d33f6abe354fc298ebe41833b90235e8a2bf), so this time it should pass: [test]

jimmidyson:
Thanks @domcleal. We're running this patch in production & it's shaved over a minute off our provisioning time (before we had to increase the hard-coded timeout at https://github.com/theforeman/foreman/blob/develop/lib/proxy_api/resource.rb#L10, otherwise provisioning timed out).

lzap:
On it, I had the issue opened already :-)

lzap:
I can confirm it works for ISC DHCP, also when I give it empty list of subnets, it does not load them thus return IP addresses. +1

domcleal:
Thanks @jimmidyson, merged as cabb60b for Foreman 1.5.1.

PR #159, opened 2014-05-19
refs #5793 - add pkg:generate_source rake task to create tar.bz2

domcleal:
Will resubmit to all core projects soon, to produce a standard way of creating package "sources".


dmitri-d:
seems to be working, lgtm!

lzap:
it works fine, but can we make the lengthy `pkg:generate_source` as a default task so it will work with this too:

```
rake pkg
```

+1 nice

domcleal:
@lzap generally it's going to be for computers rather than us so I'm not too concerned about the length.. can do though.  I was trying to name it so it's unlikely to conflict with anything anybody else wants (e.g. foreman core already has pkg:deb).

lzap:
Hmmm that's okay then. But we sometimes release tarballs, not quite
often.

LZ

On Fri, May 23, 2014 at 05:20:36AM -0700, Dominic Cleal wrote:

> @lzap generally it's going to be for computers rather than us so I'm not too concerned about the length.. can do though.  I was trying to name it so it's unlikely to conflict with anything anybody else wants (e.g. foreman core already has pkg:deb).
> 
> ---
> 
> Reply to this email directly or view it on GitHub:
> https://github.com/theforeman/smart-proxy/pull/159#issuecomment-44002369

## 

Later,

 Lukas "lzap" Zapletal
 irc: lzap #theforeman

GregSutcliffe:
:+1: I'd merge but I don't have commit-bit - someone should do the honours :)

lzap:
Pushed as
- 9f019d1 refs #5793 - add pkg:generate_source rake task to create tar.bz2 <Dominic Cleal>

Thanks!

PR #162, opened 2014-05-29
refs #5987 - remove unused packaging files

domcleal:
https://github.com/theforeman/foreman-packaging/tree/rpm/develop/foreman-proxy


lzap:
Pushed as 0007366, thank you. Nice hash!

PR #168, opened 2014-06-20
fixes #4699 - support Puppet environments API for directory envs

domcleal:
First use of Puppet's API...
- https://github.com/puppetlabs/puppet/blob/master/api/docs/http_api_index.md
- https://github.com/puppetlabs/puppet/blob/master/api/docs/http_environments.md


domcleal:
A positive piece of feedback was received on the bug report for the rebased 1.5 version of this patch :)

test/puppet/api_request_test.rb
```diff
@@ -0,0 +1,47 @@
...
+    File.expand_path(File.join(File.dirname(__FILE__), '..', 'fixtures', 'authentication'))
+  end
+
+  def test_get_environments
+    Proxy::Puppet::Plugin::settings.stubs(:puppet_url).returns('http://localhost:8140')
+    stub_certs
+    stub_request(:get, 'http://localhost:8140/v2.0/environments').to_return(:body => '{"environments":{}}')
+    result = Proxy::Puppet::EnvironmentsApi.new.find_environments
+    assert_kind_of(Hash, result)
+  end
```
dmitri-d, line 29: I think it would be useful to verify more details about data that's coming back: there's no validation that data received  from puppet affects data returned by the method...

test/puppet/api_request_test.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  def test_ssl_config
+    Proxy::Puppet::Plugin::settings.stubs(:puppet_ssl_ca).returns(File.join(fixtures, 'puppet_ca.pem'))
+    Proxy::Puppet::Plugin::settings.stubs(:puppet_ssl_cert).returns(File.join(fixtures, 'foreman.example.com.cert'))
+    Proxy::Puppet::Plugin::settings.stubs(:puppet_ssl_key).returns(File.join(fixtures, 'foreman.example.com.key'))
+    stub_request(:get, 'https://puppet.example.com:8140/v2.0/environments')
+    Proxy::Puppet::EnvironmentsApi.new.find_environments
+  end
```
dmitri-d, line 37: If I'm reading this correctly, I think you need to use mocks instead of stubs, as the former will raise errors if not called...

dmitri-d:
lgtm. Also: nice tests! I wonder if we should create an abstraction for https connections: chef and puppet each create their own (not that it's a lot of code).

test/puppet/api_request_test.rb
domcleal, line 37: Thanks, fixed to use expects.

test/puppet/api_request_test.rb
domcleal, line 29: Fixed to compare the return value against the hash that the stub_request returned.

domcleal:
Yeah, I wondered about that, but since the (required) SSL, headers and URL details were different I found myself diverging from the Chef request class somewhat.  I was unsure whether an abstraction would be a hindrance or a help.

test/puppet/api_request_test.rb
```diff
@@ -0,0 +1,47 @@
+require 'test_helper'
+require 'puppet_proxy/puppet_plugin'
+require 'puppet_proxy/api_request'
+require 'webmock/test_unit'
```
lzap, line 4: Isn't this new dependency?

domcleal, line 4: It's already used in the Chef request tests, and is in bundler.d/test.rb.

domcleal:
Regarding the "Needs testing" label, there are a few successful users on http://projects.theforeman.org/issues/4699 so I'm fairly satisfied that it's working.

lzap:
Ah ok

lzap:
Merged as 16db7a9, thanks @domcleal for your contribution!

PR #170, opened 2014-06-23
fixes #6289: avoids concurrent downloads of files with the same name and  destination dir.

dmitri-d:


domcleal:
There seem to be some intermittent test failures, could you see if you can reproduce and debug?  It's a little concerning.

My only concern with this approach is that I don't think it'll work in a Passenger environment due to the class variables.  I don't know that it's a big problem though, since that's very much a minority deployment method.

dmitri-d:
@domcleal: duh, one test was interfering with another.

lib/proxy/http_downloads.rb
```diff
@@ -0,0 +1,45 @@
+module Proxy
+  class HttpDownloads
+    @@downloads_lock = Mutex.new
```
lzap, line 3: File lock might be better for those who run proxy in other stacks than webrick perhaps?

Finally :-) This was super annoying.

lzap:
Closed intentionally?

Since we spawn a process, I think flock can replace mutex, but not big deal. LGTM

dmitri-d:
@lzap, yeah, this one was superceded by https://github.com/theforeman/smart-proxy/pull/173

PR #171, opened 2014-06-23
fixes #6306: removed possible namespace collisions with puppet and chef

dmitri-d:


domcleal:
Merged as 69f6b3e, thanks @witlessbird!

PR #173, opened 2014-06-24
fixes #6289: avoids concurrent downloads of files with the same name and destination dir.

dmitri-d:


domcleal:
Seems to have a full run of 1.8.7 failures.

dmitri-d:
@domcleal: seems to be ok now. Could you verify that permissions and location of the lock file are ok?

lib/proxy/http_downloads.rb
```diff
@@ -0,0 +1,41 @@
...
+        lock = try_locking(dst)
+        unless lock.nil?
+          HttpDownload.new(src, dst, lock)
+          return true
+        end
+        false
+      end
+
+      def try_locking(afile)
+        f = File.open(File.expand_path(lock_filename(afile), Dir.tmpdir), File::RDWR|File::CREAT, 0600)
```
domcleal: Could the lock file simply be `afile + '.lock'` so it resides in the destination directory?

dmitri-d: @domcleal: sure, but then it will be accessible via tftp, not sure if that's a big deal though.

domcleal: I think that's fine, it's only an empty file.

domcleal: Another thought, maybe create it as a hidden file, e.g. `'.' + File.basename(afile)`.  httpd at least won't serve these by default, maybe TFTP daemons are the same?  (I'd have to check)

dmitri-d: looks like dot files are being served

dmitri-d:
@domcleal: ping?

lib/proxy/http_downloads.rb
```diff
@@ -0,0 +1,44 @@
...
+      def try_locking(afile)
+        f = File.open(lock_file(afile), File::RDWR|File::CREAT, 0600)
+        return f if f.flock(File::LOCK_EX | File::LOCK_NB) == 0
+        f.close
+        nil
+      end
+
+      def lock_file(afile)
+        dir, file = *(Pathname.new(afile).split)
+        lock = Digest::MD5.hexdigest(file.to_s) + '.lock'
```
domcleal: sorry, why an MD5 here?

dmitri-d: nothing like a bit of security(?)-by-obscurity...

lib/proxy/http_downloads.rb
```diff
@@ -0,0 +1,44 @@
...
+          HttpDownload.new(src, dst, lock)
+          return true
+        end
+        false
+      end
+
+      def try_locking(afile)
+        f = File.open(lock_file(afile), File::RDWR|File::CREAT, 0600)
+        return f if f.flock(File::LOCK_EX | File::LOCK_NB) == 0
+        f.close
```
domcleal, line 16: We seem to leak file handles here, if the lock is successful

dmitri-d:
@domcleal: you are indeed correct, /me didn't check file deletion semantics...

lib/proxy/http_downloads.rb
lzap: well it's now easy to find the spot for the cracker ;-)

lzap:
LGTM

I think we should use flocks on the files we are being downloaded for easier debugging etc. You can touch them if they don't exist yet and instruct wget to overwrite them, or to append via the `-c` option (which is I think better).

lzap:
There is still the issue when Anaconda tries to download the image while wget process starts with re-download. We might extend the lock until server returns from the build mode (this would require other changes on the foreman side). Out of scope I guess.

dmitri-d:
Not sure what you mean re: locks. The PR relies on flock, which is per process/per descriptor -- I don't see how I can pass it between processes.

lzap:
Not sure if you reply to the former or latter comment, but my idea is to call flock on the same files which are being downloaded (e.g. boot/Fedora-19-vmlinuz).

dmitri-d:
@lzap: the former; I'm not sure how flock would work on the file being downloaded:
- wget doesn't lock files (at least not that I can see)
- flocks are process/file-descriptor specific

lzap:
> @lzap: the former; I'm not sure how flock would work on the file being downloaded:
> - wget doesn't lock files (at least not that I can see)

Indeed, I mean this workflow:
- proxy creates the file (zero-sized)
- proxy creates the file lock
- wget is called with `-c` (continue) option
- proxy removes the lock when wget is done

It's for discussion, it looks maybe as an overengineering ;-) I can see
the benefit of having flock with the same file which is involved.

> - flocks are process/file-descriptor specific

AFAIK flocks do work across system processes, it's their main purpose
(to synchronize among them).

See flock(1) man page for example.

## 

Later,

 Lukas "lzap" Zapletal
 irc: lzap #theforeman

dmitri-d:
Current implementation is close to what you are describing, with the exception of creating of zero-sized file by smart-proxy. What's its purpose?

lzap:
> Current implementation is close to what you are describing, with the exception of creating of zero-sized file by smart-proxy. What's its purpose?

Yeah sorry for that, it's rather nitpicking

Instead of creating separate lock files, we could put exclusive locks on
those files directly. But to do that, you need to open them first. Thus
to create them before spawning the wget.

</nitpick>

## 

Later,

 Lukas "lzap" Zapletal
 irc: lzap #theforeman

domcleal:
It's a neater/tidier idea, using the same file and seems to work under Linux.

dmitri-d:
@domcleal, @lzap: updated to reuse the file being downloaded as the lock. I decided to truncate the file before using wget, as in some cases resuming interrupted download may result in corrupted file...

lib/proxy/http_downloads.rb
```diff
@@ -0,0 +1,51 @@
...
+          # Another instance where you'll get a garbled file if you try to use -c is if you have a lame HTTP proxy that inserts a "transfer interrupted" string into the local file.t
+          f.truncate(0)
+          return f
+        end
+        f.close
+        nil
+      end
+
+      def unlock(handle)
+        File.delete(handle.path)
```
GregSutcliffe: Given the lock file is now the actual file, will this not delete the file as soon as it's downloaded?

dmitri-d: We can't delete the lock file, as it's the same file as the one being downloaded/updated by wget. The gist is: we create an empty file, acquire a lock on it, start wget with a '-c' option and pass the file we just created as the download target. Once download is complete we remove the lock.

lib/proxy/http_downloads.rb
```diff
@@ -0,0 +1,51 @@
...
+        lock = try_locking(dst)
+        unless lock.nil?
+          HttpDownload.new(src, dst, lock)
+          return true
+        end
+        false
+      end
+
+      def try_locking(afile)
+        f = File.open(afile, File::RDWR|File::CREAT, 0600)
```
dmitri-d: I don't think '0600' is appropriate now, '0644' is probably more like it.

GregSutcliffe: +1, 644 is the way to go

dmitri-d:
pls. disregard the comment re: initial file truncation. This should do the trick for now.

I think though that this functionality should be rewritten without relying on wget -- currently I cannot determine if a process completed successfully, and therefore we have to rely on wget's resumed downloads feature, which has some drawbacks.

GregSutcliffe:
ACK

I've tested this locally, and works great - multiple downloads don't clobber, and existing files are left alone as per previous behaviour. I agree we want to re-write this in pure Ruby at some point, instead of shelling out, but it's still an improvement on the current behaviour.

[mergeable]

lib/proxy/util.rb
```diff
@@ -3,47 +3,38 @@
...
+        begin
+          Open3::popen3(cmd) do |stdin,stdout,stderr,thr|
+            # PIDs are not available under Ruby 1.8
+            pid = thr.nil? ? rand(9999) : thr.pid
+            stdout.each do |line|
+              logger.debug "[#{pid}] #{line}"
+            end
+            stderr.each do |line|
+              logger.error "[#{pid}] #{line}"
+            end
```
lzap, line 25: One more nitpick - I hate this way of logging (stdout -> debug, stderr -> err) because the way wget logs out ends up with mess (debug is hidden by default, and some ordinary output is usually written to error).

I'd vote for changing this in this PR to stdout-stderr -> debug. Or if you prefer, info.

lzap:
Nice work, finally this gets fixed (to some degree). And ability to use flock(1) tool during debugging or removing stale locks is also great. I will merge once you send your opinions about my last comments.

TBH I'd appreciate more visibility (thus stdout/stderr both in info log level), because this is still temporary implementation. In future, we really want to use pure Ruby and also track progress in the UI. For this reason, I prefer info to debug (which is turned off by default).

dmitri-d:
@lzap, @domcleal: ping

lzap:
Pushed as d44cdd9 thanks @witlessbird

PR #176, opened 2014-06-30
fixes 6436: plugins are now being loaded after core modules

dmitri-d:


domcleal:
Merged as e42766e, thanks @witlessbird!

PR #177, opened 2014-07-02
fixes #6474: smart-proxy relies on Bundler to load sinatra dependency

dmitri-d:


domcleal:
Thanks @witlessbird for the quick fix.  Merged as 9ac3e80.

PR #178, opened 2014-07-04
Use Enumerable#find instead of each where possible

sodabrew:
Following up from #175 I found two more places where `.each` is better served by `.find`.


ohadlevy:
[test]

domcleal:
Merged as aa47858 for Foreman 1.6.0, thanks @sodabrew!

PR #179, opened 2014-07-04
Replace Kernel#open with File.open

sodabrew:
Kernel#open (the global `open` method) can be used to open URLs and other interesting things. We don't want that, just the `File`s please! See http://stackoverflow.com/a/1728941/3483097 for a nice writeup.


ohadlevy:
[test]

domcleal:
Merged as a220e9a for 1.6.0, cheers @sodabrew.

PR #181, opened 2014-07-07
Update the tests for Proxy::Util::CommandTask

sodabrew:
The `wget` test was hanging for me - I'm running OS X, and `127.0.0.2` is not a default alias of localhost. I couldn't figure out why the test was using `wget` and neither could the FIXME author. So in this update, use `true` and `false` to test $? return codes 0 and 1, respectively.

That said - Ruby 1.8.7 is heading out the door. If Proxy::Util::CommandTask cannot capture $? on current versions of Ruby, that's problem!

### DIFF
--- a/lib/proxy/util.rb
+++ b/lib/proxy/util.rb
@@ -14,17 +14,21 @@ module Proxy::Util
       logger.debug "Starting task: #{cmd}"
       @task = Thread.new(cmd) do |cmd|
         begin
+          status = nil
           Open3::popen3(cmd) do |stdin,stdout,stderr,thr|
             # PIDs are not available under Ruby 1.8
-            pid = thr.nil? ? rand(9999) : thr.pid
+            pid = thr.nil? ? '(unknown)' : thr.pid
             stdout.each do |line|
               logger.debug "[#{pid}] #{line}"
             end
             stderr.each do |line|
               logger.debug "[#{pid}] #{line}"
             end
+            # In Ruby 1.8, popen3 always reports an error code of 0 in $?.
+            # In Ruby >= 1.9, call thr.value to wait for a Process::Status object.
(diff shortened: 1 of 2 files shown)


sodabrew:
Second commit should return subprocess exit status in all Rubies. I can squash the commits if this looks good.

domcleal:
I've added you to our GitHub org, so if you set membership to public (https://github.com/orgs/theforeman/members) then Jenkins will test your PRs automatically.  Thanks for the contributions!

[test]

sodabrew:
Sadly, Ruby 1.8 always returns 0 for popen3 - that's what the bug report link was about it just took me three times reading it to grok - but at least I've fixed status handling for newer Rubies!

Cool thanks! Let's try this!

[test]

lib/proxy/util.rb
```diff
@@ -20,17 +20,21 @@ def initialize(cmd)
...
           end
           stderr.each do |line|
             logger.error "[#{pid}] #{line}"
           end
+          # In Ruby 1.8, popen3 always reports an error code of 0 in $?.
+          # In Ruby >= 1.9, call thr.value to wait for a Process::Status object.
+          status = thr.value unless thr.nil?
         end
-        $?
+        status ? status.exitstatus : $?
```
lzap: Beware, AFAIK existstatus is not set until you close the file handle.

lzap: Oh I've misread the indentation, you are safe :-)

lzap:
Sorry you will need to rebase :-( I've just merged Dmitri's patch which refactored this completely. http://git.io/zoPmbQ

sodabrew:
Rebased!

domcleal:
Merged as bbafbc0 for 1.6.0, thanks again @sodabrew!

PR #182, opened 2014-07-08
fixes #6522 - move request code out of chef module

mmilata:
The patch also removes foreman_ssl_ca, foreman_ssl_cert and
foreman_ssl_key config options as they don't appear to be used or
documented. Existing ssl_\* options are used instead.

(Please let me know if you'd prefer to track this change in separate ticket.)

### DIFF
--- a/lib/proxy/request.rb
+++ b/lib/proxy/request.rb
@@ -3,9 +3,8 @@ require 'net/https'
 require 'uri'
 
 # TODO: need settings validation on startup, otherwise we get a 500 error due to missing/wrong config settings when api is accessed
-# TODO: shouldn't SSL settings use ssl_certificate, ssl_ca_file, and ssl_private_key as opposed to foreman_ssl_ca, foreman_ssl_cert, and foreman_ssl_key?
 
-module Proxy::Chef
+module Proxy::HttpRequest
   class ForemanRequest
     def send_request(path, body)
       uri              = URI.parse(Proxy::SETTINGS.foreman_url.to_s)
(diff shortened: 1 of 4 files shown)


domcleal:
The foreman_\* settings _were_ used by the request class, I don't follow why you say they weren't?

They're particularly important IMO because the CA, cert and key that are used to contact are not necessarily the same as the key that the proxy is running under.  This would be typical if Foreman's web interface was running with a cert signed by a local CA, while the proxy is using Puppet certs.

[test]

mmilata:
I had assumed that since the options are not documented in the example config and googling them did not give any result then nobody is probably using them. Perhaps that's a wrong thing to assume.

> This would be typical if Foreman's web interface was running with a cert signed by a local CA, while the proxy is using Puppet certs.

Is that even possible? Both sides are verified in https requests from Foreman to proxy, wouldn't the verification fail if Foreman's certificate was signed by local CA but the proxy tried to verify it using the Puppet CA?

Or is it possible for Foreman to have different set of certificates for the web interface and communication with proxy?

mmilata:
Perhaps the following setup would work:

```
Foreman:
certificate: signed by local CA
CA certificate: puppet CA

Proxy:
certificate: signed by Puppet CA
CA certificate: local CA
```

but that would work even for the requests initiated by the proxy. What am I missing?

domcleal:
The number of foreman_chef users is probably low at the moment, so I wouldn't expect to find much.

Using different sets of certificates is entirely possible.  The web interface's certs are defined by httpd and mod_ssl typically, while the certs used for proxy communications are just for outgoing connections via rest-client and are defined under Adminster>Settings.

mmilata:
I see. What do you think about falling back to the `ssl_*` options in case the `foreman_ssl_*` options are not present in the configuration?

domcleal:
I think that's fine, but would suggest adding foreman_ssl_\* to the example configs at least so they're more visible.

mmilata:
I have updated the branch. Please review.

domcleal:
[test] Looks good to me, @witlessbird?

lib/proxy/request.rb
```diff
@@ -3,9 +3,8 @@
 require 'uri'
 
 # TODO: need settings validation on startup, otherwise we get a 500 error due to missing/wrong config settings when api is accessed
-# TODO: shouldn't SSL settings use ssl_certificate, ssl_ca_file, and ssl_private_key as opposed to foreman_ssl_ca, foreman_ssl_cert, and foreman_ssl_key?
 
-module Proxy::Chef
+module Proxy::Request
```
dmitri-d: Perhaps call it HttpRequest or HttpsRequest?

dmitri-d:
Looks good to me!

mmilata:
@witlessbird I've renamed the module.

domcleal:
[test]

domcleal:
Thanks @mmilata!  Merged as 6108539 for Foreman 1.6.

PR #183, opened 2014-07-08
added support for testing of plugins

dmitri-d:
The idea is to add 'smart_proxy' as a development dependency for a plugin, and then `require 'smart_proxy_for_testing'` in tests. gemspec may need a bit more love, esp. around dependencies though.

### DIFF
--- a/lib/smart_proxy_for_testing.rb
+++ b/lib/smart_proxy_for_testing.rb
@@ -1,0 +1,20 @@
+require 'proxy'
+require 'checks'
+require 'proxy/log'
+require 'proxy/settings'
+require 'proxy/settings/plugin'
+require 'proxy/settings/global'
+require 'proxy/util'
+require 'proxy/http_downloads'
+require 'proxy/helpers'
+require 'proxy/plugin'
+require 'proxy/error'
+
+Proxy::SETTINGS = ::Proxy::Settings::Global.new(:log_file => './logs/test.log', :log_level => 'DEBUG')
+Proxy::VERSION = File.read(File.join(File.dirname(__FILE__), '../VERSION')).chomp
+
+class ::Proxy::Plugin
+  def self.load_test_settings(a_hash)
+    @settings = ::Proxy::Settings::Plugin.new(plugin_default_settings, a_hash)
+  end
+end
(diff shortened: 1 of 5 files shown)


test/test_helper.rb
```diff
@@ -5,27 +5,7 @@
...
 
-Proxy::SETTINGS = Proxy::Settings.load_global_settings(File.expand_path("fixtures/test_settings.yml", File.dirname(__FILE__)))
-Proxy::VERSION = File.read(File.join(File.dirname(__FILE__), '../VERSION')).chomp
-
-class ::Proxy::Plugin
-  def self.load_test_settings(a_hash)
-    @settings = ::Proxy::Settings::Plugin.new(plugin_default_settings, a_hash)
-  end
-end
+require 'smart_proxy_for_testing'
```
lzap, line 11: Nit pick - vim editor will fix the missing EOL.

lzap:
LGTM, I haven't tested.

domcleal:
For Foreman core we eventually settled on doing the opposite, running all tests from within the main project.  Is it possible to do it that way around?

dmitri-d:
@domcleal: that's probably mostly due to our chosen approach being "the rails way"... I thought that something along the lines of what I did would be more traditional/developer-friendly: we are not mandating on how to run or execute tests. Also, plugin developers have access to core modules from tests should they need to.

lib/smart_proxy_for_testing.rb
```diff
@@ -0,0 +1,20 @@
...
+require 'proxy/settings/plugin'
+require 'proxy/settings/global'
+require 'proxy/util'
+require 'proxy/http_downloads'
+require 'proxy/helpers'
+require 'proxy/plugin'
+require 'proxy/error'
+
+Proxy::SETTINGS = ::Proxy::Settings::Global.new(:log_file => './logs/test.log', :log_level => 'DEBUG')
+Proxy::VERSION = File.read(File.join(File.dirname(__FILE__), '../VERSION')).chomp
```
mmilata, line 14: The `VERSION` file is not included in the gem.

mmilata:
Tested it with `foreman_proxy_abrt` plugin and besides the missing file it works fine.

dmitri-d:
@mmilata: thanks for testing; added VERSION file to the gemspec.

domcleal:
Ok, thanks.

Do you have a ticket #?

dmitri-d:
@mmilata: I started docs on plugins (including development) if you are curious: http://projects.theforeman.org/projects/foreman/wiki/Smart-Proxy_Plugins

dmitri-d:
@domcleal: ticket: http://projects.theforeman.org/issues/6772

domcleal:
Thanks @witlessbird, merged as dc635bf.

PR #185, opened 2014-07-10
DHCP API tests

sodabrew:
Following up from https://github.com/theforeman/smart-proxy/pull/175#issuecomment-47989979 this begins tests for the DHCP API.

I'll also work on tests for ISC subnet parsing, probably a separate PR.


test/dhcp/dhcp_api_test.rb
```diff
@@ -0,0 +1,45 @@
...
+  def app
+    Proxy::DhcpApi.new
+  end
+
+  def setup
+    Proxy::DhcpPlugin.load_test_settings(
+      :enabled => true,
+      :dhcp_vendor => 'isc',
+      :dhcp_config => './test/dhcp/dhcp.conf',
+      :dhcp_leases => './test/dhcp/dhcp.leases',
```
sodabrew: Or come to think of it, these are empty files right now. Instead of stubbing a subnet instance, could I put a subnet into the configs?

lzap: Yes, put them into ./test/fixtures - maybe create a subdirectory there.

lzap:
It failed in our Jenkins.

domcleal:
When you get it working, please just use "refs #6412" in the commit message so it gets linked together.  Thanks!

sodabrew:
Actually, I don't understand how I'm causing this failure, any ideas how this unrelated thing is related?

```
[210/223] RunnerTest#test_popen = 0.00 s
  2) Failure:
test_popen(RunnerTest) [smart-proxy/test/puppet/runner_test.rb:11]:
<0> expected but was
<2>.
```

domcleal:
That's really strange, it happens reliably here too, though with an exit code of '1'.

My best guess is that the test is broken, since Proxy::Puppet::Runner#popen isn't designed to wait for the exit code, that's what the shell_command method's meant to do.  I think $? might actually be from some other command... maybe the new test is inadvertently executing a command and changing $? before the (broken?) runner test is run.

sodabrew:
It's the exit status of the ping command from the /:subnet/unused_ip query
in my test! On Mac OS X, ping returns 2. I bet the standard Linux ping
returns 1 in this case:

``` text
EXIT STATUS
     The ping utility exits with one of the following values:
     0       At least one response was heard from the specified host.
     2       The transmission was successful but no responses were received.
```

sodabrew:
My commit 4820557 has a revamped Puppet::Runner test..

Edit: Huh, Ruby 1.8.7 doesn't return exit status from popen either?
Edit again: nope, I passed a string where an array was expected. I will break this out to a separate PR.

sodabrew:
Here's a neat failure: `get /192.168.122.0` returns inconsistent time format in the leases section:
`"starts"=>"Sat Jul 12 10:08:29 UTC 2014"` vs. `"starts"=>"2014-07-12 10:08:29 UTC"`

Edit: actually not that interesting, Ruby 1.8.7 vs. 1.9.3+ behavior of Time.to_s.

sodabrew:
There is one real open issue now, per 11d45f9, the Record.compare_options is looking at too many keys and failing to match, for example:

``` ruby
record parsed from config file:
{:mac=>"00:11:bb:cc:dd:ee", :subnet=>192.168.122.0/255.255.255.0, :ip=>"192.168.122.1", :hostname=>"test.example.com"}
record processed from the the API:
{"hostname"=>"test.example.com", "ip"=>"192.168.122.1", "mac"=>"00:aa:bb:cc:dd:ee", "splat"=>[], "captures"=>["192.168.122.0"]}
```

Edit: I want to emphasize that this bug has been causing me no end of pain in my deployment environment; I have to keep juggling whether a DHCP reservation is listed in static configs or in Foreman, but it cannot be both. Turns out Foreman should have been quietly not-erroring when I add new hosts to Foreman with pre-existing identical DHCP reservations! _argh_

sodabrew:
In 898fbbde9e7aeb1c5acca2a37a31e1f833e00f76 I am cleaning the args passed to `addRecord` before matching against an existing record. This fixes the ability to match and ignore pre-existing identical records.

Please note that you're not getting a "real" hash at Server.addRecord - it's Sinatra's hash with indifferent access: https://github.com/sinatra/sinatra/blob/master/lib/sinatra/base.rb#L1038

The argument should really be named params, to indicate that it's coming from over the wire, or the dhcp_api module should have the cleanup code and pass a clean options hash into Server.addRecord.

sodabrew:
Help me out, I am not understanding something. This post returns success, but the record isn't showing up anywhere? I don't see the record in the output from `get /192.168.122.0` after the successful `post`:

``` ruby
  def test_api_create_record_new
    params = {
      "hostname" => "ten.example.com",
      "ip"       => "192.168.122.10",
      "mac"      => "10:10:10:10:10:10",
      "network"  => "192.168.122.0/255.255.255.0",
      # TODO: test some optional vendor options here
    }
    post "/192.168.122.0", params
    assert last_response.ok?, "Last response was not ok: #{last_response.status} #{last_response.body}"
    expected = ''
    assert_equal expected, last_response.body

    # Repeat the load subnet test, make sure the new entry is in there
    get "/192.168.122.0"
    puts last_response.body

    # HUH WHY ISN'T THE NEW RECORD HERE!?
```

lzap:
Hmmm I might be shooting in the dark, but don't you use fixtured configuration? Isn't the setup method called in between the post and the get call? I have to admit that would be nonstandard, xUnit run setup between tests and not calls, but try to debug that.

sodabrew:
I had the same thought about the config being reset - so I put a `get /192.168.122.0` into the same test as the post call, and the entry is not there: 9e75c656df429cce2b52f481824e264cf63f4e31

test/dhcp/dhcp_api_test.rb
```diff
@@ -0,0 +1,184 @@
...
+      "network"  => "192.168.122.0/255.255.255.0",
+      # TODO: test some optional vendor options here
+    }
+    post "/192.168.122.0", params
+    assert last_response.ok?, "Last response was not ok: #{last_response.status} #{last_response.body}"
+    expected = ''
+    assert_equal expected, last_response.body
+
+    # We stubbed out the real omcmd, so fake it:
+    File.open("test/tmp/dhcpd_extra_leases.tmp", File::RDWR|File::CREAT) do |f|
```
domcleal: Should be "dhcp_extra_leases.tmp" (not "dhcpd") to match the include in dhcp.leases

test/test_helper.rb
```diff
@@ -24,6 +24,11 @@
 Proxy::SETTINGS = Proxy::Settings.load_global_settings(File.expand_path("fixtures/test_settings.yml", File.dirname(__FILE__)))
 Proxy::VERSION = File.read(File.join(File.dirname(__FILE__), '../VERSION')).chomp
 
+# Make sure that tests put their temp files in a controlled location
+# Clear temp file before each test run
+ENV['TMPDIR'] = 'test/tmp'
+%x{rm test/tmp/*.tmp}
```
domcleal: Nitpick, this prints a line to stderr on each test invocation if there aren't any files, maybe redirect it or use Ruby file builtins

domcleal:
The temp file has state between tests, which isn't great, can we reset it between individual tests and set up state if needed (e.g. for the deletion test)?  It might cause headaches for people running individual tests, or if minitest decides to run them in a different order one day.

sodabrew:
Thanks for merging #187, I'll rebase this now. Yes, the temp file has some state between tests, and there is some assumption that the tests run in order. I could merge related tests into larger test methods if that works better for you?

sodabrew:
Turns out, minitest already runs tests in random order! I added a sequence number to each test, and set minitest to run alphabetically. That resolved one set of test failures.

Good catch on the mismatched dhcp_leases file names. That resolved the other test failure.

Next steps: I will squash and reorganize the commits later today.

sodabrew:
Squashed and reorged the commits. Ready for review please!

lzap:
Remember to push :-)

sodabrew:
I did. The commits above are squashed (there are 7 commits, each with a distinct area of work) and reorged (I changed the ordering of the commits to avoid interdependence).

sodabrew:
I rebased and squashed down to 4 commits, and added Redmine tickets for each.

modules/dhcp/dhcp/providers/server/isc.rb
```diff
@@ -230,21 +224,29 @@ def filter_log log
...
 
-    def read_config file
+    def read_config file, ignore_includes=false
       logger.debug "Reading config file #{file}"
       config = []
       File.readlines(file).each do |line|
-        if /^include\s+"(.*)"\s*;/ =~ line.strip
+
+        line.strip! # remove left and right whitespace
+        next if line[0,1] == "#" # remove comments
```
dmitri-d: line.start_with("#") is a bit better...

modules/dhcp/dhcp/providers/server/isc.rb
```diff
@@ -230,21 +224,29 @@ def filter_log log
...
+        next if line.length == 0 # remove blank lines
```
dmitri-d: line.empty?

sodabrew:
I made the .gitignore / .gitkeep changes and line.start_with? and line.empty? changes.

sodabrew:
Rebased and resolved changes by dc635bf to `test/test_helper.rb`

sodabrew:
Hey, uh tests seem to be broken:

``` shell
Don't know how to build task 'jenkins:unit'
```

Which is weird, because I clearly see that Rakefile includes tasks/jenkins.rake and that defines namespace :jenkins, task :unit.

Edit: Ah, but there's a silent rescue LoadError in jenkins.rake. This must be failing:

``` ruby
   testtool = RUBY_VERSION =~ /^1\.8/ ? 'test_unit' : 'minitest'
   require "ci/reporter/rake/#{testtool}"
```

lzap:
For testers interested: http://koji.katello.org/koji/taskinfo?taskID=132591

lzap:
I can confirm this works fine, once @domcleal confirms, I can merge. [test]

lzap:
Jenkins should be fixed now, so re- [test]

sodabrew:
Rebased to pick up the CI fix in 0e616f1.

There's a conversation thread on https://github.com/sodabrew/smart-proxy/commit/a5e43bb8670d4d9c3028490aa1b733ceeab8b730#commitcomment-7171435 that isn't getting auto-linked here. [In this thread the only open issue about the PR at the moment? I want to make sure I have the code in its best possible state.]

sodabrew:
Ping - I'd really like to land this in time for 1.6 please!

domcleal:
Thanks for the reminder, it looks good!  I've responded to the discussion you linked to, that's the only outstanding bit IMO.

lzap:
Since this is now tested and reviewed, I suggest to merge this as is and ask @sodabrew to refactor the method a bit to be more clean.

domcleal:
I'm happy to merge the other three commits now if @sodabrew wants, but since the commit in question is a refactor anyway, I'd say we sort it now.

sodabrew:
Thanks! Please review 7512acd - if it resolves the modified argument issue then I'll squash it in.

domcleal:
Looks good to me!

domcleal:
#194 will fix the test failures

sodabrew:
I missed a case anyhow :) I yanked the Chef test fix off this branch.

lzap:
:+1: Thanks for this contribution!

domcleal:
[test]

domcleal:
Thanks @sodabrew!  Merged as 247c5b5..3925ae4 for Foreman 1.6.

PR #186, opened 2014-07-11
fixes #6588 - don't load user settings when testing default settings

domcleal:


dmitri-d:
Oops. Thanks for fixing this, lgtm.

lzap:
These were pretty stable tests :-)

Merged as a52a4f6 thanks Dom and Dmitri.

PR #188, opened 2014-07-17
fixes #6661: trailing slash is not required in urls

dmitri-d:


ohadlevy:
merged as 9933928 thanks @witlessbird

PR #172, opened 2014-06-23
Add support for 64-bit MAC addresses, fixes #6341

treydock:
Expands validation of MAC addresses to include 64bit MAC addresses.  This is related to theforeman/foreman#1470.

PR #189, opened 2014-07-21
Fixes #6689 - Allow nil mac address for TFTP API calls that do not require a MAC

treydock:
This fixes a regression from #172 which broke TFTP API calls that did not require a MAC address, such as `/create_default`.


domcleal:
Merged as bdc3b51 for Foreman 1.6.0.  Thanks for the fix @treydock!

PR #190, opened 2014-07-25
Fixes #6790: Restrict ci_reporter gem to less than 2.0.0 to fix CI.

sodabrew:
Same as https://github.com/theforeman/foreman/blob/ec0912fc73a276a988128e987203ad5117b97998

### DIFF
--- a/bundler.d/test.rb
+++ b/bundler.d/test.rb
@@ -3,7 +3,7 @@ group :test do
   gem 'rake'
   gem 'rack-test'
   gem 'single_test'
-  gem 'ci_reporter', '>= 1.6.3'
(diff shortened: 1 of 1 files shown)


sodabrew:
Ping @eLobato since you made the Foreman change that I'm cribbing here :)

dLobatog:
Sorry, I think only @domcleal @ohadlevy @lzap have merge access here. Thanks @sodabrew !

dLobatog:
Merged as https://github.com/theforeman/smart-proxy/commit/0e616f1b0be11f51f19d08e407fa987b3ea5abc7 , thanks @sodabrew !

Apparently I had merge access but I didn't know! I changed your commit to point to [6790](http://projects.theforeman.org/issues/6790)

PR #191, opened 2014-07-27
implements #6788 puppet scanner filter support

endyman:
implements #6788 puppet scanner filter support (kudos to Daniel Bäurer, too)

With

```
:puppet_include_modules: "role_*"
```

only modules following a certain naming schema will be included, the module path will be assembled like this:

$MODULEPATH/puppet_include_modules/manifests/*.pp

With

```
:puppet_ignore_modulefolder:
 - services
 - modules
```

directories in the Modulepath will be excluded:

if $MODULEPATH contains services or modules remove it from list to scan for puppet
modules

### DIFF
--- a/config/settings.d/puppet.yml.example
+++ b/config/settings.d/puppet.yml.example
@@ -37,3 +37,13 @@
 # Override use of Puppet's API to list environments, by default it will use only if
 # environmentpath is given in puppet.conf, else will look for environments in puppet.conf
 #:puppet_use_environment_api: true
+
+# Filter puppet class import and include only modules matching the
+# puppet_include_modules expression
+#:puppet_include_modules: "role_*"
+# Filter puppet class import and exclude directories from module path listed
+# in puppet_ignore_modulefolder.
+#:puppet_ignore_modulefolder:
+# - services
+# - modules
+
--- a/modules/puppet_proxy/class_scanner.rb
+++ b/modules/puppet_proxy/class_scanner.rb
@@ -9,9 +9,17 @@ module Proxy::Puppet
 
         parser = Puppet::Parser::Parser.new Puppet::Node::Environment.new
 
-        Dir.glob("#{directory}/*/manifests/**/*.pp").map do |filename|
-          scan_manifest File.read(filename), parser, filename
-        end.compact.flatten
+        # include modules by their name_schema
+        if Proxy::Puppet::Plugin.settings.puppet_include_modules
+          Dir.glob("#{directory}/#{Proxy::Puppet::Plugin.settings.puppet_include_modules}/manifests/**/*.pp").map do |filename|
+            scan_manifest File.read(filename), parser, filename
+          end.compact.flatten
+        else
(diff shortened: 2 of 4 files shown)


arnisoph:
+1

works great.

config/settings.d/puppet.yml.example
```diff
@@ -37,3 +37,12 @@
...
 #:puppet_use_environment_api: true
+
+# Filter puppet class import and include only modules matching the
+# puppet_include_modules expression
+#:puppet_include_modules: "role_*"
+# Filter puppet class import and exclude directories from module path listed
+# in puppet_ignore_modulefolder.
+#:puppet_ignore_modulefolder:
+# - services
+# - modules
```
arnisoph: a newline here would be nice

timogoebel:
tested, works flawlessly and improves the overall importing experience a lot.

ohadlevy:
nice, btw: are you aware we have a similar concept in foreman itself? see config/ignored_environments under http://theforeman.org/manuals/1.5/index.html#4.2.2Classes

endyman:
I thought that the ignored environments only filter post scan - so the idea was to filter earlier and on a module (not class) basis.

Danie:
It seams that puppet_ignore_modulefolder is missing in PR? Have a look at patch file "lib/proxy/puppet/environment.rb".

config/settings.d/puppet.yml.example
dmitri-d: Doesn't look like puppet_ignore_modulefolder setting is being used...

dmitri-d:
@endyman: could you add tests pls?

config/settings.d/puppet.yml.example
endyman: thanks for the notice  I just added the missing code

dmitri-d:
On the second thought I think it might be better to implement these filters as filters on the api. @domcleal, @ohadlevy: what do you think?

domcleal:
Maybe, and then we can use the ignored_environments config on the Foreman server passed through to the proxy?

dmitri-d:
@domcleal: nods

endyman:
By filter on the API you mean passing a filter to the proxy ?
Because one of the key features of this patch is to prevent scanning of classes - The ignored environments just filters post scan on the foreman side.

Is there already something like filters in use elsewhere on the smart-proxy rest api where I can take a look? 

From my understanding we would use the foreman setting instead of the smart-proxy setting and pass it to the smart proxy but still do the filtering in the scanner right ?

endyman:
http://projects.theforeman.org/issues/8210 implements a more generic puppet scanner cache which addresses scan times. The filtering aspect of this PR can then be implemented using the ignored_environments.yml. So from my perspective this PR would now result in redundant functionality.

The Advantage is, that general scantimes should be better using the cache implementation

The Drawback is that the initial import takes longer and all classes are returned to foreman - so some overhead on the wire.

@timogoebel  @bechtoldt  what do you guys think?

timogoebel:
I agree, @endyman. The cache looks very promising and should lead to similar results.

PR #195, opened 2014-08-05
fixes #6589 - add trusted_hosts check back

domcleal:


dmitri-d:
Could be more of a personal style thing (I do prefer composition over inheritance) -- I think having a helper instead of a common base class might be useful, as trusted hosts enforcement becomes a matter of inclusion of the helper...

dmitri-d:
[test]

domcleal:
I agree I think.  I can imagine situations where it may not be wanted, and composition will help.  My only worry is people forgetting to include it, which I hope we can avoid.  I've updated this to use a DSL extension (http://www.sinatrarb.com/extensions.html).

modules/bmc/bmc_api.rb
```diff
@@ -1,6 +1,10 @@
+require 'sinatra/trusted_hosts'
+
 module Proxy::BMC
   class Api < ::Sinatra::Base
     helpers ::Proxy::Helpers
+    register ::Sinatra::TrustedHosts
```
dmitri-d: instead of registering the extension with each parent class, you can do it directly with the base (probably around https://github.com/theforeman/smart-proxy/blob/develop/lib/smart_proxy.rb#L39):

<pre>
::Sinatra::Base.register ::Sinatra::TrustedHosts
</pre>

dmitri-d:
tested, lgtm with the exception of that one nitpick.

modules/bmc/bmc_api.rb
domcleal: I've tried this, but the only difficulty then is tests, as the extensions aren't registered.  I could also register them in test_helper.rb - do you think that's better?

dmitri-d: There's a test-specific version of smart-proxy (https://github.com/theforeman/smart-proxy/blob/develop/lib/smart_proxy_for_testing.rb), you could register the extension there?

domcleal: Indeed, I was just unsure whether it's worse to have code duplicated between smart_proxy (untested, but production) + smart_proxy_for_testing, or across the API classes (which at least we know is always tested).

dmitri-d: Hmm, there's that. We could replace smart_proxy_for_testing with something like smart_proxy_common and use that in both production and tests (Proxy::SETTINGS can be defined in test_helper).

domcleal: Looking at it more, I don't want to overcomplicate it.  At least it'll fail-fast if somebody did remove the trusted hosts stuff from smart_proxy.rb.  PR's updated with smart_proxy_for_testing.

dmitri-d:
merged, thanks @domcleal!

PR #197, opened 2014-08-11
fixes #7015 - to_json passes options through correctly, remove array wrapping

domcleal:
The update of ffi-yajl triggers this as it starts checking for something in the options hash, but the bug is in our code which had `{}` passed in as options, but then passed `[{}]` into the next to_json call.

This is a test blocker, again.


domcleal:
http://ci.theforeman.org/job/test_proxy_develop_pull_request/441/

GregSutcliffe:
Merged as d710548, thanks!

PR #199, opened 2014-08-13
fixes #7062: facts module uses correct path for 'requires'

dmitri-d:


modules/facts/facts.rb
```diff
@@ -1,7 +1,7 @@
 include ::Proxy::Log
 begin
   require "facter"
-  require 'facts_plugin'
+  require 'facts/facts_plugin'
```
domcleal: Maybe it's better to record if loading 'facter' was successful, and only require facts_plugin if it was known to be OK, so we don't mask errors in the loading of facts_plugin.

dmitri-d: yeah, good point.

dmitri-d:
@domcleal: ping?

lzap:
I am still testing this.

lzap:
And it works in discovery image. :+1: Thanks for prompt fix @witlessbird !

lzap:
Any chance getting this merged so I have it in tomorrows nightly?

dmitri-d:
merged in at 4bfe725

PR #201, opened 2014-08-15
Refs #4464 - Prefer selinux wrappers

lzap:
This is needed to start "puppet ca" command in the correct puppetca_t domain.

RHEL6 has `/usr/bin/puppetca`

RHEL7 currently has no such wrapper and I will be pushing fix to EPEL7 and
Fedora with wrapper (they are named start-puppet-xyz so I will follow this
convention)

Puppet Enterprise is a question mark for now since there are no wrappers - no
SELinux support?

### DIFF
--- a/modules/puppetca/puppetca_main.rb
+++ b/modules/puppetca/puppetca_main.rb
@@ -102,7 +102,8 @@ module Proxy::PuppetCa
         raise "SSL/CA unavailable on this machine"
       end
 
-      default_path = ["/opt/puppet/bin", "/opt/puppet/sbin"]
+      # prefer selinux wrappers or enterprise binaries
+      default_path = ["/usr/bin/start-puppet-ca", "/usr/bin/puppetca", "/opt/puppet/bin", "/opt/puppet/sbin"]
 
       # puppetca is the old method of using puppet cert which is new in puppet 2.6
       if Puppet::PUPPETVERSION.to_i < 3


modules/puppetca/puppetca_main.rb
```diff
@@ -102,7 +102,8 @@ def find_puppetca
         raise "SSL/CA unavailable on this machine"
       end
 
-      default_path = ["/opt/puppet/bin", "/opt/puppet/sbin"]
+      # prefer selinux wrappers or enterprise binaries
+      default_path = ["/usr/bin/start-puppet-ca", "/usr/bin/puppetca", "/opt/puppet/bin", "/opt/puppet/sbin"]
```
dmitri-d, line 106: the new items in the default_path are executables, correct? If so, this won't work, as 'which' expects a list of directories as the second parameter.

lzap:
We might not need that, do not merge... Working on other solution.

modules/puppetca/puppetca_main.rb
lzap, line 106: Right, I need to fix this.

lzap:
Okay I will rewrite this. We need this for both RHEL6 and RHEL7, wrappers are being added into EPEL6 and EPEL7. I am working on this:
- https://bugzilla.redhat.com/show_bug.cgi?id=1131398
- https://bugzilla.redhat.com/show_bug.cgi?id=1131400

We need to find and prefer those binaries:
- RHEL7 - /usr/bin/start-puppet-ca
- RHEL6 - /usr/sbin/puppetca

domcleal:
What about upstream (Puppet Labs) packages?  Changes should be made there first, or at least in parallel.

/usr/sbin/puppetca is already present in 2.7.x btw.  (This isn't about operating system versions, it's about major versions of Puppet.)

lzap:
Yeah thanks, I noticed that it is in `/usr/sbin`. No change needed.

I am not doing any changes upstream now. I doubt they accept it at all, we need to carry those wrappers. Perhaps inside foreman-proxy or -selinux package. Of course it would be great if Puppet is able to set the domain explicitly via libselinux bindings directly in Ruby.

domcleal:
This should be addressed with the Puppet dev community then, not first or only in Foreman.  Foreman or EPEL shouldn't be shipping Puppet's SELinux policy.

lzap:
I am afraid this is special case when it does not make much sense to
push this upstream. The way puppet boots would have been refactored
while we can achieve the very same result with wrappers (or runuser
which is an alternative).

But in general, I do agree I should try this. I just don't have time for
this. Fedora and EPEL have been using wrappers for very long time now, I
don't see any issue with this approach - until this is solved upstream.
Again, this is long-term run there.

What I plan to achieve with this patch is to try one existing wrapper
(/usr/sbin/puppetca) - that's for EPEL6 puppet, and one new wrapper
(/usr/bin/start-puppet-ca) which is for Fedora/EPEL7 puppet.

For all other distros this is not relevant, until the support is
upstream.

What I can try is I can file a PR adding similar wrappers into upstream
(not the proper patch modifying bootstrap code):

https://tickets.puppetlabs.com/browse/PUP-3094
https://github.com/puppetlabs/puppet/pull/2997

LZ

## 

Later,
 Lukas #lzap Zapletal

domcleal:
It's no use supporting only EPEL and Fedora packages, that's my point, it needs to be present in all packages in common usage.

lzap:
How should I proceed when they don't accept the patch then?

domcleal:
The patch doesn't look like it's been refused, however I'd suggest continuing to work with them to come to an acceptable solution.

lzap:
Just for the record, discussion is going on here:

https://groups.google.com/forum/#!topic/puppet-dev/VTb2o-Nuxkw

It looks like we will be writing proper selinux context switching into the upstream puppet so no more wrappers. So I will rework this patch to use existing wrappers for EPEL6 and EPEL7. Not sure what to do with Enterprise Puppet version. Not supported by default?

Opinions?

lzap:
Ok I _really_ want this to get moving. It looks like the most feasible option is to carry the wrappers with the proxy until this is resolved in the upstream, what you think?

For EPEL puppets, this is the current state. For RHEL6 puppetca wrapper has wrong file type:

```
[root@nightly ~]# ll -Z /usr/sbin/puppetd
-rwxr-xr-x. root root system_u:object_r:puppet_exec_t:s0 /usr/sbin/puppetd
[root@nightly ~]# ll -Z /usr/sbin/puppetca
-rwxr-xr-x. root root system_u:object_r:bin_t:s0       /usr/sbin/puppetca
```

For RHEL7 unfortunately the situation is very same:

```
[root@seven ~]# ll -Z /usr/bin/start-puppet-master
-rwxr-xr-x. root root system_u:object_r:puppetmaster_exec_t:s0 /usr/bin/start-puppet-master
[root@seven ~]# ll -Z /usr/bin/start-puppet-ca
-rwxr-xr-x. root root system_u:object_r:bin_t:s0       /usr/bin/start-puppet-ca
```

I am going to fix EPELs but we need to support other versions of Puppet (PuppetLabs) and older versions which will unlikely will have _any_ wrappers.

_Edited_ the RHEL7 bit which was incorrectly pasted.

lzap:
FYI Here is a "tracker" bug that added missing `start-puppet-ca` wrapper (which was removed from puppet by mistake). It has two blocker bugs for RHEL6 and RHEL7 to correctly set the exec file types. It slipped to 6.7 and 7.2 currently... https://bugzilla.redhat.com/show_bug.cgi?id=1131398

lzap:
I am not going to further work on getting appropriate SELinux support in Puppet, I am closing this PR. We will likely need similar patch when RHEL8 comes out or if we add Fedora support (newer versions already have the startup wrappers already).

PR #202, opened 2014-08-21
Implements #7181: added initial support for rubocop

dmitri-d:


domcleal:
Please use "fixes" so redmine and the PR bot understand :)

dmitri-d:
That explains it!

daviddavis:
@witlessbird looks good. I usually just run `rubocop` from the command line instead of having a rake task. That way you can pass options in. If you're looking to exclude certain directories you can do that in the rubocop config:

https://github.com/Katello/katello/blob/master/.rubocop.yml#L8

tasks/jenkins.rake
```diff
@@ -12,6 +12,13 @@ begin
       task :minitest  => [:pre_ci, "ci:setup:minitest"]
       task :test_unit => [:pre_ci, "ci:setup:testunit"]
     end
+
+    task :rubocop do
+      system("bundle exec rubocop \
+        --require rubocop/formatter/checkstyle_formatter \
+        --format RuboCop::Formatter::CheckstyleFormatter \
+        --no-color --out rubocop.xml")
```
daviddavis, line 20: I had to add an extra line here:

https://github.com/theforeman/foreman/pull/1666/files#diff-f95d0e467c98d6b455e88bbf8b01cff1R23

.rubocop_todo.yml
```diff
@@ -0,0 +1,486 @@
...
+# Offense count: 1
+Lint/UselessAccessModifier:
+  Enabled: false
+
+# Offense count: 22
+Lint/UselessAssignment:
+  Enabled: false
+
+# Offense count: 5
+Metrics/BlockNesting:
```
domcleal: I've updated the Jenkins job and I think it's failing due to this:

<pre>
Warning: unrecognized cop Metrics/BlockNesting found in /var/lib/workspace/workspace/test_proxy_develop_pr_rubocop/.rubocop_todo.yml
Warning: unrecognized cop Metrics/BlockNesting found in /var/lib/workspace/workspace/test_proxy_develop_pr_rubocop/.rubocop.yml
</pre>

dmitri-d: What version of rubocop is Jenkins using? If it's below 0.25.0 it won't have 'Metrics' namespace... I'll probably need to regenerate the todo config.

domcleal: 0.24.1 - as specified in your Gemfile :grinning:

dmitri-d:
@domcleal: shouldn't fail now...

daviddavis:
Do you support Ruby 1.9.2? Rubocop 0.24.1 doesn't support it. You'll probably need to add something like `if RUBY_VERSION >= "1.9.3"` to your gem files.

domcleal:
Yes, 1.9.2's supported for the proxy as Ubuntu 12.04 ships with it.

bundler.d/development.rb
```diff
@@ -5,4 +5,5 @@
   gem 'rdoc'
   gem 'single_test'
   gem 'pry'
+  gem 'rubocop' if  RUBY_VERSION > "1.9.2"
```
domcleal: Should this be pinned to a "major" version (0.24.x, 0.25.x etc)?  It's likely to cause fairly regular breakages otherwise I fear.  I think core went with 0.24.1 as it's in line with Katello(?).

dmitri-d:
@domcleal: pinned rubocop to .24.1 and rebased.

domcleal:
@witlessbird I'm expecting it to fail on those Metrics ones again, you may need to remove them

dmitri-d:
@domcleal: regenerated .rubocop_todo using .0.24.1

daviddavis:
:+1: from me.

domcleal:
Thanks @witlessbird, merged as cdd01ff for 1.7.

PR #203, opened 2014-08-22
fixes #7235: dhcp tests are now order-independent

dmitri-d:


lzap:
Can you drop few words? I am not sure I follow the case. What was the reproducer and error message?

dmitri-d:
@lzap: try executing  dhcp_api tests only, without the rest. They'll fail as they don't have all the dependencies they require.

lzap:
Ok, how do I run it please? I must have been doing something wrong:

```
be ruby -Ilib -Itest test/dhcp/dhcp_api_test.rb 
Puma 1.6.3 starting...
* Min threads: 0, max threads: 16
* Environment: test
* Listening on tcp://localhost:4567
== Sinatra/1.4.2 has taken the stage on 4567 for development with backup from Puma
```

domcleal:
That's a bug in our tests where they should be requiring "sinatra/base" rather than "sinatra".

dmitri-d:
@lzap: use: rake test TEST=test/dhcp/dhcp_api_test.rb

lzap:
Tests run fine. Looks good to me. Can you please update on what @domcleal recommends?

domcleal:
Just FYI, there are some failures on develop currently due to http://projects.theforeman.org/issues/7274, looking into it now.

dmitri-d:
I'll take a look at sinatra requires today...

dmitri-d:
@lzap, @domcleal: removed "require 'sinatra'" from tests, as we handle this in test_helper.

lzap:
Works for me, thanks. Merged as 63f9e26 for Foreman develop.

PR #205, opened 2014-08-22
moved facter loading into facter plugin descriptor

dmitri-d:


dmitri-d:
This works, although I wonder if handling 'facter' as a plugin dependency (i.e. putting it into 'facts' bundler group) is better.

lzap:
Dmitri, we need a ticket, some better explanation.

I think this could be `Refs #7062` as I understand this as a definitive solution to that problem.

We do not want to have puppet and facter as bundler dependencies because we have users with bundler installations (not bundler_ext RPM/DEB) who want to use Puppet Enterprise or other Puppet distributions.

It works on my environment. I think we can merge, someone else?

dmitri-d:
@lzap: created a ticket. Any chance you can merge this?

lzap:
Yeah sorry, github does not send out pushes via email. Pushed as f655a4e - thanks @witlessbird !

PR #198, opened 2014-08-13
fixes #7060: puppet gem is no longer required if puppetca is disabled.

dmitri-d:

PR #206, opened 2014-08-22
fixes #7240: puppetca now loads correctly if puppet module is disabled

dmitri-d:


domcleal:
I can't reproduce this, could you add more details (preferably to the bug report) to show what the problem and symptoms are?

domcleal:
Gotcha, thanks for the example.

Merged as dae892a for Foreman 1.6, thanks @witlessbird.

lzap:
This has something to do with https://github.com/theforeman/smart-proxy/pull/198

Dmitri can you please explain? The previous patch works fine for me in Foreman Discovery LiveCD.

lzap:
Ok I assume you guys sorted it out :-)

domcleal:
It's kind of the reverse of 198 I think, when puppet's disabled but the puppetca module is enabled.  If that's the case, and you're in a bundler environment (i.e. not a bundler_ext package install) then it won't be able to find the puppet library as there was no puppetca bundler group defined.

(In bundler_ext, if you have puppet installed then it'll probably load just fine, it's only bundler that's strict about library loading.)

dmitri-d:
Quite possible, I'm not sure how strict bundler_ext is about groups (internally group names are being passed around). Will need to check it...

domcleal:
It's not even really about the groups, I think it's that bundler_ext doesn't "fence" library loading in the same way bundler does.  If you start up the app and require puppet (which puppetca_main.rb did, IIRC, which is why I couldn't reproduce) then it'll load it anyway.  Under "pure" bundle exec, even if you had the gem installed but the group wasn't active in Gemfile.lock, it will pretend it can't load it.

lzap:
Yep, bundler_ext really does "dumb" loading of what you `require`. And you need to specify which groups to load (or load them all).

PR #207, opened 2014-08-27
fixes #7274 - generate date format expectation in tests to handle Ruby/JSON differences

domcleal:
ffi-yajl 1.1.0 changed the date format, so this simply regenerates the expected value depending on the current environment, as we already know it changes between Ruby versions too.


domcleal:
http://ci.theforeman.org/job/test_proxy_develop_pull_request/460/ tests it, all green.

lzap:
Nice hack. Thanks.

lzap:
Pushed as ce9a075. Thanks.

PR #208, opened 2014-08-29
fixes #7297 - corrected name of DNS namespace in virsh error handlers

domcleal:


dmitri-d:
This looks good as a fix. Virsh provider should be refactored not to have dependencies on modules though. It looks like the only thing shared between dhcp and dns is Proxy::Virsh#virsh method?

domcleal:
Yeah, I can move the error class into here/core to make more sense.

dmitri-d:
I created a ticket for the refactoring: http://projects.theforeman.org/issues/7298. 

As an alternative to moving error classes into the core, we could move dns/dhcp-specific code into corresponding modules...

domcleal:
Ah yes, I see what you mean now.  I'll have a closer look later and maybe just do the refactoring now.

domcleal:
Refactoring's complete, I moved those specific DNS/DHCP methods back into the correct modules and used a Proxy::Virsh::Error class to raise errors from the shared code to the modules.

dmitri-d:
looks good, will give a spin later today...

dmitri-d:
merged at b32f2f7, ty @domcleal!

PR #209, opened 2014-09-04
fixes #7351: plugins no longer crash smart-proxy if they are missing either of rackup paths

dmitri-d:


GregSutcliffe:
Tested, verified, merged as b340269. Thanks!

PR #210, opened 2014-09-07
Allows chef node and client deletion

ares:
These changes are required for new foreman-chef feature allowing user to delete nodes and clients (host certificates in puppet terminology) upon host deletion in foreman. I'm starting to wonder whether we should extract chef into a separate proxy plugin soon.

### DIFF
--- a/config/settings.d/chef.yml.example
+++ b/config/settings.d/chef.yml.example
@@ -2,7 +2,22 @@
 :enabled: false
 # :chef_authenticate_nodes: true
 # :chef_server_url: "https://chef.example.net"
-# smart-proxy client node needs to have some admin right on chef-server
+
+# smart-proxy client node needs to have admin rights on chef-server
 # in order to retrive all nodes public keys
 # :chef_smartproxy_clientname: 'host.example.net'
 # :chef_smartproxy_privatekey: '/etc/chef/client.pem'
+
+# by default ssl verification of request to you chef server is enabled,
+# you're supposed to install CA certificate yourself
+# this usually consist of two steps
+#   download the CA cert to /etc/pki/tls/certs/, e.g. ca-cert-root.pem
+#   ln -s ca-cert-root.pem $( openssl x509 -hash -noout -in ca-cert-root.pem )".0"
+# you can use self-signed certificate (see below) or disable verification
+# which is definitely not recommended for production
+# :chef_ssl_verify: true
+
+# if you're using self-signed certificate for you chef server, you can specify
+# the certificate file here or leave it empty (default)
+# :chef_ssl_pem_file: '/etc/chef/chef.example.com.pem'
+
--- a/modules/chef_proxy/chef_api.rb
+++ b/modules/chef_proxy/chef_api.rb
@@ -1,5 +1,7 @@
 require 'proxy/request'
 require 'chef_proxy/authentication'
+require 'chef_proxy/resources/node'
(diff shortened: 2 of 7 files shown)


mmoll:
:+1: to plugin extraction

config/settings.d/chef.yml.example
```diff
@@ -2,7 +2,22 @@
...
+#   download the CA cert to /etc/pki/tls/certs/, e.g. ca-cert-root.pem
+#   ln -s ca-cert-root.pem $( openssl x509 -hash -noout -in ca-cert-root.pem )".0"
+# you can use self-signed certificate (see below) or disable verification
+# which is definitely not recommended for production
+# :chef_ssl_verify: true
+
+# if you're using self-signed certificate for you chef server, you can specify
+# the certificate file here or leave it empty (default)
+# :chef_ssl_pem_file: '/etc/chef/chef.example.com.pem'
+
```
dmitri-d, line 23: You should define default values in chef_plugin.rb for the settings here.

config/settings.d/chef.yml.example
```diff
@@ -2,7 +2,22 @@
 :enabled: false
 # :chef_authenticate_nodes: true
 # :chef_server_url: "https://chef.example.net"
```
dmitri-d, line 4: Is there a reasonable default setting for this url? Otherwise I think there should be a check that the url is reachable...

modules/chef_proxy/chef_api.rb
```diff
@@ -15,15 +17,59 @@ class Api < ::Sinatra::Base
...
     end
 
     post "/reports" do
+      logger.debug 'report upload request received'
       Proxy::Chef::Authentication.new.authenticated(request) do |content|
         Proxy::HttpRequest::Reports.new.post_report(content)
       end
     end
+
+    get "/chefproxy/nodes/:fqdn" do
```
dmitri-d, line 33: hmm. Chef module is mounted under 'api'; You are adding another prefix here -- 'chefproxy'. I think we should have just one, 'chefproxy' (or just 'chef'?).

modules/chef_proxy/resources/base.rb
```diff
@@ -0,0 +1,20 @@
...
+        )
+        @connection.ssl_verify = Proxy::Chef::Plugin.settings.chef_ssl_verify
+        self_signed = Proxy::Chef::Plugin.settings.chef_ssl_pem_file
+        if !self_signed.nil? && !self_signed.empty?
+          @connection.ssl_pem_file = self_signed
+        end
+      end
+    end
+  end
+end
```
dmitri-d, line 20: No checks if config values have been initialized; This won't be an issue, however, if you define default values for these settings.

modules/chef_proxy/chef_api.rb
ares, line 33: We have `/api/reports` and `/api/hosts/facts` that are for uploading data from chef-client to foreman. The API is on same path as it is on foreman so in chef-client you can use either of them. The `/api/chefproxy/*` is for the other direction - foreman communicating to chef-server.

ares, line 33: ah I think I see what you mean, should we map it as two sinatra apps resulting in chefproxy being mounted to `/chef`?

ares:
Closing this one, will open another removing chef related code, since I extracted it to a chef plugin https://github.com/theforeman/smart_proxy_chef

PR #211, opened 2014-09-08
Fixes #7352 - Missing DNS record should return 404 not 400 for DELETE

shlomizadok:
If DNS entry is not found, return 404

### DIFF
--- a/lib/proxy/virsh.rb
+++ b/lib/proxy/virsh.rb
@@ -17,12 +17,12 @@ module Proxy::Virsh
         return e.parent.attributes["ip"]
       end
     end
-    raise Proxy::DNS::Error.new("Cannot find DNS entry for #{host}")
+    raise Proxy::Dns::NotFound.new("Cannot find DNS entry for #{host}")
   rescue Exception => e
-    msg = "DNS virsh provider error: unable to retrive virsh info: #{e}"
+    msg = "DNS virsh provider error: unable to retrieve virsh info: #{e}"
     logger.error msg
     logger.debug xml if defined?(xml)
-    raise Proxy::DNS::Error, msg
+    raise Proxy::Dns::Error, msg
   end
 
   def virsh *params
@@ -51,7 +51,7 @@ module Proxy::Virsh
       "--xml", "'<host ip=\"#{ip}\"><hostname>#{hostname}</hostname></host>'",
       "--live", "--config"
   rescue Exception => e
-    raise Proxy::DNS::Error, "Failed to update DNS: #{e}"
+    raise Proxy::Dns::Error, "Failed to update DNS: #{e}"
   end
 
   def virsh_update_dhcp command, mac, ip, name
--- a/modules/dns/dns_api.rb
+++ b/modules/dns/dns_api.rb
@@ -63,6 +63,8 @@ module Proxy::Dns
       begin
         dns_setup({:fqdn => fqdn, :value => value, :type => type})
         @server.remove
+      rescue Proxy::Dns::NotFound => e
+        log_halt 404, e
       rescue => e
         log_halt 400, e
       end
--- a/modules/dns/dns.rb
+++ b/modules/dns/dns.rb
@@ -2,6 +2,7 @@ require 'dns/dns_plugin'
 
 module Proxy::Dns
   class Error < RuntimeError; end
+  class NotFound < RuntimeError; end
   class Collision < RuntimeError; end
   class Record
     include Proxy::Log


domcleal:
Should be #7352, please also amend the commit message as well as the title.

lib/proxy/virsh.rb
```diff
@@ -17,12 +17,12 @@ def find_ip_for_host host
...
       end
     end
-    raise Proxy::DNS::Error.new("Cannot find DNS entry for #{host}")
+    raise Proxy::Dns::NotFound.new("Cannot find DNS entry for #{host}")
   rescue Exception => e
-    msg = "DNS virsh provider error: unable to retrive virsh info: #{e}"
+    msg = "DNS virsh provider error: unable to retrieve virsh info: #{e}"
     logger.error msg
     logger.debug xml if defined?(xml)
-    raise Proxy::DNS::Error, msg
```
domcleal, line 25: duplicates #208

lib/proxy/virsh.rb
```diff
@@ -17,12 +17,12 @@ def find_ip_for_host host
...
+    raise Proxy::Dns::NotFound.new(msg)
```
domcleal: I don't think that's appropriate?

lib/proxy/plugin.rb
```diff
@@ -123,11 +123,11 @@ def bundler_group
...
 
   def http_rackup
-    File.read(self.class.get_http_rackup_path) unless self.class.get_http_rackup_path.nil?
+    self.class.get_http_rackup_path.nil? ? "" : File.read(self.class.get_http_rackup_path)
   end
 
   def https_rackup
-    File.read(self.class.get_https_rackup_path) unless self.class.get_https_rackup_path.nil?
+    self.class.get_https_rackup_path.nil? ? "" : File.read(self.class.get_https_rackup_path)
   end
```
dmitri-d, line 131: These changes are already in develop. Pls. rebase.

lib/proxy/virsh.rb
dmitri-d: Agree with @domcleal.

lib/proxy/plugin.rb
shlomizadok, line 131: :+1:

modules/dns/dns_api.rb
```diff
@@ -63,6 +63,8 @@ def dns_setup(opts)
       begin
         dns_setup({:fqdn => fqdn, :value => value, :type => type})
         @server.remove
+      rescue Proxy::Dns::NotFound => e
```
dmitri-d, line 66: This will produce different results depending on the dns provider. 'nsupdate' and 'nsupdate_gss' will return 400, while 'virsh' will return a 404. I don't think this is a good idea...

lib/proxy/virsh.rb
shlomizadok: Not sure I understand the comment. Do you think it should return an error here?

modules/dns/dns_api.rb
ohadlevy, line 66: @witlessbird somehow i think we lost it on refactoring the proxy, but originally, the proxy reported 404 when trying to delete a non existing record, and was used in foreman -> https://github.com/theforeman/foreman/blob/develop/lib/proxy_api/dns.rb#L23

lib/proxy/virsh.rb
domcleal: Yes, it shouldn't return a "404 not found" if there is a general error retrieving stuff from libvirt, or it'll obscure a real bug.

modules/dns/dns_api.rb
domcleal, line 66: Sure, but Dmitri's point stands that we should check for consistency between all providers - nsupdate doesn't look like it's been changed, it may return 400 as he said.

GregSutcliffe:
@shlomizadok The commit message is incorrect, it currently suggests you are adding the functionality of returning 400, when in fact thats the problem. It should read something like "Return 404 when a DNS entry is not found"

shlomizadok:
@GregSutcliffe :+1:

lib/proxy/virsh.rb
```diff
@@ -17,12 +17,12 @@ def find_ip_for_host host
...
     end
-    raise Proxy::DNS::Error.new("Cannot find DNS entry for #{host}")
+    raise Proxy::Dns::NotFound.new("Cannot find DNS entry for #{host}")
   rescue Exception => e
-    msg = "DNS virsh provider error: unable to retrive virsh info: #{e}"
+    msg = "DNS virsh provider error: unable to retrieve virsh info: #{e}"
     logger.error msg
     logger.debug xml if defined?(xml)
-    raise Proxy::DNS::Error, msg
+    raise Proxy::Dns::Error, msg
```
ohadlevy, line 25: note that libvirt will raise an exception if there are no xml entries for any hosts, this will catch this exception and not the not found above.

modules/dns/dns_api.rb
dmitri-d, line 66: Nods, we should return the same error independent of the provider (client doesn't know which provider is being used, so they'll have to be defensive in treating our responses).

I'm not sure how much work it would be to add this disambiguation of possible errors to nsupdate providers, as we are parsing plain text there.

ohadlevy, line 66: :+1:

shlomizadok:
Closing this PR and restarting fresh.

PR #212, opened 2014-09-13
Remove chef related code

ares:
Code was extracted to separate chef plugin. See
https://github.com/theforeman/smart_proxy_chef


dmitri-d:
I'm ok with this, @domcleal?

dmitri-d:
@ares: do you need a review of the extracted plugin, or is it a verbatim copy?

domcleal:
I'm fine in principle with the move.

ares:
@witlessbird it's a copy with additional featues, if you're interested you can go over https://github.com/theforeman/smart_proxy_chef and add comments there

ares:
We should put some warning to release notes, users have to install the plugin if they relied on this functionality. The plugin is not yet packed though.

domcleal:
Indeed, but should also be handled through packaging (obsoletes).

dmitri-d:
@ares: I left you a comment in the closed PR at https://github.com/theforeman/smart_proxy_chef. It mostly looks good to me, the only issue is that you are using an 'api' namespace in this module, which I think sounds too generic and can potentially cause conflicts (or modules not being loaded) down the line. I assume this namespace was used by chef module before the whole modularization?

ares:
@witlessbird thanks, the `/api` was there before modularization (and I believe is still there until we merge this PR) and the path is there so we have same URL if we point chef-clients to foreman directly or smart-proxy (which optionally does the authentication). I will move it later, probably when I write-up some new installation docs or work on installer support for this.

domcleal:
@ares could you rebase, assign a redmine ticket and let's get this into 1.7 please

ares:
I'll try to build package for the plugin first since we would drop packaging for this functionality

ares:
I didn't make it. @domcleal I know it's past 1.7 branching but this does not add feature (it removes it) so maybe we could still get it into 1.7.

rvrignaud:
I tested this PR with smart-proxy and smart_proxy_chef installed. We have same functional level. I would be I favor of merging this PR to avoid unnecessary complexity.

domcleal:
@ares I'm happy to have this in 1.7, but only on the basis that the packages are sorted out ASAP after merging.  (Before RC2 - if longer, then I'll probably revert in 1.7-stable as we need to ship something functioning.)

bundler.d/chef.rb
```diff
@@ -1,3 +0,0 @@
-group :chef do
-  gem 'chef', ">= 11.6.2"
```
domcleal, line 2: mixlib-shellout is also in test.rb because of chef, it should be removed

ares, line 2: mixlib-shellout removed

domcleal:
Thanks @ares, merged as ed82c92.

PR #213, opened 2014-09-23
fixes #7596 - validate plugin dependencies when loading, various fixes

domcleal:


test/plugin_test.rb
```diff
@@ -20,4 +20,51 @@ def test_http_rackup_returns_empty_string_with_missing_rackup_path
...
+    end
+  end
+
+  def test_unsatisified_dependency
+    eval "class TestPlugin5 < Proxy::Plugin; plugin :test5, '1.5'; end"
+    eval "class TestPlugin6 < Proxy::Plugin; plugin :test6, '1.0'; requires :test5, '> 2.0'; end"
+    assert_nothing_raised do
+      TestPlugin5.new.validate_dependencies!(TestPlugin5.dependencies)
+    end
+    assert_raise(::Proxy::PluginVersionMismatch) do
```
dmitri-d, line 39: this bit is quite similar to `test_core_dependency`. Otoh, there's no test that verifies that a plugin with satisfiable dependency is valid...

dmitri-d:
looks good, just the small issue in tests.

test/plugin_test.rb
domcleal, line 39: Yeah, I've cleared it up now into two separate tests, one for satisfied deps and one for unsatisfied.  core_dependency was meant to test the case with a -develop suffix, so I've added a comment above.

domcleal:
I had to change the tests from using eval to simply declaring the classes after a very strange issue.  When using find_plugin, I'd get back a different (unnamespaced) class object to the one I'd get when just using the constant in the test.

Updated.

dmitri-d:
merged in at 20b9f96. thanks, @domcleal!

PR #215, opened 2014-09-30
Fixes #7749 - allow using logger in log_halt helper

ares:


domcleal:
dimatha reports a successful test via IRC.  I'd like to know when this broke, so we can include it in 1.6.1 if necessary there.. I suspect though that it's develop-only.

ares:
I think this got broken during modularization but only on old sinatra, or do you want me to find exact sinatra version?

domcleal:
Nah, not worried about the sinatra version - I was more curious if it was broken in 1.6.0 or nightly only.  Don't worry, it'll go into 1.6.1.

domcleal:
Thanks @ares, merged as 2fe082e.

PR #217, opened 2014-10-07
fixes #7822 - forbid HTTPS requests with no client SSL certificate

domcleal:
Please review as a priority, and test heavily.  It seems that OpenSSL correctly refuses connections when the certificate presented isn't trusted by the CA that the server is using (note, we don't have separate client/server CAs), and the problem is only when no client cert is presented.

I can't see a way to check that relying on OpenSSL for bad verification is always solid.  The closest I've found is that the socket that WEBrick sees has an OpenSSL verification result attached (http://ruby-doc.org/stdlib-1.9.3/libdoc/openssl/rdoc/OpenSSL/SSL/SSLSocket.html) but I can't see a way to get at that.  In any case, it only seems to reach the smart proxy when there's either no cert or it validates and in both cases the verify_result was "OK".

Lastly, I'm looking at more tests in the 7822-ssl-verify branch which test Proxy::Launcher's https_app by spinning up the rack/WEBrick app and then make requests, but this will take more refactoring.  I will probably follow up with this in another ticket, and only to develop (not 1.6 etc).

No CVE assigned yet - will add it to commit message before it's merged.

### DIFF
--- a/lib/sinatra/ssl_client_verification.rb
+++ b/lib/sinatra/ssl_client_verification.rb
@@ -1,0 +1,13 @@
+require 'sinatra/base'
+
+::Sinatra::Base.helpers ::Proxy::Helpers
+::Sinatra::Base.helpers ::Proxy::Log
+::Sinatra::Base.before do
+  if ['yes', 'on', '1'].include? request.env['HTTPS'].to_s
+    if request.env['SSL_CLIENT_CERT'].to_s.empty?
+      log_halt 403, "No client SSL certificate supplied"
+    end
+  else
+    logger.debug('require_ssl_client_verification: skipping, non-HTTPS request')
+  end
+end
--- a/lib/smart_proxy.rb
+++ b/lib/smart_proxy.rb
@@ -24,6 +24,7 @@ Proxy::BundlerHelper.require_groups(:default)
 
 require 'rack-patch' if Rack.release < "1.3"
 require 'sinatra-patch'
+require 'sinatra/ssl_client_verification'
 require 'sinatra/trusted_hosts'
 
 module Proxy
--- a/lib/smart_proxy_for_testing.rb
+++ b/lib/smart_proxy_for_testing.rb
@@ -11,6 +11,7 @@ require 'proxy/plugin'
 require 'proxy/error'
 
 require 'sinatra/base'
+require 'sinatra/ssl_client_verification'
 require 'sinatra/trusted_hosts'
 
 Proxy::SETTINGS = ::Proxy::Settings::Global.new(:log_file => './logs/test.log', :log_level => 'DEBUG')
--- a/test/sinatra/ssl_client_verification_test.rb
+++ b/test/sinatra/ssl_client_verification_test.rb
@@ -1,0 +1,38 @@
+require 'test_helper'
+require 'json'
+require 'sinatra/base'
+
+ENV['RACK_ENV'] = 'test'
+
+class SSLClientVerificationTest < Test::Unit::TestCase
+  include Rack::Test::Methods
+
+  def app
+    TestApp.new
+  end
+
+  def test_http
+    get '/test'
+    assert last_response.ok?
+  end
+
+  ['yes', 'on', '1'].each do |yes|
+    define_method("test_https_no_cert_https_#{yes}") do
+      get '/test', {}, {'HTTPS' => yes}
+      assert last_response.forbidden?
+    end
+  end
+
+  def test_https_cert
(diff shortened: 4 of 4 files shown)


domcleal:
@witlessbird fixed the handling of different HTTPS values, thanks (https://github.com/domcleal/smart-proxy/commit/84f5b4cd5ffda1a4a1f0bdcdeeb937b2f8acaa3a#commitcomment-8064426)

lib/sinatra/ssl_client_verification.rb
```diff
@@ -0,0 +1,20 @@
...
+module Sinatra
+  module SSLClientVerification
+    def require_ssl_client_verification
+      helpers ::Proxy::Helpers
+      helpers ::Proxy::Log
+
+      before do
+        if ['yes', 'on', '1'].include? request.env['HTTPS'].to_s
+          if request.env['SSL_CLIENT_CERT'].to_s.empty?
+            log_halt 403, "No client SSL certificate supplied"
```
dmitri-d: Is there a plan to test the cert chain here?

dmitri-d: nm, we have OpenSSL::SSL::VERIFY_PEER in the https config.

lib/smart_proxy.rb
```diff
@@ -38,6 +39,7 @@ module Proxy
   ::Sinatra::Base.use ::Proxy::LoggerMiddleware # instead, we have our own logging middleware
   ::Sinatra::Base.use ::Rack::CommonLogger, ::Proxy::Log.logger
   ::Sinatra::Base.set :env, :production
+  ::Sinatra::Base.register ::Sinatra::SSLClientVerification
```
dmitri-d: As we are most likely going to need ssl client cert verification for all modules accessible over https, I wonder if instead of using a helper it would be better (as in less repetitious) to have a global filter, defined around here.

domcleal: Yeah, I'd be happier with that too I think.

mmoll:
Not a really excessive test, but I'm getting the expected responses now, according to my input.
- just curl -k -> No client SSL certificate supplied
- curl with cert and key from another CA -> tlsv1 alert unknown ca
- curl with cert and key from the same CA -> JSON output

lib/smart_proxy.rb
domcleal: How do you think it should be implemented?  I don't see how to add a filter globally, without either modifying Sinatra::Base (which I'm a little wary of, as they warn against it), and creating a base class for all API impls.

dmitri-d:
tested, looks good.

lib/smart_proxy.rb
dmitri-d: sry, just noticed the comment. I was thinking of: 

```
::Sinatra::Base.before do
...
end
```

sometime during loading. Is that what Sinatra devs warn against? Source looks pretty benign -- all this will do is add this filter to the list of `:before` filters.

domcleal: Ah, that looks fine.  I'll give it a go.

domcleal: Updated, how's that?  It probably needs a re-test, sorry.

lib/sinatra/ssl_client_verification.rb
```diff
@@ -0,0 +1,13 @@
+require 'sinatra/base'
+
+::Sinatra::Base.helpers ::Proxy::Helpers
+::Sinatra::Base.helpers ::Proxy::Log
+::Sinatra::Base.before do
```
dmitri-d, line 5: looks good, although it might be a bit more readable if you used a proc here instead...

dmitri-d:
[test]

domcleal:
The test failure looks valid, let me look into it.  gem update perhaps.

lib/sinatra/ssl_client_verification.rb
domcleal, line 5: Is that possible?  I can't get it to work..

dmitri-d, line 5:
```
::Sinatra::Base.before(&Proc.new { verify_client_cert })
def verify_client_cert
...
end
```

Having tried that though, I can say it doesn't make any difference in clarity whatsoever. My apologies.

domcleal, line 5: hehe

dmitri-d:
tested with the updated filter, looks good!

domcleal:
Test failure's addressed in #218 btw.

dmitri-d:
[test]

dmitri-d:
Merged at 52f0bac. Thanks, @domcleal!

PR #223, opened 2014-10-13
Let's re-use http(s) connection to foreman for subsequent requests

isimluk:


domcleal:
[test]

lzap:
Looks good to me, just for the record: `foreman_url` is a setting that is currently not set with the installer and it is missing from the example configuration. We have a ticket to set that automatically.

Can you please create a refactor issue in our tracker http://projects.theforeman.org/projects/smart-proxy/issues/new and squash into one commit in the format of:

```
Fixes #XYZ - promote uri and http ...
```

Thanks

isimluk:
I have filed issue #7922. Thanks for review!

dmitri-d:
Did you see performance issues around Net:HTTP instantiation? I'd very much prefer not carry around state, if possible. An issue with this refactor (not a big one atm) is that ForemanRequest is no longer thread-safe.

isimluk:
I did not saw performance issues yet. However, creating new connection per request (given all the ssl validations) may cause performance issues.

Do you need a thread safe ForemanRequest? Wouldn't it be better idea to have a ForemanRequest instance per thread?

isimluk:
The reason, why I started this refactoring is that I want to decouple the send_request function. And that will certainly include storing a state somewhere. The problem I have is that send_request does a lot of things and it may not fit well to all the usages (smart proxy plugin). My plug-in for instance wants to set content-encoding header and different content-type.

lzap:
I don't see the point @witlessbird, the class is instntiated per request, it will not survive multiple requests at all. Do you think that this gets reused when in passenger mode?

dmitri-d:
@isimluk:fair enough. Can you break up #http into dedicated initializer and accessor? Then the aceessor can be something like:

```
def http
  @http ||= http_init
end
```

isimluk:
Sure. What would be rubyistics name for such initializer?

isimluk:
Err., I see http_init :-D

isimluk:
done;

dmitri-d:
[test]

isimluk:
Rebased patch on top of develop branch. (Was conflicting with rubocop fix at 25691fff37c8ce8eaa7c62ac47a60befd6bfa329).

dmitri-d:
[test]

dmitri-d:
[test]

dmitri-d:
merged at 8efc06c; thanks, @isimluk!

PR #224, opened 2014-10-16
Refs #969 - Proxy-side changes for serving templates from the proxy

dustints:
An update to @GregSutcliffe's original PR (https://github.com/theforeman/smart-proxy/pull/100/files). Ports his original feature to the new plugin api.

### DIFF
--- a/config/settings.d/templates.yml.example
+++ b/config/settings.d/templates.yml.example
@@ -1,0 +1,16 @@
+# Set this to true if the Proxy should handle template requests on behalf of Foreman
+:enabled: true
+
+
+# This plugin also requires that :foreman_url: be set in the main settings.yml
+# This lets the plugin know how to obtain the templates from foreman.
+
+
+# This allows the proxy to define how hosts that are being provisioned where to
+# obtain the templates from.
+# The default protocol is http on port 80 unless otherwise specified in the url.
+# Examples:
+# smart-proxy.example.com:8080
+# https://smart-proxy.example.com
+# http://1.2.3.4:3000
+#:template_url: https://smart-proxy.example.com:8443
--- a/lib/proxy/request.rb
+++ b/lib/proxy/request.rb
@@ -1,19 +1,50 @@
 require 'net/http'
 require 'net/https'
 require 'uri'
-
-# TODO: need settings validation on startup, otherwise we get a 500 error due to missing/wrong config settings when api is accessed
+require 'cgi'
 
 module Proxy::HttpRequest
-  class ForemanRequest
-    def send_request(path, body)
-      path = [uri.path, path].join('/') unless uri.path.empty?
-      req = Net::HTTP::Post.new(URI.join(uri.to_s, path).path)
+  class ForemanRequestFactory
+    def initialize(base_uri)
+      @base_uri = base_uri
+    end
+
+    def query_string(input={})
+      input.map{|k,v| "#{CGI.escape(k.to_s)}=#{CGI.escape(v)}"}.join("&")
+    end
+
+    def create_get(path, query={})
+      uri = uri(path)
+      req = Net::HTTP::Get.new("#{uri.path || '/'}?#{query_string(query)}")
+      req = add_headers(req)
+      req
+    end
+
+    def uri(path)
+      URI.join(@base_uri.to_s, path)
+    end
+
+    def add_headers(req)
       req.add_field('Accept', 'application/json,version=2')
       req.content_type = 'application/json'
-      req.body         = body
+      req
+    end
 
-      http.request(req)
+    def create_post(path, body)
+      req = Net::HTTP::Post.new(uri(path).path)
+      req = add_headers(req)
+      req.body = body
+      req
+    end
+  end
+
+  class ForemanRequest
+    def send_request(request)
+      http.request(request)
+    end
+
+    def request_factory
+      ForemanRequestFactory.new(uri)
     end
 
     def uri
@@ -51,13 +82,13 @@ module Proxy::HttpRequest
 
   class Facts < ForemanRequest
     def post_facts(facts)
-      send_request('/api/hosts/facts',facts)
+      send_request(request_factory.create_post('/api/hosts/facts',facts))
     end
   end
 
   class Reports < ForemanRequest
     def post_report(report)
-      send_request('/api/reports',report)
+      send_request(request_factory.create_post('/api/reports',report))
     end
   end
 end
--- a/lib/smart_proxy.rb
+++ b/lib/smart_proxy.rb
@@ -44,6 +44,7 @@ module Proxy
   require 'root/root'
   require 'facts/facts'
   require 'dns/dns'
+  require 'templates/templates'
   require 'tftp/tftp'
   require 'dhcp/dhcp'
   require 'puppetca/puppetca'
@@ -76,7 +77,9 @@ module Proxy
     def http_app
       return nil if SETTINGS.http_port.nil?
       app = Rack::Builder.new do
-        ::Proxy::Plugins.enabled_plugins.each {|p| instance_eval(p.http_rackup)}
+        ::Proxy::Plugins.enabled_plugins.each do |p|
+          instance_eval(p.http_rackup)
+        end
       end
 
       Rack::Server.new(
--- a/modules/templates/handler.rb
+++ b/modules/templates/handler.rb
@@ -1,0 +1,26 @@
+require "proxy/util"
+require 'proxy/request'
+require 'uri'
+
+module Proxy::Templates
+
+  class Handler < ::Proxy::HttpRequest::ForemanRequest
+    extend Proxy::Log
+
+    def get_template(kind, token)
+      request = request_factory.create_get("/unattended/#{kind}", :token=> token)
+      res = send_request(request)
+
+      # You get a 201 from the 'built' URL
+      raise "Error retrieving #{kind} for #{token} from #{uri.host}: #{res.class}" unless ["200", "201"].include?(res.code)
+      Proxy::Log.logger.info "Template: request for #{kind} using #{token} at #{uri.host}"
+      res.body
+    end
+
+    def self.get_template kind, token
+      @handler ||= Handler.new
+      @handler.get_template(kind,token)
+    end
+  end
+
+end
--- a/modules/templates/templates_api.rb
+++ b/modules/templates/templates_api.rb
@@ -1,0 +1,22 @@
+require 'templates/handler'
+class Proxy::TemplatesApi < Sinatra::Base
+  helpers ::Proxy::Helpers
+  authorize_with_trusted_hosts
+
+  # When template feature is used, foreman uses this end-point to provide basse url for hosts to fetch templates.
+  # It will also modify the rendering of the foreman_url specified in the templates.
+  get "/templateServer" do
+    begin
+      content_type :json
+      {"templateServer" => (Proxy::Templates::Plugin.settings.template_url || "")}.to_json
+    rescue => e
+      log_halt 400, e
+    end
+  end
+
+  get "/:kind" do |kind|
+    log_halt(500, "Failed to retrieve #{kind} template for #{params[:token]}: ") do
+      Proxy::Templates::Handler.get_template(kind, params[:token])
+    end
+  end
+end
--- a/modules/templates/http_config.ru
+++ b/modules/templates/http_config.ru
@@ -1,0 +1,5 @@
+require 'templates/templates_api'
+
+map "/unattended" do
+  run Proxy::TemplatesApi
+end
--- a/modules/templates/templates.rb
+++ b/modules/templates/templates.rb
@@ -1,0 +1,1 @@
+require 'templates/templates_plugin'
--- a/modules/templates/templates_plugin.rb
+++ b/modules/templates/templates_plugin.rb
@@ -1,0 +1,21 @@
+module Proxy::Templates
+  class Plugin < ::Proxy::Plugin
+    http_rackup_path File.expand_path("http_config.ru", File.expand_path("../", __FILE__))
+    https_rackup_path File.expand_path("http_config.ru", File.expand_path("../", __FILE__))
+
+    default_settings :enabled => true
+    plugin :templates, ::Proxy::VERSION
+
+    after_activation do
+      unless Proxy::SETTINGS.foreman_url
+        logger.warn "missing :foreman_url: from configurations. 'templates' module is disabled"
+        ::Proxy::Plugins.disable_plugin(:templates)
+      end
+
+      unless Proxy::Templates::Plugin.settings.template_url
+        logger.warn "missing :templates_url: from configurations. 'templates' module is disabled"
+        ::Proxy::Plugins.disable_plugin(:templates)
+      end
+    end
+  end
+end
--- a/test/templates/template_api_test.rb
+++ b/test/templates/template_api_test.rb
@@ -1,0 +1,33 @@
+require 'test_helper'
+require 'json'
+require 'templates/templates_api'
+require 'templates/templates'
+
+class TemplateApiTest < Test::Unit::TestCase
+  include Rack::Test::Methods
+
+  def app
+    Proxy::TemplatesApi.new
+  end
+
+  def setup
+    Proxy::Templates::Plugin.settings.stubs(:template_url).returns("someproxy:8443")
+    Proxy::SETTINGS.stubs(:foreman_url).returns("someproxy:8443")
+    @args = { :token => "test-token" }
+  end
+
+  def test_api_can_return_templateserver
+    get "/templateServer"
+    assert last_response.ok?, "Last response was not ok"
+    data = JSON.parse(last_response.body)
+    assert_equal("someproxy:8443", data["templateServer"].to_s)
+  end
+
+  def test_api_can_ask_for_a_template
+    Proxy::Templates::Handler.stubs(:get_template).returns("An API template")
+    get "/provision", @args
+    assert last_response.ok?, "Last response was not ok"
+    assert_match("An API template", last_response.body)
+  end
+
+end
--- a/test/templates/template_test.rb
+++ b/test/templates/template_test.rb
@@ -1,0 +1,22 @@
+require 'test_helper'
+require 'uri'
+require 'net/http'
+require 'mocha'
+require 'templates/templates_plugin'
+require 'templates/handler'
+require 'webmock/test_unit'
+
+
+class TemplateTest < Test::Unit::TestCase
+  def setup
+    @foreman_url = 'https://foreman.example.com'
+    Proxy::SETTINGS.stubs(:foreman_url).returns(@foreman_url)
+  end
+
+  def test_template_requests_return_data
+    @expected_body = "my template"
+    stub_request(:get, @foreman_url+'/unattended/provision?token=test-token').to_return(:status => 200, :message => "OK", :body => @expected_body)
+    result =  Proxy::Templates::Handler.get_template('provision', 'test-token')
+    assert_equal(@expected_body, result)
+  end
+end
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -1,4 +1,6 @@
 require "test/unit"
+require 'fileutils'
+
 $: << File.join(File.dirname(__FILE__), '..', 'lib')
 $: << File.join(File.dirname(__FILE__), '..', 'modules')
 


sodabrew:
Oh sweet! Glad to see someone pick this up, I've had interest but no time recently.

Why do lib/proxy/template.rb and modules/templates/handler.rb have nearly identical contents? What's the difference between a Proxy::Template and a Proxy::Templates?

dustints:
@sodabrew thanks! that was an extra file I forgot to remove from https://github.com/theforeman/smart-proxy/pull/100

dustints:
can't tell exactly which cop of rubocop is failing. does anyone know?

dLobatog:
@dustint-rh Rubocop was just updated in https://github.com/theforeman/smart-proxy/commit/8aa4e0f94bdb0aaff7fc61f95fd55b06a72e98c1, before that tests wouldn't run, I think that could be the reason,

[test] to trigger the run again

modules/templates/handler.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    # Gets a template from Foreman
+    def self.get_template kind, token
+      # Parse the Foreman URI for a connection
+      uri              = URI.parse("#{Proxy::Templates::Plugin.settings.foreman_url}/unattended/#{kind}?token=#{token}")
+      req              = Net::HTTP::Get.new(uri.request_uri)
+      http             = Net::HTTP.new(uri.host, uri.port)
+      http.use_ssl     = uri.scheme == 'https'
+      # TODO: handle CA properly; for now we'll accept self-signed certs
+      http.verify_mode = OpenSSL::SSL::VERIFY_NONE if http.use_ssl?
```
sodabrew: What's the plan for validating the central Foreman's certificate?

modules/templates/templates_api.rb
```diff
@@ -0,0 +1,22 @@
...
+    begin
+      content_type :json
+      {"templateServer" => (Proxy::Templates::Plugin.settings.template_url || "")}.to_json
+    rescue => e
+      log_halt 400, e
+    end
+  end
+
+  get "/:kind" do |kind|
+    log_halt 403, "Proxy not configured to handle templates" unless Proxy::Templates::Plugin.settings.enabled
```
dmitri-d: This is redundant, as this url won't be available if the module is disabled.

modules/templates/templates_api.rb
```diff
@@ -0,0 +1,22 @@
...
+    log_halt 403, "No URI specified for :foreman_url:" unless Proxy::Templates::Plugin.settings.foreman_url
+    log_halt(404, "Failed to retrieve #{kind} template for #{params[:token]}: ") {
```
dmitri-d: Probably 403 is better here.

modules/templates/handler.rb
```diff
@@ -0,0 +1,27 @@
...
+      res              = http.start { |http| http.request(req) }
+
+      # You get a 201 from the 'built' URL
+      raise "Error retrieving #{kind} for #{token} from #{uri.host}: #{res.class}" unless ["200", "201"].include?(res.code)
+      logger.info "Template: request for #{kind} using #{token} at #{uri.host}"
+      res.body
+    end
+  end
+
+end
```
dmitri-d, line 26: There's a PR that makes it possible to reuse ForemanRequest (#223), which I expect is going to be merged shortly. Would be less duplication if you used that.

modules/templates/templates_api.rb
GregSutcliffe: Surely "page not found" is more appropriate than "permission denied" when the proxy cannot find a template for the token in question?

modules/templates/handler.rb
GregSutcliffe: I guess we should copy what the enc/report-processor do? They also need to initiate connections to Foreman

modules/templates/templates_api.rb
dmitri-d: I think either is a bit of a stretch here. I wonder if 500 is more appropriate? As for 403 ('Forbidden') vs. 404 (rfc2616): "If the request method was not HEAD and the server wishes to make public why the request has not been fulfilled, it SHOULD describe the reason for the refusal in the entity. If the server does not wish to make this information available to the client, the status code 404 (Not Found) can be used instead.".

The way I'm reading this is that if there's a reason (other than wrong url) why a certain url can not be accessed, 403 is more appropriate.

modules/templates/handler.rb
dustints: @sodabrew, @GregSutcliffe 
Updated to use ForemanRequest (#223) as suggested below. This makes the user define 
foreman_url,
foreman_ssl_cert,
foreman_ssl_key 
in the main settings.yml to use ssl

modules/templates/handler.rb
dustints, line 26: @witlessbird updated to use ForemanRequest (#223) but had to make some modifications to handle http GET requests.

modules/templates/templates_api.rb
dustints: changed to 500

dustints:
rebased

dmitri-d:
[test]

dmitri-d:
@dustint-rh: could you squash your commits pls?

config/settings.d/templates.yml.example
```diff
@@ -0,0 +1,10 @@
+# Set this to true if the Proxy should handle template requests on behalf of Foreman
+:enabled: true
+# Where is Foreman?
+# no need to specify http(s) here as templates are always retrieved on http
```
sodabrew: It looks like Addressable::URI in combination with Net::HTTP.new would allow a bare IP or hostname, or bare IP or hostname : port, in additional to a fully specified URI. I might suggest preventing non-fully-qualified URIs?

Suggested replacement commnet, because "templates are always retrieved on http" is a false statement (you can retrieve by https).

``` ruby
# The default protocol is http on port 80 unless otherwise specified in the url.
# Examples:
# foreman.example.com:3000
```

lib/proxy/request.rb
```diff
@@ -1,13 +1,63 @@
...
+    end
+
+    def add_headers(req)
+      req.add_field('Accept', 'application/json,version=2')
+      req.content_type = 'application/json'
+      req
+    end
+
+    def create_post(path, body)
+      req = Net::HTTP::Post.new(absolute_path(path))
```
dustints: fixme: absolute_path(path) to uri(path)

lib/proxy/request.rb
```diff
@@ -1,13 +1,63 @@
...
 module Proxy::HttpRequest
+  class ForemanRequestFactory
+    def initialize(base_uri)
+      @base_uri = base_uri
+    end
+
+    def create_get(path, query)
+      uri = uri(path)
+      uri.query_values = query
+      req = Net::HTTP::Get.new(uri)
```
dustints: fixme: call to add_headers

dmitri-d:
[test]

config/settings.d/templates.yml.example
dustints: @sodabrew updated the comment.
and removed the `:foreman_url` option since we will be using the one from settings.yml

dustints:
@witlessbird, @sodabrew, @GregSutcliffe squashed and updated to get passing tests.

lib/proxy/request.rb
```diff
@@ -1,23 +1,57 @@
 require 'net/http'
 require 'net/https'
+require 'addressable/uri'
```
dmitri-d: Why do you require this dependency?

lib/proxy/request.rb
```diff
@@ -1,23 +1,57 @@
...
 require 'uri'
```
dmitri-d, line 3: Is this still needed if 'addressable/uri' is used?

lib/proxy/request.rb
```diff
@@ -26,7 +60,8 @@ def http
 
     private
     def http_init
-      http             = Net::HTTP.new(uri.host, uri.port)
+      port             = uri.port || 443 if uri.scheme == 'https'
```
dmitri-d: `|| 443 if uri.scheme == 'https'` is redundant.

lib/proxy/request.rb
```diff
@@ -26,7 +60,8 @@ def http
...
+      http             = Net::HTTP.new(uri.host, port || uri.port)
```
dmitri-d: `http  = Net::HTTP.new(uri.host, uri.port)` is sufficient?

lib/smart_proxy.rb
```diff
@@ -44,6 +44,7 @@ module Proxy
   require 'root/root'
   require 'facts/facts'
   require 'dns/dns'
+  require 'templates/templates'
```
dmitri-d, line 47: Are we going to ship this in the core? @domcleal, @ohadlevy

ohadlevy, line 47: I think it make sense, so +1 from my side.

domcleal, line 47: Probably a plugin first in its state, it's what's happening for other features.

lib/proxy/request.rb
dustints: @witlessbird in addressable::uri, it allows you to pass a hash to set the query instead you having to manually url encode your query. (mostly just for L16 below)

modules/templates/templates_api.rb
```diff
@@ -0,0 +1,21 @@
...
+    begin
+      content_type :json
+      {"templateServer" => (Proxy::Templates::Plugin.settings.template_url || "")}.to_json
+    rescue => e
+      log_halt 400, e
+    end
+  end
+
+  get "/:kind" do |kind|
+    log_halt 403, "No URI specified for :foreman_url:" unless Proxy::Templates::Plugin.settings.foreman_url
```
dmitri-d: I don't think we formally agreed on this, but I think (and I tried to follow this logic in other parts of smart-proxy) it's better to fail-fast: if a required piece of configuration is missing, the module shouldn't be loaded at all, and an appropriate message should be logged.

dmitri-d: @domcleal, what do you think?

lib/proxy/request.rb
dustints: @witlessbird Addressable::URI doesn't seem as smart as URI when it comes to setting the right port.

```
> Addressable::URI.parse('https://google.com').port
=> nil

> URI.parse('https://www.google.com').port
=> 443
```

modules/templates/templates_api.rb
```diff
@@ -0,0 +1,21 @@
+require 'templates/handler'
+class Proxy::TemplatesApi < Sinatra::Base
+  helpers ::Proxy::Helpers
+  authorize_with_trusted_hosts
+
+  get "/templateServer" do
```
dmitri-d, line 8: Probably worth adding a comment re: why we need this api point.

lib/proxy/request.rb
```diff
@@ -1,23 +1,57 @@
...
-      req = Net::HTTP::Post.new(URI.join(uri.to_s, path).path)
+  class ForemanRequestFactory
+    def initialize(base_uri)
+      @base_uri = base_uri
+    end
+
+    def create_get(path, query={})
+      uri = uri(path)
+      uri.query_values = query
+      req = Net::HTTP::Get.new(uri.path)
```
dustints: uri.path should be uri.request_uri instead of uri.path

modules/templates/templates_api.rb
GregSutcliffe, line 8: Indeed. For clarity, it's because hosts on a private provisioning network may have a different name/ip to contact the proxy than the one Foreman knows about. eg

```
host -- 10.1.1.0/24 -- proxy -- 172.20.10.0/24 -- foreman
foreman talks to proxy.company.lan on 172.20.10.5
host talks to proxy.provision.company.lan on 10.1.1.1
```

So we can't simply make this a Setting in the UI - the proxy itself must be responsible for declaring where the hosts should get templates from. Foreman can then use this info in `renderer#foreman_url` to return an appropriate template to the host (via the proxy).

Dustin, can you add an abbreviated form of the above to the PR?

dustints, line 8: @GregSutcliffe yup, will do

lib/proxy/request.rb
dmitri-d: what about http/port 80?

modules/templates/templates_api.rb
GregSutcliffe, line 8: Forgot to mention, this is very similar to what we do in the TFTP api, where we have a /bootServer endpoint, which the _DHCP_ orchestration in Foreman uses to write next-ip data in the host reservations.

EDIT: sorry, /serverName - https://github.com/theforeman/smart-proxy/blob/develop/modules/tftp/tftp_api.rb#L69-L71

lib/smart_proxy.rb
dmitri-d, line 47: I think I'm with @domcleal on this one -- I think the original plan was to ship all modules as gems...

modules/templates/templates_api.rb
dustints: reminder to also sub in favor of Proxy::SETTINGS.foreman_url over Proxy::Templates::Plugin.settings.foreman_url

Gemfile
```diff
@@ -3,6 +3,7 @@ source 'http://rubygems.org'
 gem 'json'
 gem 'sinatra'
 gem 'rack', '>= 1.1'
+gem 'addressable'
```
GregSutcliffe: Is this essential? We try to keep proxy deps to a minimum

config/settings.d/templates.yml.example
```diff
@@ -0,0 +1,11 @@
...
+:enabled: true
+
+# Where to obtain templates?
+# what is the base url to obtain templates when provisioning.
+# The default protocol is http on port 80 unless otherwise specified in the url.
+# Examples:
+# foreman.example.com:3000
+# https://foreman.example.com
+# http://1.2.3.4:3000
+#:template_url: https://capsulefqdn:8443
```
GregSutcliffe: You also need to copy `:foreman_url` as well as `:template_url` from my original PR. One is the URL the proxy uses to contact foreman, the other is the URL that should be written into the templates so that the host knows where to find the proxy.

modules/templates/handler.rb
```diff
@@ -0,0 +1,26 @@
...
+require 'proxy/request'
+require 'uri'
+
+module Proxy::Templates
+
+  class Handler < ::Proxy::HttpRequest::ForemanRequest
+    extend Proxy::Log
+
+    def get_template(kind, token)
+      request = request_factory.create_get("/unattended/#{kind}", :token=> token)
```
GregSutcliffe, line 11: This doesn't work, i get host as `nil`. If i add logging of the uri object to request_factory.create_request I get:

```
I, [2014-10-29T17:18:52.945476 #12172]  INFO -- : uri.scheme = 
I, [2014-10-29T17:18:52.945558 #12172]  INFO -- : uri.host   = 
I, [2014-10-29T17:18:52.945635 #12172]  INFO -- : uri.port   = 
I, [2014-10-29T17:18:52.945752 #12172]  INFO -- : uri.path   = /unattended/provision
```

Not good. I tried changing this line to `"#{Proxy::Templates::Plugin.settings.foreman_url}/unattended/#{kind}"` which correctly gave:

```
I, [2014-10-29T17:20:05.681750 #12604]  INFO -- : uri.scheme = http
I, [2014-10-29T17:20:05.681812 #12604]  INFO -- : uri.host   = 192.168.122.1
I, [2014-10-29T17:20:05.681878 #12604]  INFO -- : uri.port   = 3000
I, [2014-10-29T17:20:05.681934 #12604]  INFO -- : uri.path   = /unattended/provision
```

but i'm still getting connection refused. Foreman is definitely listening on 192.168.122.1:3000 - investigating the connection refused now.

lib/proxy/request.rb
GregSutcliffe: This just caught me out too, can you update please?

GregSutcliffe: Actually, in my testing uri.host is still getting unset because we're only passing the path. why are we not using `Net::HTTP::Get.new(uri)` ?

modules/templates/templates_api.rb
GregSutcliffe: +1 for fail fast, i've spent the last hour debugging a connection refused because of this :(

lib/smart_proxy.rb
GregSutcliffe, line 47: +1, we will cut this into a gem once the code is ready and working.

lib/proxy/request.rb
GregSutcliffe: Ah i see, the ForemanRequest class references :foreman_url in the main Settings file, but that setting hasn't been added in this PR, so it's assumed to be nil and fails. See comments below about fail-fast for missing settings, and please add the example setting to the PR.

config/settings.d/templates.yml.example
GregSutcliffe: Scratch this, and see my comment below about adding this to the main config file as per ForemanRequest

modules/templates/templates_api.rb
dustints: @GregSutcliffe, @witlessbird is there an example of fail fast?

lib/proxy/request.rb
dustints: if the uri is 'https://google.com/'
the request it would make to the server would be
GET https://google.com/ 
instead of 
GET /

updating since we're moving back to URI instead of Addressable::URI

i saw that it was present in settings.yml in develop branch...so thought it would be  to use it.
going to use a fail fast( couldn't find an existing example yet).

modules/templates/handler.rb
dustints, line 11: the first one looks right after the create_request
because it gets the scheme, host, port form the already created http connection.

dustints, line 11: (minus the missing query params) that is fixed now.

GregSutcliffe, line 11: Yeah, my confusion was the :foreman_url setting moving to the main settings file. Once I set that the rest of the code started working :)

lib/proxy/request.rb
dustints, line 3: removed addressable for uri

config/settings.d/templates.yml.example
```diff
@@ -0,0 +1,11 @@
...
+:enabled: true
+
+# This allows the proxy to define how hosts that are being provisioned where to
+# obtain the templates from.
+# The default protocol is http on port 80 unless otherwise specified in the url.
+# Examples:
+# capsule.example.com:8080
+# https://capsule.example.com
+# http://1.2.3.4:3000
+#:template_url: https://capsulefqdn:8443
```
GregSutcliffe: nitpick: "capsule" shouldn't be used upstream, it's the downstream productname. "smart_proxy_fqdn" would be better.

config/settings.d/templates.yml.example
```diff
@@ -0,0 +1,11 @@
...
+:enabled: true
+
+# This allows the proxy to define how hosts that are being provisioned where to
+# obtain the templates from.
+# The default protocol is http on port 80 unless otherwise specified in the url.
+# Examples:
+# smart-proxy.example.com:8080
+# https://smart-proxy.example.com
+# http://1.2.3.4:3000
+#:template_url: https://capsulefqdn:8443
```
GregSutcliffe: [done] could you add a comment here reminding users to ensure the :foreman_url setting is set in the config file? I forgot to do that, and it caused me no end of trouble :)

modules/templates/templates_api.rb
```diff
@@ -0,0 +1,24 @@
...
+    begin
+      content_type :json
+      {"templateServer" => (Proxy::Templates::Plugin.settings.template_url || "")}.to_json
+    rescue => e
+      log_halt 400, e
+    end
+  end
+
+  get "/:kind" do |kind|
+    #TODO remove to fail fast.
```
dmitri-d: @dustints, @GregSutcliffe -- an example of disabling a plugin on startup: https://github.com/theforeman/smart-proxy/blob/develop/modules/facts/facts_plugin.rb#L8

lib/smart_proxy.rb
GregSutcliffe, line 47: Given the oddball situation that would occur if this were a plugin (code merged to Foreman core depending on a separately packaged proxy plugin), and given the Foreman renderer code would need rewriting for the core side code to be a plugin, I'm changing my vote to align with Ohad - this should be merged in to the smart-proxy rather than packaged separately. I think it's sufficiently useful to users to segregate networks to have it merged in (but disabled) by default.

modules/templates/templates_api.rb
GregSutcliffe: @witlessbird perfect, looks like what we need. @dustints can you oblige? :)

dustints:
@GregSutcliffe, @witlessbird  yup, just updated

modules/templates/templates_api.rb
GregSutcliffe: @dustints nice!

```
W, [2014-11-03T17:44:42.091978 #4895]  WARN -- : missing :foreman_url: from configurations. 'templates' module is disabled
W, [2014-11-03T17:44:42.092099 #4895]  WARN -- : missing :templates_url: from configurations. 'templates' module is disabled
```

GregSutcliffe:
@dustints looks good, I'll give the whole thing a run through tomorrow, and then I think we're done.

modules/templates/templates_api.rb
dmitri-d: This TODO can be removed now.

lib/proxy/request.rb
```diff
@@ -1,19 +1,52 @@
 require 'net/http'
 require 'net/https'
 require 'uri'
+require 'cgi'
 
 # TODO: need settings validation on startup, otherwise we get a 500 error due to missing/wrong config settings when api is accessed
```
dmitri-d: This TODO can be removed now that plugin validates presence of foreman_url .

dmitri-d:
@dustints, @GregSutcliffe: lgtm.

lib/proxy/request.rb
GregSutcliffe: I'll remove on merge

config/settings.d/templates.yml.example
ekohl: Still a mention of capsule here btw.

lib/smart_proxy.rb
ekohl, line 47: Makes sense, especially since it doesn't pull in additional dependencies.

dustints:
@GregSutcliffe @ekohl @witlessbird 
removed TODOs and removed 'capsule' from config template.

GregSutcliffe:
Merged as a53d835, with a slight change to ensure the plugin is disabled by default. Thanks @dustints!

PR #226, opened 2014-10-22
Set default modulepath to '/etc/puppet/modules:/usr/share/puppet/modules'

soliverr:
When I setted up puppetmaster by [theforeman/puppet module](https://forge.puppetlabs.com/theforeman/puppet), I got default puppetmaster configuration without _environments_ (there is always _production_ enviromment by default).

As I can see, the Linux distributors placed packaged puppet modules in _/usr/share/puppet/modules/_ directory.

But _smart-proxy_ search modules in _/etc/puppet/modules_ only by default.

It may be usefull to add _/usr/share/puppet/modules/_ into default search path.

### DIFF
--- a/modules/puppet_proxy/environment.rb
+++ b/modules/puppet_proxy/environment.rb
@@ -58,7 +58,7 @@ class Proxy::Puppet::Environment
       end
       if env.values.compact.size == 0
         # fall back to defaults - we probably don't use environments
-        env[:production] = conf[:main][:modulepath] || conf[:master][:modulepath] || '/etc/puppet/modules'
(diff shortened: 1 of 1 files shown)


dmitri-d:
@domcleal, @GregSutcliffe: what do you guys think?

domcleal:
I think it's correct, matching the same default as Puppet.  Though I really wouldn't recommend relying on it - explicit configuration is preferable, and will be required in Puppet 4 with the move to directory environments.

[test]

soliverr:
I have updated my puppetmaster to 3.7.2 version and configured it to use directory environments. Now _smart-proxy_ behaves as expected, Foreman imports all modules sucessfully.

Really, this patch does not make a sense.

Thank you.

PR #227, opened 2014-10-22
fixes #7352 - missing DNS record returns 404

shlomizadok:


modules/dns/providers/virsh.rb
```diff
@@ -48,7 +48,7 @@ def find_ip_for_host host
           return e.parent.attributes["ip"]
         end
       end
-      raise Proxy::Dns::Error.new("Cannot find DNS entry for #{host}")
+      raise Proxy::Dns::NotFound.new("Cannot find DNS entry for #{host}")
```
dmitri-d, line 58: Not going to work here, as the exception will be recast as `Proxy::Dns::Error`.

dmitri-d:
@shlomizadok: ping?

dmitri-d:
looks good; @shlomizadok, could you add tests that verify we throw `Proxy::Dns::NotFound` exception when record isn't found?

shlomizadok:
@witlessbird , added some tests for a beginning. and all are :green_apple:

dmitri-d:
merged at 3f4ee4b. Thanks, @shlomizadok!

PR #228, opened 2014-10-23
fixes #7832 - add SSL verification tests for production Rack apps

domcleal:
Split out the launcher into its own file, then we can generate a rack app
from an integration test.  We launch the app and then can make real requests
against it, in order to test SSL.

---

There's still a bit of request logging on stdout I've not managed to squash yet...


domcleal:
Seems tests are really unreliable under Ruby 1.8.7, I need to look into exactly why.  Various SSL errors are popping up during the connection process.

dmitri-d:
looks good. Not sure what's going on with tests though; It might be that the webrick/sinatra doesn't have enough time to initialize?

domcleal:
Yeah, might be.  I don't think there's much harm in increasing the timeout, since it's only going to be hit in an error case.

domcleal:
PR rebased and updated.

Intermittent errors are fixed, probably as Ruby 1.8.7 is no longer supported or tested. The refactoring previously needed is mostly in develop already (e.g. 335a161), so the patch is much smaller. The stderr logging noise has been fixed in the PR.

test/sinatra/ssl_client_verification_integration_test.rb
```diff
@@ -0,0 +1,78 @@
...
+  def test_https_no_cert
+    launch_https
+    http = Net::HTTP.new('localhost', @settings.https_port)
+    http.use_ssl = true
+    http.verify_mode = OpenSSL::SSL::VERIFY_NONE
+    res = http.get('/')
+    assert_kind_of Net::HTTPForbidden, res
+  end
+
+  def test_https_wrong_cert
```
dmitri-d: A nitpick: might be helpful to add a comment as to why the client with the certificate used in this test should be rejected.

dmitri-d:
Thanks for the update, Looks very good. Let me know re: comment, I'll merge it then.

test/sinatra/ssl_client_verification_integration_test.rb
domcleal: Thanks, I've updated the name of the test to state that the cert's from a different authority.

dmitri-d:
Merged, thanks @domcleal.

PR #229, opened 2014-10-24
Fixes #8006 - added STDOUT proxy log option

lzap:
Although we have it documented in our example file, this was never working:

```
```

Although one can set this to `/dev/stdout`, this does not work on foreman
discovery image under systemd environment because stdout cannot be reopened
there (Permission error). I would like have the possibility to setup STDOUT
this way, because I want to capture the output into systemd journal in the
discovery image environment.

By default, systemd captures stdout/stderr and adds entries to the journal
under the foreman-proxy tag. Therefore on Fedoras and RHEL7+ we can perhaps
check "proper syslog support" for foreman-proxy as well.

### DIFF
--- a/lib/proxy/log.rb
+++ b/lib/proxy/log.rb
@@ -11,8 +11,13 @@ module Proxy
     end
 
     def self.logger
-      # We keep the last 6 10MB log files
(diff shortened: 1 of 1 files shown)


dmitri-d:
looks good, will test in a bit.

dmitri-d:
merged in at 3172820. thank you, @lzap!

PR #233, opened 2014-11-05
fixes #8282: sslv3 is disabled

dmitri-d:


dmitri-d:
you can test the patch by enabling ssl and executing
`nmap --script ssl-enum-ciphers -p 8443 localhost`

lib/smart_proxy.rb
```diff
@@ -27,6 +27,94 @@
 require 'sinatra/ssl_client_verification'
 require 'sinatra/trusted_hosts'
 
+# Workaround for ruby CVE-2014-3566: creates safe defaults for SSLContext
+# also see https://www.ruby-lang.org/en/news/2014/10/27/changing-default-settings-of-ext-openssl/
```
ohadlevy: Can we move it to another file?

domcleal:
This should probably be configurable, it isn't the first, and won't be the last time that people will need to change ciphers or protocols in response to research.

dmitri-d:
@domcleal: the first bit starting at line https://github.com/witlessbird/smart-proxy/blob/poodle-8282/lib/smart_proxy.rb#L36 was taken verbatim from http://svn.ruby-lang.org/cgi-bin/viewvc.cgi?revision=48121&view=revision, affects openssl default settings, and eventually (when the upstream fix makes it into most of rubys?) should be removed. Do we really care about sslv3 (or v2 for that matter) that much? There isn't a way to enable/disable a particular cipher in Webrick -- I'm not keen on introducing a custom SSL option for that, so we'll have to enable/disable this patch.

dmitri-d:
Another approach might be not to touch ciphers at all, but disable SSLv3-related options in SSLContext, will test that now.

dmitri-d:
Nope, ciphers need to be restricted.

ohadlevy:
@domcleal :+1: from my side

ohadlevy:
merged as 6ddd00d thanks @witlessbird

PR #235, opened 2014-11-07
Fixes #8320 - shutdown shell provider concurrency issue

lzap:
Unfortunately the `shutdown` command does not support finer granularity than
minutes. So we will waste one minute everytime here. Better ideas?

We need to make sure the response gets returned back. Maybe to spawn a shell
with sleep 5s && shutdown...

### DIFF
--- a/modules/bmc/shell.rb
+++ b/modules/bmc/shell.rb
@@ -30,25 +30,35 @@ module Proxy
       end
 
       def powercycle
-        # search for sudo
-        sudo = which("sudo")
+        sudo = which('sudo')
+        shutdown = which('shutdown')
 
         unless sudo
-          logger.warn "sudo binary was not found - aborting"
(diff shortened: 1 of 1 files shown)


lzap:
Ok after some time playing with a shell wrapper, I found that the most reliable way is to actually spawn new thread, let the Sinatra to finish the request (giving it 5 seconds grace period) and calling shutdown normally.

lzap:
Testers wanted, now, who wants to reboot the dev machine, huh? ;-)

lzap:
Tip for reviewer: To test this, use this command: `curl -3 -H "Accept:application/json" -H "Content-Length:0" -k -X PUT http://192.168.122.52:8443/bmc/ignored/chassis/power/cycle`

dmitri-d:
lgtm.

dmitri-d:
merged at 81c679e. Thanks, @lzap!

PR #236, opened 2014-11-10
Fixes #8210 - Adding caching for smart-proxy puppet classes using Stefan Julin's implementation of cache

orrabin:


domcleal:
`Fixes #8210`, missing # prevents Redmine linking.

(I'd also move the attribution to Stefan into the second part of the commit message, it's rather long.  "smart-proxy" is redundant too.)

orrabin:
@domcleal sorry, added # to comment

modules/puppet_proxy/puppet_cache.rb
```diff
@@ -0,0 +1,64 @@
+require 'json'
+
+module Proxy::Puppet
+  class PuppetCache
+    extend Proxy::Log
+    class << self
+      def read_cache directory, environment, parser, scanner
+        logger.info("Running scan_directory on #{environment}: #{directory}")
+
+        cachefile= "#{Proxy::SETTINGS.cache_location}/cache_#{environment}.json"
```
dmitri-d: pls. use `File.expand_path("cache_#{environment}.json", Proxy::SETTINGS.cache_location)` instead

dmitri-d: also, cache_location shouldn't be a global setting, but rather a puppet-proxy module one. Could you add a sensible default value for it to https://github.com/theforeman/smart-proxy/blob/develop/modules/puppet_proxy/puppet_plugin.rb and make a note of it in https://github.com/theforeman/smart-proxy/blob/develop/config/settings.d/puppet.yml.example?

dmitri-d:
Could you add tests to cover modules/puppet_proxy/puppet_cache.rb, in particular cache expiration bits?

orrabin:
@witlessbird can I add the tests directly to the class_scanner tests or do you want it to have it's own test? it would be easier to use the class scanners but I can stub them if needed

config/settings.d/puppet.yml.example
```diff
@@ -41,3 +41,7 @@
 # Override use of Puppet's API to list environments, by default it will use only if
 # environmentpath is given in puppet.conf, else will look for environments in puppet.conf
 #:puppet_use_environment_api: true
+
+# Cache options
+#:use_cache: false
+#:cache_location: '~/.cache/smart-proxy/puppet_cache'
```
lzap: While this is sane default, I think we do not want to use `/usr/share/foreman` for the production instance. Therefore I think we should patch our installer to set this value to something like `/var/lib/foreman/cache` or better `/var/cache/foreman/puppet`.

orrabin: I'll set it here and in the defaults, `/var/cache/foreman/puppet` is the best path?

domcleal: It shouldn't be 'foreman' at all, this is the smart proxy.

I think it'd be best to have it default (or even be hardcoded) to the installation dir + '/cache/', then in the packaging we can symlink /usr/share/foreman-proxy/cache to /var/cache/foreman-proxy in a similar way to how we do other directories (inc in Foreman core).

lzap: > I think it'd be best to have it default (or even be hardcoded) to the
> installation dir + '/cache/', then in the packaging we can symlink
> /usr/share/foreman-proxy/cache to /var/cache/foreman-proxy in a
> similar way to how we do other directories (inc in Foreman core).

This sounds like a plan.

Just make the default value relative to the installation dir ("cache/")
and then raise a packaging PR to create the symlink. It's one liner, I
can help you with that.

LZ

## 

Later,
 Lukas #lzap Zapletal

dmitri-d: agree with @domcleal.

dmitri-d:
@orrabin: I think a suite dedicated to puppet_cache would be a proper place for cache-specific tests. You can also add tests that verify that class scanners do in fact use cache, and those should go into corresponding class scanner suites.

config/settings.d/puppet.yml.example
```diff
@@ -41,3 +41,7 @@
 # Override use of Puppet's API to list environments, by default it will use only if
 # environmentpath is given in puppet.conf, else will look for environments in puppet.conf
 #:puppet_use_environment_api: true
+
+# Cache options
+#:use_cache: false
+#:cache_location: 'cache/'
```
orrabin: @lzap, @domcleal please make sure this is what you meant and then I'll open up a packaging issue to symlink /usr/share/foreman-proxy/cache to /var/cache/foreman-proxy

domcleal: @orrabin yep.  I'd suggest adding cache/.gitkeep to git so the directory already exists for people, and perhaps make caching enabled by default?  (In the puppet_plugin.rb, and set this .example file up to match it.)

orrabin:
@domcleal I added cache/.gitkeep and opened the issue: http://projects.theforeman.org/issues/8390

@witlessbird I added tests for puppet_cache.
I needed to make some changes to help stub as we discussed so this needs re-review

modules/puppet_proxy/puppet_plugin.rb
```diff
@@ -3,7 +3,7 @@ class Plugin < Proxy::Plugin
     http_rackup_path File.expand_path("http_config.ru", File.expand_path("../", __FILE__))
     https_rackup_path File.expand_path("http_config.ru", File.expand_path("../", __FILE__))
   
-    default_settings :puppet_provider => 'puppetrun', :puppetdir => '/etc/puppet'
+    default_settings :puppet_provider => 'puppetrun', :puppetdir => '/etc/puppet', :use_cache => true, :cache_location => 'cache/'
```
dmitri-d: This cache location path is relative to the current working directory, which is problematic as it can change. Lets put cache directory in the root directory of the smart-proxy installation by default; you can use something like: `File.expand_path("cache", File.expand_path("../../", __FILE__))`.

test/puppet/class_scanner_eparser_test.rb
```diff
@@ -141,9 +146,34 @@ class foreman::install (
...
   end
 
-  def test_should_handle_import_in_a_manifest
+  def test_should_handle_import_in_a_manifest_without_cache
     return unless Puppet::PUPPETVERSION.to_f >= 3.2
-    klasses =  Proxy::Puppet::ClassScannerEParser.scan_directory('./test/fixtures/modules_include')
+    Proxy::Puppet::Plugin.load_test_settings(:use_cache => false)
+
+    klasses =  Proxy::Puppet::ClassScannerEParser.scan_directory('./test/fixtures/modules_include', "example_env")
+    assert_kind_of Array, klasses
```
dmitri-d: Not sure if this is a useful check -- I think size check is sufficient...

test/puppet/class_scanner_eparser_test.rb
```diff
@@ -141,9 +146,34 @@ class foreman::install (
...
+    assert_equal 2, klasses.size
+
+    klass = klasses.find {|k| k.name == "sub::foo" }
+    assert klass
```
dmitri-d, line 157: `assert klasses.any? { |k| k.name == 'sub::foo' }`

dmitri-d, line 157: nm, you make additional asserts on klass. You can use `#any?` in the assert below though.

test/puppet/class_scanner_test.rb
```diff
@@ -60,12 +60,25 @@ class foreman::params {
     assert_equal "foreman", klass.module
   end
 
-  def test_should_scan_a_dir
-    klasses =  Proxy::Puppet::ClassScanner.scan_directory('/tmp/no_such_dir')
+  def test_should_scan_a_dir_without_cache
```
dmitri-d: I know you didn't write this test, but -- what is being tested here? That ClassScanner returns an empty array for a non-existent directory? If that's useful then let's rename the test, otherwise let's kill it.

test/puppet/class_scanner_test.rb
```diff
@@ -60,12 +60,25 @@ class foreman::params {
...
+    Proxy::Puppet::Plugin.load_test_settings(:use_cache => true)
+    Proxy::Puppet::PuppetCache.stubs(:read_from_cache).returns('./test/fixtures/modules_include' =>
+                                                               { 'testinclude' =>
+                                                               {:timestamp => Time.now,
+                                                                :manifest => [[Proxy::Puppet::PuppetClass.new('testinclude')],
+                                                                              [Proxy::Puppet::PuppetClass.new('testinclude::sub::foo')]] }})
+    Proxy::Puppet::PuppetCache.stubs(:write_to_cache)
+    klasses =  Proxy::Puppet::ClassScanner.scan_directory('./test/fixtures/modules_include', "example_env")
+    assert_kind_of Array, klasses
+    assert !klasses.empty?
```
dmitri-d: How do you know that the cache has been used? You'll either need to verify the contents of `:manifest`, or (probably easier) use a mock instead of a stub. The latter would allow you to verify whether a given call was made as expected.

test/puppet/puppet_cache_test.rb
```diff
@@ -0,0 +1,70 @@
...
+    Proxy::Puppet::PuppetCache.stubs(:write_to_cache)
+  end
+
+  def test_should_refresh_cache_when_dir_is_not_in_cache
+    Proxy::Puppet::PuppetCache.stubs(:read_from_cache).returns({})
+    cache = Proxy::Puppet::PuppetCache.scan_directory_with_cache('./test/fixtures/modules_include', 'example_env', nil, @scanner)
+
+    assert_kind_of Array, cache
+    # scan_manifest is called twice and always returns 2 puppet classes
+    assert_equal 4, cache.size
```
dmitri-d: That doesn't sound right -- depending on whether cache is used Proxy::Puppet::ClassScanner will return two or four classes (see https://github.com/theforeman/smart-proxy/pull/236/files#diff-c81c1889c5ab9d5f600d7c22c4f43e5bR157)?

test/puppet/puppet_cache_test.rb
```diff
@@ -0,0 +1,70 @@
...
+
+    klass = cache.find { |k| k.name == "sub::foo" }
+    assert cache
+    assert_equal "testinclude", klass.module
+
+    klass = cache.find { |k| k.name == "testinclude" }
+    assert klass
+  end
+
+  def test_should_not_refresh_cache_when_dir_is_updated
```
dmitri-d: The name is a bit confusing.  test_should_not_refresh_cache_when_cache_is_more_recent perhaps?

test/puppet/puppet_cache_test.rb
orrabin: this has to do with the stub for `scan_manifest`
when cache is refreshed the scanner is stubed so scan_manifest always returns two classes, since `scan_manifest` is called twice the result will have 4 classes.
The reason for calling `scan_manifest` twice is that it is called here:

``` ruby
  modulemanifest = Dir.glob("#{path}/manifests/**/*.pp").map do |filename|
              scanner.scan_manifest File.read(filename), parser, filename
            end
```

and there are two file names that return from `Dir.glob`
when the cache doesn't need refresh then scan_minifest won't be called resulting in just two classes - those stubed in the cache

config/settings.d/puppet.yml.example
```diff
@@ -41,3 +41,7 @@
 # Override use of Puppet's API to list environments, by default it will use only if
 # environmentpath is given in puppet.conf, else will look for environments in puppet.conf
 #:puppet_use_environment_api: true
+
+# Cache options
+#:use_cache: true
+#:cache_location: 'cache/'
```
orrabin: @witlessbird can the example stay relative after changing the default to `File.expand_path("cache", File.expand_path("../../", __FILE__))`?

test/puppet/class_scanner_test.rb
```diff
@@ -60,12 +60,24 @@ class foreman::params {
     assert_equal "foreman", klass.module
   end
 
-  def test_should_scan_a_dir
-    klasses =  Proxy::Puppet::ClassScanner.scan_directory('/tmp/no_such_dir')
+  def test_should_scan_and_return_empty_array_when_directory_does_not_exist
```
orrabin, line 62: @witlessbird I'm not sure how useful this is but I changed the name. 
I can remove it or change it to a scan test without use of the cache for an existing dir, what do you prefer?

test/puppet/puppet_cache_test.rb
dmitri-d: Somehow I missed of stubbing of `scan_manifest` in setup. You can have a stub/mock return different values on subsequent invocations: `@scanner.stubs(:scan_manifest).returns([[Proxy::Puppet::PuppetClass.new('testinclude')])
         .then.returns([Proxy::Puppet::PuppetClass.new('testinclude::sub::foo')]])`. See http://gofreerange.com/mocha/docs/Mocha/Expectation.html#then-instance_method for another example. 
I'm not sure it's worth stubbing/mocking `scan_manifest` though -- we are touching the filesystem in this test anyway, and any test double makes tests fragile.

config/settings.d/puppet.yml.example
dmitri-d: I think `/var/lib/smart-proxy` would be a better place for this. You can also make a note about the default location.

orrabin:
@witlessbird  thanks for the comments and the help with mocking.
I've fixed the cache location and the tests

modules/puppet_proxy/puppet_cache.rb
```diff
@@ -0,0 +1,75 @@
...
+          JSON.parse(File.read(cachefile))
+        else
+          {}
+        end
+      end
+
+      def write_to_cache cache, environment
+        cachefile = File.expand_path("cache_#{environment}.json", Proxy::Puppet::Plugin.settings.cache_location)
+        tmpfile = cachefile + '.tmp'
+        File.open(tmpfile, 'w') { |file| file.write(cache.to_json) }
```
dmitri-d: This fails if the cache directory doesn't exist.

modules/puppet_proxy/puppet_plugin.rb
```diff
@@ -3,7 +3,8 @@ class Plugin < Proxy::Plugin
     http_rackup_path File.expand_path("http_config.ru", File.expand_path("../", __FILE__))
     https_rackup_path File.expand_path("http_config.ru", File.expand_path("../", __FILE__))
   
-    default_settings :puppet_provider => 'puppetrun', :puppetdir => '/etc/puppet'
+    default_settings :puppet_provider => 'puppetrun', :puppetdir => '/etc/puppet', :use_cache => true,
+                     :cache_location => File.expand_path("cache", File.expand_path("../../", __FILE__))
```
dmitri-d: looks like the path to cache dir needs another '.../' -- it expands to <smart_proxy_dir>/modules/cache, instead of <smart_proxy_dir>/cache.

modules/puppet_proxy/class_scanner.rb
```diff
@@ -1,17 +1,14 @@
...
 module Proxy::Puppet
-  class ClassScanner
+  class ClassScanner < ClassScannerBase
     class << self
       # scans a given directory and its sub directory for puppet classes
       # returns an array of PuppetClass objects.
-      def scan_directory directory
-
+      def scan_directory directory, environment
         parser = Puppet::Parser::Parser.new Puppet::Node::Environment.new
```
dmitri-d: 'parser' appears to be a transient parameter: it's not being used in ClassScannerBase#scan_directory, but rather is passed on to ClassScanner.scan_manifest/ClassScannerEParser.scan_manifest. You can instantiate correct parser in those methods instead.

dmitri-d:
Sorry I noticed these three issues just now. But other than these, looks good!

modules/puppet_proxy/puppet_plugin.rb
orrabin: how can I test this? 
when I run it locally the path is evaluated to where I expected

dmitri-d: hmm. I noticed this when I hit '/environments/:environment/classes' when cache dir doesn't exist...

modules/puppet_proxy/puppet_plugin.rb
```diff
@@ -2,8 +2,9 @@ module Proxy::Puppet
   class Plugin < Proxy::Plugin
     http_rackup_path File.expand_path("http_config.ru", File.expand_path("../", __FILE__))
     https_rackup_path File.expand_path("http_config.ru", File.expand_path("../", __FILE__))
-  
-    default_settings :puppet_provider => 'puppetrun', :puppetdir => '/etc/puppet'
+
+    default_settings :puppet_provider => 'puppetrun', :puppetdir => '/etc/puppet', :use_cache => true,
+                     :cache_location => File.expand_path("cache", File.expand_path("../../", __FILE__))
```
orrabin: @witlessbird didn't mean to remove the comments from here. 
I'll try to check this again and see when cache is created in models/cache

modules/puppet_proxy/puppet_plugin.rb
```diff
@@ -2,8 +2,9 @@ module Proxy::Puppet
   class Plugin < Proxy::Plugin
     http_rackup_path File.expand_path("http_config.ru", File.expand_path("../", __FILE__))
     https_rackup_path File.expand_path("http_config.ru", File.expand_path("../", __FILE__))
-  
-    default_settings :puppet_provider => 'puppetrun', :puppetdir => '/etc/puppet'
+
+    default_settings :puppet_provider => 'puppetrun', :puppetdir => '/etc/puppet', :use_cache => true,
+                     :cache_location => File.expand_path("cache", File.expand_path("../../../", __FILE__))
```
orrabin, line 7: @witlessbird I didn't notice I changed `puppet.yml` setting locally for testing
checked this again and it was missing a `../`

orrabin:
@witlessbird I fixed all your comments, can you think of any more tests I should add to avoid the errors you found? (I stubbed write_to_cache so cache existence wan't tested)

modules/puppet_proxy/puppet_cache.rb
```diff
@@ -0,0 +1,77 @@
...
+
+        if File.exist?(cachefile)
+          JSON.parse(File.read(cachefile))
+        else
+          {}
+        end
+      end
+
+      def write_to_cache cache, environment
+        if File.directory?(Proxy::Puppet::Plugin.settings.cache_location)
```
dmitri-d: Cache file will not be written to if cache_location directory doesn't exist or isn't a directory. I think you need to verify if the directory exists and create it if it doesn't.

modules/puppet_proxy/puppet_cache.rb
```diff
@@ -0,0 +1,77 @@
...
+          cachefile = File.expand_path("cache_#{environment}.json", Proxy::Puppet::Plugin.settings.cache_location)
+          tmpfile = cachefile + '.tmp'
+          File.open(tmpfile, 'w') { |file| file.write(cache.to_json) }
```
dmitri-d, line 76: Just realized that this will be an issue during multiple concurrent requests to '/environments/blah/classes'. I think synchronizing this whole block is too coarse -- no point for waiting on end of write to a cache file for environment 'A' if we are creating cache for environment 'B'. Might be worth using a file lock: attempt to acquire a lock on a temp file, proceed with an update if acquired, and sleep-wait if not. You can find an example of file locking here: https://github.com/theforeman/smart-proxy/blob/develop/lib/proxy/http_downloads.rb

modules/puppet_proxy/puppet_cache.rb
```diff
@@ -0,0 +1,87 @@
...
+        end
+      end
+
+      def write_to_cache cache, environment
+        cache_dir = Proxy::Puppet::Plugin.settings.cache_location
+        FileUtils.mkdir_p(cache_dir) unless File.directory?(cache_dir)
+
+        cachefile = File.expand_path("cache_#{environment}.json", cache_dir)
+
+        lock =  Proxy::HttpDownloads.try_locking(cachefile)
```
orrabin: I'd like to move this and `unlock` to some helper but I need it both in `lib/proxy/http_downloads.rb` and `modules/puppet_proxy/puppet_cache.rb`. where is the best place to put it?

modules/puppet_proxy/puppet_cache.rb
```diff
@@ -0,0 +1,88 @@
...
+        while lock.nil?
+         sleep 1
```
orrabin: @witlessbird how much time should this wait? I don't want to delay the response that much, maybe skipping the `write_to_cache` is better then waiting?

modules/puppet_proxy/puppet_cache.rb
```diff
@@ -0,0 +1,88 @@
...
+        if File.exist?(cachefile)
+          JSON.parse(File.read(cachefile))
+        else
+          {}
+        end
+      end
+
+      def write_to_cache cache, environment
+        cache_dir = Proxy::Puppet::Plugin.settings.cache_location
+        FileUtils.mkdir_p(cache_dir) unless File.directory?(cache_dir)
```
orrabin, line 69: I'd like to lock this too, should I change the `try_lock` to support directories which means a change to `Proxy::HttpDownloads` or write a new `try_lock_dir`?

dmitri-d, line 69: Not sure if it's worth it -- FileUtils.mkdir_p skips dir creation if it already exists?

modules/puppet_proxy/puppet_cache.rb
dmitri-d: maybe somthing like Proxy::FileLock? I'd place it in `lib/proxy/file_lock.rb`

modules/puppet_proxy/puppet_cache.rb
dmitri-d: That's a very good question, I did wonder about this too. Chances are that the changes being written to the file are the same we just found, so yeah, let's try skipping the write?

modules/puppet_proxy/puppet_cache.rb
orrabin, line 69: It skips creation if the dir exists so I'll leave this for now

modules/puppet_proxy/puppet_cache.rb
orrabin: even if the changes are different it will try to write again next time the cache is used so I prefer to skip writing and not wait

orrabin:
@witlessbird I fixed all your comments

dmitri-d:
Merged in at 51c455b. Thanks, @orrabin!

PR #237, opened 2014-11-16
Fixes #8347 - facts feature should not be enabled by default

orrabin:


lzap:
To give you some background - facts_api is only used on the discovery image, this should been turned off by default. Anyway, the plugin is not loaded when `facter` gem is missing.

LGTM except the bundler.d change, we don't need this for discovery I think. But if @domcleal don't mind adding this new group let's have it there (we will likely need a packaging change tho).

dmitri-d:
The idea is to use the facter gem that is available on the system, not to install ours. Could you remove  `bundler.d/facts.rb`? Looks good otherwise.

domcleal:
That's what bundler_ext is for, and this is the same pattern as in core.  It looks correct to me.

orrabin:
@domcleal I removed `bundler.d/facts.rb` before seeing your comment.
@witlessbird what if facter gem isn't available on the system? shouldn't we install?

dmitri-d:
If facter isn't available we don't load the plugin: https://github.com/theforeman/smart-proxy/blob/develop/modules/facts/facts_plugin.rb#L8

lzap:
Yes @witlessbird is right, I forgot. Looks sane, thank you. Merging as fc67f91.

PR #239, opened 2014-12-01
Fixes #8538 - ISC dhcp plugin now respets binding

lzap:


dmitri-d:
@lzap: could you add tests that exercise 'binding state' and 'binding next state' set to 'abandoned'? @GregSutcliffe: mind taking another look at this?

lzap:
> @lzap: could you add tests that exercise 'binding state' and 'binding next state' set to 'abandoned'? @GregSutcliffe: mind taking another look at this? 

Actually "abandoned" is not a valid lease state for ISC DHCP (there are
only "free" and "active"). The state was never used and I removed it
from the code.

There is still a bug in abandoned leases handling, but this is a bigger
issue that has something to do with unused IPs. I filed another issue
for this particular problem:

http://projects.theforeman.org/issues/8557

We need to correctly parse the abandoned leases (it's not a state, but a
different statement) and also make sure that unused_ip code does not
return such a leases. Different story.

## 

Later,
 Lukas #lzap Zapletal

GregSutcliffe:
@witlessbird sure, i'll get to this today.

test/dhcp/dhcp_api_test.rb
```diff
@@ -54,13 +54,20 @@ def test_api_02_get_network
...
+          "ip"     => "192.168.122.2",
+          "mac"    => "00:aa:bb:cc:dd:ee",
+          "starts" => date_format("Sat Jul 12 10:08:29 UTC 2014"),
+          "ends"   => nil,
+          "state"  => "active"
+        },{
+          "ip"     => "192.168.122.89",
+          "mac"    => "ec:f4:bb:c6:ca:fe",
+          "starts" => "2014-10-16 12:59:40 UTC",
+          "ends"   => "2199-01-01 00:00:01 UTC",
```
GregSutcliffe: You forgot `date_format()` on these two lines, so the tests fail on 1.8.7

modules/dhcp/providers/server/isc.rb
```diff
@@ -92,7 +92,13 @@ def loadSubnetData subnet
           body.split(";").each do |data|
             opts.merge! parse_record_options(data)
           end
-          next if opts[:state] == "free" || opts[:state] == "abandoned" || opts[:mac].nil?
+          next if opts[:mac].nil?
+          # delete expired (explicitly) expired leases
+          if (opts[:state] == "free" || (opts[:next_state] == "free" && opts[:ends] && opts[:ends] < Time.now))
```
GregSutcliffe: Rubocop failure: Don't use parentheses around the condition of an `if`.

GregSutcliffe:
A few small test failures, but otherwise looks good! I'll try to test it first thing tomorrow, so if you can fix the test failures, we're golden!

lzap:
Thanks, pushed.

lzap:
Ok so I pushed again, I found another scenario with conflicts. This fixes both problems and adds unit tests for both. Please re-test.

lzap:
Correction: lease were not overlapping by time, we just did not check dupes.

lzap:
Warning: When/If backporting to Staypuft version, note this line:

```
conf = format((@config+@leases).split("\n"))
```

Other than that, manual merge is straightforward.

lzap:
Staypuft backport for the record is: http://sprunge.us/iREI

lzap:
We still does not handle { deleted; } statement correctly. Need to drop few more use cases.

lzap:
Our ISC parsing was in _bad shape_.

lzap:
Just to put us on the same level @ohadlevy, from the dhcpd.leases man page:

_Format_

Lease descriptions are stored in a format that is parsed by the same recursive descent parser used to read the dhcpd.conf(5) and dhclient.conf(5) files. Lease files can contain lease declarations, and also group and subgroup declarations, host declarations and failover state declarations. Group, subgroup and host declarations are used to record objects created using the OMAPI protocol.
The lease file is a log-structured file - whenever a lease changes, the contents of that lease are written to the end of the file. This means that it is entirely possible and quite reasonable for there to be two or more declarations of the same lease in the lease file at the same time. In that case, the instance of that particular lease that appears last in the file is the one that is in effect.

Group, subgroup and host declarations in the lease file are handled in the same manner, except that if any of these objects are deleted, a rubout is written to the lease file. This is just the same declaration, with { deleted; } in the scope of the declaration. When the lease file is rewritten, any such rubouts that can be eliminated are eliminated. It is possible to delete a declaration in the dhcpd.conf file; in this case, the rubout can never be eliminated from the dhcpd.leases file.

lzap:
Ok I've added one more test for deleted reservation handling. This was handled correctly, so I only covered it with a test. We still do not handle "aborted" statement, but I don't see this as something important so this was filed as a different issue (see above).

Ready for review and testing!

test/dhcp/dhcp_api_test.rb
```diff
@@ -147,42 +168,6 @@ def test_api_08_create_record_new
...
-      } ],
-      "leases"       => [ {
-        "ip"         => "192.168.122.2",
-        "mac"        => "00:aa:bb:cc:dd:ee",
-        "starts"     => date_format("Sat Jul 12 10:08:29 UTC 2014"),
-        "ends"       => nil,
-        "state"      => "active",
-      } ]
-    }
-    assert_equal expected, data
```
GregSutcliffe, line 185: why delete all this? this is the check that a new record was actually written to the file....

test/dhcp/dhcp_api_test.rb
```diff
@@ -86,7 +107,7 @@ def test_api_04_get_network_unused_ip_again
     get "/192.168.122.0/unused_ip"
     assert last_response.ok?, "Last response was not ok: #{last_response.status} #{last_response.body}"
     data = JSON.parse(last_response.body)
-    expected = {"ip"=>"192.168.122.4"}
+    expected = {"ip"=>"192.168.122.5"}
```
GregSutcliffe: this change would be unecessary if you didn't use 192.168.122.4 in your new reservation data below.

GregSutcliffe:
Couple of minor test points, but the code looks solid. I'm going to attempt to run this on my proxy now, testing ISC is always fun.

test/dhcp/dhcp_api_test.rb
GregSutcliffe, line 185: ah, this just duplicates https://github.com/theforeman/smart-proxy/pull/239/files#diff-b596f53a47adeb92f29cac23dc40f021L47 - fine.

modules/dhcp/providers/server/isc.rb
```diff
@@ -204,9 +232,15 @@ def find_record_by_hostname subnet, hostname
       end
     end
 
+    def find_record_by_ip subnet, ip
+      subnet.records.find do |v|
```
GregSutcliffe, line 236: This is wrong, you should do `subnet.leases.find`. If you have a lease file with both a reservation and an expired lease for the same IP, then this will delete both. An example lease file would be:

```
host quux.elysium.emeraldreverie.org {
  dynamic;
  hardware ethernet 52:54:00:31:a3:97;
  fixed-address 172.20.10.53;
        supersede server.next-server = ac:14:0a:01;
        supersede host-name = "quux.elysium.emeraldreverie.org";
}
lease 172.20.10.53 {
  starts 0 2014/04/06 15:01:31;
  ends 0 2014/04/06 16:00:09;
  tstp 0 2014/04/06 16:00:09;
  cltt 0 2014/04/06 15:01:31;
  binding state free;
  hardware ethernet 00:1d:60:a5:b1:2a;
}
```

Using `subnet.records`, the reservation will be parsed, added to `subnet` and then the lease parsing will hit line 107 (because state is 'free') and come here via line 108. This method will return the reservation, and the it will be deleted, even though it's a valid reservation.

Could you fix and add a test?

GregSutcliffe:
One small issue with accidentally deleting reservations, and then we can merge :)

modules/dhcp/providers/server/isc.rb
GregSutcliffe, line 236: Actually, find_record_by_subnet is probably a useful helper later on, perhaps we should do

```
record = find_record_by_ip(subnet, ip).find{|r| r.type == :lease}
```

at line 108?

lzap, line 236: Good catch. Hmm, `find` only returns the first occurance, this will not work. As I just learned, there is a helper method `kind` that compares records by it's class. I've used this approach, added your example to the test case. Thanks.

One day, I will rewrite DHCP from scratch. I have couple of ideas how to make this fast.

ohadlevy, line 236: please share!?

GregSutcliffe:
Merged as a7c8c34, thanks @lzap!

sodabrew:
Nice work, @lzap!

PR #241, opened 2014-12-18
Fixes #8748 - Allow per-module enabling of http(s)

GregSutcliffe:
Defaults to http=off and https=on for all except the templates module which is the reverse. Not added to RootPlugin as that should always be on anyway.

I'm wondering if it makes sense to do something more condensed like `:enabled = http|https|true(both)|false(none)` ?

### DIFF
--- a/lib/proxy/plugin.rb
+++ b/lib/proxy/plugin.rb
@@ -66,12 +66,20 @@ class ::Proxy::Plugin
       @after_activation_blk = blk
     end
 
+    def http_enabled?
+      [true,'http'].include?(self.settings.enabled)
+    end
+
     def http_rackup_path(path)
-      @get_http_rackup_path = path
+      @get_http_rackup_path = path if http_enabled?
+    end
+
+    def https_enabled?
+      [true,'https'].include?(self.settings.enabled)
     end
 
     def https_rackup_path(path)
-      @get_https_rackup_path = path
+      @get_https_rackup_path = path if https_enabled?
     end
 
     def dependencies
--- a/config/settings.d/bmc.yml.example
+++ b/config/settings.d/bmc.yml.example
@@ -1,5 +1,7 @@
 ---
+# Can be true, false, or http/https to enable just one of the protocols
 :enabled: false
+
 # Available providers:
 # - freeipmi / ipmitool - requires the appropriate package installed, and the rubyipmi gem
 # - shell - for local reboot control (requires sudo access to /sbin/shutdown for the proxy user)
--- a/config/settings.d/dhcp.yml.example
+++ b/config/settings.d/dhcp.yml.example
@@ -1,5 +1,7 @@
 ---
+# Can be true, false, or http/https to enable just one of the protocols
 :enabled: false
+
 # valid vendors:
 #   - isc
 #   - native_ms (Microsoft native implementation)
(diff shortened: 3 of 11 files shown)


domcleal:
This needs a migration of some sort, as I'd guess a user upgrading is simply going to get http/https both set to nil by default and their services won't be configured?

Also, the default configs seem to assume HTTPS, which isn't the case for the other default settings.

GregSutcliffe:
@domcleal good point - perhaps enabling more options in :enabled is the way to go, then we don't have to do migrations?

modules/bmc/bmc_plugin.rb
```diff
@@ -1,8 +1,8 @@
 module Proxy::BMC
   class Plugin < Proxy::Plugin
-    http_rackup_path File.expand_path("http_config.ru", File.expand_path("../", __FILE__))
-    https_rackup_path File.expand_path("http_config.ru", File.expand_path("../", __FILE__))
-  
+    http_rackup_path File.expand_path("http_config.ru", File.expand_path("../", __FILE__)) if Proxy::BMC::Plugin.settings.http
```
dmitri-d: I think the check should be done inside http_rackup_path/https_rackup_path methods.

dmitri-d:
@GregSutcliffe: I like your idea of multi-value `enable`, I think enabling a module is a bit too verbose atm.

modules/bmc/bmc_plugin.rb
GregSutcliffe: thats a good idea, but how would I know which setting file to check (i.e whats the plugin name)? as far as i can see, http_rackup_path doesn't care what the plugin namespace is, and while I can guess it from the filename (File.expand_path... has /modules/dhcp/ in it for example) the namespace isn't consisent (Proxy::TFTP::Plugin vs Proxy::DhcpPlugin for example)

domcleal: You're _inside_ Proxy::BMC::Plugin already, so you have `settings`

GregSutcliffe: Nevermind, I'm being stupid, it's an inherited method, so self.settings works fine :facepalm:

GregSutcliffe:
@domcleal @witlessbird thanks, you're right the new version is much shorter :)

Thoughts on tests for this? I guess I need to add some stuff to plugin_test.rb, declaring the new enabled options, but the issue is that plugin is always enabled (it shows up in /features) but only responds to one of the protocols - do I need to stub and then make a request and ensure i get a 404 for the disabled protocol?

dmitri-d:
lgtm (tested). On a related note, should the installer configure modules with 'enabled: https' instead of 'enabled: true'?

dmitri-d:
Merged in at 5984657. Thanks, @GregSutcliffe!

lib/proxy/plugin.rb
```diff
@@ -66,12 +66,20 @@ def after_activation(&blk)
       @after_activation_blk = blk
     end
 
+    def http_enabled?
+      [true,'http'].include?(self.settings.enabled)
+    end
+
     def http_rackup_path(path)
-      @get_http_rackup_path = path
+      @get_http_rackup_path = path if http_enabled?
```
ohadlevy, line 74: this is broken, as http_rackup_path is called before the plugin method is called, settings will always be empty and the test does not actually test the real scenario :( took some time to figure out :)

PR #245, opened 2015-01-12
Fixes #8905 - bind_port config option

lzap:


domcleal:
Incidentally, do we set RACK_ENV in our packaging correctly for production use?

dmitri-d:
@domcleal: we are overriding RACK_ENV at https://github.com/theforeman/smart-proxy/blob/develop/lib/smart_proxy.rb#L42 and setting it to :production. 

Having said that:
- :env has been declared obsolete and should be replaced with :environment 
- I wonder if we should rely on RACK_ENV instead.

dmitri-d:
@lzap: testing...

domcleal:
@witlessbird yeah, presumably that's not entirely working since 1.4 (although I've not noticed it personally).  If this PR's merged and setting RACK_ENV becomes necessary, then packaging (init scripts?) will need updating too.

dmitri-d:
@lzap: works as advertised. Could you define a default value for `:host` in https://github.com/theforeman/smart-proxy/blob/develop/lib/proxy/settings/global.rb and mention it in the comments in `settings.yml.example`?

config/settings.yml.example
```diff
@@ -30,7 +30,9 @@
 # Uncomment and modify if you want to change the default pid file '/var/run/foreman-proxy/foreman-proxy.pid'
 #:daemon_pid: /var/run/foreman-proxy/foreman-proxy.pid
 
-# HTTP ports configuration
+# host and ports configuration
+# host to bind ports to (default: 0.0.0.0 for production, localhost for development)
```
domcleal: only true under certain versions of sinatra?

dmitri-d: also, we set environment to production, so it always defaults to 0.0.0.0. As I suggested in my other comment, I think it would be better to define the default value in global settings, and override it in the config file, as with other settings.

domcleal: Yeah, though I think the point here is that sometimes it _isn't_ defaulting to 0.0.0.0?  like sinatra's ignoring https://github.com/lzap/smart-proxy/blob/binding-config-8905/lib/smart_proxy.rb#L42

dmitri-d: Right, that too. I wonder if it's related to our use of `:env` there instead of `:environment`.

dmitri-d:
@lzap: ping?

config/settings.yml.example
lzap: Ok so I think we agree that default value is always `0.0.0.0` for production so I will remove the comment and just leave there the default commented value. Development is not that important and people can just override.

lzap:
Rebased.

domcleal:
That looks more sensible, thanks.

May I make a nitpick and ask that `'*'` becomes the default value?  This way it'll listen on IPv6 interfaces too.

lzap:
> May I make a nitpick and ask that '*' becomes the default value?  This way it'll listen on IPv6 interfaces too.

Changed, in the example YAML as well.

## 

Later,
 Lukas #lzap Zapletal

config/settings.yml.example
```diff
@@ -30,7 +30,9 @@
 # Uncomment and modify if you want to change the default pid file '/var/run/foreman-proxy/foreman-proxy.pid'
 #:daemon_pid: /var/run/foreman-proxy/foreman-proxy.pid
 
-# HTTP ports configuration
+# host and ports configuration
+# host to bind ports to (Sinatra 1.4+ defaults to localhost for development setup)
```
domcleal: The Sinatra version comment's obsolete now, right?

dmitri-d: Could you list possible values for the `bind_host` parameter in the comment?

dmitri-d:
lgtm, other than a nitpick re: possible values

config/settings.yml.example
lzap: I don't think so. What do you want me to put in there?

lzap:
@witlessbird possible values? Can you elaborate?

dmitri-d:
@lzap -- something like: `# possible values: localhost, 0.0.0.0`

lzap:
Ok modified.

config/settings.yml.example
```diff
@@ -30,7 +30,9 @@
 # Uncomment and modify if you want to change the default pid file '/var/run/foreman-proxy/foreman-proxy.pid'
 #:daemon_pid: /var/run/foreman-proxy/foreman-proxy.pid
 
-# HTTP ports configuration
+# host and ports configuration
+# host to bind ports to (possible values: *, localhost, 0.0.0.0)
+#:bind_host: *
```
dmitri-d: Don't think you can use a `*` in a yaml file as a value. It probably can be escaped, but I'm not sure how. Maybe just keep `localhost` and `0.0.0.0`?

domcleal: It has to be quoted.

lzap: > It has to be quoted.

K

## 

Later,
 Lukas #lzap Zapletal

dmitri-d:
lgtm!

dmitri-d:
merged in at 9ea2ac7. Thanks, @lzap!

PR #246, opened 2015-01-14
Refactor trusted_hosts handling

lazyfrosch:
On HTTPS we will get the FQDN from the client certificate and check against the
list.

While on HTTP we will perform both reverse DNS and forward DNS lookup to verify
the client may talk to us.

Refs #7849

### DIFF
--- a/lib/proxy/helpers.rb
+++ b/lib/proxy/helpers.rb
@@ -1,3 +1,6 @@
+require 'openssl/x509'
+require 'resolv'
+
 module Proxy::Helpers
   include Proxy::Log
 
@@ -19,4 +22,51 @@ module Proxy::Helpers
     logger.debug exception.backtrace.join("\n") if exception.is_a?(Exception)
     halt code, message
   end
+
+  # read the HTTPS client certificate from the environment and extract its CN
+  def https_cert_cn
+    certificate_raw = request.env['SSL_CLIENT_CERT'].to_s
+    log_halt 403, 'could not read client cert from environment' if certificate_raw.empty?
+
+    begin
+      certificate = OpenSSL::X509::Certificate.new certificate_raw
+      if certificate.subject && certificate.subject.to_s =~ /CN=([^\s\/,]+)/i
+        $1
+      else
+        log_halt 403, 'could not read CN from the client certificate'
+      end
+    rescue OpenSSL::X509::CertificateError => e
+        log_halt 403, "could not parse the client certificate\n\n#{e.message}"
+    end
+  end
+
+  # reverse lookup an IP address while verifying it via forward resolv
+  def remote_fqdn(forward_verify=true)
+    ip = request.env['REMOTE_ADDR']
+    log_halt 403, 'could not get remote address from environment' if ip.empty?
+
+    begin
+      dns = Resolv.new
+      fqdn = dns.getname(ip)
+    rescue Resolv::ResolvError => e
+      log_halt 403, "unable to resolve hostname for ip address #{ip}\n\n#{e.message}"
+    end
+
+    unless forward_verify
+      fqdn
+    else
+      begin
+        forward = dns.getaddresses(fqdn)
+      rescue Resolv::ResolvError => e
+        log_halt 403, "could not forward verify the remote hostname - #{fqdn} (#{ip})\n\n#{e.message}"
+      end
+
+      if forward.include?(ip)
+        fqdn
+      else
+        log_halt 403, "untrusted client has no matching forward DNS lookup - #{fqdn} (#{ip})"
+      end
+    end
+  end
+
 end
--- a/lib/sinatra/trusted_hosts.rb
+++ b/lib/sinatra/trusted_hosts.rb
@@ -1,22 +1,27 @@
-require 'resolv'
-
 module Sinatra
   module TrustedHosts
     def authorize_with_trusted_hosts
       helpers ::Proxy::Helpers
 
       before do
-        # When :trusted_hosts is given, check the reverse DNS entry of the remote IP and ensure it's listed
-        if Proxy::SETTINGS.trusted_hosts
-          begin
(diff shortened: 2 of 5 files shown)


lazyfrosch:
Tested in my environment, with both HTTPS from an IP that was no reverse lookup, and connection from localhost via trusted_hosts and ip.

domcleal:
[test] although we'll probably have to update the trusted hosts test.

lazyfrosch:
Are you using a patched version of "test-unit"?

I can't find a method "sorted" nowhere for that gem?

My test fails like this:

```
.../test/unit/test-suite-creator.rb:51:in `collect_test_names': undefined method `sort_test_names_in_sorted_order' for #<Test::Unit::TestSuiteCreator:0x00000003de8d68> (NoMethodError
```

domcleal:
No, I've not seen that before!

dmitri-d:
Looks good, a couple of notes:
- could you extract branches of the condition at https://github.com/theforeman/smart-proxy/pull/246/files#diff-d00662426e39c5a083182f33fc43ccb2R19 into methods? A bit hard to read as it is atm..
  - could you add some tests for both http and https (let me know if you need help with this)

lib/sinatra/trusted_hosts.rb
```diff
@@ -6,17 +7,59 @@ def authorize_with_trusted_hosts
...
+              log_halt 403, "Untrusted client has no matching forward DNS lookup - #{remote_fqdn} (#{remote_ip}). If it's to be a trusted host, ensure it has a reverse DNS and a matching forward entry." +
+              "\n\n" + "#{e.message}"
+            end
+
           end
+
           if !Proxy::SETTINGS.trusted_hosts.include?(remote_fqdn.downcase)
             log_halt 403, "Untrusted client #{remote_fqdn.downcase} attempted to access #{request.path_info}. Check :trusted_hosts: in settings.yml"
           end
+
```
lzap, line 24: Why do we just throw the `remote_fqdn` out on line 37. I understand that client can send any certificate he wants with arbitrary CN, but at least for correct queries both CA name and reverse lookup should both match.

lazyfrosch, line 24: Reverse Lookup is not necessarily valid at all times, e.g. I have a VPN between all my root servers, there is no valid reverse-forward lookup.

And also in other areas like dynamic ip addressing might reverse lookup be more or less a nice to have.

The certificate is still validated if it is signed by the Puppet CA.

dmitri-d:
Re: failing tests (also see https://github.com/sinatra/sinatra/issues/951) : Looks like we either need to temporary pin rack to `~> 1.5.0`, or use sinatra from master:HEAD. @domcleal -- I think the former is a bit easier, considering that I don't know how temporary "temporary" is.

domcleal:
Can we pin to < 1.6?  We have some older versions in circulation...

dmitri-d:
Yeah, I think so, will check (and do a PR).

dmitri-d:
https://github.com/theforeman/smart-proxy/pull/249 should fix tests failures.

lazyfrosch:
Yeah, now actual errors. Will do an update these days.

dmitri-d:
[test]

lazyfrosch:
So, I've moved the actual X509 and Resolv stuff into the helpers, we now have a smaller TrustedHosts class.

I've added tests to validate resolving and cert handling.

The forward verify can also be disabled by config, in special environments where this is not possible.

Please review.

domcleal:
[test]

domcleal:
[test] some intermittent GitHub errors.

lazyfrosch:
I addressed the rubocop warnings...

[test] - nice try, I know :D

test/sinatra/trusted_hosts_test.rb
```diff
@@ -28,27 +28,69 @@ def test_trusted_hosts_empty
...
     Proxy::SETTINGS.expects(:trusted_hosts).at_least_once.returns(['host.example.org'])
     Resolv.any_instance.expects(:getname).with('10.0.0.1').raises(Resolv::ResolvError, 'no name')
     get '/test', nil, 'REMOTE_ADDR' => '10.0.0.1'
     assert last_response.forbidden?
   end
 
+  def test_trusted_hosts_https_matches
+    Proxy::SETTINGS.expects(:trusted_hosts).at_least_once.returns(['host.example.org'])
+    OpenSSL::X509::Certificate.stubs(:new).returns(OpenSSL::X509::Certificate)
+    OpenSSL::X509::Certificate.expects(:subject).twice.returns('CN=host.example.org,...')
```
dmitri-d: you can probably relax the number of calls constraint to `at_least_once` here.

dmitri-d:
[test]

dmitri-d:
[test]

dmitri-d:
lgtm. Will test this in a bit. @domcleal: ping?

lib/proxy/helpers.rb
```diff
@@ -19,4 +22,51 @@ def log_halt code=nil, exception=nil
...
+
+    begin
+      certificate = OpenSSL::X509::Certificate.new certificate_raw
+      if certificate.subject && certificate.subject.to_s =~ /CN=([^\s\/,]+)/i
+        $1
+      else
+        log_halt 403, 'could not read CN from the client certificate'
+      end
+    rescue OpenSSL::X509::CertificateError => e
+        log_halt 403, "could not parse the client certificate\n\n#{e.message}"
```
domcleal, line 39: Nitpick, some additional whitespace here.

domcleal:
:+1:  looks great!

dmitri-d:
tested, lgtm.

dmitri-d:
merged in at 30aff66. Thanks, @lazyfrosch!

domcleal:
Thanks @lazyfrosch, this will be a great help for users without rDNS set up!

lazyfrosch:
@domcleal glad you like it :)

PR #247, opened 2015-01-14
Remove implicit test ordering from DHCP API test

lazyfrosch:
I had an conflict with test-unit on my system.

test-unit uses alphabetic by default, only minitest supports
"sorted", which is default there.

refs #8948 - Problem with implicit test ordering from DHCP API test

### DIFF
--- a/test/dhcp/dhcp_api_test.rb
+++ b/test/dhcp/dhcp_api_test.rb
@@ -11,11 +11,6 @@ ENV['RACK_ENV'] = 'test'
 class DhcpApiTest < Test::Unit::TestCase
   include Rack::Test::Methods
 
-  # Run tests in alphabetical order for this file
-  def self.test_order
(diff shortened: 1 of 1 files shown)


domcleal:
Makes sense!

[test]

dmitri-d:
lgtm

domcleal:
The only thing to do is to amend the message to start "fixes #8949", but that can also be done on merge.

lazyfrosch:
amended, I'm used to mostly use refs, noted in mind.

dmitri-d:
merged in at e3020e6. Thanks, @lazyfrosch!

domcleal:
The difference between fixes/refs for us is how Redmine's configured - it references and closes the ticket with fixes, but just adds the commit with refs.

dmitri-d:
@lazyfrosch: hah, beat you to it.

PR #255, opened 2015-02-05
Fixes #9245 - Add override setting to allow mcollective puppet runs to r...

logicminds:
- previously we could set the mcollective user via puppet_user setting. However, this could be misleading
   to the user as its not obvious where it was used in the smart proxy.  With the new mcollective user
   setting its pretty clear what the intention of the user is for.  In setups like Puppet Enterprise
   there is usually a user for mco and puppet which we have to keep separate and thus require a different user.
- This adds a new setting to the puppet config called :mcollective_user:

### DIFF
--- a/config/settings.d/puppet.yml.example
+++ b/config/settings.d/puppet.yml.example
@@ -33,6 +33,9 @@
 # Which user to invoke sudo as to run puppet commands
 #:puppet_user: root
 
+# If you want to override the puppet_user above just for mco commands
+#:mcollective_user: peadmin
+
 # URL of the puppet master itself for API requests
 #:puppet_url: https://puppet.example.com:8140
 # SSL certificates used to access the puppet master API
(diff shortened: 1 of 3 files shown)


dmitri-d:
[test]

logicminds:
I am going to refactor this slightly to allow for more universal mcollective run.  The hostname will not always be 100% accurate so we should rely on the ip instead.

https://github.com/logicminds/smart-proxy/blob/819c8805e33a2a967406c39e05f82fe5379f6849/modules/puppet_proxy/mcollective.rb#L26

``` shell
 /usr/bin/sudo -u peadmin /opt/puppet/bin/mco puppet runonce --with ipaddress=#{ipaddress}
```

logicminds:
Actually after thinking about this, I am going to keep as is.  If I added the ipaddress filter it would bind everyone to use the facts metadata plugin which some people might not have.

This is ready to merge as is.

logicminds:
@witlessbird is this ok to merge?

dmitri-d:
@logicminds: I'll take another look at this tomorrow, and will merge then. As a rule authors don't merge their own PRs.

dmitri-d:
merged in at 75f5871. Thanks, @logicminds!

PR #256, opened 2015-02-06
Fixes #7008: dns providers are now plugins

dmitri-d:


dmitri-d:
Ping @domcleal, @GregSutcliffe: settings are now in a dedicated file; Provider has access to main modules' settings via Plugin.main_module_settings call. Renamed provider_target into main_module. A miscellany of  other changes.

config/settings.d/dns.yml.example
```diff
@@ -8,6 +8,9 @@
 #   virsh
 #   dnscmd
 #:dns_provider: nsupdate
+#:nsupdate
```
dmitri-d: dns_server and dns_ttl shouldn't actually be here.

config/settings.d/dns_nsupdate.yml.example
```diff
@@ -0,0 +1,6 @@
+---
+# Can be true, false, or http/https to enable just one of the protocols
```
domcleal: would this remain true for providers?  Do providers obey enabled/disabled (and what happens if you set :dns_provider to a disabled provider?)

config/settings.d/dns_nsupdate.yml.example
```diff
@@ -0,0 +1,6 @@
...
+:enabled: true
+
+:dns_server: dns.domain.com
```
domcleal: I like all of this, but we need to think about migrations, again.

lib/proxy/plugin.rb
```diff
@@ -110,6 +134,14 @@ def default_settings(a_hash = {})
       @plugin_default_settings.merge!(a_hash)
     end
 
+    def main_module_settings
+      if main_module && provider_factory_proc
+        @main_module_settings ||= ::Proxy::Plugins.find_plugin(main_module).settings
+      else
+        raise ::Proxy::PluginMisconfigured, "Calling '::Proxy::Plugin.provider_target_settings' is unsupported for non-provider plugins.'"
```
domcleal: provider_target_settings?

modules/dns_nsupdate/dns_nsupdate_plugin.rb
```diff
@@ -0,0 +1,13 @@
+module ::Proxy::Dns::Nsupdate
+  class Plugin < ::Proxy::Plugin
+    plugin :nsupdate, ::Proxy::VERSION
+    default_settings :enabled => true, :dns_server => 'localhost', :dns_ttl => 86400
+
+    requires :dns, '~> 1.8.0'
```
domcleal: Maybe for core providers we should just use ::Proxy::VERSION so it's a 1:1 match.

lib/proxy/plugin.rb
```diff
@@ -110,6 +134,14 @@ def default_settings(a_hash = {})
       @plugin_default_settings.merge!(a_hash)
     end
 
+    def main_module_settings
+      if main_module && provider_factory_proc
```
domcleal: `if provider?` ?

modules/dns_nsupdate/dns_nsupdate_plugin.rb
```diff
@@ -0,0 +1,13 @@
+module ::Proxy::Dns::Nsupdate
+  class Plugin < ::Proxy::Plugin
+    plugin :nsupdate, ::Proxy::VERSION
```
domcleal: :dns_nsupdate?

domcleal:
I like it!

lzap:
Can we give at least some context when issue number is missing? :-)

I see `virsh` provider removed but I don't see the code being sent back. The same for other plugins... I guess this is WIP?

dmitri-d:
apologies, @lzap -- I treated this as a PoC, hence the general roughness. The idea is to convert module providers (like those found in dns and dhcp modules) into pugins, much like modules themselves. This would simplify development of new providers by the community (something that came up at fosdem).

domcleal:
Use http://projects.theforeman.org/issues/7008

dmitri-d:
@domcleal, @GregSutcliffe: a whole bunch of changes:
- providers and plugins are now separate classes, with a common mix-in
- simplified definition and registration of providers (mostly the same as plugins now)
- re-added virsh and dnscmd providers (still need to sort out nsupdate_gss)
- added configuration migration (for both pre-modularized and modularized smart-proxy configs)

config/settings.d/dns.yml.example
```diff
@@ -1,19 +1,10 @@
 ---
-# Can be true, false, or http/https to enable just one of the protocols
```
domcleal, line 2: Did you mean to remove those?

config/settings.d/dns.yml.example
```diff
@@ -1,19 +1,10 @@
...
 # Valid providers:
-#   nsupdate
-#   nsupdate_gss (for GSS-TSIG support)
-#   virsh
-#   dnscmd
+#   dns_nsupdate
+#   dns_nsupdate_gss (for GSS-TSIG support)
+#   dns_virsh
+#   dns_dnscmd
 #:dns_provider: nsupdate
```
domcleal: dns_nsupdate

domcleal:
I'm fine with the direction you've taken it, though I haven't done an in-depth review, it looks reasonable.

modules/dns/dns_api.rb
```diff
@@ -5,36 +5,8 @@ class Api < ::Sinatra::Base
...
-      when "virsh"
-        require 'dns/providers/virsh'
-        @server = Proxy::Dns::Virsh.new(opts.merge(
-          :virsh_network => Proxy::SETTINGS.virsh_network
-        ))
-      else
-        log_halt 400, "Unrecognized or missing DNS provider: #{Proxy::Dns::Plugin.settings.dns_provider || "MISSING"}"
-      end
+      factory = ::Proxy::Plugins.find_provider_factory(Proxy::Dns::Plugin.settings.dns_provider)
+      @server = factory.call(opts)
```
lzap: I wonder if it makes sense to cache the server to prevent finding it every single request:

```
@@server = factory.call(opts)
```

lzap:
I like that.

modules/dns/dns_api.rb
dmitri-d: Possibly, but I think it's better to delegate all responsibility for creation/returning of dns providers (in this case) to the factory itself, as only it knows all the nitty-gritty involved, including potential side effects and such.

config/settings.d/dns_dnscmd.yml.example
```diff
@@ -0,0 +1,8 @@
+---
+# Can be true, false
+:enabled: true
+
+# use this setting if you are managing a dns server which is not localhost though this proxy
```
GregSutcliffe, line 6: This is identical to the comment in nsupdate - can we make this easier to distinguish when I would use either?

extra/migrate_dns_settings.rb
```diff
@@ -0,0 +1,80 @@
+#!/usr/bin/ruby
+#
+# Script to migrate foreman-proxy settings.yml
+# to plugin form
+#
+# Greg Sutcliffe <gsutclif@redhat.com> 2014
```
GregSutcliffe: Pretty sure I didn't write this... ;)

config/settings.d/dns_dnscmd.yml.example
```diff
@@ -0,0 +1,8 @@
+---
+# Can be true, false
```
lzap: Nitpicking: I think it is obvious, but if you prefer to have it... ;-)

extra/migrate_dns_settings.rb
```diff
@@ -0,0 +1,80 @@
...
+# When running as a script
+if __FILE__ == $0 then
+  orig_file = ARGV[0] || '/etc/foreman-proxy/settings.d/dns.yml'
+  data      = YAML.load_file(orig_file)
+
+  output,unknown = migrate_dns_configuration(data.dup)
+  exit(1) unless should_save?(data, output) # it looks like we already have the new style settings
+  write_to_files(output,unknown)
+  exit(0)
+end
```
lzap: I would appreciate some help screen, or at least what to provide in comments in this script.

I would name this with `_1.9` suffix so it is obvious what version this is meant for.

lzap: Oh it looks like it is idempotent, no version needed then.

lzap:
There are still few minor comments by Greg, are you going to amend once again, or is this the final version?

domcleal:
This only touches DNS, it still needs expanding to the rest of the modules if the impl there is correct.

domcleal:
Migrations also need looking at and testing in more detail, as it seems to only load the main settings.yml file by default, but also accepts an argument.  Packaging may need updating to call it with the other config files (if that's how it's meant to be used) or it needs expanding to load other per-module files.

lzap:
@witlessbird can you briefly describe what your expectations are at this moment? to test this with DNS only?

domcleal:
Correction to my comment about migrations, I see there's an additional script... so we're going to get one for each module that's converted and call them all in sequence from packages?

config/settings.d/dns_dnscmd.yml.example
dmitri-d: Hah, I suppose, but in other places we also allow `http` and `https` values. I left the comment to keep it clear.

extra/migrate_dns_settings.rb
dmitri-d: I can add some help info here, although this is currently executed during package install...

dmitri-d:
@lzap: I started with DNS, but the idea is to modify other modules (DHCP is probably the next one in line) too. I was planning on creating dedicated PRs as I moved along.

@domcleal: Not sure re: migration scripts; we could do script per module, or update existing scripts to handle more modules. Whichever will be easier to use/maintain, I guess.

And finally, yep, I'm going to address @GregSutcliffe's comments.

domcleal:
Doing the modules in stages sounds fine to me, as long as they get done :smile:   The "puppet" module would be a good one to do too, as it has a lot of "runner" providers - though I'd like to ensure the internal command running API is good first.

I'm a bit concerned about having so many migration scripts, though I admit it will probably work.  While I really don't wish to overcomplicate things, I feel like we could do with AR's DB migrations for config files...

lzap:
Are those migration scripts really necessary since we use puppet to deploy Foreman. We can ask users just to rerun installer.

domcleal:
@lzap we support package upgrades usually, and in fact our installer simply doesn't work through major version updates right now so that can't be mandatory

dmitri-d:
@domcleal: I like the idea of AR DB migrations-like approach for configuration migrations, I'll give it some thought.

dmitri-d:
Added comments in dns provider config files and updated comments in extra/migrate_dns_settings.rb

lzap:
Ok testing migration and operation now.

config/settings.d/dns_nsupdate.yml.example
```diff
@@ -0,0 +1,16 @@
+---
+#
+# Configuration file for 'nsupdate' dns provider
+#
+# Can be true, false
+:enabled: true
```
lzap: Noticed one thing - we have dns_provider setting and enabled/disabled for all. Can't we just assume that the only enabled provider is the one user wants to use? And error out if the number of enabled providers is not 1.

lzap: I mean users need to do two steps to configure a provider, what happens when they setup disabled one? We should cover this with tests. I'd prefer to simplify it a bit more effectively getting rid of the dns.yml.

config/settings.d/dns_nsupdate.yml.example
```diff
@@ -0,0 +1,16 @@
...
+
+#:dns_key: /etc/rndc.key
+# use this setting if you are managing a dns server which is not localhost though this proxy
+#:dns_server: dns.domain.com
+# use this setting if you want to override default TTL setting (86400)
+#:dns_ttl: 86400
```
lzap: On the other hand the TTL setting is a good candidate to stay in the dns.yml global file. Virsh does not have it, but that's not the right reason to copy it everywhere. I mean virsh is very limited provider used for development. There is no TTL there...

lzap: To wrap this up:
- get rid of the dns_provider
- move dns_ttl to the dns.yaml file

lzap:
So my initial testing:

```
ruby extra/migrate_dns_settings.rb config/settings.d/dns.yml
```

Created dns.yml in my working directory, but did not create those subconfiguration files.

config/settings.d/dns_nsupdate.yml.example
dmitri-d: - Agree re: dns_ttl, and it's an easy fix.
- I think I'd actually prefer to keep dns_provider setting, but get rid of enable/disable in providers: this way no need to handle conflicts between providers.
- Will take a look at the migration issue.

lzap: Cool, that makes sense too.

Can you please file packaging changes required to get this working? I will do testing on live instances. Assuming there are changes necessary (the second migration script wont get triggered currently during package upgrade if I read that correctly).

dmitri-d:
a bunch more changes:
- only provider specified in use_provider config setting will be enabled (no need to edit provider config files to enable/disable individual providers)
- providers are now being configured before main modules (allows to disable the main module if provider isn't available/enabled)
- ttl setting now resides in dns plugin

@lzap: there are two migration scripts: one for old (pre-modules) style of config -- `migrate_settings.rb`, another for the current one -- `migrate_dns_settings.rb`.

dmitri-d:
also, see https://github.com/theforeman/foreman-packaging/pull/603 for rpm packaging-related changes

extra/migrate_settings.rb
```diff
@@ -83,6 +94,18 @@ def migrate(data)
   return output, data
 end
 
+def migrate_dns_provider_name(aname)
+  if recognized_dns_provider_name?(aname)
```
dLobatog: nitpick: (might be me)

``` ruby
return aname unless recognized_dns_provider_name?(aname)
'dns_' + aname 
```

reads a bit better

lib/proxy/pluggable.rb
```diff
@@ -0,0 +1,107 @@
...
+
+    def settings
+      @settings ||= Proxy::Settings.load_plugin_settings(plugin_default_settings, settings_file)
+    end
+
+    def initialize_after(*module_names)
+      @initialize_after ||= []
+      if module_names.empty?
+        to_return = @uses_provider? @initialize_after + [settings.use_provider] : @initialize_after
+        return to_return.map(&:to_sym)
```
dLobatog: nitpick - pointless return

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -17,22 +22,22 @@ def initialize options = {}
...
+            msg = "Added DNS entry #{@fqdn} => #{@value}"
+            cmd = "/RecordAdd #{zone} #{@fqdn}. A #{@value}"
+            execute(cmd, msg)
+          end
+        when "PTR"
+          if name = dns_find(@value)
+            raise(Proxy::Dns::Collision, "#{@value} is already used by #{name}") unless name == @fqdn
+          else
+            # TODO: determine reverse zone names, #4025
+            return true
```
dLobatog: nitpick: pointless return

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -41,16 +46,16 @@ def create
...
+        when "A"
+          raise Proxy::Dns::NotFound.new("Cannot find DNS entry for #{@fqdn}") unless dns_find(@fqdn)
+          zone = @fqdn.sub(/[^.]+./,'')
+          msg = "Removed DNS entry #{@fqdn} => #{@value}"
+          cmd = "/RecordDelete #{zone} #{@fqdn}. A /f"
+          execute(cmd, msg)
+        when "PTR"
+          # TODO: determine reverse zone names, #4025
+          raise Proxy::Dns::NotFound.new("Cannot find DNS entry for #{@value}") unless dns_find(@value)
+          return true
```
dLobatog: nitpick: pointless return

lib/proxy/provider.rb
```diff
@@ -0,0 +1,30 @@
+class ::Proxy::Provider
+  include ::Proxy::Pluggable
+  include ::Proxy::Log
+
+  class << self
+    attr_reader :provider_factory_proc
+
+    def plugin(plugin_name, aversion, attrs)
```
dLobatog, line 8: Should this class inherit from Proxy::Plugin in some way? https://github.com/theforeman/smart-proxy/pull/256/files#diff-cc107af997731be1551c20e4af1d19b2R114 shows that aside from provider_factory this class is effectively the same.

dLobatog:
Generally :+1: , I couldn't find any glitches when testing it

lib/proxy/provider.rb
dmitri-d, line 8: I went with composition instead of inheritance -- all common logic is in `::Proxy::Pluggable`...

extra/migrate_settings.rb
dmitri-d: rolled back changes to migrate_settings and migration_test, as those are going to be replaced by the new settings migrations (kept dns_settings_migrations for now as a reference, will convert to settings migration once the PR has been merged)

dmitri-d:
@eLobato: fixed things you pointed out, except the inheritance from Proxy::Plugin bit (pls. see my comment above).

domcleal:
Could you rebase this and use the new migration system please?

dmitri-d:
Nods, that's my plan...

dmitri-d:
Rebased. Dns settings migration is in a dedicated class now. Use `rake migrate_settings` to run migrations.

lib/proxy/provider.rb
```diff
@@ -0,0 +1,30 @@
...
+  end
+
+  def configure_plugin
+    logger.info("'#{plugin_name}' settings were initialized with default values: %s" % log_used_default_settings) unless settings.used_defaults.empty?
+    validate!
+    ::Proxy::Plugins.plugin_enabled(plugin_name, self)
+    ::Proxy::BundlerHelper.require_groups(:default, bundler_group)
+    after_activation
+  rescue Exception => e
+    logger.error("Couldn't enable plugin #{plugin_name}: #{e}:#{e.backtrace.join('/n')}")
```
domcleal: `"\n"` (double quotes needed, wrong slash)

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -1,12 +1,17 @@
...
-  class Dnscmd < Record
+module Proxy::Dns::Dnscmd
+  class Record < ::Proxy::Dns::Record
     include Proxy::Log
     include Proxy::Util
     attr_reader :resolver
 
+    def self.record(attrs = {})
+      new(attrs.merge(:server => ::Proxy::Dns::Dnscmd::Plugin.settings.dns_server,
+                      :ttl => ::Proxy::Dns::Plugin.settings.dns_ttl))
```
domcleal, line 12: The migration moves dns_ttl into the nsupdate providers, should it add it also to dnscmd, or keep it in the main dns config?

modules/dns_nsupdate/dns_nsupdate_gss_main.rb
```diff
@@ -26,4 +26,4 @@ def nsupdate cmd
       super
     end
   end
-end
+end
```
domcleal: Nitpick, if you can configure your IDE/editor to keep end of file newlines, it'd be appreciated.

extra/migrations/20150611000000_migrate_dns_settings.rb
```diff
@@ -0,0 +1,73 @@
...
+    output, unknown = migrate_dns_configuration(to_migrate)
+    copy_original_configuration_except(path("settings.d", "dns.yml"))
+    write_to_files(output, unknown)
+  end
+
+  def known_dns_options
+    {
+        :enabled            => [:dns],
+        :dns_provider       => [:dns],
+        :dns_key            => [:dns_nsupdate, :dns_nsupdate_gss],
```
domcleal: I don't think nsupdate_gss needs the dns_key, as it would only rely on Kerberos.

config/settings.d/dns_virsh.yml.example
```diff
@@ -0,0 +1,4 @@
+---
+#
+# Configuration file for libvirtd-specific 'virsh' dns provider
+#
```
domcleal, line 4: Perhaps just note that the only config is currently in the main settings.yml (the libvirt network name, IIRC)

modules/dns_dnscmd/dns_dnscmd_main.rb
domcleal, line 12: Disregard, will put another comment in above.

extra/migrations/20150611000000_migrate_dns_settings.rb
```diff
@@ -0,0 +1,73 @@
...
+        :dns_server         => [:dns_nsupdate, :dns_nsupdate_gss, :dns_dnscmd],
+        :dns_ttl            => [:dns_nsupdate, :dns_nsupdate_gss],
```
domcleal: dns_ttl is in the main dns.yml file, not the per-provider examples.  The dnscmd provider also tries to access it from dns.yml.

lib/proxy/plugin.rb
```diff
@@ -167,18 +144,4 @@ def configure_plugin
     logger.error("Couldn't enable plugin #{plugin_name}: #{e}:#{e.backtrace.join('/n')}")
```
domcleal: ditto about backtrace

domcleal:
I can't pinpoint precisely where to fix this, but when you enter an incorrect dns_provider name, the error message is a bit cryptic:

```
ERROR -- : Couldn't enable plugin dns: One or more dependencies in [:dns_virs] have not been loaded.
```

Do you have an example standalone plugin (e.g. gem) that adds a new provider just to see how well it works as an extension mechanism?

Rakefile
```diff
@@ -37,3 +41,22 @@ Rake::RDocTask.new(:rdoc) do |rdoc|
...
+  migrations_dir_path = File.join(app_dir, "extra", "migrations")
+  migrations_state_file_path = File.join(app_dir, "config", "migration_state")
+  FileUtils.touch(migrations_state_file_path)
+
+  Dir.mktmpdir do |working_dir|
+    ::Proxy::Migrator.new(
+        working_dir, migrations_dir_path, config_src_path, modules_config_src_path,
+        ::Proxy::Migrations.new(migrations_state_file_path).load_past_migrations!).migrate
+
+    FileUtils.mv(File.join(app_dir, "config"), File.join(app_dir, "config_#{Time.now.strftime("%Y%m%d%H%M%S")}"))
```
domcleal, line 59: Could you add this path to gitignore please?

dmitri-d:
@domcleal: fixed the issues you pointed out. Failures to load dependencies are being logged better now: failure to find a module is different from module not being enabled. I'll do an example gem too.

domcleal:
I wrote this silly example gem, which maintains a hosts file: https://github.com/domcleal/smart_proxy_dns_hosts.  Seems to work pretty well!

modules/dns_nsupdate/dns_nsupdate_gss_plugin.rb
```diff
@@ -0,0 +1,17 @@
+module ::Proxy::Dns::NsupdateGSS
+  class Plugin < ::Proxy::Provider
+    plugin :dns_nsupdate_gss, ::Proxy::VERSION,
+           :factory => proc { |attrs| ::Proxy::Dns::NsupdateGSS::Record.record(attrs) }
+
+    default_settings :dns_server => 'localhost', :dns_ttl => 86_400,
```
domcleal: Should this default be in Proxy::Dns::Plugin itself?

domcleal: Sorry, the _dns_ttl_ default specifically, since it's in the main dns.yml config.

modules/dns_nsupdate/dns_nsupdate_plugin.rb
```diff
@@ -0,0 +1,15 @@
+module ::Proxy::Dns::Nsupdate
+  class Plugin < ::Proxy::Provider
+    plugin :dns_nsupdate, ::Proxy::VERSION,
+           :factory => proc { |attrs| ::Proxy::Dns::Nsupdate::Record.record(attrs) }
+
+    default_settings :dns_server => 'localhost', :dns_ttl => 86_400,
```
domcleal: ditto re placement of dns_ttl default

modules/dns_nsupdate/dns_nsupdate_gss_plugin.rb
dmitri-d: Not sure I understand you. Are you asking whether the `dns_ttl` default value should refer to the one defined in `Proxy::Dns::Plugin`? I suppose that's acceptable in this case -- providers exist in the context of the main module, and I think such a dependency is ok.

dmitri-d:
@domcleal: nm, I added a default value for `dns_ttl` in the main module, and removed it from both `dns_nsupdate` and `dns_nsupdate_gss`. Both of the submodules were referring to the setting in the main module anyway (which makes sense).

domcleal:
Merged as d905c67 for Foreman 1.10, thanks @witlessbird!  I'll probably make a start on converting the Route53 PR to a plugin.

PR #257, opened 2015-02-09
If dst file name contains a slash - we don't use a hyphen

DimmKirr:
If dst file name contains a slash - we don't use a hyphen (file goes into a subdirectory)

Additionally fixes #9299

### DIFF
--- a/modules/tftp/server.rb
+++ b/modules/tftp/server.rb
@@ -95,12 +95,11 @@ module Proxy::TFTP
   def self.fetch_boot_file dst, src
     # If dst file name contains a slash - we don't use a hyphen (file goes into a subdirectory)
     if dst.include? "/"      
-      filename    = dst + src.split("/")[-1]
+      filename  = dst + src.split("/")[-1]
     else
-      filename    = dst + '-' + src.split("/")[-1]
+      filename  = dst + '-' + src.split("/")[-1]
     end
 
-    filename    = dst + '-' + src.split("/")[-1]
     destination = Pathname.new(File.expand_path(filename, Proxy::TFTP::Plugin.settings.tftproot)).cleanpath
     tftproot    = Pathname.new(Proxy::TFTP::Plugin.settings.tftproot).cleanpath
     raise "TFTP destination outside of tftproot" unless destination.to_s.start_with?(tftproot.to_s)
--- a/modules/tftp/server.rb
+++ b/modules/tftp/server.rb
@@ -93,6 +93,13 @@ module Proxy::TFTP
   end
 
   def self.fetch_boot_file dst, src
+    # If dst file name contains a slash - we don't use a hyphen (file goes into a subdirectory)
+    if dst.include? "/"      
+      filename    = dst + src.split("/")[-1]
+    else
+      filename    = dst + '-' + src.split("/")[-1]
+    end
+
     filename    = dst + '-' + src.split("/")[-1]
     destination = Pathname.new(File.expand_path(filename, Proxy::TFTP::Plugin.settings.tftproot)).cleanpath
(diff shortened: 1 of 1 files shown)


modules/tftp/server.rb
```diff
@@ -93,6 +93,13 @@ def pxeconfig_file mac
...
 
   def self.fetch_boot_file dst, src
+    # If dst file name contains a slash - we don't use a hyphen (file goes into a subdirectory)
+    if dst.include? "/"      
+      filename    = dst + src.split("/")[-1]
+    else
+      filename    = dst + '-' + src.split("/")[-1]
+    end
+
     filename    = dst + '-' + src.split("/")[-1]
```
dmitri-d: The changes you made are going to be overridden here.

dmitri-d:
Thanks for the PR, @kireevco! In addition to the comment above, could you file a ticket explaining the issue and change the commit message to be in the format `Fixes #<issue number>: short description here`.

dLobatog:
@witlessbird The reasoning is here, this PR is part of a few other changes to community templates and core http://projects.theforeman.org/issues/9299

DimmKirr:
Oops, my bad. Of course it shouldn't be like that. Let me fix everything.

DimmKirr:
Looks like my commit message didn't work for a bot again... Next time I'll use 
Fixes #9299: fix copy-paste override issue

dmitri-d:
[test]

modules/tftp/server.rb
```diff
@@ -93,7 +93,13 @@ def pxeconfig_file mac
...
 
   def self.fetch_boot_file dst, src
-    filename    = dst + '-' + src.split("/")[-1]
+    # If dst file name contains a slash - we don't use a hyphen (file goes into a subdirectory)
+    if dst.include? "/"      
+      filename  = dst + src.split("/")[-1]
+    else
+      filename  = dst + '-' + src.split("/")[-1]
+    end
+
```
dmitri-d, line 102: I'm afraid this is still not going to work: on line 109 https://github.com/kireevco/smart-proxy/blob/develop/modules/tftp/server.rb#L109, the resulting directory tree won't have the last directory, if I correctly understand the directory layout suggested in #9299. 

Is the directory per architecture required for pxe booting of Windows? Could we get away with our current approach, i.e. prefixing boot files with the os and architecture name? For example, the names of boot files for centos 6.4 for x86 are: CentOS-6.4-x86_64-initrd.img and CentOS-6.4-x86_64-vmlinuz.

DimmKirr, line 102: Yes, it never worked with the hyphens, because windows was looking for specific files, and they didn't exist, since they had arch prefixes with hyphens. (bcd-x86, for example).
I currently have this code in my production and It does create a directory tree that is desired.

I get something like boot/windows-x64/moot.wim, boot/windows-x86/boot.wim, etc.

Are you testing with my patch to windows operating system model In foreman?

If you could, please explain what's not correct in the directory structure, so I can fix It.

dmitri-d:
Sorry, my bad, the directory tree will get created properly. I'll test this first thing tomorrow. Any chance you could add a test for this change (and squash commits)?

dmitri-d:
[test]

dmitri-d:
@kireevco: There is a different PR that fixes this issue, but retains backwards compatibility. Could you take a look: https://github.com/theforeman/smart-proxy/pull/258

DimmKirr:
Looks good to me. Let's use that one instead of https://github.com/theforeman/smart-proxy/pull/257
If someone else could test this, it would be great too!

dmitri-d:
closing in favour of https://github.com/theforeman/smart-proxy/pull/258

PR #258, opened 2015-02-12
Fixes #9299: prefixes ending with a '/' do not get appended a '-'

dmitri-d:


dmitri-d:
This is based on https://github.com/theforeman/smart-proxy/pull/257 by @kireevco. The original was not backwards-compatible. @kireevco: does this work for you? @domcleal: ping

domcleal:
Fine by me, if it's OK with @kireevco.

DimmKirr:
Sure, totally fine with me. Can you explain a little bit what was incompatible and what changed?

dmitri-d:
@kireevco: The difference is that in this PR a '-' is not being used between a prefix and a filename only when the prefix ends with a '/'. This is to preserve the current behaviour when a prefix can contain both subdirectories and a part of a filename.

DimmKirr:
I'm confused, I felt like everything in the else block that I left was specifically for That purpose, no?

dmitri-d:
It's currently possible to have a prefix that has sub-directories in it, for example 'dir1/not_a_dir'. With changes in your PR the resulting filename would be 'dir1/not_a_dir/filename', while it should be 'dir1/not_a_dir-filename'.

DimmKirr:
With 'dir1/not_a_dir-filename' windows won't be able to use the files (since they will have prefixes), and that defeats the purpose. 

Ideally we should switch to directories for everything at some point, it won't hurt and will be more organized, IMHO.

dmitri-d:
In the PR I'm proposing, should a prefix end with a '/' it will be treated like a path with directories only, and a '-' won't be used. This covers both existing behaviour and additional behaviour required by windows. You need to make sure that when you define pxe_prefix for windows, you terminate it with a '/'.

DimmKirr:
Got it. let me test it and I'll change another PR for foreman core

## 

Sincerely,
 Dmitry Kireev
 cel.: +1 (818) 850-5577
 email, gtalk: dmitry@kireev.co
dmitry@kireev.co skype: kireev.co
 aim: 336719228

dmitri-d:
@domcleal: mind merging this in?

domcleal:
Merged as a729746, thanks @kireevco and @witlessbird!

PR #259, opened 2015-02-19
Fixes #9457 - fixed file permissions of debug script

lzap:
Katello is fine, we need just to fix here.


domcleal:
Merged as db48412, thanks @lzap!

PR #260, opened 2015-02-20
Fixes #9482 - introduced path patterns to authorize_with_trusted_hosts

lzap:
Foreman discovery proxy plugin needs to specify for which paths it will
authorize. Since Sinatra supports patters for filters
(http://www.sinatrarb.com/intro.html#Filters) it is relatively easy to
implement this so plugins can take advantage of this. This change will NOT
break the API, by default filters will apply to all paths (nil).

For https://github.com/theforeman/smart_proxy_discovery/pull/6

### DIFF
--- a/lib/sinatra/authorization.rb
+++ b/lib/sinatra/authorization.rb
@@ -1,9 +1,9 @@
 module Sinatra
   module Authorization
-    def authorize_with_trusted_hosts
+    def authorize_with_trusted_hosts pattern=nil
       helpers ::Proxy::Helpers
 
-      before do
+      before pattern do
         # When :trusted_hosts is given, we check the client against the list
         # HTTPS: test the certificate CN
         # HTTP: test the reverse DNS entry of the remote IP
@@ -26,11 +26,11 @@ module Sinatra
       end
     end
 
-    def authorize_with_ssl_client
+    def authorize_with_ssl_client pattern=nil
       helpers ::Proxy::Helpers
       helpers ::Proxy::Log
 
-      before do
+      before pattern do
         if ['yes', 'on', '1'].include? request.env['HTTPS'].to_s
           if request.env['SSL_CLIENT_CERT'].to_s.empty?
             log_halt 403, "No client SSL certificate supplied"


domcleal:
Wouldn't it perhaps be better to group authorisation differences by module, rather than individual routes?  Much less scope for error that way.  (I mean, this is my whole reason for hating the idea of any client ever talking to the smart proxy.)

Is there a minimum Sinatra version required?

lzap:
Two modules look like a nicer solution, yeah. But I wonder how to map two separate API modules into one namespace:

``` ruby
require 'smart_proxy_discovery/discovery_api'

map '/discovery' do
  run Proxy::Discovery::Api
end
```

Sinatra changelog says: 1.2.0+ / 2011-03-03

lzap:
@witlessbird do you think it is possible? ^

http://www.sinatrarb.com/intro.html#Dynamic%20Application%20Creation

lzap:
@domcleal how about this - let's do per-request and we can refactor this for Discovery 2.1 where we can change base URLs (changes needs to be done both in Plugin and Image to have this consistent like `/discovery_host2foreman` and `/discovery_foreman2host`)

dmitri-d:
This would work, but I too would prefer either grouping similarly authenticated methods in modules (although this might require dedicated mounting points), or simply invoking authorization check inside specific controller methods.

dmitri-d:
@lzap -- re: dynamic app creation: this is what smart-proxy relies on, but instead of composing apps using sinatra api, we rely on rack instead.

domcleal:
> Sinatra changelog says: 1.2.0+ / 2011-03-03

For the record, we currently support back to 1.0.  (I don't necessarily mind doing newer if we have to, we can package it.)

lzap:
@witlessbird my question was more if I can do something like:

``` ruby
require 'smart_proxy_discovery/discovery_api'

map '/discovery' do
  run Proxy::DiscoveryModuleA::Api
  run Proxy::DiscoveryModuleA::Api
end
```

I think I need to try...

domcleal:
Since they both install filters, that might affect the whole API?

dmitri-d:
@lzap: I'm quite certain you can. Not sure how filters would interact between modules (if at all), will need to take a look at the code.

lzap:
It looks like filters are per App module and not global, so it should work. I will try that on Monday then.

lzap:
Just for the record what I googled out: https://github.com/sinatra/sinatra/issues/750 (will try this later on). Thanks. Modules look safer than this.

dmitri-d:
On the second thought, this may not work with earlier versions of Sinatra -- I seem to remember issues around last module taking precedence (won't matter if we decide to upgrade to sinatra 1.2 though).

lzap:
Ok after some research I found it is not possible to do multiple `run` calls as it is stored in a hash.

Anyway I found a more sane way of splitting the Sinatra app into two. Closing this in fav of: https://github.com/theforeman/smart_proxy_discovery/pull/7

PR #263, opened 2015-03-10
Fixes #9709 -  allow the bmc module to log all impi calls

logicminds:


logicminds:
I have created a mega PR #264 with a bunch of other related fixes.

logicminds:
@witlessbird It used to be public but someone else made it private. Not sure if they had a good reason but check the git logs.  It would have been much easier to test agains public.

PR #264, opened 2015-03-11
Bmc fixes

logicminds:
- Fixes 9709
- Fixes 9710
- Fixes 9711 
- Fixes 9740

### DIFF
--- a/config/settings.d/bmc.yml.example
+++ b/config/settings.d/bmc.yml.example
@@ -6,9 +6,3 @@
 # - freeipmi / ipmitool - requires the appropriate package installed, and the rubyipmi gem
 # - shell - for local reboot control (requires sudo access to /sbin/shutdown for the proxy user)
 #:bmc_default_provider: freeipmi
-#
-# Setting the log level below will cause rubyipmi to output logs to /tmp/rubyipmi.log
-# If you want the Rubyipmi logs to go into the smart_proxy logs you will need to set the main log level in settings.yaml to DEBUG
-# and disable the provider_log_level below
-# valid options are WARN, DEBUG, ERROR, FATAL, INFO, UNKNOWN
-#:provider_log_level: DEBUG
--- a/config/settings.d/bmc.yml.example
+++ b/config/settings.d/bmc.yml.example
@@ -6,3 +6,9 @@
 # - freeipmi / ipmitool - requires the appropriate package installed, and the rubyipmi gem
 # - shell - for local reboot control (requires sudo access to /sbin/shutdown for the proxy user)
 #:bmc_default_provider: freeipmi
+#
+# Setting the log level below will cause rubyipmi to output logs to /tmp/rubyipmi.log
+# If you want the Rubyipmi logs to go into the smart_proxy logs you will need to set the main log level in settings.yaml to DEBUG
+# and disable the provider_log_level below
+# valid options are WARN, DEBUG, ERROR, FATAL, INFO, UNKNOWN
+#:provider_log_level: DEBUG
--- a/modules/bmc/bmc_api.rb
+++ b/modules/bmc/bmc_api.rb
@@ -24,7 +24,18 @@ module Proxy::BMC
 
     # returns host operations
     get "/:host" do
-      { :actions => %w[chassis lan] }.to_json
+      { :actions => %w[chassis lan test] }.to_json
+    end
+
+    # runs a test against the bmc device and returns true if the connection was successful
+    get "/:host/test" do
+      bmc_setup
+      begin
+        result = @bmc.test
+        { :action => :test, :result => result }.to_json
+      rescue => e
+        log_halt 400, e
+      end
     end
 
     # returns chassis operations
--- a/modules/bmc/bmc_api.rb
+++ b/modules/bmc/bmc_api.rb
@@ -1,25 +1,32 @@
+require 'bmc/ipmi'
+
 module Proxy::BMC
   class Api < ::Sinatra::Base
     helpers ::Proxy::Helpers
     authorize_with_trusted_hosts
     authorize_with_ssl_client
-
     # All GET requests will only read ipmi data, no changes
     # All PUT requests will update information on the bmc device
 
-    # FIXME: this is broken... 
-    get "" do
-      # return list of available options
+    # return list of available options
+    get "/" do
+      res = ['providers', 'providers/installed', 'host']
+      { :available_resources => res}.to_json
     end
 
-    # FIXME: this is broken... Returns a list of bmc providers
+    # Returns a list of bmc providers
     get "/providers" do
-      { :providers => Proxy::BMC.providers }.to_json
+      { :providers => Proxy::BMC::IPMI.providers + ['shell'] }.to_json
     end
 
-    # FIXME: this is broken... Returns a list of installed providers
+    # Returns a list of installed providers
     get "/providers/installed" do
-      { :installed_providers => Proxy::BMC.installed_providers? }.to_json
+      { :installed_providers => Proxy::BMC::IPMI.providers_installed + ['shell'] }.to_json
+    end
+
+    # returns a helpful message that the user should supply a hostname
+    get "/host" do
+      { :message => "You need to supply the hostname or ip of the actual bmc device"}.to_json
     end
 
     # returns host operations
@@ -216,62 +223,61 @@ module Proxy::BMC
       end
     end
 
-    private
+    # returns a provider type by validating the given type.  If for some reason the type is invalid
+    # this function will try and find the first available provider that could be used.  If there are no providers available
+    # lets halt and notify the user.
+    def find_ipmi_provider(provider_type)
+      # check to see if provider is given and if no default provider is set, search for installed providers
+      unless Proxy::BMC::IPMI.installed?(provider_type)
+        # check if provider_type is a valid type
+        if Proxy::BMC::IPMI.providers.include?(provider_type)
+          logger.warn "#{provider_type} specified but it is not installed"
+        else
+          logger.warn "Invalid BMC type: #{provider_type}, must be one of #{Proxy::BMC::IPMI.providers.join(',')}"
+        end
+        if Proxy::BMC::IPMI.providers_installed.length > 0
+          provider_type = Proxy::BMC::IPMI.providers_installed.first
+          logger.warn "Using #{provider_type} as the default BMC provider"
+        else
+          log_halt 400, "No BMC providers are installed, please install at least freeipmi or ipmitool"
+        end
+      end
+      provider_type
+    end
 
     def bmc_setup
       # Either use the default provider or allow user to specify provider in request
       provider_type ||= params[:bmc_provider] || Proxy::BMC::Plugin.settings.bmc_default_provider
       provider_type.downcase! if provider_type
-      if log_level = ::Proxy::BMC::Plugin.settings.provider_log_level
-        # using the log_level below will cause rubyipmi to create its own log, instead of using the proxy log
-        begin
-          log_level = ::Logger.const_get(log_level.upcase)
-        rescue
-          logger.warn("Incorrect bmc log level supplied #{log_level}, ignoring setting")
-          log_level = nil
-        end
+      # unless the provider is shell find a suitable provider
+      if provider_type != 'shell'
+        provider_type = find_ipmi_provider(provider_type)
       end
 
       case provider_type
-      when /ipmi/
-        require 'bmc/ipmi'
+        when 'freeipmi', 'ipmitool'
 
-        raise "unauthorized" unless auth.provided?
-        raise "bad_authentication_request" unless auth.basic?
+          raise "unauthorized" unless auth.provided?
+          raise "bad_authentication_request" unless auth.basic?
 
-        username, password = auth.credentials
+          username, password = auth.credentials
 
-        # setting the log level must come before IPMI initialization
-        # if the user supplies the provider_log_level all provider logs will be sent to
-        # the provider specific log instead of the smart proxy logs
-        # if the user doesn't supply the provider log level, all logs will be sent to the smart proxy logs
-        # this was configured to make easier specific log output for BMC without having to see other logs
-        if log_level
-          Proxy::BMC::IPMI.log_level = log_level
-        else
           # this causes rubyipmi to use the supplied logger, most actions in rubyipmi only output during Logger::DEBUG
           Proxy::BMC::IPMI.logger = logger
-        end
 
-        # check to see if provider is given and no default provider is set, search for installed providers
-        unless Proxy::BMC::IPMI.installed?(provider_type)
-          # check if provider_type is a valid type
-          if Proxy::BMC::IPMI.providers.include?(provider_type)
-            log_halt 400, "#{provider_type} is not installed, please install a ipmi provider"
-          else
-            log_halt 400, "Unrecognized or missing bmc provider type: #{provider_type}"
-          end
-        end
-
-        # all the use of the http auth basic header to pass credentials
-        args = { :host     => params[:host], :username     => username,
-                 :password => password,      :bmc_provider => provider_type }
-        @bmc = Proxy::BMC::IPMI.new(args)
-      when "shell"
-        require 'bmc/shell'
-        @bmc = Proxy::BMC::Shell.new
-      else
-        log_halt 400, "Unrecognized or missing BMC type: #{provider_type}"
+          # we use the http auth basic header to pass credentials
+          # The idea here is to pass the credentials on the command line
+          # to execute ipmi commands and has nothing to do with authorization
+          # of using smart-proxy. Its simply a tunnel to pass credentials through,
+          # since we are essentially remotely executing ipmi commands using Rubyipmi.
+          args = { :host     => params[:host], :username     => username,
+                   :password => password,      :bmc_provider => provider_type }
+          @bmc = Proxy::BMC::IPMI.new(args)
+        when "shell"
+          require 'bmc/shell'
+          @bmc = Proxy::BMC::Shell.new
+        else
+          log_halt 400, "Invalid BMC type: #{provider_type}"
       end
     rescue => e
       log_halt 400, e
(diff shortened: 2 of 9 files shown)


logicminds:
This will require rubyipmi 0.9.1 which was just released today.

logicminds:
[test]

logicminds:
An example of what this provider can be seen here.

http://projects.theforeman.org/projects/smart-proxy/wiki/API#BMC-Controls  (Troubleshooting section)

modules/bmc/ipmi.rb
```diff
@@ -5,22 +5,41 @@ module Proxy
...
-        else
-          Rubyipmi.provider_installed?
+        rescue
+          false
         end
       end
 
-      def self.providers_installed?
-        return Rubyipmi.providers_installed?
+      def self.providers_installed
```
dmitri-d, line 26: This is a bit confusing -- the convention is that methods with a `?` suffix return a boolean, but here it appears to return a list...

modules/bmc/bmc_api.rb
```diff
@@ -1,30 +1,48 @@
...
     get "/providers" do
-      { :providers => Proxy::BMC.providers }.to_json
+      { :providers => Proxy::BMC::IPMI.providers << 'shell' }.to_json
     end
 
-    # FIXME: this is broken... Returns a list of installed providers
+    # Returns a list of installed providers
     get "/providers/installed" do
-      { :installed_providers => Proxy::BMC.installed_providers? }.to_json
+      { :installed_providers => Proxy::BMC::IPMI.providers_installed << 'shell' }.to_json
```
dmitri-d: this modifies w/e list Proxy::BMC::IPMI.providers_installed returns; `Proxy::BMC::IPMI.providers_installed + ['shell']` is safer.

modules/bmc/bmc_api.rb
```diff
@@ -208,50 +217,80 @@ class Api < ::Sinatra::Base
...
+      # check to see if provider is given and if no default provider is set, search for installed providers
+      unless Proxy::BMC::IPMI.installed?(provider_type)
+        # check if provider_type is a valid type
+        if Proxy::BMC::IPMI.providers.include?(provider_type)
+          logger.warn "#{provider_type} specified but it is not installed"
+        else
+          logger.warn "Invalid BMC type: #{provider_type}, must be one of #{Proxy::BMC::IPMI.providers.join(',')}"
+        end
+        if Proxy::BMC::IPMI.providers_installed.length > 0
+          provider_type = Proxy::BMC::IPMI.providers_installed.first
```
dmitri-d, line 252: if you do `return Proxy::BMC::IPMI.providers_installed.first` you'll avoid modifying local state.

modules/bmc/bmc_api.rb
```diff
@@ -208,50 +217,80 @@ class Api < ::Sinatra::Base
...
-                 :password => password,      :bmc_provider => provider_type }
-        @bmc = Proxy::BMC::IPMI.new(args)
-      when "shell"
-        require 'bmc/shell'
-        @bmc = Proxy::BMC::Shell.new
-      else
-        log_halt 400, "Unrecognized or missing BMC type: #{provider_type}"
+          # all the use of the http auth basic header to pass credentials
+          args = { :host     => params[:host], :username     => username,
+                   :password => password,      :bmc_provider => provider_type }
```
dmitri-d, line 287: This is a bit weird: you are using http authorization header to pass in username/password, although they are not being used by smart-proxy for authorization (client certs and trusted hosts list are).

modules/bmc/ipmi.rb
logicminds, line 26: Yes, I know.  I noticed this the other day in Rubyipmi and I plan to fix it but it will have to wait since its a breaking change.  I have updated this in smart-proxy but this this will most likely be fixed in 0.11.0 of rubyipmi.

modules/bmc/bmc_api.rb
logicminds, line 287: The idea here is to pass the credentials on the command line to execute ipmi commands and has nothing to do with authorization of using smart-proxy.  Its simply a tunnel to pass credentials through, since we are essentially remotely executing ipmi commands using Rubyipmi.  At this time Rubyipmi does not support certificates so neither does the bmc smart-proxy module.  I think it would be awesome to rely on a certificate but a lot more work would be needed to make this happen which is completely out of scope for this PR.  Furthermore, passing credentials is the easiest thing to do and using a ssl cert would probably be the next line of business for this module and Rubyipmi.

logicminds:
@witlessbird regarding the private method.  It appears after some testing that that even when I remove the private decorator all methods are still marked private.  I am not sure why the class is still setting the methods private so there is something that I don't understand.  This looks to be the case will all the other smart proxy modules as well.  So as far as refactoring this to have a public bmc_setup method I am not sure this is possible unless its removed from the class and place inside a Proxy::BMC module which might not be ideal and might introduce other issues like authorization. 

``` ruby
[3] pry(#<BmcApiTest>)> Proxy::BMC::Api.new.bmc_setup
NoMethodError: undefined method `bmc_setup' for #<Sinatra::Wrapper:0x007fa5619ebef8>
from (pry):3:in `test_api_recovers_from_missing_provider'
[4] pry(#<BmcApiTest>)>
 Proxy::BMC::Api.new.send(:bmc_setup)
NoMethodError: undefined method `bmc_setup' for #<Sinatra::Wrapper:0x007fa5632969a8>
from (pry):5:in `test_api_recovers_from_missing_provider'
```

dmitri-d:
sinatra overrides `::Sinatra::Base.new`, as a result you don't actually get an instance of your controller back. Call `::Proxy::BMC::Api.new!` instead.

modules/bmc/bmc_api.rb
```diff
@@ -1,30 +1,48 @@
+require 'bmc/ipmi'
```
lzap, line 1: You are adding global require here, this will cause issues on discovery-image. Please load dependencies only when ipmi is enabled by configuration later on, not here.

config/settings.d/bmc.yml.example
```diff
@@ -6,3 +6,9 @@
 # - freeipmi / ipmitool - requires the appropriate package installed, and the rubyipmi gem
 # - shell - for local reboot control (requires sudo access to /sbin/shutdown for the proxy user)
 #:bmc_default_provider: freeipmi
+#
+# Setting the log level below will cause rubyipmi to output logs to /tmp/rubyipmi.log
+# If you want the Rubyipmi logs to go into the smart_proxy logs you will need to set the main log level in settings.yaml to DEBUG
+# and disable the provider_log_level below
+# valid options are WARN, DEBUG, ERROR, FATAL, INFO, UNKNOWN
+#:provider_log_level: DEBUG
```
lzap: Can you reword the whole description, it's a bit confusing. Use words comment/uncomment to describe desired behavior. Describe current state (when it's commented) and what happens when you uncomment.

Is really `/tmp` the best sane default value? I'd vote for something like `/var/log/smart-proxy/rubyimpi.log` instead.

I think the default configuration should be to use Smart Proxy logger with INFO/WARN level to see warnings and errors there. We can extend our `foreman-debug` script to collect this information then: https://github.com/theforeman/smart-proxy/blob/develop/extra/foreman-debug-proxy

lzap:
Do you have an advice how to test this without BMC-enabled hardware? I guess there could be some mock mode or something in the library or IPMI stack.

modules/bmc/bmc_api.rb
dmitri-d, line 1: It's actually ok here, as bmc_api will only get loaded if the module was enabled. https://github.com/theforeman/smart-proxy/blob/develop/lib/smart_proxy.rb#L100 is the bit that takes care of this...

logicminds:
@lzap About the only thing that I came up with is a tool that feeds back STDOUT and STDERR along with an error code given STDIN.  But I did not finish writing that tool and its still a private repo.

config/settings.d/bmc.yml.example
logicminds: I think I am just going to remove this feature so it doesn't confuse anyone.

logicminds:
I could probably add a mock provider to rubyipmi which could then be used in smart-proxy when the mock provider is selected.  I don't know when I would have time to do this though.

logicminds:
[test]

logicminds:
@witlessbird @lzap Jenkins seems to be failing tests, but I can get them to fail on my system.  Is this a known issue with Jenkins and doing forced updates?

logicminds:
This is ready to merge, I have addressed all the issues discussed above and refactored the code.

domcleal:
I think it's because we don't have freeipmi installed on our slaves.  We installed ipmitool as the existing tests assume that's present, but I also get failures locally without freeipmi:

<pre>
[  3/309] BmcApiTest#test_api_bmc_setup_returns_new_ipmi_proxy_given_freeipmiwhich: no ipmipower in (/home/dcleal/.rvm/gems/ruby-2.1.0@proxy/bin:/home/dcleal/.rvm/gems/ruby-2.1.0@global/bin:/home/dcleal/.rvm/rubies/ruby-2.1.0/bin:/home/dcleal/.rvm/bin:/home/dcleal/bin:/ho
me/dcleal/go/bin:/home/dcleal/.gem/home/bin:/home/dcleal/code/puppet/puppet/bin:/home/dcleal/code/facter/bin:/usr/lib64/qt-3.3/bin:/usr/lib64/ccache:/usr/local/bin:/usr/local/sbin:/usr/bin:/usr/sbin:/bin:/sbin/code/augeas/augeas-sandbox:/sbin)
 = 0.01 s
  1) Error:
BmcApiTest#test_api_bmc_setup_returns_new_ipmi_proxy_given_freeipmi:
NoMethodError: undefined method `accept?' for nil:NilClass
    /home/dcleal/code/foreman/smart-proxy/lib/proxy/helpers.rb:20:in `log_halt'
    /home/dcleal/code/foreman/smart-proxy/modules/bmc/bmc_api.rb:283:in `rescue in bmc_setup'
    /home/dcleal/code/foreman/smart-proxy/modules/bmc/bmc_api.rb:250:in `bmc_setup'
    /home/dcleal/code/foreman/smart-proxy/test/bmc/bmc_api_test.rb:89:in `test_api_bmc_setup_returns_new_ipmi_proxy_given_freeipmi'
    /home/dcleal/.rvm/rubies/ruby-2.1.0/lib/ruby/2.1.0/minitest/unit.rb:1265:in `run'
    /home/dcleal/.rvm/rubies/ruby-2.1.0/lib/ruby/2.1.0/test/unit/testcase.rb:17:in `run'
    /home/dcleal/.rvm/rubies/ruby-2.1.0/lib/ruby/2.1.0/minitest/unit.rb:940:in `block in _run_suite'
    /home/dcleal/.rvm/rubies/ruby-2.1.0/lib/ruby/2.1.0/minitest/unit.rb:933:in `map'
    /home/dcleal/.rvm/rubies/ruby-2.1.0/lib/ruby/2.1.0/minitest/unit.rb:933:in `_run_suite'
    /home/dcleal/.rvm/rubies/ruby-2.1.0/lib/ruby/2.1.0/test/unit.rb:663:in `block in _run_suites'
    /home/dcleal/.rvm/rubies/ruby-2.1.0/lib/ruby/2.1.0/test/unit.rb:661:in `each'
    /home/dcleal/.rvm/rubies/ruby-2.1.0/lib/ruby/2.1.0/test/unit.rb:661:in `_run_suites'
    /home/dcleal/.rvm/rubies/ruby-2.1.0/lib/ruby/2.1.0/minitest/unit.rb:884:in `_run_anything'
    /home/dcleal/.rvm/rubies/ruby-2.1.0/lib/ruby/2.1.0/minitest/unit.rb:1092:in `run_tests'
    /home/dcleal/.rvm/rubies/ruby-2.1.0/lib/ruby/2.1.0/minitest/unit.rb:1079:in `block in _run'
    /home/dcleal/.rvm/rubies/ruby-2.1.0/lib/ruby/2.1.0/minitest/unit.rb:1078:in `each'
    /home/dcleal/.rvm/rubies/ruby-2.1.0/lib/ruby/2.1.0/minitest/unit.rb:1078:in `_run'
    /home/dcleal/.rvm/rubies/ruby-2.1.0/lib/ruby/2.1.0/minitest/unit.rb:1066:in `run'
    /home/dcleal/.rvm/rubies/ruby-2.1.0/lib/ruby/2.1.0/test/unit.rb:27:in `run'
    /home/dcleal/.rvm/rubies/ruby-2.1.0/lib/ruby/2.1.0/test/unit.rb:780:in `run'
    /home/dcleal/.rvm/rubies/ruby-2.1.0/lib/ruby/2.1.0/test/unit.rb:372:in `block (2 levels) in autorun'
    /home/dcleal/.rvm/rubies/ruby-2.1.0/lib/ruby/2.1.0/test/unit.rb:33:in `run_once'
    /home/dcleal/.rvm/rubies/ruby-2.1.0/lib/ruby/2.1.0/test/unit.rb:371:in `block in autorun'
[  8/309] BmcApiTest#test_api_can_get_installed_providerswhich: no ipmipower in (/home/dcleal/.rvm/gems/ruby-2.1.0@proxy/bin:/home/dcleal/.rvm/gems/ruby-2.1.0@global/bin:/home/dcleal/.rvm/rubies/ruby-2.1.0/bin:/home/dcleal/.rvm/bin:/home/dcleal/bin:/home/dcleal/go/bin:/ho
me/dcleal/.gem/home/bin:/home/dcleal/code/puppet/puppet/bin:/home/dcleal/code/facter/bin:/usr/lib64/qt-3.3/bin:/usr/lib64/ccache:/usr/local/bin:/usr/local/sbin:/usr/bin:/usr/sbin:/bin:/sbin/code/augeas/augeas-sandbox:/sbin)
 = 0.01 s
  2) Failure:
BmcApiTest#test_api_can_get_installed_providers [/home/dcleal/code/foreman/smart-proxy/test/bmc/bmc_api_test.rb:157]:
&lt;["freeipmi", "ipmitool", "shell"]&gt; expected but was
&lt;["ipmitool", "shell"]&gt;.
</pre>

I'm rolling it out now, will retest in a bit.

dmitri-d:
I'm seeing a local test failure when both freeipmi and ipmitool have been installed:

```
[ 28/309] BmcApiTest#test_api_recovers_from_nil_provider = 0.07 s                                                      
  1) Failure:
test_api_recovers_from_nil_provider(BmcApiTest) [/home/wb/sandbox/smart-proxy/test/bmc/bmc_api_test.rb:122]:
Last response was not ok
```

domcleal:
[test]

domcleal:
Unit tests pass now on Jenkins, I wonder why it fails for @witlessbird.  However there are some rubocop warnings, see the link on http://ci.theforeman.org/job/test_proxy_develop_pr_rubocop/200/

dmitri-d:
I looked into the failure I'm seeing:
- in test_api_recovers_from_nil_provider, https://github.com/logicminds/smart-proxy/blob/bmc_fixes/modules/bmc/bmc_api.rb#L41 returns an instance of Proxy::BMC::Shell
- Proxy::BMC::Shell doesn't implement #test method, and throws NotImplementedError
- NotImplementedError is not a StandardError, therefore rescue at https://github.com/logicminds/smart-proxy/blob/bmc_fixes/modules/bmc/bmc_api.rb#L43 doesn't catch it.

Not sure why in my case test gets an instance of Proxy::BMC::Shell. More importantly, implementation of either `Proxy::BMC::Base` or  `Proxy::BMC::Api` has to change, as it is currently possible for hard to troubleshoot failures to occur.

I think either api has to catch `Exception`, not just `StandardError`, or interface compliance check in `Proxy::BMC::Base` has to be done differently.

logicminds:
> On Mar 19, 2015, at 5:20 AM, Dmitri Dolguikh notifications@github.com wrote:
> 
> I looked into the failure I'm seeing:
> 
>   • in test_api_recovers_from_nil_provider, https://github.com/logicminds/smart-proxy/blob/bmc_fixes/modules/bmc/bmc_api.rb#L41 returns an instance of Proxy::BMC::Shell

I can reproduce this when I enable :bmc_default_provider: shell in the config/bmc.yml file.  So this is probably what you have in your config file.

>   • Proxy::BMC::Shell doesn't implement #test method, and throws NotImplementedError

Easy fix could catch.

>   • NotImplementedError is not a StandardError, therefore rescue at https://github.com/logicminds/smart-proxy/blob/bmc_fixes/modules/bmc/bmc_api.rb#L43 doesn't catch it.
> I think something else is going on here because the error is still of exception class. NotImplementedError <  ScriptError < Exception.

I have been mocking the settings provider value for some tests, but not all tests, so the default would apply if the value is not mocked.

> Not sure why in my case test gets an instance of Proxy::BMC::Shell. More importantly, implementation of either Proxy::BMC::Base or Proxy::BMC::Api has to change, as it is currently possible for hard to troubleshoot failures to occur.

You are getting an instance of shell because that is what you have set as your shell provider and because nil was expected in the test but I didn’t mock the settings value it was picking up your settings value.
Adding this as a mock Proxy::BMC::Plugin.settings.stubs(:bmc_default_provider).returns(nil) fixes the test and ensures nothing is picked up from the settings file.

dmitri-d:
@logicminds: `rescue => e` form will only catch StandardError exceptions. NotImplementedError doesn't have StandardError in its inheritance chain, so none of the exceptions thrown in the ::Proxy::BMC::Base will be caught.

logicminds:
@witlessbird  makes sense now.  https://robots.thoughtbot.com/rescue-standarderror-not-exception

So than we could assume all other methods that the shell provider does not implement will also suffer from this same issue.  Should we file a new ticket for this issue, since its out of scope?

logicminds:
[test]

I just fixed all rubocop warnings and pushed another rubyipmi gem version to fix some style issues.  I also opened issue 9840 because it was unrelated but still important for future provider support.  I would expect all tests to pass on any system.

dmitri-d:
merged in at e917190, 3fba6e1, ec00558, 0d70f58, 369793b. Thanks, @logicminds!

PR #267, opened 2015-03-12
Fixes #9740 - support foreman_url with path

ares:


config/settings.yml.example
```diff
@@ -22,6 +22,8 @@
 # (default: true)
 #:forward_verify: true
 
+# if you run foreman on sub-uri (e.g. http://127.0.0.1:3000/foreman/) you must
+# specify the trailing slash
```
dmitri-d: This works, but I think it would be less error prone if we automatically added the missing trailing slash when it's missing...

ares:
better? `[-1..-1]` is required by Ruby 1.8 IIRC

lib/proxy/request.rb
```diff
@@ -7,6 +7,7 @@ module Proxy::HttpRequest
   class ForemanRequestFactory
     def initialize(base_uri)
       @base_uri = base_uri
+      @base_uri += '/' if @base_uri[-1..-1] != '/'
```
dmitri-d, line 10: I look at 1.8.7 docs, String#end_with? is available: http://ruby-doc.org/core-1.8.7/String.html#method-i-end_with-3F

dmitri-d:
I like this much better, see my comment though.

dmitri-d:
[test]

dmitri-d:
@domcleal, @ares: closing this one in favour of #269.

PR #268, opened 2015-03-19
Add TFTP option tftp_fetch_boot_file true/false to disable fetching boot files

sodabrew:
I have a separate mechanism for controlling /var/lib/tftpboot installer binaries, so I'd rather not re-fetch these files on every host build.

In prior versions, Foreman continued to build hosts even though it did not have permissions to overwrite my pxe boot files. In 1.7, which I've just installed, the permissions issue causes a build rollback.

If this looks reasonable I'll file up the tickets and add tests.

### DIFF
--- a/config/settings.d/tftp.yml.example
+++ b/config/settings.d/tftp.yml.example
@@ -5,3 +5,6 @@
 #:tftproot: /var/lib/tftpboot
 # Defines the TFTP Servername to use, overrides the name in the subnet declaration
 #:tftp_servername: tftp.domain.com
+
+# Optionally disable fetching TFTP boot files
+#:tftp_fetch_boot_files: false
--- a/modules/tftp/server.rb
+++ b/modules/tftp/server.rb
@@ -99,11 +99,15 @@ module Proxy::TFTP
     tftproot    = Pathname.new(Proxy::TFTP::Plugin.settings.tftproot).cleanpath
     raise "TFTP destination outside of tftproot" unless destination.to_s.start_with?(tftproot.to_s)
 
-    # Ensure that our image directory exists
-    # as the dst might contain another sub directory
-    FileUtils.mkdir_p destination.parent
+    if Proxy::TFTP::Plugin.settings.tftp_fetch_boot_files
+      # Ensure that our image directory exists
+      # as the dst might contain another sub directory
+      FileUtils.mkdir_p destination.parent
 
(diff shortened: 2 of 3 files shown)


dmitri-d:
I think smart-proxy is the wrong place for disabling of pxe boot file fetching -- this is a pretty dumb api end point that shouldn't have this sort of control over file downloads. A better spot would in foreman itself, see tftp orchestration https://github.com/theforeman/foreman/blob/develop/app/models/concerns/orchestration/tftp.rb#L38.

Having said that:
- wget that we use for downloads attempts to download files if they changed only, although I don't know how reliable this is
- ideally, we should switch to an in-process downloader and rely on 'If-Modified-Since' and/or 'If-None-Match' headers to avoid re-downloading of files.

sodabrew:
I was thinking about this more overnight, and I agree that this doesn't make sense in the SP.

Here's what I propose to work on next:
- New SP API: `/tftp/check_boot_file` - checks if a filename exists, returns basic size/date info.
- New Foreman UI: In the Operating Systems definition, add a "Should download TFTP boot files?" checkbox.
- Alter Foreman Orchestration:  "Should download TFTP boot files?"
  - If true: use `/tftp/fetch_boot_file`.
  - If false: use `/tftp/check_boot_file`.

dmitri-d:
A new SP API: /tftp/check_boot_file might be useful (can't decide yet), but I think smart-proxy needs to be smart enough not to download files that haven't changed. Again, wget supposedly does that, but I'd like to confirm. If not, then it might be a good time to replace wget with an in-process downloader. 

The rest sounds good -- mind creating redmine tickets for those?

domcleal:
On the subject of wget and being smart enough, it generally seems to avoid downloading the files on subsequent requests, but always creates a zero byte file when an error occurs which has to be manually to trigger another download.  Some sort of check action including the size/checksum may help solve that too.

It can't seem to distinguish between a file that already exists and one that's been updated remotely too: http://projects.theforeman.org/issues/5069

I'd also like to see synchronous downloads so errors are reported to Foreman, but it's not a popular idea as it could tie up the smart proxy (http://projects.theforeman.org/issues/3034).

dmitri-d:
The last two things are definitely possible if we decide to go with our own implementation (as opposed to wget). I'd prefer an async download, but also expose a dedicated resource for tracking of download progress/status...

sodabrew:
I'll put together the redmine tickets (and links to existing tickets) and some PRs next week. Thanks again for the feedback and direction!

PR #269, opened 2015-03-20
Fixes #9740 - support foreman_url with path

dmitri-d:
This is based on ares' implementation


domcleal:
Merged as 95cd9c8, thanks @witlessbird and @ares.

PR #270, opened 2015-03-20
Fixes #9852 - REST API violation in BMC smart proxy API

logicminds:
```
* authentication errors now return 401 and proper message
```

### DIFF
--- a/test/bmc/bmc_api_test.rb
+++ b/test/bmc/bmc_api_test.rb
@@ -21,6 +21,45 @@ class BmcApiTest < Test::Unit::TestCase
     authorize user, pass
   end
 
+  def test_api_throws_401_error_when_auth_is_not_provided
+    Proxy::BMC::IPMI.stubs(:providers_installed).returns(['freeipmi'])
+    Proxy::BMC::Plugin.settings.stubs(:bmc_default_provider).returns('freeipmi')
+    auth = mock()
+    auth.expects(:provided?).returns(false)
+    Proxy::BMC::Api.any_instance.stubs(:auth).returns(auth)
+    test_args = { :bmc_provider => 'freeipmi' }
+    get "/#{host}/lan/gateway", test_args
+    assert_equal 'unauthorized', last_response.body
+    assert_equal 401, last_response.status
+  end
+
+  def test_api_does_not_throw_401_error_when_auth_is_provided_and_in_basic_format
+    Proxy::BMC::IPMI.stubs(:providers_installed).returns(['freeipmi'])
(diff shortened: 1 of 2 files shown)


logicminds:
This code fixes the issue by sending a proper http code back to notify the user via a 401 unauthorized code.  Rubyipmi API doesn't support connection failure reasons so this doesn't address someone sending the wrong credentials.  Although this might be added in a later version of Rubyipmi.

test/bmc/bmc_api_test.rb
```diff
@@ -21,6 +21,48 @@ def setup
     authorize user, pass
   end
 
+  def test_api_throws_401_error_when_auth_is_not_provided
+    Proxy::BMC::IPMI.stubs(:providers_installed).returns(['freeipmi'])
+    Proxy::BMC::Plugin.settings.stubs(:bmc_default_provider).returns('freeipmi')
+    auth = Object.new
+    auth.stubs(:provided?).returns(false)
+    auth.stubs(:basic?).returns(true)
+    auth.stubs(:credentials).returns(nil)
```
dmitri-d: I think you only need to stub auth.provided? in this test... Also, please use a mock instead of a stub there...

test/bmc/bmc_api_test.rb
```diff
@@ -21,6 +21,48 @@ def setup
...
+    assert_equal 200, last_response.status
+  end
+
+  def test_api_throws_401_error_when_auth_is_not_basic
+    Proxy::BMC::IPMI.stubs(:providers_installed).returns(['freeipmi'])
+    Proxy::BMC::Plugin.settings.stubs(:bmc_default_provider).returns('freeipmi')
+    auth = Object.new
+    auth.stubs(:provided?).returns(true)
+    auth.stubs(:basic?).returns(false)
+    auth.stubs(:credentials).returns('name', 'password')
```
dmitri-d: Again, you only need to stub/mock auth.provided and auth.basic? methods. I'd use a mock for the latter.

test/bmc/bmc_api_test.rb
logicminds: auth.basic was removed.  Although I am unsure of how to mock without stubbing?  Can you give me an example?  I am now using mock() instead of Object.new but thats all I changed.

dmitri-d:
@logicminds: you can use `expects(:blah)` instead of `stubs(:blah)` if you want to verify that a method was called. Mocha checks that all mocks were satisfied at the end of a test.

logicminds:
@witlessbird replaced the stubs with mocks.  This also helped find a few unnecessary mocks that I removed as well.

dmitri-d:
merged in at 46cf170. Thanks, @logicminds!

PR #272, opened 2015-04-01
Fixes #10003: Support for config file migrations

dmitri-d:
to use:  ruby migrate_settings.rb config_src_dir working_dir_path migrations_dir_path migrations_state_file_path

where: 
- config_src_dir: dir containing current/old config files (won't be modified)
- working_dir_path: dir where migration results will be stored; individual migrations store results in subdirectories; final results will be in 'results' subdir.
- migrations_dir_path: dir containing migrations (I can/should probably locate this automatically)
- migrations_state_file_path: file containing migration state, can be non-existent if this is the first run

@domcleal, @lzap: ping


lzap:
Little bit busy, on my radar. Sorry.

extra/migrate_settings.rb
```diff
@@ -1,108 +1,150 @@
...
     end
   end
 end
 
-# When running as a script
 if __FILE__ == $0 then
-  orig_file = ARGV[0] || '/etc/foreman-proxy/settings.yml'
-  data      = YAML.load_file(orig_file)
+  if ARGV.size < 4
+    p "Usage: migrate.rb <config_src_dir> <working_dir_path> <migrations_dir_path> <migrations_state_file_path>"
```
domcleal: Could we have more defaults for these, so it can work from the smart-proxy root directory with minimal arguments, using ./config, ./tmp, ./extra/migrations/ and some additional dir (add .gitkeep)?

dmitri-d: I was trying to avoid cli arguments parsing, hence the spartan interface. Would no arguments/fully specified arguments solution work?

domcleal: Ah okay.  getopts is quite easy :)  But yes, that would also be fine.

dmitri-d: I guess getopts is simple enough. Will add that.

domcleal:
To add another complexity, a smart proxy installation might have plugins installed too, so we need to be able to migrate those configs too.

I think we can get around it with some packages, if we repackaged the migrations into the main smart proxy migrations directory, but for source installations, I suppose we actually need to be loading migrations from the gem installation path or from the plugin registration.  Any thoughts on that?

dmitri-d:
I thought about automatically including migrations for non-default plugins. A couple of the issues involved: 
- finding out locations for migrations. Actually including migrations is as easy as specifying another directory in the search/load path.
- persisted state. Right now the assumption is that the history is linear, so I can simply discard all migrations before the migration that was executed last. I don't think this is going to work with third party plugins, and we'll need to treat migrations as a tree (a branch per plugin). At this point inter-branch ordering issues may enter the picture.

After considering the latter point I figured that migrations should be done by plugins/plugins packages themselves (they can reuse the main script, but point it to the correct dirs and files). 

Current approach will simply copy config files of third party plugins without any changes applied. User will then need to update non-core plugins by hand...

dmitri-d:
I suppose that I could store the complete history of executed migrations to address issues with third-party plugin migrations. Still, provided that most upgrades will be done via package managers, I'd avoid this bit altogether.

domcleal:
Okay, thanks for the info, I'm alright to leave it as a packaging problem for now for plugins, but we might have to watch and reconsider it in the future.

Keeping a full history would be closer to how Rails works, where I don't think you really worry about branching, but simply run migrations in date order including any missing ones.

extra/migrate_settings.rb
```diff
@@ -1,108 +1,150 @@
...
+  end
```
domcleal: missing an `exit` inside this conditional, it tries to continue execution

extra/migrate_settings.rb
```diff
@@ -1,108 +1,150 @@
...
+
+      load_migrations(migrations_dir_path)
+
+      executed_migrations = []
+      migrations = @migrations.since_last(last_migration)
+      migrations.each do |migration|
+        m = migration.new(working_dir_path)
+
+        create_migration_dirs(m.src_dir, m.dst_dir)
+        if  migration == migrations.first
```
domcleal: nitpick, extra space after `if`

extra/migrate_settings.rb
```diff
@@ -1,108 +1,150 @@
...
-  data      = YAML.load_file(orig_file)
+  if ARGV.size < 4
+    p "Usage: migrate.rb <config_src_dir> <working_dir_path> <migrations_dir_path> <migrations_state_file_path>"
+  end
 
-  output,unknown = migrate(data.dup)
-  exit(1) if data == output[:settings] # it looks like we already have the new style settings
-  write_to_files(output,unknown)
+  config_src_dir = ARGV[0]
+  working_dir_path = ARGV[1]
```
domcleal: The working dir is designed to hold both temporary (working) files and the results, right?  So the caller would run this script and then copy the results back over the real /etc/ dir?  And then probably rm -rf the wd...

domcleal: (including the migration state file, since that's assumed to be `wd/results/migration_state`)

extra/migrate_settings.rb
```diff
@@ -1,108 +1,150 @@
...
+  migrations_dir_path = ARGV[2]
+  migrations_state_file_path = ARGV[3]
+
+  ::Proxy::Migrator.new(working_dir_path, migrations_dir_path, config_src_dir, migrations_state_file_path).migrate
   exit(0)
-end
+end
```
domcleal: nitpick, please configure your editor to keep newlines :)

extra/migrate_settings.rb
```diff
@@ -1,108 +1,150 @@
...
+      @migrations << migration_class
+    end
 
-def modules
-  [ :settings, :tftp, :dns, :dhcp, :puppet, :puppetca, :bmc, :chef, :realm ]
-end
+    def since_last(last_migration = nil)
+      return @migrations if last_migration.nil? || last_migration.empty?
+      last_migration_idx = @migrations.map(&:name).index(last_migration)
+      starting_migration_idx =  last_migration_idx.nil? ? 0 : last_migration_idx + 1
```
domcleal: nitpick, two spaces after `=`

extra/migrations/27032015000000_migrate_monolithic_config.rb
```diff
@@ -0,0 +1,112 @@
+#
```
domcleal: The filename here is perhaps the wrong way around (DDMMYYYY) if we're sorting the filenames, should it be YYYYMMDD?

dmitri-d: Both monotonically increase, don't think it makes a difference (I find the "normal" format a bit easier on the eyes though).

dmitri-d:
@domcleal: added cli options parser, together with default values for all dirs/files. Migrations state now contains all migrations -- thought it was easier to add this now, rather than later.

I'm a bit torn between replacing original files (convenience) vs. creating a copy of everything (correctness, etc). I figured as the main mode is going to be from install scripts, the latter is ok, as we are moving files around anyway...

extra/migrations/27032015000000_migrate_monolithic_config.rb
domcleal: If we add a 2105201500000 though, it'll run before this one...

dmitri-d:
Oh, a minor convenience point -- it now prints dirs/files it's going to use to stdout.

extra/migrations/27032015000000_migrate_monolithic_config.rb
dmitri-d: oh, damn, lexicographical comparison !ftw.

dmitri-d: yeah, let's use the format you suggested.

dmitri-d:
- renamed monolithic config migration to use YYYYMMDD as file name prefix.
- migration state contains all past migrations now, not just the ones that were loaded during the last migration.

dmitri-d:
[test]

extra/migrate_settings.rb
```diff
@@ -1,108 +1,199 @@
...
-    :customrun_args     => :puppet,
-    :customrun_cmd      => :puppet,
-    :freeipa_remove_dns => :realm
-  }
-end
+    def load_past_migrations!
+      raise "Migration state file #{@migration_state_file_path} can't be found" unless File.exist?(@migration_state_file_path)
+      @past_migrations = File.read(@migration_state_file_path).split('\n')
+      self
+    end
```
lzap, line 19: Uncaught IO errors, I think nice error with emptying the array would be nice.

extra/migrate_settings.rb
```diff
@@ -1,108 +1,199 @@
...
-      modules.each do |mod|
-        next unless option.to_s =~ /^#{mod.to_s}/
-        output[mod][option] = value
-        data.delete(option)
+    def persist_migrations_state(result_dir_path)
+      migration_state = File.open(File.join(result_dir_path, "migration_state"), "w")
+      migration_state.write((past_migrations + migrations.map(&:to_s)).uniq.join("\n"))
+    ensure
+      migration_state.close unless migration_state.nil?
+    end
```
lzap, line 36: Ditto.

extra/migrate_settings.rb
```diff
@@ -1,108 +1,199 @@
...
+        m = migration.new(working_dir_path)
+
+        create_migration_dirs(m.src_dir, m.dst_dir)
+        if migration == migrations.first
+          copy_original_configuration(m.src_dir)
+        else
+          copy_previous_migration_results(executed_migrations.last.dst_dir, m.src_dir)
+        end
+
+        m.migrate
```
lzap, line 127: Good idea might be to put this into ensure block isolating each migration. Those containing errors would be skipped, error printed out and user could fix it later re-running the migration process again.

lzap, line 127: I suppose each migration is idependent, if not then this is better of course.

test/migrations/3_test_migration_three.rb
```diff
@@ -0,0 +1,4 @@
+class TestMigrationThree < ::Proxy::Migration
+  def migrate
+  end
+end
```
lzap, line 4: Do you intend to keep those?

test/migrations/migration_test.rb
```diff
@@ -0,0 +1,24 @@
...
+  def setup
+    @module = Module.new
+
+    @migration = ::Proxy::Migrations.new("dummy", ["#{@module}::TestMigrationOne", "#{@module}::TestMigrationTwo"])
+    ::Proxy::Migration.inject_migrations_instance(@migration)
+
+    @module.module_eval(File.read('./test/migrations/1_test_migration_one.rb'))
+    @module.module_eval(File.read('./test/migrations/2_test_migration_two.rb'))
+    @module.module_eval(File.read('./test/migrations/3_test_migration_three.rb'))
+  end
```
lzap, line 14: Oh I see. Sure.

lzap:
Few minor, one valid comments. Tests failed fyi.

I plan to test this on RHEL6 and RHEL7. If you have some RPM scratch builds that would be great, otherwise I can just test from git.

extra/migrate_settings.rb
dmitri-d, line 127: I don't think an assumption that migrations are independent of the final state of preceding migrations can be made -- I'm thinking upgrade to an upgrade scenario. If one migration fails, we need to abort the rest of migrations, hence the lack of error handling. Having said that, it would probably be useful to have meaningful error messages...

dmitri-d:
It appears that tests failed to due to a jenkins failure, re-running.

dmitri-d:
[test]

extra/migrate_settings.rb
dmitri-d, line 19: No point in continuing if an exception occurred, hence no cleaning up. Re: nice error -- I don't really want to wrap or catch/re-raise exceptions (as the whole thing should be killed on exception anyway). I could add logging, but not sure if it's going to be that useful either...

extra/migrate_settings.rb
lzap, line 127: In that case you need to encapsulate this line bellow

```
@migrations.persist_migrations_state(result_dir_path)
```

into `ensure` block. Because if an error occurs, all migrations that finished cleanly before the exception occured will never be persisted ergo executed multiple times once user resubmit the process.

I am still for catching/rethrowing the error and printing some _nice message_ to the user how to proceed. Stacktrace is maybe nice for engineer, but not for end user.

extra/migrate_settings.rb
lzap, line 19: These post-java guys... ;-) My idea was if the file was broken, you can start over with clean array. But no biggie.

lzap:
Ok looks good, tests are green. One comment in regard to the storing migrations.

dmitri-d:
@lzap: fixed the issue you pointed out!

extra/migrate_settings.rb
```diff
@@ -1,108 +1,200 @@
...
+
+  parser.parse!(args)
+  result
 end
 
-# When running as a script
 if __FILE__ == $0 then
-  orig_file = ARGV[0] || '/etc/foreman-proxy/settings.yml'
-  data      = YAML.load_file(orig_file)
+  options = parse_cli_options(ARGV)
```
lzap, line 235: Nit pick, this should error out when no options are provided, otherwise:

```
[lzap@lzapx smart-proxy-test]$ ruby extra/migrate_settings.rb
extra/migrate_settings.rb:17:in `load_past_migrations!': Migration state file /home/lzap/work/smart-proxy-test/config/migration_state can't be found (RuntimeError)
```

extra/migrate_settings.rb
```diff
@@ -1,108 +1,200 @@
...
-    :puppetdir          => :puppetca,
-    :ssldir             => :puppetca,
-    :sudo_command       => :puppetca,
-    :customrun_args     => :puppet,
-    :customrun_cmd      => :puppet,
-    :freeipa_remove_dns => :realm
-  }
-end
+    def load_past_migrations!
+      raise "Migration state file #{@migration_state_file_path} can't be found" unless File.exist?(@migration_state_file_path)
```
lzap, line 16: Should we create the state file automatically when it is not present?

extra/migrate_settings.rb
dmitri-d, line 235: Hmm, I think the default location is sensible? I could add an empty migration state there though...

extra/migrate_settings.rb
lzap, line 16: The same for the target directory?

lzap, line 16: Taking back, target directory _is_ created automatically.

dmitri-d, line 16: Hmm. Not sure, tbh. I feel that if the state file doesn't exist at the specified (or default) location, the user should be notified -- could be a wrong path, for example. Otherwise it's possible to execute a wrong set of migrations and not notice it (may or may not be an issue, depending on the implementation of migrations). @domcleal -- thoughts?

lzap:
Ok it works, but I have two issues:

```
[lzap@lzapx smart-proxy-test]$ ruby extra/migrate_settings.rb -s config -t config_result -m ms
using: 
 config files dir: /home/lzap/work/smart-proxy-test/config,
 working dir: /home/lzap/work/smart-proxy-test/config_result,
 migrations dir: /home/lzap/work/smart-proxy-test/extra/migrations,
 migrations state path: /home/lzap/work/smart-proxy-test/ms
[lzap@lzapx smart-proxy-test]$ find . -name migration_state
./config_result/result/migration_state
[lzap@lzapx smart-proxy-test]$ cat ms
```

First, the `migration_state` file was set to `ms` but it is empty and result was stored in `./config_result/result/migration_state` instead. We need this to be working because on RPM packaged systems we will override this into `/etc/foreman` most likely, or `/var/lib`. It must not be in `/usr/share/foreman` for sure.

Second, I see some output of settings on stdout. I think this is not necessary. I would like to see migration file or class names being applied instead.

lzap:
For the record, tested with: https://github.com/theforeman/smart-proxy/blob/1.5-stable/config/settings.yml.example

extra/migrate_settings.rb
domcleal, line 16: I agree with @witlessbird here, since it's an input file we should raise an error when it's missing.  We can ship an empty file in the packages and mark it as a config?

dmitri-d:
re: migrations state file -- it's only being read, but not written to. This is the same for the rest of migrations -- the final results will be located in results/ directory, nothing is being overwritten in the filesystem. My thinking was that we will be relying on install scripts to move/copy files into their locations for that ...

 I'm printing out various dirs on startup, mostly as a reminder, in case default values were used. Will add migrations too.

dmitri-d:
@lzap: now prints out migrations as they are being executed.

extra/migrate_settings.rb
```diff
@@ -1,108 +1,202 @@
...
+      @migrations.persist_migrations_state(result_dir_path)
+    end
+
+    def create_migration_dirs(src_dir, dst_dir)
+      FileUtils.mkdir_p(src_dir)
+      FileUtils.mkdir_p(File.join(dst_dir, "settings.d"))
+    end
+
+    def copy_original_configuration(src_dir)
+      FileUtils.cp_r(File.join(config_src_dir, '.'), src_dir)
```
domcleal: This doesn't quite work with our packaging, as the settings.d dir can be changed by `:settings_directory: /etc/foreman-proxy/settings.d` in the settings.yml file.  You might need to open the settings.yml, check for settings_directory and copy that path too.

extra/migrate_settings.rb
```diff
@@ -1,108 +1,202 @@
...
+    opts.on('-s', '--src_dir SRC_DIR_PATH', 'configuration files directory path (read-only)') do |src_dir|
+      result[:src_dir] = path(src_dir)
+    end
+    opts.on('-t', '--tmp_dir TMP_DIR_PATH', 'working directory path') do |tmp_dir|
+      result[:tmp_dir] = path(tmp_dir)
+    end
+    opts.on('-m', '--migrations_dir MIGRATIONS_DIR_PATH', 'migrations directory path (read-only)') do |migrations_dir|
+      result[:migrations_dir] = path(migrations_dir)
+    end
+    opts.on('-ms', '--migration_state MIGRATION_STATE_PATH', 'path to the file storing executed migrations (read-only)') do |ms_path|
```
domcleal: `-ms` isn't a valid short form, it needs to be one character:

```
11:25 $ ruby extra/migrate_settings.rb -ms /tmp/migration_state                                                                         
extra/migrate_settings.rb:17:in `load_past_migrations!': Migration state file /home/dcleal/code/foreman/smart-proxy/s can't be found (RuntimeError)
```

extra/migrate_settings.rb
```diff
@@ -1,108 +1,202 @@
...
+        create_migration_dirs(m.src_dir, m.dst_dir)
+        if migration == migrations.first
+          copy_original_configuration(m.src_dir)
+        else
+          copy_previous_migration_results(@migrations.executed_migrations.last.dst_dir, m.src_dir)
+        end
+
+        m.migrate
+
+        if migration == migrations.last
```
domcleal: If there aren't any migrations to run, the results don't get copied.  Should we call copy_to_results_dir outside of the loop?

Edit: if not, I'll need to handle this case explicitly - currently we just copy results, no matter what.

domcleal:
I tried updating our RPMs to use this, ran into a few issues noted above, but otherwise looks pretty good.  https://github.com/theforeman/foreman-packaging/pull/664 is the RPM PR, will duplicate for Debian once one works.

dmitri-d:
- changed how location of `settings.d` is specified: directly using `-d` cli option, determined using `settings_directory` option in `settings.yml`, or default value of `app_dir/config/settings.d` will be used.
- fixed short form options
- if no migrations were executed, results dir will contain original unchanged settings files.
- path to the main settings file can now be specified directly

domcleal:
Nitpick: add config/migration_state to .gitignore.

extra/migrate_settings.rb
```diff
@@ -1,108 +1,230 @@
...
+
+    def persist_migrations_state(result_dir_path)
+      migration_state = File.open(File.join(result_dir_path, "migration_state"), "w")
+      migration_state.write((past_migrations + executed_migrations.map {|m| m.class.name}).uniq.join("\n"))
+    ensure
+      migration_state.close unless migration_state.nil?
+    end
+
+    def modules_configuration_exist?
+      past_migrations.include?("MigrateMonolithicConfig")
```
domcleal: There's a bug when you're migrating the existing (say 1.8) config which _does_ have a settings.d/ directory but this migration hasn't run (since our state file is empty).

This results in us trying to run MigrateMonolithicConfig, but copying the original config files into the source dir for the migration skips settings.d, because it runs this exist? method, which returns false.  We do actually need to copy the settings.d/ directory, if it exists, into the source dir else we lose it.

<pre>
✔ ~/code/foreman/smart-proxy [:74ed262|✚ 1…5] 
16:06 $ ll config/migration_state
-rw-rw-r--. 1 dcleal dcleal 0 May 22 11:00 config/migration_state
✔ ~/code/foreman/smart-proxy [:74ed262|✚ 1…5] 
16:06 $ ruby extra/migrate_settings.rb 
using:
  config file: /home/dcleal/code/foreman/smart-proxy/config/settings.yml,
  modules config dir: /home/dcleal/code/foreman/smart-proxy/config/settings.d,
  working dir: /home/dcleal/code/foreman/smart-proxy/tmp,
  migrations dir: /home/dcleal/code/foreman/smart-proxy/extra/migrations,
  migrations state path: /home/dcleal/code/foreman/smart-proxy/config/migration_state
      
Running migrations...
migrate_monolithic_config
copy_original_configuration("/home/dcleal/code/foreman/smart-proxy/tmp/migrate_monolithic_config/src", false)
✔ ~/code/foreman/smart-proxy [:74ed262|✚ 1…5] 
16:07 $ ll tmp/result/
total 12
-rw-rw-r--. 1 dcleal dcleal   23 May 27 16:07 migration_state
drwxrwxr-x. 2 dcleal dcleal 4096 May 27 15:59 settings.d
-rw-rw-r--. 1 dcleal dcleal  481 May 27 16:07 settings.yml
✔ ~/code/foreman/smart-proxy [:74ed262|✚ 1…5] 
16:07 $ ll tmp/result/settings.d/
total 0
✔ ~/code/foreman/smart-proxy [:74ed262|✚ 1…5] 
16:07 $ ll tmp/migrate_monolithic_config/*
tmp/migrate_monolithic_config/dst:
total 8
drwxrwxr-x. 2 dcleal dcleal 4096 May 27 15:59 settings.d
-rw-rw-r--. 1 dcleal dcleal  481 May 27 16:07 settings.yml

tmp/migrate_monolithic_config/src:
total 4
-rw-r--r--. 1 dcleal dcleal 442 May 27 16:07 settings.yml
✔ ~/code/foreman/smart-proxy [:74ed262|✚ 1…5] 
16:07 $ ll tmp/migrate_monolithic_config/dst/settings.d/
total 0
✔ ~/code/foreman/smart-proxy [:74ed262|✚ 1…5] 
</pre>

dmitri-d:
`settings.d` is now always being copied if it's present.

domcleal:
Thanks @witlessbird, this is mergeable, but I'm going to just finish off the packaging.

extra/migrate_settings.rb
```diff
@@ -1,108 +1,228 @@
...
+        @migrations.executed_migrations << m
+      end
+
+      if @migrations.executed_migrations.empty?
+        copy_original_configuration(result_dir_path)
+      else
+        copy_to_results_dir(@migrations.executed_migrations.last.dst_dir, result_dir_path)
+      end
+    ensure
+      @migrations.persist_migrations_state(result_dir_path)
```
domcleal: This can hide exceptions when they're thrown at the start of this method, as result_dir_path doesn't exist until the mkdir_p call on line 114:

<pre>
# ruby /usr/share/foreman-proxy/extra/migrate_settings.rb 
/usr/share/foreman-proxy/extra/migrate_settings.rb:34:in `initialize': No such file or directory - /usr/share/foreman-proxy/tmp/result/migration_state (Errno::ENOENT)
        from /usr/share/foreman-proxy/extra/migrate_settings.rb:34:in `open'
        from /usr/share/foreman-proxy/extra/migrate_settings.rb:34:in `persist_migrations_state'
        from /usr/share/foreman-proxy/extra/migrate_settings.rb:140:in `migrate'
        from /usr/share/foreman-proxy/extra/migrate_settings.rb:224
</pre>

extra/migrate_settings.rb
```diff
@@ -1,108 +1,228 @@
...
+    end
+
+    def load_migrations(migrations_dir_path)
+      migration_files = Dir.glob(File.join(migrations_dir_path, "*.rb")).sort
+      ::Proxy::Migration.inject_migrations_instance(@migrations)
+      migration_files.each { |file| require file }
+    end
+
+    def migrate
+      raise "Migrations dir '#{migrations_dir_path}' doesn't exist" unless Dir.exist?(migrations_dir_path)
```
domcleal: `Dir.exist?` is throwing an error on Ruby 1.8.7, suggest you perhaps use `File.directory?`

extra/migrations/20150327000000_migrate_monolithic_config.rb
```diff
@@ -0,0 +1,112 @@
...
+#
+# Greg Sutcliffe <gsutclif@redhat.com> 2014
+
+require 'fileutils'
+require 'yaml'
+
+class MigrateMonolithicConfig < ::Proxy::Migration
+  def migrate
+    # Skip the migration if we are dealing with modularized proxy
+    if Dir.exist?(File.join(src_dir, "settings.d"))
```
domcleal: ditto

extra/migrate_settings.rb
domcleal: Are you able to add any test coverage of this method, so it gets run on multiple Rubies?

dmitri-d:
@domcleal: fixed issues you found when running on 1.8.7; refactored `Migrator#migrate` method and added tests for paths validation. Will add a test for a migration execution tomorrow.

dmitri-d:
now with moar tests!

domcleal:
Merged as 73ca26d, thanks @witlessbird!

PR #273, opened 2015-04-02
Fixes #7543 - allow additional rubyipmi connection options to be passed through

logicminds:
- allows the user to pass additional rubyipmi options that were not exposed through smart proxy before.
- replaces some stubs with mock in other tests

### DIFF
--- a/modules/bmc/ipmi.rb
+++ b/modules/bmc/ipmi.rb
@@ -65,7 +65,7 @@ module Proxy
       end
 
       def connect(args = { })
-        Rubyipmi.connect(args[:username], args[:password], args[:host], args[:bmc_provider])
+        Rubyipmi.connect(args[:username], args[:password], args[:host], args[:bmc_provider], args[:options] || {})
       end
 
       # returns boolean true if connection to device is successful
--- a/test/bmc/bmc_api_test.rb
+++ b/test/bmc/bmc_api_test.rb
@@ -100,14 +100,14 @@ class BmcApiTest < Test::Unit::TestCase
   end
 
   def test_api_bmc_setup_returns_new_ipmi_proxy_given_ipmitool
-    api = Proxy::BMC::Api.new!
     Proxy::BMC::Plugin.settings.stubs(:bmc_default_provider).returns('freeipmi')
-    auth = Object.new
+    api = Proxy::BMC::Api.new!
+    auth = mock()
     auth.stubs(:provided?).returns(true)
     auth.stubs(:basic?).returns(true)
     auth.stubs(:credentials).returns('username','password')
     api.stubs(:auth).returns(auth)
-    api.stubs(:find_ipmi_provider).returns('ipmitool')
+    api.expects(:find_ipmi_provider).with('ipmitool').returns('ipmitool')
     api.stubs(:params).returns(:bmc_provider => 'ipmitool', :host => :host)
     result = api.bmc_setup
     assert_kind_of(Proxy::BMC::IPMI,result)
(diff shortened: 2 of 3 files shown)


modules/bmc/ipmi.rb
```diff
@@ -65,7 +65,8 @@ def bootdevice=(args={ :device => nil, :reboot => false, :persistent => false })
       end
 
       def connect(args = { })
-        Rubyipmi.connect(args[:username], args[:password], args[:host], args[:bmc_provider])
+        args[:options] ||= {}
+        Rubyipmi.connect(args[:username], args[:password], args[:host], args[:bmc_provider], args[:options])
```
dmitri-d: you can use `args[:options] || {}` instead of modifying the passed in hashmap.

test/bmc/bmc_api_test.rb
```diff
@@ -64,10 +64,10 @@ def test_api_bmc_setup_returns_new_ipmi_proxy_given_ipmitool
...
     Proxy::BMC::Plugin.settings.stubs(:bmc_default_provider).returns('freeipmi')
     Proxy::BMC::Plugin.settings.stubs(:provider_log_level).returns(nil)
-    auth = Object.new
-    auth.stubs(:provided?).returns(true)
-    auth.stubs(:basic?).returns(true)
-    auth.stubs(:credentials).returns('username','password')
+    auth = mock()
+    auth.expects(:provided?).returns(true)
+    auth.expects(:basic?).returns(true)
+    auth.expects(:credentials).returns('username','password')
```
dmitri-d: My understanding is that this test verifies that `ipmitool` is being used when `bmc_provider` is set appropriately. If this is the case, I would stick with stubs for authorization (not really important for this test whether auth calls have been made), but would probably use a mock for `find_ipmi_provider`, together with verifying parameters passed to it...

test/bmc/bmc_api_test.rb
```diff
@@ -79,10 +79,10 @@ def test_api_bmc_setup_returns_new_ipmi_proxy_given_freeipmi
...
```
dmitri-d: same as above comment.

logicminds:
@witlessbird adjusted the tests as noted above.  Should be ready to go.

dmitri-d:
merged in at f4f52e0. Thanks, @logicminds!

lzap:
What kind of options do you usually pass in here? And will this be escaped properly?

I am asking because I am trying to reproduce this BZ https://bugzilla.redhat.com/show_bug.cgi?id=1206361 and it looks like this could solve the problem.

lzap:
Ok now I see it's related. @logicminds could you please file another PR referring this issue number with extra documentation in the example config file what can be set? I see that instead of "-L" one needs to use "privilege" key value. It is worth documenting this.

domcleal:
This isn't configuration AFAICT, it's passed into the BMC request from the API consumer.  See http://projects.theforeman.org/issues/7543#note-10 and also http://projects.theforeman.org/issues/10102.

dmitri-d:
Indeed, see test at https://github.com/theforeman/smart-proxy/pull/273/files#diff-510c048f716a460d0d956bec115a3213R138 for an example.

lzap:
Is there a related Foreman Core patch then which allows sending required options?

lzap:
No, I assume. Created http://projects.theforeman.org/issues/10112

lzap:
I am testing this and I must be doing something wrong. Help:

```
curl -u root:xyz -3 -H "Accept:application/json" -X GET -d '{"options":{"driver":"lanplus"}}' http://localhost:8000/bmc/1.2.3.4/chassis/power/status
{"action":"status","result":null}
```

Is the input format correct? Is that a JSON?

It works only if I hardcode in the codebase like this:

```
params[:options] = { :driver => "lan20" }
```

lzap:
Tried both `-d '{"options":{"driver":"lanplus"}}'` and `-d '{"options":{"driver":"lan20"}}'` and it looks like it does not work with the input I am giving it. Help @logicminds :-)

domcleal:
You probably need to specify `Content-Type: application/json` if you're passing JSON in as a body.  Passing Accept indicates what you want to receive.

logicminds:
I'll update the smart proxy wiki with instructions as well.  Additionally I am going to make lan20 the default in rubyipmi which would require older devices to pass in the driver type.  Additional we could make a default driver type setting in smart proxy. While auto detect does work I have only ever tested on my hp system and only assume it works on others.  @lzap

logicminds:
Rubyipmi 0.9.3 was just released which fixes an issue where the options hash keys are converted to symbols.  This should fix the issue.  @domcleal

lzap:
Ok thanks I can confirm that rubyipmi 0.10.0 fixes the issue. Here is the command I used:

```
# curl -u root:xyz -3 -H "Accept: application/json" -H "Content-Type: application/json" -X GET -d '{"options":{"driver":"lan20"}}' http://localhost:8000/bmc/1.2.3.4/chassis/power/status
{"action":"status","result":"on"}
```

lzap:
Also I see that "lan20" is now the default. This will break on older systems. The question is if we want to put this into 1.8. I guess so.

logicminds:
Yea, worst case scenario is we make our own default within smart-proxy instead of relying on Rubyipmi defaults by just controlling some default options to Rubyipmi.  But the majority of people would have to be running lan 2.0 by now.  If someone is running an older system it would be 10 years old by now.

PR #275, opened 2015-04-10
Fixes #10116 - BMC is not able to use parameters passed in through body

logicminds:
- increase rubyipmi version to 0.10.0 which defaults to lan20 driver
- fixes a issue where parameters in a JSON encoded body were being ignored
- adds content type json to response header

### DIFF
--- a/lib/proxy/helpers.rb
+++ b/lib/proxy/helpers.rb
@@ -40,6 +40,27 @@ module Proxy::Helpers
     end
   end
 
+  # parses the body as json and returns a hash of the body
+  # returns empty hash if there is a json parse error or body is empty
+  # request.env["CONTENT_TYPE"] must contain application/json in order for the json to be parsed
+  def parse_json_body
+    json_data = {}
+    # if the user has explicitly set the content_type then there must be something worth decoding
+    # we use a regex because it might contain something else like: application/json;charset=utf-8
+    # by default the content type will probably be set to "application/x-www-form-urlencoded" unless the
+    # user changed it.  If the user doesn't specify the content type we just ignore the body since a form
+    # will be parsed into the request.params object for us by sinatra
+    if request.env["CONTENT_TYPE"] =~ /application\/json/
+      begin
+        body_parameters = request.body.read
+        json_data = JSON.parse(body_parameters)
+      rescue => e
+        log_halt 415, "Invalid JSON content in body of request: \n#{e.message}"
+      end
+    end
+    json_data
+  end
+
   # reverse lookup an IP address while verifying it via forward resolv
   def remote_fqdn(forward_verify=true)
     ip = request.env['REMOTE_ADDR']
--- a/modules/bmc/bmc_api.rb
+++ b/modules/bmc/bmc_api.rb
@@ -8,6 +8,10 @@ module Proxy::BMC
     # All GET requests will only read ipmi data, no changes
     # All PUT requests will update information on the bmc device
 
+    before do
+      content_type :json
+    end
+
     # return list of available options
     get "/" do
       res = ['providers', 'providers/installed', 'host']
@@ -260,7 +264,7 @@ module Proxy::BMC
 
     def bmc_setup
       # Either use the default provider or allow user to specify provider in request
-      provider_type ||= params[:bmc_provider] || Proxy::BMC::Plugin.settings.bmc_default_provider
+      provider_type = params['bmc_provider'] || body_parameters['bmc_provider'] || Proxy::BMC::Plugin.settings.bmc_default_provider
       provider_type.downcase! if provider_type
       # unless the provider is shell find a suitable provider
       if provider_type != 'shell'
@@ -269,12 +273,9 @@ module Proxy::BMC
 
       case provider_type
         when 'freeipmi', 'ipmitool'
-
           log_halt 401, "unauthorized" unless auth.provided?
           log_halt 401, "bad_authentication_request, credentials are not in auth.basic format" unless auth.basic?
-
           username, password = auth.credentials
-
           # this causes rubyipmi to use the supplied logger, most actions in rubyipmi only output during Logger::DEBUG
           Proxy::BMC::IPMI.logger = logger
 
@@ -283,7 +284,8 @@ module Proxy::BMC
           # to execute ipmi commands and has nothing to do with authorization
           # of using smart-proxy. Its simply a tunnel to pass credentials through,
           # since we are essentially remotely executing ipmi commands using Rubyipmi.
-          args = { :host     => params[:host], :username     => username, :options => params[:options],
+
+          args = { :host     => params[:host], :username     => username, :options => body_parameters['options'],
                    :password => password, :bmc_provider => provider_type }
           @bmc = Proxy::BMC::IPMI.new(args)
         when "shell"
@@ -294,6 +296,15 @@ module Proxy::BMC
       end
     rescue => e
       log_halt 400, e
+    end
+
+    # gets any parameters that were parsed into the body
+    # we have to parse the JSON body in the request and merge into the params object for later use
+    # if the user does not set the content type to application/json we will just assume there is garbage in the body
+    # also if the user decides to do http://127.0.0.1/bmc/192.168.1.6/test?bmc_provider=freeipmi as well as pass in
+    # a json encode body with the parameters, all of these items will be merged together
+    def body_parameters
+      @body_parameters ||= parse_json_body.merge(params)
     end
 
     def auth
--- a/bundler.d/bmc.rb
+++ b/bundler.d/bmc.rb
@@ -1,3 +1,3 @@
 group :bmc do
-  gem 'rubyipmi', '>= 0.9.2'
+  gem 'rubyipmi', '>= 0.10.0'
 end
--- a/modules/bmc/ipmi.rb
+++ b/modules/bmc/ipmi.rb
@@ -65,7 +65,12 @@ module Proxy
       end
 
       def connect(args = { })
-        Rubyipmi.connect(args[:username], args[:password], args[:host], args[:bmc_provider], args[:options] || {})
+        if args[:options].instance_of?(Hash)
+          options = args[:options]
+        else
+          options = {}     # catches nil and empty string cases
+        end
+        Rubyipmi.connect(args[:username], args[:password], args[:host], args[:bmc_provider], options)
       end
 
       # returns boolean true if connection to device is successful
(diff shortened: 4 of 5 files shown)


logicminds:
[test]

modules/bmc/bmc_api.rb
```diff
@@ -8,6 +8,30 @@ class Api < ::Sinatra::Base
...
+        begin
+
+          body_parameters = request.body.read
+          unless body_parameters.nil? || body_parameters.empty?
+            json_data = JSON.parse(body_parameters)
+            # convert all keys to symbols
+            json_data.keys.each do |key|
+              json_data[(key.to_sym rescue key) || key] = json_data.delete(key)
+            end
+            params.merge!(json_data)
```
lzap: Frankly, I don't like merging anything into params hash at all. This can bring security issues in the future (if we authorize users per host against URL then this is vulnerable to passing a different host accessing arbitrary system).

Why don't you create a helper method that parses the body and returns what you need?

Other than this one :+1: It took me an hour on Friday to realize this actually does not work...

dmitri-d: I second lzap. Let's treat incoming parameters/body/etc as immutable and provide convenience methods to retrieve them or change their representation.

logicminds:
I ended up changing a few other things as well.
1. moved code to helper method as suggested
2. moved body_parameters to a method call
3. merged params into body_parameters instead of other way around (as suggested).  This maintains backward compatibility with bmc_provider being passed in through non body params.
4. Some tests were updated to use get/put calls since the request object is now mandatory on some tests and its easier to mock the entire thing with get/put calls than by stubbing individual methods.
5. Removed conversion to symbols since its quicker to just keep as is.

logicminds:
My only other comment about this code is that the GET requests ideally should be changed to POST requests since GET requests don't normally have a body associated with the request.  The problem is just maintaining backward compatibility.  It works fine passing a body in a get request but as mentioned [here](http://stackoverflow.com/questions/978061/http-get-with-request-body) there are some drawbacks.  This wasn't required before because we were not passing a body, but now that we are we might need to use the solutions in the [url](http://stackoverflow.com/questions/978061/http-get-with-request-body).  So I am going to open a new ticket regarding this issue since the other modules in smart_proxy might be using the same GET requests instead of a SEARCH request.  @lzap @witlessbird 

Also, since the logging of Rubyipmi commands was added in a previous commit you can now use curl commands to test some of this stuff without actually having a IPMI device.  The commands will fail because the host cannot be reached, but you at least see the output of the rendered command.

``` shell
$ curl -vks --header "Content-Type: application/json" -d '{"bmc_provider":"ipmitool", "options":{"privilege":"USER"}}' 'http://127.0.0.1:8443/bmc/192.168.1.21/chassis/power/on' -X PUT -u admin:password

localhost - - [13/Apr/2015:16:45:18 PDT] "PUT /bmc/192.168.1.21/chassis/power/status HTTP/1.1" 200 52
- -> /bmc/192.168.1.21/chassis/power/status
D, [2015-04-13T16:45:32.564089 #23113] DEBUG -- : require_ssl_client_verification: skipping, non-HTTPS request
D, [2015-04-13T16:45:32.571708 #23113] DEBUG -- : /usr/local/bin/ipmitool -H 192.168.1.21 -U admin -L USER -I lanplus -f /var/folders/h6/v6nv76td37s7vqj902_z59kh0000gn/T/20150413-23113-1c9vroz power status
D, [2015-04-13T16:45:43.624149 #23113] DEBUG -- : "/usr/local/bin/ipmitool -H 192.168.1.21 -U admin -L USER -I lanplus -f /var/folders/h6/v6nv76td37s7vqj902_z59kh0000gn/T/20150413-23113-1c9vroz power status"
D, [2015-04-13T16:45:43.624230 #23113] DEBUG -- : IPMI LAN send command failed
Error: Unable to establish IPMI v2 / RMCP+ session
IPMI LAN send command failed
IPMI LAN send command failed
Error: Unable to establish IPMI v2 / RMCP+ session
IPMI LAN send command failed
IPMI LAN send command failed
Error: Unable to establish IPMI v2 / RMCP+ session


```

lzap:
I am developing the UI part of this thing. Will test once I have it done.

lzap:
Yeah create a ticket for `GET`, I think we can refactor later. Btw I saw an IPMI simulator software somewhere. Should I bother trying it? Also I am in the middle of writing a "ssh" provider that can be used for testing (will simply ssh and reboot/shutdown - no power on support tho :)

dmitri-d:
I would like to start by saying that converting calls indiscriminately to POSTs (or SEARCHes) is a terrible idea: there's a pretty specific semantics associated with all http methods and violating it isn't the way to go forward. 

I see no discussion of handling of body data in GET requests in http spec (http://tools.ietf.org/html/rfc2616#page-53) -- it's undefined (aka vendor-specific). I could argue that since our platform (webrick + sinatra) supports passing of parameters in GETs, we could leave things as they stand.

Another option is to reconsider this module api, and make it more restful. A possible approach is to make `chassis`,  `lan`,  and `test` top level resources, and treat impi providers, hosts, etc as filter parameters, for example:

`GET /:host/chassis/power/?:action?` would become
 `GET /chassis/power/:action?provider=:provider_id&host=:host_id...` or

`GET /:host/lan/?:action?` would become
 `GET /lan/:action?provider=:provider_id&host=:host_id...`

PUT requests can be handled similarly. I'm not sure at this point how well this would work for passing in of a parameter hash. What kind of parameters are those? How many are going to be passed in at a time?

Depending on the type and quantity of parameters in that hash, we might need to introduce a concept of a session, which isn't ideal and would require some storage on the smart-proxy side (something we had before, albeit as an experimental feature).

logicminds:
Note this pull request is ready and I don't intend on changing anything else.   The discussion above is for a possible future enhancement that is not required for this pull request.

lib/proxy/helpers.rb
```diff
@@ -40,6 +40,25 @@ def https_cert_cn
...
+  def parse_json_body
+    json_data = {}
+    if request.env["CONTENT_TYPE"] == "application/json"
+      begin
+        body_parameters = request.body.read
+        unless body_parameters.nil? || body_parameters.empty?
+          json_data = JSON.parse(body_parameters)
+        end
+      rescue => e
+        logger.debug("Invalid JSON content in body of request: #{e.message}")
```
dmitri-d: do you just swallow the exception here? Should probably respond with a 400?

lib/proxy/helpers.rb
```diff
@@ -40,6 +40,25 @@ def https_cert_cn
...
+      end
+    else
+      logger.warn("Request header content_type not set to application/json")
```
dmitri-d: You probably want to notify the caller that you expect json payload in the body.

lib/proxy/helpers.rb
logicminds: The idea was to swallow the error and assume the user would view the logs should something go wrong with the request.  But it could also be useful to return 400 as well.  I'll go ahead and change to 400 since I was on the fence about this already.

lib/proxy/helpers.rb
logicminds: Would this restriction break anything around foreman?  Does foreman submit correct content_type?  I'll go ahead and update this as suggested.

logicminds:
I updated the code to return error [415](http://tools.ietf.org/html/rfc7231#section-6.5.13).

So if the content of the body is invalid when json is supplied, error 415 will be returned.
If the user passes in the wrong content-type when supplying json in the body we do not error out because the bmc api can still rely on application/x-www-form-urlencoded parameters and some defaults so its technically not an error.   About the only way for me to return a useful error is to parse every single request coming through the bmc controller even if the user does not supply a body full of parameters.  I have kept this error handling out because I think running the json parser on every request just to determine a correct content-type matches with a json body seems like a waste of resources.

So if a user does this and forgets the content-type header: 

Invalid call that will result in the body data being ignored, and the use of the default bmc_provider being used.

``` shell
curl -vks -d '{"bmc_provider":"ipmitool", "options":""}' 'http://127.0.0.1:8443/bmc/192.168.1.21/chassis/power/on?bmc_provider=freeipmi' -X PUT -u admin:password

```

valid call without json body

``` shell
'http://127.0.0.1:8443/bmc/192.168.1.21/chassis/power/on?bmc_provider=freeipmi' -X PUT -u admin:password

```

The defaults will be used which the default content-type is "application/x-www-form-urlencoded".    Sinatra sees this as

``` ruby
request.params
=> {"bmc_provider"=>"freeipmi",
 "{\"bmc_provider\":\"ipmitool\", \"options\":\"\"}"=>nil}

```

dmitri-d:
Ok, this makes sense. Might be worth leaving a comment in the code and/or docs regarding this behaviour though.

logicminds:
I'll update the docs and have already commented on most of this in the helper code.

lzap:
Still not done with my provider, hopefuly this week to test this.

lzap:
The provider is ready: https://github.com/theforeman/smart-proxy/pull/278

lzap:
The code looks good, testing.

logicminds:
@witlessbird unless ther are any objections this should be good to merge.

lzap:
Sorry to let you wait. So in my testing with rubyipmi 0.10 I still see lanplus as the default one:

```
curl ... /bmc/192.168.1.21/chassis/power/on -d '{"bmc_provider":"ipmitool","options":{"privilege":"USER"}}'
DEBUG -- : /usr/bin/ipmitool -H 192.168.1.21 -U admin -L USER -I lanplus -f /tmp/20150429-30376-e154q power status
```

Is this expected? I would expect lan20 the default one.

lzap:
When I explicitly specify driver lan20, I still see lanplus. Something does not work. This is my helper tool for easier curl: https://github.com/lzap/bin-public/blob/master/req-proxy

```
~/bin/public/req-proxy /bmc/192.168.1.21/chassis/power/on -d '{"bmc_provider":"ipmitool","options":{"driver":"lan20"}}' -X PUT

/usr/bin/ipmitool -H 192.168.1.21 -U admin -I lanplus -f /tmp/20150429-30522-fmuoz9 power status
```

logicminds:
@lzap its working as expected.  The lan20 driver is the generic name that gets mapped the the provider specific driver type.  If you change providers you will see a different driver type on the command line even though you still specify lan20.

https://github.com/logicminds/rubyipmi/blob/master/lib/rubyipmi/freeipmi/connection.rb#L27
https://github.com/logicminds/rubyipmi/blob/master/lib/rubyipmi/ipmitool/connection.rb#L31

lzap:
Merged as 6ebc1ccd9e993d377eba9052470380d29a29a, thank you!

PR #278, opened 2015-04-23
Fixes #10250 - simple SSH BMC provider

lzap:
Also removed nonsense test which always fail on my configuration (I have a
different TFTP root set on my dev setup).

### DIFF
--- a/config/settings.d/bmc.yml.example
+++ b/config/settings.d/bmc.yml.example
@@ -5,4 +5,15 @@
 # Available providers:
 # - freeipmi / ipmitool - requires the appropriate package installed, and the rubyipmi gem
 # - shell - for local reboot control (requires sudo access to /sbin/shutdown for the proxy user)
+# - ssh - limited remote control (status, reboot, turn off)
 #:bmc_default_provider: freeipmi
+
+# default user and ssh key for ssh provider
+#:bmc_ssh_user: root
+#:bmc_ssh_key: /usr/share/foreman/.ssh/id_rsa
+
+# individual commands for actions
+#:bmc_ssh_powerstatus: "true"
+#:bmc_ssh_powercycle: "shutdown -r +1"
+#:bmc_ssh_poweroff: "shutdown +1"
+#:bmc_ssh_poweron: "false"
--- a/lib/proxy/util.rb
+++ b/lib/proxy/util.rb
@@ -91,6 +91,29 @@ module Proxy::Util
     end
   end
 
+  def shell_command(cmd, wait = true)
+    begin
+      c = popen(cmd)
+      unless wait
+        Process.detach(c.pid)
+        return 0
+      end
+      Process.wait(c.pid)
+    rescue Exception => e
+      logger.error("Exception '#{e}' when executing '#{cmd}'")
+      return false
+    end
+    logger.warn("Non-null exit code when executing '#{cmd}'") if $?.exitstatus != 0
+    $?.exitstatus == 0
+  end
+
+  def popen(cmd)
+    # 1.8.7 note: this assumes that cli options are space-separated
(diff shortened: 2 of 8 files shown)


logicminds:
Can you provide test coverage for the new ssh provider?   What user will the ssh command login as?  If not the root user you will need to use sudo on the remote machine to run those commands.  This most likely will not work on windows and non linux hosts.  I am also wondering if we should give the ability to pass in a different ssh key file reference (use this key location instead of the default).  

On a side note I wonder if we should add support for fence agents which would provide much more coverage of various devices if needed. https://access.redhat.com/articles/28603

lzap:
Rebased, added user, individual commands configuration, tests.

config/settings.d/bmc.yml.example
```diff
@@ -5,4 +5,15 @@
...
+# - ssh - limited remote control (status, reboot, turn off)
 #:bmc_default_provider: freeipmi
+
+# default user and ssh key for ssh provider
+#:bmc_ssh_user: root
+#:bmc_ssh_key: /usr/share/foreman/.ssh/id_rsa
+
+# individual commands for actions
+#:bmc_ssh_powerstatus: "true"
+#:bmc_ssh_powercycle: "shutdown -r +1"
```
dmitri-d, line 17: Do all of these (`powercycle` and `poweroff` in particular) need to be settings? Can they be moved into the provider itself?

lib/proxy/util.rb
```diff
@@ -91,6 +91,29 @@ def escape_for_shell(command)
     end
   end
 
+  def shell_command(cmd, wait = true)
```
dmitri-d, line 94: Could you open another PR to update puppetrun providers to use this method instead the one defined in modules/puppet_proxy_common/runner.rb?

test/bmc/bmc_api_ssh_test.rb
```diff
@@ -0,0 +1,47 @@
...
+  include Rack::Test::Methods
+
+  def app
+    Proxy::BMC::Api.new
+  end
+
+  def setup
+    @host = "somehost"
+    @args = { :bmc_provider => "ssh" }
+    Proxy::BMC::Plugin.settings.stubs(:poweron).returns('echo poweron')
```
dmitri-d: Please use `Proxy::BMC::Plugin.load_test_settings()` instead of stubbing (less pain down the road).

dmitri-d (changes requested):

config/settings.d/bmc.yml.example
lzap: As @logicminds stated above, there are some fencing utilities which works from shell, so you could possibly configure them here and use the ssh provider with them.

lib/proxy/util.rb
lzap: Absolutely, can you ping me after you merge this so I can fork it easily :-)

test/bmc/bmc_api_ssh_test.rb
lzap: Added.

dmitri-d:
[test]

dmitri-d:
[test]

dmitri-d:
Unhappy rubocop is unhappy.

lzap:
Done!

dmitri-d:
Merged as d959942. Thanks, @lzap!

mmoll:
anybody up for a corresponding PR to puppet-foreman_proxy? :)

lzap:
Michael Moll píše v Út 04. 10. 2016 v 12:28 -0700:

> anybody up for a corresponding PR to puppet-foreman_proxy? :)

I use this mainly for development, I don't see many use cases in the
production environment.

## 

Later,
 Lukas "lzap" Zapletal

PR #279, opened 2015-04-24
Fixes #10259 - Template proxy does not lookup by MAC address

whefter:
The native template API (/unattended) can lookup by MAC address (PXElinux?mac=aa:bb:cc:dd:ee:ff) in addition to lookup by token (this is very useful for provisioning Windows, for example). The template proxy feature, so far, can not.

Attached is a pull request with my attempt to fix this issue.

### DIFF
--- a/modules/templates/handler.rb
+++ b/modules/templates/handler.rb
@@ -7,23 +7,25 @@ module Proxy::Templates
   class Handler < ::Proxy::HttpRequest::ForemanRequest
     extend Proxy::Log
 
-    def get_template(kind, token, static = false)
-      opts = { :token => token,
-               :url   => Proxy::Templates::Plugin.settings.template_url
+    def get_template(kind, params, static = false)
+      opts = {
+        :url => Proxy::Templates::Plugin.settings.template_url
       }
+      opts[:mac]    = params[:mac] if params.has_key?('mac')
+      opts[:token]  = params[:token] if params.has_key?('token')
       opts[:static] = static if static
       request = request_factory.create_get("/unattended/#{kind}", opts)
       res = send_request(request)
 
       # You get a 201 from the 'built' URL
-      raise "Error retrieving #{kind} for #{token} from #{uri.host}: #{res.class}" unless ["200", "201"].include?(res.code)
-      Proxy::Log.logger.info "Template: request for #{kind} using #{token} at #{uri.host}"
+      raise "Error retrieving #{kind} for #{opts.to_json} from #{uri.host}: #{res.class}" unless ["200", "201"].include?(res.code)
+      Proxy::Log.logger.info "Template: request for #{kind} using #{opts.to_json} at #{uri.host}"
       res.body
     end
 
-    def self.get_template kind, token, static = false
+    def self.get_template kind, params, static = false
       @handler ||= Handler.new
-      @handler.get_template(kind,token, static)
+      @handler.get_template(kind, params, static)
     end
   end
 
--- a/modules/templates/templates_api.rb
+++ b/modules/templates/templates_api.rb
@@ -14,8 +14,8 @@ class Proxy::TemplatesApi < Sinatra::Base
   end
 
   get "/:kind" do |kind|
-    log_halt(500, "Failed to retrieve #{kind} template for #{params[:token]}: ") do
-      Proxy::Templates::Handler.get_template(kind, params[:token], params[:static])
+    log_halt(500, "Failed to retrieve #{kind} template for #{params.to_json}: ") do
+      Proxy::Templates::Handler.get_template(kind, params, params[:static])
     end
   end
 end
--- a/test/templates/template_test.rb
+++ b/test/templates/template_test.rb
@@ -17,15 +17,25 @@ class TemplateTest < Test::Unit::TestCase
 
   def test_template_requests_return_data_and_contain_template_url
     @expected_body = "my template"
+    @args = { :token => "test-token" }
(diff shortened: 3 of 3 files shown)


domcleal:
[test]

I should note that Foreman itself [doesn't support MAC lookups](http://projects.theforeman.org/issues/8239), that's added by foreman_bootdisk and was only really intended for that...

modules/templates/handler.rb
```diff
@@ -7,23 +7,25 @@ module Proxy::Templates
...
     extend Proxy::Log
 
-    def get_template(kind, token, static = false)
-      opts = { :token => token,
-               :url   => Proxy::Templates::Plugin.settings.template_url
+    def get_template(kind, params, static = false)
+      opts = {
+        :url => Proxy::Templates::Plugin.settings.template_url
       }
+      opts[:mac]    = params[:mac] if params.has_key?('mac')
```
dmitri-d, line 14: You are expecting string keys, but symbols are being sent instead (also the reason for test failures. Also, you can use `rake test` to execute tests locally).

modules/templates/templates_api.rb
```diff
@@ -14,8 +14,8 @@ class Proxy::TemplatesApi < Sinatra::Base
   end
 
   get "/:kind" do |kind|
-    log_halt(500, "Failed to retrieve #{kind} template for #{params[:token]}: ") do
-      Proxy::Templates::Handler.get_template(kind, params[:token], params[:static])
+    log_halt(500, "Failed to retrieve #{kind} template for #{params.to_json}: ") do
+      Proxy::Templates::Handler.get_template(kind, params, params[:static])
```
dmitri-d, line 18: You are passing params already, why extract `static` parameter from the hash?

test/templates/template_test.rb
```diff
@@ -17,15 +17,25 @@ def setup
 
   def test_template_requests_return_data_and_contain_template_url
     @expected_body = "my template"
+    @args = { :token => "test-token" }
```
dmitri-d, line 20: No need for an instance variable here (in tests below too).

test/templates/template_test.rb
```diff
@@ -17,15 +17,25 @@ def setup
...
-    result = Proxy::Templates::Handler.get_template('provision', 'test-token', 'true')
+    result = Proxy::Templates::Handler.get_template('provision', @args, 'true')
+    assert_equal(@expected_body, result)
+  end
+
+  def test_template_requests_with_macaddress
+    @expected_body = "my template"
+    @args = { :mac => "aa:bb:cc:dd:ee:ff" }
+    stub_request(:get, @foreman_url+'/unattended/provision?static=true&mac=aa:bb:cc:dd:ee:ff&url='+@template_url).to_return(:status => [200, 'OK'], :body => @expected_body)
+    result = Proxy::Templates::Handler.get_template('provision', @args)
```
dmitri-d, line 38: expected url has `static=true`, but when you are making the call, `static` parameter is set to default value of `false`

modules/templates/handler.rb
```diff
@@ -7,23 +7,25 @@ module Proxy::Templates
...
+      opts[:token]  = params[:token] if params.has_key?('token')
       opts[:static] = static if static
```
dmitri-d, line 16: see my comment re: passing `static` together with the rest of hash.

test/templates/template_test.rb
```diff
@@ -17,15 +17,25 @@ def setup
...
     stub_request(:get, @foreman_url+'/unattended/provision?token=test-token&url='+@template_url).to_return(:status => [200, 'OK'], :body => @expected_body)
-    result = Proxy::Templates::Handler.get_template('provision', 'test-token')
+    result = Proxy::Templates::Handler.get_template('provision', @args)
     assert_equal(@expected_body, result)
   end
 
   def test_template_requests_pass_static_flag_to_foreman
     @expected_body = "my template"
+    @args = { :token => "test-token" }
     stub_request(:get, @foreman_url+'/unattended/provision?static=true&token=test-token&url='+@template_url).to_return(:status => [200, 'OK'], :body => @expected_body)
```
logicminds, line 29: Just a note, your not required to use stub_request if you use the following mixin `include Rack::Test::Methods`

With this mixin you can just use get, post, put, ... like the following example

``` ruby
test_args = { :bmc_provider => 'freeipmi' }
get "/#{host}/lan/gateway", test_args
assert_equal 'unauthorized', last_response.body
assert_equal 401, last_response.status
```

Where last_response is automatically generated by the mixin.

Also note, if you want to pass a body you would do:

``` ruby
get "/#{host}/lan/gateway", test_args.to_json`, "CONTENT_TYPE" => "application/json" 
```

When you need to verify body contents

``` ruby
data = JSON.parse(last_response.body)
assert_match(/true|false/, data["result"].to_s)
```

whefter:
Thanks everyone, guess the PR betrays once more that I'm new to Ruby.
- The symbols vs. strings issue: As I was writing this, I was having it give me debug output and the passed parameters showed up as strings - not symbols - e.g. {'mac' => aa:bb...}. Tests didn't complain, either.
- The instance variable issue: 'expected_body' was an instance variable as well, so I suspected a style thing

I wrote this under some time pressure and, as stated, the tests passed, for some reason... I'll look into this with more time tomorrow.

lzap:
Hello, any update on this one @whefter ?

lzap:
I am closing this in favor of https://github.com/theforeman/smart-proxy/pull/287. Please carry on in the review. Solved all the comments except the `stub_request` one. Thanks.

PR #283, opened 2015-05-05
Fixes #10324 - Slow serial lookup in Puppet CA module

jcmcken:
I'm not able to test this in our production environment, but a quick `Benchmark.measure` with dummy data seems to suggest what you would assume, that `Set#include?` look-up is much faster than `Array#include?`.


dmitri-d:
[test]

dmitri-d:
Merged in 0a83088. Thanks, @jcmcken!

PR #285, opened 2015-05-21
Fixes #10378: smart-proxy service starts on windows now

dmitri-d:


dmitri-d:
@claer: could you give this a try?

Reedler:
Service won't start now getting Error 1053 Service did not respond to the start or control request in a timely fashion

dmitri-d:
@Reedler: Do you have `smart-proxy\log` directory? If so, what's in `windows-service.log` file? Otherwise, could you create it and try again?

dmitri-d:
@claer, @Reedler: ping?

bin/smart-proxy-win-service
```diff
@@ -1,40 +1,35 @@
...
     end
   end
 
-  daemon = Daemon.new
-  daemon.mainloop
-rescue Exception => err
-  File.open(logfile, (File::APPEND|File::CREAT|File::WRONLY)){ |f| f.puts " ***Daemon failure #{Time.now} err=#{err}" }
+  SmartDaemon.mainloop
+rescue Exception => e
+  logger.error("Daemon failure: #{e}  #{e.backtrace}")
```
domcleal: Nitpick, usually `.join("\n")` a backtrace as it's an array.

extra/register-service.rb
```diff
@@ -23,11 +21,6 @@
   puts 'The acount can be local or a domain account. If it is a domain account then use the domain\account syntax'
   user  = ask("Run this service as this user? ") {|u| u.default = default_user}
   pass1 = ask("Enter the user's password ") {|p| p.echo = "x"}
-  pass2 = ask("Reenter the password "){|p| p.echo = "x"}
```
domcleal, line 26: Why remove this check?

extra/register-service.rb
```diff
@@ -1,14 +1,12 @@
-#!/usr/bin/env ruby
-
 if RUBY_PLATFORM !~ /mingw/
   puts "To install this service on Unix please create a startup script"
 else
   require 'rubygems'
   require 'highline/import'
   require 'win32/service'
-  require 'rbconfig'
```
domcleal, line 9: needed for 1.8 compatibility?

bundler.d/windows.rb
```diff
@@ -1,5 +1,4 @@
 group :windows do
   gem 'highline', :platform => :mingw
   gem 'win32-service', :platforms => :mingw
-  gem "win32-open3", :platforms => :mingw_18
```
domcleal: Also needed in some providers under 1.8?

extra/register-service.rb
dmitri-d, line 26: Not sure it's worth it?  It's quite annoying needing to enter the password twice here as it's pretty easy to change service password if you get it wrong. It's also an easy issue to spot...

dmitri-d:
@domcleal: I fixed the issues you pointed out, but will need to test them on windows on 1.8.7

domcleal:
Seems there are a couple more issues with 1.8.7, so perhaps we should ignore it for the time being.  highline needs pinning to < 1.7.0 and win32-service needs some pinning as it uses require_relative in its current release, I didn't try getting past that errors.

On 1.9.3 and 2.1 x86 on 2012R2 x64 though, I'm having issues registering the service and wondered if you'd experienced it.  A segfault occurs on the Service.stop line, and if I comment out the stop/delete lines, it still segfaults when creating the new service: http://fpaste.org/237972/66412014/

domcleal:
I wonder if you saw the same segfault issue at any point @witlessbird, or if I should give it another go?

dmitri-d:
I haven't. I'll give it another go though, now that I have a windows VM.

dmitri-d:
@domcleal: rebased. Also tested on windows (2012r2) on ruby 1.9.3 and 2.0.0. Fixed an issue with bundler not detecting `Gemfile` and crashing when running as a service (had to launch ruby with `-C`). Works now!

domcleal:
Were you able to register the service?  My paste has been GCed now, but the service registration script was simply segfaulting.

dmitri-d:
Nods, I was, on both platforms.

dmitri-d:
Forgot to mention -- I was using ruby-installer, other windows-based ruby platforms most likely won't work.

helge000:
@witlessbird This does not work for me.
**Note:** running smart proxy from cmd works.
1. Register Service works - cool feature: User registrations and netsh permissions!
2. User name / password - works

Windows cannot start the service: 

> Error 1053: The service did not respond to the start or control request in a timely fashion

I have this issue when I create the service using sc.exe; tried this with `ruby` and `rubyw`

```
sc create smart_proxy binPath= "C:\ruby\bin\ruby.exe c:\smart-proxy\bin\smart-proxy "
```

BTW; I use a very basic config; features are all disabled:

```
:bind_host: '::'
:http_port: 8000
:log_file: STDOUT
```

extra/register-service.rb
```diff
@@ -1,52 +1,49 @@
...
+default_user = ENV["USERNAME"]
+default_user = ENV["USERDOMAIN"] + '\\' + default_user if ENV["USERDOMAIN"]
+
+puts "This service must run as a user with permission to execute the netsh dhcp script"
+puts 'The acount can be local or a domain account. If it is a domain account then use the domain\account syntax'
+user  = ask("Run this service as this user? ") {|u| u.default = default_user}
+pass1 = ask("Enter the user's password ") {|p| p.echo = "x"}
+
+begin
+  Service.stop("smart proxy")   rescue nil
```
helge000, line 30: We should safe ourselves some hassle and use `smart_proxy` or `smart-proxy` here

dmitri-d:
@helge000: What's the command used to start the service? Can you manually launch smart-proxy via command used by the service?

helge000:
> Can you manually launch smart-proxy via command used by the service?

@witlessbird Indeed, forgot to mention that. Command line runs fine in CMD. Two other things:
- Running the proxy as `NT AUTORITY\SYSTEM`
- [NSSM](https://nssm.cc/) service, using the same command, works as expected.

dmitri-d:
@helge000: I'll see if I can replicate the issue today or monday...

dmitri-d:
@helge000: what are you using for the command line when configuring the service via nssm? Can you connect to the service after it starts?

helge000:
@witlessbird (Currently I have issues with VM- clipboard sync from, so no real output from `nssm get`):
- NSSM uses `ruby bin/smart-proxy`; I see it running in tasklist, netstat is showing an open port (I just assume all is fine at this point)
- `bin/smart-proxy-win-service` does **not** work for me: the command runs, silently exits and does not fork in any way; 
- The `Error 1053` seems unrelated to this issue since NSSM is able to run both commands; while the service created by win32-service fails _while_ executing the command

bin/smart-proxy-win-service
```diff
@@ -1,40 +1,38 @@
...
 include Win32
 
-begin
-  logfile = "../../tmp/proxy.log"
-  $stdout.reopen(logfile, "a")
-  $stdout.sync = true
-  $stderr.reopen($stdout)
-  puts "#{Time.now}: Service is starting"
+def logger
+  @logger ||= Logger.new(File.expand_path("win-service.log", File.expand_path("../../log", __FILE__)))
```
helge000: @witlessbird The issue is here: For some reason beyond me `@logger` is not initialized and `log/` does not exist. Changing this for instance to `File.expand_path("win-service.log", File.expand_path("../", __FILE__)))` works.

dmitri-d: Nods, 'log' directory needs to exist. I suppose I could create it if it doesn't exist though.

bin/smart-proxy-win-service
```diff
@@ -1,40 +1,38 @@
...
+      # the daemon is about to exit.
+      logger.info("Service is terminating")
     end
 
     def service_stop
-      puts "#{Time.now}: Service stopped"
-      exit!(true)
-    end
+      logger.info("exit")
+      exit!
```
helge000: This does not work well: `1067: the process terminated unexpectedly`

bin/smart-proxy-win-service
helge000: I think the reason for a fail safe is that it is fail save :). IMHO it shout be changed in way to go directly to the smart-proxy root folder.

Also, can explain think why `@logger` is not there? Is this not read from settings.yml? This way, we could use the same log file; witch would be the best solution.

domcleal:
Back to WoC to address @helge000's comments.

dmitri-d:
@helge000, @domcleal: added creation of the log dir, fixed exception occurring on shutdown.

domcleal:
I'm getting a lot further than I did last time I tried this as the service now registers correctly, but I can't get smart-proxy-win-service to start successfully.

When running `bundle exec ruby bin/smart-proxy-win-service`, there are some harmless bundler messages and it exits a few seconds later.  win-service.log shows this:

<pre>
I, [2015-11-23T15:11:12.202993 #3832]  INFO -- : Service is initializing
E, [2015-11-23T15:11:14.218741 #3832] ERROR -- : Daemon failure: Service_Main thread exited abnormally  C:/Ruby21/lib/ruby/gems/2.1.0/gems/win32-service-0.8.7/lib/win32/daemon.rb:282:in `mainloop'
C:/Ruby21/lib/ruby/gems/2.1.0/gems/win32-service-0.8.7/lib/win32/daemon.rb:217:in `mainloop'
c:/smart-proxy/bin/smart-proxy-win-service:36:in `&lt;main>'
</pre>

The main proxy.log file shows only a "log created" message, even at debug level.  I don't see the service_main log message in either file.

dmitri-d:
@domcleal: I found a couple of issues that were preventing service startup on windows: a missing 'logger' requirement in bin/smart-proxy-win-service, and a signal handler for unsupported on windows signal (TTIN). Service starts up ok after that. Could you give this another try?

domcleal:
Merged as 8e19663, thanks @witlessbird.

PR #287, opened 2015-05-28
Fixes #10259 - Templates are now correctly proxied

lzap:
I have amended my changes from the previous review: https://github.com/theforeman/smart-proxy/pull/279

### DIFF
--- a/modules/templates/handler.rb
+++ b/modules/templates/handler.rb
@@ -7,23 +7,27 @@ module Proxy::Templates
   class Handler < ::Proxy::HttpRequest::ForemanRequest
     include Proxy::Log
 
-    def get_template(kind, token, static = false)
-      opts = { :token => token,
-               :url   => Proxy::Templates::Plugin.settings.template_url
+    def get_template(kind, params)
+      opts = {
+        :url => Proxy::Templates::Plugin.settings.template_url
       }
-      opts[:static] = static if static
+      # mac parameter will be only searched when foreman_bootdisk plugin is enabled on the foreman side
+      opts[:mac]    = params[:mac] if params.has_key?(:mac)
+      opts[:token]  = params[:token] if params.has_key?(:token)
+      opts[:static] = params[:static] if params.has_key?(:static)
       request = request_factory.create_get("/unattended/#{kind}", opts)
       logger.debug "retrieving a template from %s/%s" % [Proxy::Templates::Plugin.settings.template_url, request.path]
       res = send_request(request)
 
       # You get a 201 from the 'built' URL
-      raise "Error retrieving #{kind} for #{token} from #{uri.host}: #{res.class}" unless ["200", "201"].include?(res.code)
+      raise "Error retrieving #{kind} for #{opts.inspect} from #{uri.host}: #{res.class}" unless ["200", "201"].include?(res.code)
+      Proxy::Log.logger.info "Template: request for #{kind} using #{opts.inspect} at #{uri.host}"
       res.body
     end
 
-    def self.get_template kind, token, static = false
+    def self.get_template(kind, params)
       @handler ||= Handler.new
-      @handler.get_template(kind,token, static)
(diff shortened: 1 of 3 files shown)


domcleal:
Foreman's API doesn't support MAC address lookup, that's only foreman_bootdisk, so I'm not sure this should be here.  The PR also cannot be merged.

lzap:
Since this is a public API, I don't see particular reason why not to have it here. Amended.

lzap:
I mean I don't see any issues with having it like this. Could create extra API tho.

domcleal:
I'm not sure I understand your point?  MAC isn't an option supported by Foreman, so we shouldn't carry support in the smart proxy for it..

lzap:
Oh now I understand. Foreman without bootdisk will not return the correct data to proxy, won't it?

domcleal:
Right, Foreman without bootdisk doesn't look up a template with this parameter.

lzap:
Ok, how about this: I add a parameter and this lookup will be turned off by default. Bootdisk users will be asked to turn this feature on here.

domcleal:
I don't follow, an extra request parameter?

dmitri-d:
I would very much like avoid a configurable API parameter. looking at the foreman side, the `mac` would be ignored in the absence of bootdisk. Could we put a comment next to https://github.com/theforeman/smart-proxy/pull/287/files#diff-40b456408d24baac53fa8e0cf3315631R14 saying that this is only available in the presence of bootdisk?

lzap:
Yeah sorry I mean configuration option (settings.d/templates.yaml).

lzap:
Added the comment...

lzap:
I really don't know how to approach this to be honest other than explicitly disabling the feature via YAML settings. I can do some testing what happens when bootdisk is not installed, I'd expect just 404 which is not bad at all. We can maybe catch the exception if there is any and issue some proper error message.

lzap:
For the record, we need this to enable the indirect communication patch for bootdisk: https://github.com/theforeman/foreman_bootdisk/pull/9

dmitri-d:
@lzap, yeah, worth a test. I don't think anything will break and the 'mac' parameter will be ignored. I think.

lzap:
Ok Foreman without Bootdisk returns 404:

```
2015-06-30 17:01:45 [app] [I] Started GET "/unattended/iPXE?url=http%3A%2F%2F192.168.122.1%3A8448" for 127.0.0.1 at 2015-06-30 17:01:45 +0200
2015-06-30 17:01:45 [app] [I] Processing by UnattendedController#iPXE as TEXT
2015-06-30 17:01:45 [app] [I]   Parameters: {"url"=>"http://192.168.122.1:8448", "unattended"=>{}}
2015-06-30 17:01:45 [app] [I] unattended: unable to find a host that matches the request from 127.0.0.1
2015-06-30 17:01:45 [app] [I] Filter chain halted as :get_host_details rendered or redirected
2015-06-30 17:01:45 [app] [I] Completed 404 Not Found in 109.2ms (ActiveRecord: 1.2ms)
```

Which is properly handled in proxy:

```
lzapx.brq.redhat.com - - [30/Jun/2015:17:01:45 CEST] "GET /unattended/iPXE?mac=52:54:00:da:75:61 HTTP/1.1" 500 213
- -> /unattended/iPXE?mac=52:54:00:da:75:61
```

dmitri-d:
Alternatively you could create a dedicated smart-proxy module for this. On the foreman side you can either use a different pxe linux template (pointing to the new api end point) or override `default_template_url` helper.

lzap:
That looks like total overrkill, how about this: I move the ability to search for a host by mac into Foreman Core? @domcleal

lzap:
Ok I am closing this patch as it looks like generic proxying will be better. It will also help to solve problem when Anaconda http headers are currently not being handed over (I think).

PR #290, opened 2015-06-15
refs #8046: Add possibility for dhcp providers to provide own method to load single records

Klaas-:
I recently started working on a infoblox dhcp provider and it takes ages to load single dhcp records without a possiblity to not load whole subnets on requesting a single record :)

### DIFF
--- a/modules/dhcp/dhcp_api.rb
+++ b/modules/dhcp/dhcp_api.rb
@@ -81,7 +81,7 @@ class Proxy::DhcpApi < ::Sinatra::Base
   get "/:network/:record" do
     begin
       content_type :json
-      record = load_subnet[params[:record]]
+      record = @server.find_record(params[:record])
       log_halt 404, "Record #{params[:network]}/#{params[:record]} not found" unless record
       record.options.to_json
     rescue => e


dmitri-d:
[test]

modules/dhcp/server.rb
```diff
@@ -75,6 +75,9 @@ def find_subnet value
     end
 
     def find_record record
+      if self.respond_to? :loadRecord
```
dmitri-d: `loadRecord` method doesn't exist. Did you forget to include all of the changes?

Klaas-:
hey, no its just that new dhcp providers (like the one I'm working on - https://github.com/Klaas-/smart-proxy/tree/infoblox_support) can use loadRecord to allow a single record to be loaded without going through the entire subnet - it may be a good idea to implement it for msdhcp too I've read that they have simmilar problems when loading of a bigger subnet takes ages :)

dmitri-d:
@Klaas-: In that case overload `find_record` in your provider(s). We'll revisit this If we get to the point when most/all providers define `loadRecord`.

Klaas-:
find_record isn't used in dhcp_api.rb even though I didn't add the find_record function. I couldn't figure out a way on how to overwrite the dhcp_api.rb function from a dhcp server provider

modules/dhcp/dhcp_api.rb
```diff
@@ -81,7 +81,7 @@ def load_subnet
   get "/:network/:record" do
     begin
       content_type :json
-      record = load_subnet[params[:record]]
+      record = @server.find_record(params[:record])
```
dmitri-d, line 84: This should work. I'll do a some testing to confirm and then merge.

dmitri-d:
[test]

dmitri-d:
merged in at fa3178a. Thanks, @Klaas-!

PR #291, opened 2015-06-15
Fixes #10527 - DHCP/TFTP support for Cisco POAP

fernandocarolo:
This adds support for Cisco POAP at the DHCP and TFTP proxies, working together with pull request #2464 for Foreman.

### DIFF
--- a/modules/dhcp/providers/server/isc.rb
+++ b/modules/dhcp/providers/server/isc.rb
@@ -46,6 +46,7 @@ module Proxy::DHCP
 
       statements += solaris_options_statements(options)
       statements += ztp_options_statements(options)
+      statements += poap_options_statements(options)
 
       omcmd "set statements = \"#{statements.join(" ")}\""      unless statements.empty?
       omcmd "create"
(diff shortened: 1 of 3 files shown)


dmitri-d:
[test]

dmitri-d:
merged in at 17d3726. Thanks, @fernandocarolo!

DebtFromAbove:
@fernandocarolo I am interested in implementing this exact same functionality for Arista switches that I use. They have a feature similar to Junos called ZTP. However, studying both your patch and the Junos ZTP patch code, I believe the actual implementation is much closer, if not identical to your patch for POAP. The arista switch requests an address via DHCP. The server replies with an address and provides the boot-filename option which is a TFTP link to a script which contains instructions on how to wget the 'startup-config' and optionally pull down and install new firmware and extensions.

The script pulled from TFTP is not python but a list of CLI commands, as detailed on: 
https://eos.arista.com/ztp-set-up-guide/
The page explains quite well and briefly how it pulls down its config using DHCP and TFTP.

Here is an example of that script from the website above:

```
#!/usr/bin/Cli -p2
enable
copy http://192.168.111.2/ztp-startup-config flash:startup-config
copy http://192.168.111.2/EOS.swi flash:EOS.swi
config
boot system flash:EOS.swi
```

My question is do you think your patch (unmodified) would be suitable to implement configuration via Arista ZTP?

I believe I could use the "provision.erb" template from the community templates associated with the POAP patch to create this initial CLI script file. It may even be possible to use the next stage "finish" template as a skeleton for dynamically creating the 'startup-config' file for the switch. Do you think this assertion is correct?

many thanks

PR #292, opened 2015-06-17
fixes #10768: failures when connecting to foreman are now being logged

dmitri-d:


lib/proxy/request.rb
```diff
@@ -39,8 +39,13 @@ def create_post(path, body)
   end
 
   class ForemanRequest
+    include Proxy::Log
+
     def send_request(request)
       http.request(request)
+    rescue Exception => e
+      logger.error("Error when accessing %s: %s" % [uri.to_s, e])
```
domcleal: Should this print more of the full request URI out, not just the base URI?

domcleal: I'd also suggest logging the exception backtrace, unless that's going to happen further up the stack?

dmitri-d:
@domcleal: logging the complete request url now. I don't think it makes sense to log the backtrace there, this is to log the type of an error (timeout, connection refused, etc) and the url to caused it. I think dealing with the issue itself (which may or may not include logging of the backtrace) should be done closer to where the request was made...

domcleal:
Okay, and that makes sense, but I can't work out precisely what the original bug report is complaining about, because the templates module already uses log_halt and appears to log resolution errors just fine:

<pre>
D, [2015-06-29T12:16:34.856407 #10827] DEBUG -- : retrieving a template from http://192.168.122.1:8443//unattended/provision?token=a&url=http%3A%2F%2F192.168.122.1%3A8443
E, [2015-06-29T12:16:34.861500 #10827] ERROR -- : Failed to retrieve provision template for a: getaddrinfo: Name or service not known
D, [2015-06-29T12:16:34.861790 #10827] DEBUG -- : /home/dcleal/.rvm/rubies/ruby-2.1.0/lib/ruby/2.1.0/net/http.rb:879:in `initialize'
/home/dcleal/.rvm/rubies/ruby-2.1.0/lib/ruby/2.1.0/net/http.rb:879:in `open'
/home/dcleal/.rvm/rubies/ruby-2.1.0/lib/ruby/2.1.0/net/http.rb:879:in `block in connect'
/home/dcleal/.rvm/rubies/ruby-2.1.0/lib/ruby/2.1.0/timeout.rb:67:in `timeout'
/home/dcleal/.rvm/rubies/ruby-2.1.0/lib/ruby/2.1.0/net/http.rb:878:in `connect'
/home/dcleal/.rvm/rubies/ruby-2.1.0/lib/ruby/2.1.0/net/http.rb:863:in `do_start'
/home/dcleal/.rvm/rubies/ruby-2.1.0/lib/ruby/2.1.0/net/http.rb:852:in `start'
/home/dcleal/.rvm/rubies/ruby-2.1.0/lib/ruby/2.1.0/net/http.rb:1369:in `request'
/home/dcleal/code/foreman/smart-proxy/lib/proxy/request.rb:43:in `send_request'
/home/dcleal/code/foreman/smart-proxy/modules/templates/handler.rb:17:in `get_template'
/home/dcleal/code/foreman/smart-proxy/modules/templates/handler.rb:26:in `get_template'
/home/dcleal/code/foreman/smart-proxy/modules/templates/templates_api.rb:18:in `block (2 levels) in <class:TemplatesApi>'
/home/dcleal/code/foreman/smart-proxy/lib/proxy/helpers.rb:14:in `log_halt'
/home/dcleal/code/foreman/smart-proxy/modules/templates/templates_api.rb:17:in `block in <class:TemplatesApi>'
</pre>

Does this add anything, given that 25c06dc also started logging the URL (seen in the first message)?

dmitri-d:
@domcleal: I think there is an issue here (not necessary as reported), as which request details will be logged is up to individual modules atm.  Also, perhaps it makes sense to reduce the verbosity of error logging in the templates module?

domcleal:
Not sure I follow, since the templates module has more context about the request, shouldn't it perform the logging?  I don't think its severity should be reduced.

We're only talking about one module here, and another in a plugin - which, if it doesn't log, can be fixed.

dmitri-d:
I think it does make some sense to log request error details in one place (if for no other reason than reduce the duplication/ensure important bits are logged). You are right though, we only have two modules that use ForemanRequest atm (templates and chef), and fixing both is an option...

domcleal:
And smart_proxy_discovery, I think.

I'll leave you to decide what to do with this PR and ticket?

dmitri-d:
I'd say let's merge this PR and leave modules as is.

domcleal:
I don't think it should be merged when it creates double-logging though, which it does with templates.

dmitri-d:
Lemme fix that.

dmitri-d:
I figured I'll close this PR and the ticket, as https://github.com/theforeman/smart-proxy/commit/25c06dcb097c10c2d0879eccd370182ebf283101 is sufficient.

PR #293, opened 2015-06-17
fixes #10769 - added a debug trace when downloading templates from foreman

dmitri-d:


modules/templates/handler.rb
```diff
@@ -13,11 +13,12 @@ def get_template(kind, token, static = false)
       }
       opts[:static] = static if static
       request = request_factory.create_get("/unattended/#{kind}", opts)
+      logger.debug "retrieving a template from %s/%s" % [Proxy::Templates::Plugin.settings.template_url, request.path]
       res = send_request(request)
 
       # You get a 201 from the 'built' URL
       raise "Error retrieving #{kind} for #{token} from #{uri.host}: #{res.class}" unless ["200", "201"].include?(res.code)
-      Proxy::Log.logger.info "Template: request for #{kind} using #{token} at #{uri.host}"
+      logger.info "Template: request for #{kind} using #{token} at #{uri.host}"
```
lzap: I don't like flushing security tokens via info level. Please do debug here as well.

Other than that :+1: it works fine.

dmitri-d: Removed that line altogether, as it was duplicating log at line 16.

lzap:
Merged as 25c06dcb097c10c2d0879eccd370182ebf283101, thank you!

PR #277, opened 2015-04-12
Fixes #7766 - performance enhancements for DHCP on windows

briangann:
These changes enhances the DHCP smartproxy for windows (native_ms.rb) by providing sub-second responses for most queries. (This is as fast as you can get netsh.exe to respond)

Key changes:
- Constructs a script file to be run by netsh, and parses the output.
- Invokes netsh.exe with correct parameters
- Bugfix in the report function - would crash when no option exists

Benefit:
Large DHCP scopes take a significant amount of time to enumerate with the previous method of running netsh for each query. The smartproxy times out or takes up to 6 minutes on a /22 network, even on very fast dedicated hardware.

Performance:
Usage of this modification has shown it to perform at sub-second response even with large scopes.
A /22 scope previously would take over 6 minutes to enumerate and the browser would time out or appear to be "stuck."

PR #294, opened 2015-06-18
Fixes #7766 - performance enhancements for DHCP on windows

dmitri-d:


dmitri-d:
This is in most part based on https://github.com/theforeman/smart-proxy/pull/277. I refactored the code to be more testable, added a couple of test and fixed broken tests.

dmitri-d:
@Reedler, @briangann: any chance you could give this a try?

modules/dhcp/providers/server/native_ms.rb
```diff
@@ -196,6 +230,61 @@ def execute cmd, msg=nil, error_only=false
...
+    end
+
+    # convert scriptfile output into a hash for fast queries
+    def parse_script_output_to_hash content
+      processed = {}
+      queryResults = ""
+      count = 0
+      found = false
+      anIP = nil
+      content.each do |line|
```
domcleal, line 249: I guess this is the line Oliver's referring to here: http://projects.theforeman.org/issues/7766#change-44781

briangann, line 249: I will give this a try, I am running 1.8.7 Ruby so I didn't hit that .each bug/deprecation.Code looks good to me, kudos to the rework for testability 😀BrianAt Jun 24, 2015, 7:30:49 AM, Dominic Cleal wrote:In modules/dhcp/providers/server/native_ms.rb:

> -        std_in.close  unless std_in.nil?
> -        std_out.close unless std_in.nil?
> -        std_err.close unless std_in.nil?
> -      end
> -      report msg, response, error_only
> -      response
> -    end
>   +
> -    # convert scriptfile output into a hash for fast queries
> -    def parse_script_output_to_hash content
> -      processed = {}
> -      queryResults = ""
> -      count = 0
> -      found = false
> -      anIP = nil
> -      content.each do |line|

I guess this is the line Oliver's referring to here: http://projects.theforeman.org/issues/7766#change-44781

dmitri-d:
fixed issues when running on Ruby > 1.8.7 (there were a few instances of String#each calls).

modules/dhcp/providers/server/native_ms.rb
```diff
@@ -226,23 +315,23 @@ def report msg, response, error_only
...
-          opts = SUNW.update(Standard)
-          title = opts.select {|k,v| v[:code] == optionId}.flatten[0]
-          logger.debug "found option #{title}"
-          options[title] = $1
+          #TODO: this logic is broken, as the output reports only once the vendor type
+          # making it impossible to detect if its a standard option or a custom one.
+          when /For vendor class \[([^\]]+)\]:/
+            options[:vendor] = "<#{$1}>"
+          when /OptionId : (\d+)/
+            #logger.debug "optionid parsed is " + $1
```
domcleal: leftover debug

modules/dhcp/providers/server/native_ms.rb
```diff
@@ -196,6 +230,61 @@ def execute cmd, msg=nil, error_only=false
       response
     end
 
+    # Invokes netsh.exe with a script file
+    def execute_with_script scriptfile, msg=nil, error_only=false
```
domcleal, line 214: This method's nearly identical to `execute` just above, the only difference seems to be the command and timeout - could they share the remainder of the code?

modules/dhcp/providers/server/native_ms.rb
```diff
@@ -85,39 +86,72 @@ def find_or_create_vendor_name vendor, exception
...
     end
 
+    def loadSubnetData subnet
+      super
+      all_entries = subnet_entries(subnet)
+      netsh_tmp = build_netsh_script(all_entries)
+      script_output = execute_with_script(netsh_tmp.path,"Running NETSH in script mode")
+      subnet_entries_to_records(all_entries, parse_script_output_to_hash(script_output))
+    ensure
+      netsh_tmp.unlink
```
domcleal: Check first if netsh_tmp is !nil.

modules/dhcp/providers/server/native_ms.rb
```diff
@@ -196,6 +230,61 @@ def execute cmd, msg=nil, error_only=false
...
+      count = 0
+      found = false
+      anIP = nil
+      content.each do |line|
+        if line.match(/^Options for the Reservation Address ([\w\.]+)/)
+          anIP = $1
+          found = true
+          count+=1
+        end
+        if line.match(/^Command completed/) && found
```
domcleal: You can get rid of `found` and just check for `!anIP.nil?`

modules/dhcp/providers/server/native_ms.rb
```diff
@@ -196,6 +230,61 @@ def execute cmd, msg=nil, error_only=false
...
+      processed = {}
+      queryResults = ""
+      count = 0
+      found = false
+      anIP = nil
+      content.each do |line|
+        if line.match(/^Options for the Reservation Address ([\w\.]+)/)
+          anIP = $1
+          found = true
+          count+=1
```
domcleal: Perhaps replace `count` with `processed.size` in the debug.

modules/dhcp/providers/server/native_ms.rb
```diff
@@ -85,39 +86,72 @@ def find_or_create_vendor_name vendor, exception
...
+    def build_netsh_script(subnet_entries)
+      netsh_tmp = Tempfile.new('smartproxydhcp')
+      netsh_tmp.write("dhcp server\n")
+      subnet_entries.each_value do |entry|
+        netsh_tmp.write("scope #{entry[:subnet].network} Show ReservedOptionValue #{entry[:ip]}\n")
+      end
+
+      netsh_tmp
+    ensure
+      netsh_tmp.close
```
domcleal: Also check `!netsh_tmp.nil?` in case it dies on line 1.

dmitri-d:
@domcleal: did a bunch of small refactorings you suggested. @briangann: could you give these changes a try please?

modules/dhcp/providers/server/native_ms.rb
```diff
@@ -172,21 +205,33 @@ def loadSubnets
...
-          std_in, std_out, std_err  = Open3.popen3(command)
+        timeout(timeout) do
+          logger.debug "executing: #{full_command}"
+          std_in, std_out, std_err  = Open3.popen3(full_command)
           response  = std_out.readlines
           response += std_err.readlines
         end
       rescue TimeoutError
-        raise Proxy::DHCP::Error.new("Netsh did not respond within #{tsecs} seconds")
+        raise puts("Netsh did not respond within #{timeout} seconds")
```
domcleal: `raise puts`?  Shouldn't it raise an exception?  (Plus puts returns nil, IIRC)

dmitri-d:
@domcleal: fixed.

domcleal:
Thanks, looks ready if somebody such as @briangann would be kind enough to give it another test.

dLobatog:
@witlessbird @domcleal I have a spare Windows server 2012 box I could use, but I'm not sure how to test this, if you can provide with a script that creates reservations I could help?

logicminds:
I wrote this code a long time ago when I had ms dhcp skills.  Feel free to use it or if you want I can change license or whatever.  Essentially it parses the entire dhcp dump into a ruby object.

https://github.com/logicminds/msdhcpdump/blob/master/test.rb  (this is the test code with test dump file)
@witlessbird @domcleal 

Note: this was written with the little ruby skills I had at the time so don't poke too much fun at it.

dmitri-d:
@eLobato: I don't have a script. I would set up dhcp service on that machine and tried provisioning a a few vms, or something like that...

briangann:
the new code works - it retains the bulk of the original PR.  performance looks great.

PR #295, opened 2015-06-19
Fixes #8046: Add basic infoblox dhcp support

Klaas-:
This should add basic dhcp support for infoblox appliances over the rest api using the infoblox gem >= 0.4.1 (https://github.com/govdelivery/infoblox)

### DIFF
--- a/modules/dhcp/providers/server/infoblox.rb
+++ b/modules/dhcp/providers/server/infoblox.rb
@@ -1,0 +1,148 @@
+require 'dhcp/subnet'
+require 'dhcp/record'
+require 'dhcp/record/reservation'
+require 'dhcp/server'
+require 'infoblox'
+
+module Proxy::DHCP
+  # Represents Infoblox DHCP Server (https://www.infoblox.com) uses https://github.com/govdelivery/infoblox >= 0.4.1
+  class Infoblox < Server
+    def initialize(options = {})
+      super options[:server]
+      @username = options[:username]
+      @password = options[:password]
+    end
+
+   def connect
+     ::Infoblox::Connection.new(:username => Proxy::DhcpPlugin.settings.username, :password => Proxy::DhcpPlugin.settings.password, :host => Proxy::DhcpPlugin.settings.dhcp_server)
+   end
+
+   def delRecord subnet, record
+     validate_subnet subnet
+     validate_record record
+     # TODO: Refactor this into the base class
+     raise InvalidRecord, "#{record} is static - unable to delete" unless record.deleteable?
+     # "Deleting"" a record here means just disabling dhcp
+     connection = connect
+     host = ::Infoblox::Host.find(connection, "ipv4addr" => record.ip)
+     unless host.empty?
+       # if not empty, first element is what we want to edit
+       host = host.first
+       # Select correct ipv4addr object from ipv4addrs array
+       hostip = host.ipv4addrs.find { |ip| ip.ipv4addr == record.ip }
+       hostip.configure_for_dhcp = false
+       # Send object
+       host.put
+     end
+     subnet.delete(record)
+   end
+
+    def addRecord options={}
+      logger.debug "Add Record"
+      record = super(options)
+      
+      connection = connect
+      host = ::Infoblox::Host.find(connection, "ipv4addr" => record.ip)
+      # If empty create:
+      if host.empty?
+        logger.debug "Add Record - Create"
+        # Create new host object
+        host = ::Infoblox::Host.new(:connection => connection)
+        host.name = record.name
+        host.add_ipv4addr(record.ip)
+        post = true
+      else
+        logger.debug "Add Record - Exists using first element"
+        # if not empty, first element is what we want to edit
+        host = host.first
+        post = false
+      end
+      options = record.options
+      # Overwrite values without checking
+      # Select correct ipv4addr object from ipv4addrs array
+      hostip = host.ipv4addrs.find { |ip| ip.ipv4addr == record.ip }
+      logger.debug "Add Record - record.name: #{record.name}, hostip.host #{hostip.host}, record.mac #{record.mac}, record.ip #{record.ip}"
+      logger.debug "Add Record - options[:nextServer] #{options[:nextServer]}, options[:filename] #{options[:filename]}, hostip.ipv4addr: #{hostip.ipv4addr} "
+      raise InvalidRecord, "#{record} Hostname mismatch" unless hostip.host == record.name
+      hostip.mac = record.mac
+      hostip.configure_for_dhcp = true
+      hostip.nextserver = options[:nextServer]
+      hostip.use_nextserver = true
+      hostip.bootfile = options[:filename]
+      hostip.use_bootfile = true
+      ## Test if Host Entry has correct IP
+      raise InvalidRecord, "#{record} IP mismatch" unless hostip.ipv4addr == record.ip
+      # Send object
+      post ? host.post : host.put
+      record
+    end
+    
+    def find_record(record)
+      logger.debug "loadRecord"
+      connection = connect
+      # if record is a String it can be either ip or mac, true = mac --> lookup ip
+      if record.is_a?(String) && (IPAddr.new(record) rescue nil).nil?
+        hostdhcp = ::Infoblox::HostIpv4addr.find(connection, "mac" => record).first
+        ipv4address = hostdhcp.ipv4addr
+      elsif record.is_a?(String)
+        ipv4address = record
+      end
+      ipv4address = record[:ip] if record.is_a?(Proxy::DHCP::Record)
+      ipv4address = record.to_s if record.is_a?(IPAddr)
+      host = ::Infoblox::Host.find(connection, "ipv4addr" => ipv4address).first
+      return nil if host.nil? || host.name.empty?
+      hostdhcp = ::Infoblox::HostIpv4addr.find(connection, "ipv4addr" => ipv4address).first
+      return nil unless hostdhcp.configure_for_dhcp
+      return nil if hostdhcp.mac.empty? || hostdhcp.ipv4addr.empty?
+      opts = {:hostname => host.name}
+      opts[:mac] = hostdhcp.mac
+      opts[:ip] = hostdhcp.ipv4addr
+      opts[:deleteable] = true
+      opts[:nextServer] = hostdhcp.nextserver if hostdhcp.use_nextserver
+      opts[:filename] = hostdhcp.bootfile if hostdhcp.use_bootfile
+      # Subnet should only be one, not checking that yet
+      subnet = subnets.find { |s| s.include? ipv4address}
+      Proxy::DHCP::Record.new(opts.merge(:subnet => subnet))
+    end
+
+    def loadSubnetData subnet
+      # Load network from infoblox, iterate over ips to gather additional settings
+      logger.debug "LoadSubnetData"
+      super
+      network = IPAddr.new(subnet.to_s, Socket::AF_INET)
+      connection = connect
+      # max results are currently set to work in my setup, one could calculate that setting by looking at netmask :)
+      network = ::Infoblox::Ipv4address.find(connection, "network" => "#{network}/#{subnet.cidr}", "_max_results" => "2500")
+      # Find out which hosts are in use
+      network.each do |host|
+        # next if certain values are not set
+        next if host.names.empty? || host.mac_address.empty? || host.ip_address.empty?
+        hostdhcp = ::Infoblox::HostIpv4addr.find(connection, "ipv4addr" => host.ip_address).first
+        next unless hostdhcp.configure_for_dhcp
+        opts = {:hostname => host.names.first}
+        opts[:mac] = host.mac_address
+        opts[:ip] = host.ip_address
+        # broadcast and network entrys are not deleteable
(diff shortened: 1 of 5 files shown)


modules/dhcp/providers/server/infoblox.rb
```diff
@@ -0,0 +1,145 @@
...
+      @password = options[:password]
+    end
+
+   def delRecord subnet, record
+      validate_subnet subnet
+      validate_record record
+      # TODO: Refactor this into the base class
+      raise InvalidRecord, "#{record} is static - unable to delete" unless record.deleteable?
+      # "Deleting"" a record here means just disabling dhcp
+      connection = ::Infoblox::Connection.new(:username => Proxy::DhcpPlugin.settings.username, :password => Proxy::DhcpPlugin.settings.password, :host => Proxy::DhcpPlugin.settings.dhcp_server)
```
unclebilly: You might consider creating a `connection` method... this line is repeated several times in different methods.

Klaas-: okay :)

lzap:
Hello,

shell we test against this product?

https://www.infoblox.com/downloads/software/ddi-trial

Can you specify supported products and versions in the example conf
file?

## 

Later,
 Lukas #lzap Zapletal

dmitri-d:
@Klaas-: could you add tests for this please?

dmitri-d:
[test]

domcleal:
The test failures are for style checks, which you can see at http://ci.theforeman.org/job/test_proxy_develop_pr_rubocop/276/checkstyleResult/ or by running `rubocop` locally.

Klaas-:
Hi,
was busy on another project - I fixed the style checks, my local rubocop complains about more than jenkins - I think I fixed all style problems.
@lzap all infoblox products support their wapi layer -- so yeah I think you can try that :)
@witlessbird what kind of tests do I need to write? How would one test them if they don't have an infoblox application? Do I need to write a mock implementation of the restapi?

dmitri-d:
[test]

modules/dhcp/providers/server/infoblox.rb
```diff
@@ -0,0 +1,148 @@
...
+        host = ::Infoblox::Host.new(:connection => connection)
+        host.name = record.name
+        host.add_ipv4addr(record.ip)
+        post = true
+      else
+        logger.debug "Add Record - Exists using first element"
+        # if not empty, first element is what we want to edit
+        host = host.first
+        post = false
+      end
```
dmitri-d, line 59: I'd break this up into two methods one dealing with host creation, and another with modification.

modules/dhcp/providers/server/infoblox.rb
```diff
@@ -0,0 +1,148 @@
...
+      hostip = host.ipv4addrs.find { |ip| ip.ipv4addr == record.ip }
+      logger.debug "Add Record - record.name: #{record.name}, hostip.host #{hostip.host}, record.mac #{record.mac}, record.ip #{record.ip}"
+      logger.debug "Add Record - options[:nextServer] #{options[:nextServer]}, options[:filename] #{options[:filename]}, hostip.ipv4addr: #{hostip.ipv4addr} "
+      raise InvalidRecord, "#{record} Hostname mismatch" unless hostip.host == record.name
+      hostip.mac = record.mac
+      hostip.configure_for_dhcp = true
+      hostip.nextserver = options[:nextServer]
+      hostip.use_nextserver = true
+      hostip.bootfile = options[:filename]
+      hostip.use_bootfile = true
```
dmitri-d, line 72: This bit can be extracted into a method too, which will help with the refactoring above.

modules/dhcp/providers/server/infoblox.rb
```diff
@@ -0,0 +1,148 @@
...
+      ## Test if Host Entry has correct IP
+      raise InvalidRecord, "#{record} IP mismatch" unless hostip.ipv4addr == record.ip
+      # Send object
+      post ? host.post : host.put
```
dmitri-d, line 76: You won't need this conditional if you do the refactorings I suggested.

modules/dhcp/providers/server/infoblox.rb
```diff
@@ -0,0 +1,148 @@
...
+     unless host.empty?
+       # if not empty, first element is what we want to edit
+       host = host.first
+       # Select correct ipv4addr object from ipv4addrs array
+       hostip = host.ipv4addrs.find { |ip| ip.ipv4addr == record.ip }
+       hostip.configure_for_dhcp = false
+       # Send object
+       host.put
+     end
+     subnet.delete(record)
```
dmitri-d, line 37: You should call `super` here instead...

modules/dhcp/providers/server/infoblox.rb
```diff
@@ -0,0 +1,148 @@
...
+   end
+
+    def addRecord options={}
+      logger.debug "Add Record"
+      record = super(options)
```
dmitri-d, line 42: No need to pass arguments here, `record = super` is sufficient.

modules/dhcp/providers/server/infoblox.rb
```diff
@@ -0,0 +1,148 @@
...
+      # Send object
+      post ? host.post : host.put
+      record
+    end
+    
+    def find_record(record)
+      logger.debug "loadRecord"
+      connection = connect
+      # if record is a String it can be either ip or mac, true = mac --> lookup ip
+      if record.is_a?(String) && (IPAddr.new(record) rescue nil).nil?
```
dmitri-d, line 84: You can use https://github.com/theforeman/smart-proxy/blob/develop/lib/proxy/validations.rb#L9 to check if a string is a valid mac address.

modules/dhcp/providers/server/infoblox.rb
```diff
@@ -0,0 +1,148 @@
...
+        hostdhcp = ::Infoblox::HostIpv4addr.find(connection, "mac" => record).first
+        ipv4address = hostdhcp.ipv4addr
+      elsif record.is_a?(String)
+        ipv4address = record
```
dmitri-d, line 88: probably worth validating this.

modules/dhcp/providers/server/infoblox.rb
```diff
@@ -0,0 +1,148 @@
...
+      end
+      ipv4address = record[:ip] if record.is_a?(Proxy::DHCP::Record)
+      ipv4address = record.to_s if record.is_a?(IPAddr)
```
dmitri-d, line 91: I'm quite certain `record` parameter in `find_record` is never a `Proxy::DHCP::Record` or an `IPAddr`, only a string.

modules/dhcp/providers/server/infoblox.rb
```diff
@@ -0,0 +1,148 @@
...
+      post ? host.post : host.put
+      record
+    end
+    
+    def find_record(record)
+      logger.debug "loadRecord"
+      connection = connect
+      # if record is a String it can be either ip or mac, true = mac --> lookup ip
+      if record.is_a?(String) && (IPAddr.new(record) rescue nil).nil?
+        hostdhcp = ::Infoblox::HostIpv4addr.find(connection, "mac" => record).first
```
dmitri-d, line 85: I don't know the api, but I assume that this will return a nil if dhcp record cannot be found? You'll need to handle that. Ditto for `::Infoblox::Host.find` call below

modules/dhcp/providers/server/infoblox.rb
```diff
@@ -0,0 +1,148 @@
...
+        ipv4address = hostdhcp.ipv4addr
+      elsif record.is_a?(String)
+        ipv4address = record
+      end
+      ipv4address = record[:ip] if record.is_a?(Proxy::DHCP::Record)
+      ipv4address = record.to_s if record.is_a?(IPAddr)
+      host = ::Infoblox::Host.find(connection, "ipv4addr" => ipv4address).first
+      return nil if host.nil? || host.name.empty?
+      hostdhcp = ::Infoblox::HostIpv4addr.find(connection, "ipv4addr" => ipv4address).first
```
dmitri-d, line 94: Do you need to make this call when `find_record` is called with a mac address? You should be able to use the results of the call on line 85...

modules/dhcp/providers/server/infoblox.rb
```diff
@@ -0,0 +1,148 @@
...
+      hostip.use_bootfile = true
+      ## Test if Host Entry has correct IP
+      raise InvalidRecord, "#{record} IP mismatch" unless hostip.ipv4addr == record.ip
+      # Send object
+      post ? host.post : host.put
+      record
+    end
+    
+    def find_record(record)
+      logger.debug "loadRecord"
```
dmitri-d, line 81: you should call `super` here, and check if the record has already been loaded. Otherwise you'll get an error at the end when attempting to create a record that already exists.

modules/dhcp/providers/server/infoblox.rb
```diff
@@ -0,0 +1,148 @@
...
+      # Subnet should only be one, not checking that yet
+      subnet = subnets.find { |s| s.include? ipv4address}
+      Proxy::DHCP::Record.new(opts.merge(:subnet => subnet))
+    end
+
+    def loadSubnetData subnet
+      # Load network from infoblox, iterate over ips to gather additional settings
+      logger.debug "LoadSubnetData"
+      super
+      network = IPAddr.new(subnet.to_s, Socket::AF_INET)
```
dmitri-d, line 112: can't you use `subnet.network` here instead?

modules/dhcp/providers/server/infoblox.rb
```diff
@@ -0,0 +1,148 @@
...
+      connection = connect
+      # max results are currently set to work in my setup, one could calculate that setting by looking at netmask :)
+      network = ::Infoblox::Ipv4address.find(connection, "network" => "#{network}/#{subnet.cidr}", "_max_results" => "2500")
+      # Find out which hosts are in use
+      network.each do |host|
+        # next if certain values are not set
+        next if host.names.empty? || host.mac_address.empty? || host.ip_address.empty?
+        hostdhcp = ::Infoblox::HostIpv4addr.find(connection, "ipv4addr" => host.ip_address).first
```
dmitri-d, line 120: I did a quick look at infoblox api, but couldn't find any bulk operations, is that true? You may want to check if hostdhcp isn't empty -- if the number of records is large and leases short, there's a chance that the record may not be there by the time you get here.

dmitri-d:
@Klaas- 

for https://github.com/theforeman/smart-proxy/pull/295/files#diff-5daf377f40c93007066c8742a07aa442R20 I'd have a test that verifies that `::Infoblox::Host.find` is made with correct arguments, a test to make sure host gets updated, and a test to verify that the dhcp record is removed from the subnet.

for https://github.com/theforeman/smart-proxy/pull/295/files#diff-5daf377f40c93007066c8742a07aa442R40 (assuming you do the refactorings): test to verify search call uses correct args, test to verify host record creation parameters, test to verify common method that updates host record, test to verify record creation uses post, and test to verify record modification uses put.

for https://github.com/theforeman/smart-proxy/pull/295/files#diff-5daf377f40c93007066c8742a07aa442R80: test that ::Infoblox::HostIpv4addr.find is called with correct args when looking up by mac/ip, and another test to verify that `Proxy::DHCP::Record` is being created with correct values.

for https://github.com/theforeman/smart-proxy/pull/295/files#diff-5daf377f40c93007066c8742a07aa442R108: test that ::Infoblox::Ipv4address.find on line 115 is made with correct args, then another to test that `Proxy::DHCP::Reservation`s are being populated correctly (for the sake of simplicity I'd return a single host in the network.

You'll need to use mocks/stubs here quite a bit, we use mocha (http://gofreerange.com/mocha/docs/). Feel free to ask me any questions you have re: testing, mocking, or comments I made.

modules/dhcp/providers/server/infoblox.rb
Klaas-, line 91: thats what server.rb does for find_record -- during my tests I only saw strings :) if you're certain about that we can maybe remove it from both? :)

dmitri-d, line 91: I'm working on this in a dedicated PR.

ohadlevy:
@Klaas- ping?

Klaas-:
I'm busy doing something else atm, expect some more development on this 
in 1-2 months, in the mean time all help and further improvement ideas 
are welcome :)

dmitri-d:
I doubt now that dhcp provider modularization is largely completed this PR will be accepted into the core. I would suggest releasing support for infoblox as a module. PR that converts existing dhcp providers into modules is here: https://github.com/theforeman/smart-proxy/pull/312. You can use that as a starting point, and I'm available to answer questions should you have any.

MagnusJernberg:
@Klaas- @witlessbird 
Hello, i'm looking into different ipam tools and are also interested in infoblox. we need foreman to work with a 3rd party ipam tool. 

Does the previous comments mean that this fix will not be merged and that support for infoblox needs to be added as a part of #312 ?

dmitri-d:
@MagnusJernberg: not part of https://github.com/theforeman/smart-proxy/pull/312, but rather a separate PR that implements infoblox as a provider.

dmitri-d:
Closing this PR, as infoblox support should be implemented as a dhcp module provider. @Klaas-: let me know if you need some pointers for how to create a provider...

PR #296, opened 2015-06-22
fixes #10895: moved smart_proxy_chef-specific classes to the plugin i…

dmitri-d:
…tself


domcleal:
Merged as e506345, thanks @witlessbird.

PR #297, opened 2015-07-09
fixes #10941: fixed puppet class caching

dmitri-d:


dmitri-d:
@domcleal, @orrabin: ping?

test/puppet/puppet_class_test.rb
```diff
@@ -42,4 +38,31 @@ def test_scan_directory_loads_eparser_scanner
...
+  def test_json_serialization
+    clazz = Proxy::Puppet::PuppetClass.new("foreman_proxy::install",
+                                   "namedconf_path"=>"${::dns::params::namedconf_path}",
+                                   "dnsdir"=>"${::dns::params::dnsdir}")
+
+    assert clazz.to_json.include?("\"json_class\":\"Proxy::Puppet::PuppetClass\"")
+    assert clazz.to_json.include?("\"klass\":\"foreman_proxy::install\"")
+    assert(
+      clazz.to_json.include?(
+          "\"params\":{\"namedconf_path\":\"${::dns::params::namedconf_path}\",\"dnsdir\":\"${::dns::params::dnsdir}\"")
```
domcleal: This is failing intermittently on 1.8.7.. I'd probably bet that the keys are swapping order as the iteration order isn't deterministic under 1.8.

dmitri-d:
@domcleal: fixed the tests...

modules/puppet_proxy/puppet_cache.rb
```diff
@@ -19,13 +19,13 @@ def scan_directory_with_cache directory, environment, scanner
         changed    = false
         manifest   = Dir.glob("#{directory}/*").map do |path|
           puppetmodule = File.basename(path)
-          mtime        = File.mtime(path)
+          mtime        = File.mtime(path).to_i.to_s
```
domcleal: I can't think of a disadvantage to doing this, but storing the integer in JSON would be safe enough, no?

test/puppet/puppet_cache_test.rb
```diff
@@ -1,19 +1,21 @@
...
     @scanner.stubs(:scan_manifest).returns([Proxy::Puppet::PuppetClass.new('testinclude')]).then.
-                                   returns([Proxy::Puppet::PuppetClass.new('testinclude::sub::foo')])
-    Proxy::Puppet::PuppetCache.stubs(:write_to_cache)
+        returns([Proxy::Puppet::PuppetClass.new('testinclude::sub::foo')])
   end
 
   def test_should_refresh_cache_when_dir_is_not_in_cache
-    Proxy::Puppet::PuppetCache.stubs(:read_from_cache).returns({})
+    Proxy::Puppet::PuppetCache.expects(:read_from_cache).returns({})
+    Proxy::Puppet::PuppetCache.expects(:write_to_cache)
```
domcleal: Would you mind adding `.with` to this and the test below to ensure what's being written to the cache is sane?

test/puppet/puppet_cache_test.rb
```diff
@@ -28,12 +30,13 @@ def test_should_refresh_cache_when_dir_is_not_in_cache
...
+    mtime = File.mtime(Dir.glob('./test/fixtures/modules_include/*')[0]).to_i
 
     Proxy::Puppet::PuppetCache.stubs(:read_from_cache).returns('./test/fixtures/modules_include' =>
-                                                               { 'testinclude' => { :timestamp => mtime - 1000,
-                                                                                    :manifest  => [[Proxy::Puppet::PuppetClass.new('test')],
-                                                                                                   [Proxy::Puppet::PuppetClass.new('test::sub::foo')]] }})
+                                                                   { 'testinclude' => { 'timestamp' => (mtime - 1000).to_s,
+                                                                                        'manifest'  => [[Proxy::Puppet::PuppetClass.new('test')],
+                                                                                                        [Proxy::Puppet::PuppetClass.new('test::sub::foo')]] }})
+    Proxy::Puppet::PuppetCache.expects(:write_to_cache)
```
domcleal: here

modules/puppet_proxy/puppet_cache.rb
dmitri-d: The issue is with deserialization -- an integer will be serialized to a string, and will stay a string after deserialization.

domcleal: Are you sure it'll serialise to a string?  JSON should support integers.

<pre>
irb(main):004:0> JSON.load({"mtime" => 123}.to_json)
=> {"mtime"=>123}
</pre>

dmitri-d:
@domcleal: Updated tests; you were indeed correct re: integer serialization, can't remember now what made me think that. Probably too much peering into the debugger output :/.

modules/puppet_proxy/puppet_class.rb
```diff
@@ -37,11 +37,25 @@ def name
...
-  private
   attr_reader :klass
 
   def has_module?(klass)
     !!klass.index("::")
   end
+
+  def to_json(*a)
+    {
+        'json_class' => self.class.name,
```
domcleal, line 48: I assume this triggers the deserialisation process to call json_create.  I just tried on an EL6 installation with json 1.5.5 (note, that's not easy to change) and I think we're just getting hashes rather than classes:

```
Failed to show puppet classes: undefined method `name' for #<Hash:0x7f58598147d8>
```

(I think that's in the puppet_api.)

Does this require a min version of json to work?  Could we transform the hashes into objects in our deserialisation method instead of relying on json to do it?

dmitri-d, line 48: Not entirely sure why this wasn't working under json 1.5.5. Switched to explicitly instantiating `Proxy::Puppet::PuppetClass` objects instead.

dmitri-d, line 48: Here: https://github.com/theforeman/smart-proxy/pull/297/files#diff-30a6d20a4ff1bee060c846732b335086R68

modules/puppet_proxy/puppet_cache.rb
```diff
@@ -58,7 +58,16 @@ def read_from_cache environment
...
         if File.exist?(cachefile)
-          JSON.parse(File.read(cachefile))
+          as_hash = JSON.load(File.new(cachefile))
+          # ran into issues with deserialisation in older versions of json gem (ruby 1.8.7 + json 1.5.5),
+          # and opted for the deserialisation approach below.
+          # TODO: switch back to json library based deserialisation by implementing PuppetClass.json_create
+          # when support for Ruby 1.8.7 has been dropped.
+          as_hash.values.each do |puppet_module|
+            puppet_module['manifest'] =
+                puppet_module['manifest'].map { |klass| [Proxy::Puppet::PuppetClass.from_hash(klass.first)] }
```
domcleal: I get an `undefined method 'map' for nil:NilClass` error on this line.  The iteration doesn't look right to me, at least based on what appears in my JSON cache:

<pre>
[root@foreman6 ~]# json_reformat < /var/cache/foreman-proxy/cache_production.json                                                                                                                                                                                               
{
  "/etc/puppet/environments/production/modules": {
    "ntp": {
      "timestamp": 1436879142,
      "manifest": [
        [
          {
            "params": {
              "keys_file": "${$ntp::params::keys_file}",
</pre>


`as_hash` would be the top level hash with directories as keys, so `puppet_module` would be a hash of modules rather than a single one.

I think there's a mistake in the idempotency test I wrote which isn't taking into account the `cache[directory] = lcache` line (49) above, causing this to crash in real life.

dmitri-d:
This should do it. Turns out class scanner output is a bit weird -- can contain `PuppetClass` instances, empty arrays, or arrays of empty arrays. 

PS> Please make it stop.

dmitri-d:
[test]

domcleal:
Okay, that's looking good.

> Turns out class scanner output is a bit weird -- can contain PuppetClass instances, empty arrays, or arrays of empty arrays.

Yeah, I was unsure about your change to flatten and map the manifest array, but it doesn't much matter.  It looks like an array per manifest file and then classes beneath that - while it's usually 1:1, I guess somebody thought to preserve the layout.  It only gets flattened later, so doesn't much matter.

> PS> Please make it stop.

One more.  On upgrade from the previous version with a bad cached file, I think it needs to be discarded if the data doesn't match the expected format.  I get this when requesting classes:

```
Failed to show puppet classes: Must provide puppet class name
```

Here's a copy of the cache if it helps:

```
{"/etc/puppet/environments/production/modules":{"ntp":{"manifest":[["ntp"],["ntp::config"],["ntp::install"],["ntp::service"],["ntp::params"]],"timestamp":"Wed Jul 15 11:51:20 +0100 2015","manifest":[["ntp"],["ntp::config"],["ntp::install"],["ntp::service"],["ntp::params"]],"timestamp":"Wed Jul 15 11:51:21 +0100 2015"},"stdlib":{"manifest":[["stdlib"],["stdlib::stages"]],"timestamp":"Wed Jul 15 11:51:20 +0100 2015","manifest":[["stdlib"],["stdlib::stages"]],"timestamp":"Wed Jul 15 11:51:21 +0100 2015"}}}
```

dmitri-d:
@domcleal: added version info to the cache file; cache is now being discarded if plugin version and cache version do not match.

domcleal:
That's a really good idea!

ohadlevy:
:+1:

domcleal:
Merged as 4d5fc80, thanks @witlessbird.

PR #299, opened 2015-07-13
Fixes #10877 - puppet pinned to 3.x series

lzap:
Does not work yet.


domcleal:
Merged as 9bb37a6, thanks @lzap.

PR #300, opened 2015-07-13
Fixes #10259 - Templates are now fully proxied

lzap:
It was ignoring some parameters (MAC for bootdisk) as well as headers
(HTTP_X_RHN_PROVISIONING_*).

Also removed incorrect TFTP unit test that was failing on all systems which
do not have set TFTP root to `/var/lib/tftpboot`.

Changes by Lukas Zapletal

Replaces https://github.com/theforeman/smart-proxy/pull/287


lzap:
@ohadlevy this implements what we talked about - full proxying. Also fixed RHN Anaconda headers which were totally ignored!

lzap:
Fixed Ruby 1.8.7 syntax error.

lzap:
And a Rubocop warning.

test/tftp/tftp_test.rb
```diff
@@ -11,10 +11,6 @@ def test_should_have_a_logger
     assert_respond_to @tftp, :logger
   end
 
-  def test_path_to_tftp_directory_without_tftproot_setting
```
domcleal, line 14: This looks unrelated.

ohadlevy:
@witlessbird can you review please? :+1: from my side.

ohadlevy:
@lzap is there a PR in foreman side to take usage of this functionality ?

modules/templates/templates_api.rb
```diff
@@ -14,8 +14,8 @@ class Proxy::TemplatesApi < Sinatra::Base
   end
 
   get "/:kind" do |kind|
-    log_halt(500, "Failed to retrieve #{kind} template for #{params[:token]}: ") do
-      Proxy::Templates::Handler.get_template(kind, params[:token], params[:static])
+    log_halt(500, "Failed to retrieve #{kind} template for #{params.inspect}: ") do
+      Proxy::Templates::Handler.get_template(kind, params, headers)
```
dmitri-d: `Sinatra::Helpers#headers` is for setting response headers, and returns current response headers when called without parameters: http://www.rubydoc.info/gems/sinatra/Sinatra/Helpers#headers-instance_method. Rack stores request headers in `request.env` hash, under keys starting with `HTTP_` (http://rack.rubyforge.org/doc/SPEC.html, lookup "HTTP_ Variables").

modules/templates/handler.rb
```diff
@@ -7,23 +7,24 @@ module Proxy::Templates
...
 
-    def get_template(kind, token, static = false)
-      opts = { :token => token,
-               :url   => Proxy::Templates::Plugin.settings.template_url
-      }
-      opts[:static] = static if static
+    def get_template(kind, params, headers)
+      opts = params.merge(:url => Proxy::Templates::Plugin.settings.template_url)
       request = request_factory.create_get("/unattended/#{kind}", opts)
+      headers.each do |k, v|
```
witlessbird: This modifies response headers, not request headers of `request`.

modules/templates/handler.rb
```diff
@@ -7,23 +7,24 @@ module Proxy::Templates
...
+        request.add_field(k, v)
```
dmitri-d: Probably worth modifying `ForemanRequestFactory` to accept headers as a parameter to `create_` methods (https://github.com/theforeman/smart-proxy/blob/develop/lib/proxy/request.rb#L16).

modules/templates/handler.rb
```diff
@@ -7,23 +7,24 @@ module Proxy::Templates
...
+      end
```
dmitri-d: Also, I think it would be nice to add `X-Forwarded-For` header here, since we are acting as an http proxy.

modules/templates/templates_api.rb
lzap: Not sure what you mean, I want to verbosely copy all the headers from request to response, so `HTTP_` prefix is unwanted and the helper method `headers` is not avaible in the Handler instance.

lzap:
@ohadlevy https://github.com/theforeman/foreman_bootdisk/pull/9

modules/templates/handler.rb
lzap: Sure will do.

modules/templates/handler.rb
lzap: Ok good point. I have amended one additional change because it was not working properly. Will amend the above two points later today.

modules/templates/templates_api.rb
dmitri-d: at https://github.com/theforeman/smart-proxy/pull/300#discussion-diff-34666024R18 you are calling `Sinatra::Helpers#headers` to populate headers of the request to Foreman. This method doesn't return incoming request headers, you'll get something like: 
`{"Content-Type"=>nil}`
vs. actual request headers:
`{"HTTP_USER_AGENT"=>"curl/7.32.0", "HTTP_HOST"=>"localhost:8000", "HTTP_ACCEPT"=>"*/*", "HTTP_VERSION"=>"HTTP/1.1"}`

lzap:
Amended, I am passing all three objects in the Handler instance:

```
      Proxy::Templates::Handler.get_template(kind, request, params, headers)
```

I think this is the sanest method.

modules/templates/handler.rb
```diff
@@ -7,23 +7,29 @@ module Proxy::Templates
...
-      opts = { :token => token,
-               :url   => Proxy::Templates::Plugin.settings.template_url
-      }
-      opts[:static] = static if static
-      request = request_factory.create_get("/unattended/#{kind}", opts)
-      logger.debug "retrieving a template from %s/%s" % [Proxy::Templates::Plugin.settings.template_url, request.path]
-      res = send_request(request)
+    def get_template(kind, request, params, headers)
+      url = Proxy::Templates::Plugin.settings.template_url || raise("Setting template_url not set!")
+      proxy_ip = URI.parse(url).host rescue "127.0.0.1"
```
dmitri-d: Instead of this and the line above, set a default value for `template_url` setting in https://github.com/theforeman/smart-proxy/blob/develop/modules/templates/templates_plugin.rb

lzap: Actually this must be always true, otherwise the plugin disables itself. I will remove the extra code.

lzap:
Fixed tests and all the comments. Added extra test file for ForemanRequest class.

modules/templates/handler.rb
```diff
@@ -7,23 +7,28 @@ module Proxy::Templates
...
-      logger.debug "retrieving a template from %s/%s" % [Proxy::Templates::Plugin.settings.template_url, request.path]
-      res = send_request(request)
+    def get_template(kind, request, params)
+      url = Proxy::Templates::Plugin.settings.template_url
+      proxy_ip = URI.parse(url).host
+      opts = params.clone.merge(:url => url)
+      opts.delete("kind")
+      opts.delete("splat")
+      opts.delete("captures")
+      proxy_headers = Hash[request.env.select{|k,v| k =~ /^HTTP_/ && k !~ /^HTTP_(VERSION|HOST)$/}.map{|k,v| [k[5..-1],v]}].merge("X-Forwarded-For" => "#{request.ip}, #{proxy_ip}")
```
dmitri-d: This will break on 1.8.7 (Hash#select returns an array there).

modules/templates/handler.rb
```diff
@@ -7,23 +7,28 @@ module Proxy::Templates
   class Handler < ::Proxy::HttpRequest::ForemanRequest
```
dmitri-d: Could you rename `Handler` to something like `ProxyTemplateRequest`?

modules/templates/handler.rb
```diff
@@ -7,23 +7,28 @@ module Proxy::Templates
...
 
       # You get a 201 from the 'built' URL
-      raise "Error retrieving #{kind} for #{token} from #{uri.host}: #{res.class}" unless ["200", "201"].include?(res.code)
+      raise "Error retrieving #{kind} for #{opts.inspect} from #{uri.host}: #{res.class}" unless ["200", "201"].include?(res.code)
+      Proxy::Log.logger.info "Template: request for #{kind} using #{opts.inspect} at #{uri.host}"
       res.body
     end
 
-    def self.get_template kind, token, static = false
+    def self.get_template(kind, request, params)
```
dmitri-d: I would prefer passing ip and env hash directly, it's more clear and easier to test...

modules/templates/handler.rb
```diff
@@ -7,23 +7,28 @@ module Proxy::Templates
...
-               :url   => Proxy::Templates::Plugin.settings.template_url
-      }
-      opts[:static] = static if static
-      request = request_factory.create_get("/unattended/#{kind}", opts)
-      logger.debug "retrieving a template from %s/%s" % [Proxy::Templates::Plugin.settings.template_url, request.path]
-      res = send_request(request)
+    def get_template(kind, request, params)
+      url = Proxy::Templates::Plugin.settings.template_url
+      proxy_ip = URI.parse(url).host
+      opts = params.clone.merge(:url => url)
```
dmitri-d: I don't think you need to clone `params` here, `ForemanRequestFactory` isn't going to modify the object...

lzap: Factory does not modify it, but `get_template` method does. It is removing some keys from it below.

dmitri-d: Are these the modifications below? Do you know where these keys are coming from?

modules/templates/handler.rb
lzap: I don't understand, this is a clean code and works fine.

modules/templates/handler.rb
lzap: Sure, fixed.

modules/templates/handler.rb
lzap: Hmmm I am not so sure about this one. Objects should be passed instead of it's data. I prefer nicer code to nicer tests.

modules/templates/handler.rb
lzap: Yes these are generated by Sinatra. Since I want this to be generic, I want to copy all the parameters except these. Not sure if there is more clever method than this one.

modules/templates/template_proxy_request.rb
```diff
@@ -0,0 +1,35 @@
...
+    include Proxy::Log
+
+    def get_template(kind, request, params)
+      url = Proxy::Templates::Plugin.settings.template_url
+      proxy_ip = URI.parse(url).host
+      opts = params.clone.merge(:url => url)
+      opts.delete("kind")
+      opts.delete("splat")
+      opts.delete("captures")
+      proxy_headers = Hash[request.env.select{|k,v| k =~ /^HTTP_/ && k !~ /^HTTP_(VERSION|HOST)$/}.map{|k,v| [k[5..-1],v]}].merge("X-Forwarded-For" => "#{request.ip}, #{proxy_ip}")
```
dmitri-d: Didn't realise ruby does a splat under the hood when assigning an array to multiple vars, so this works under 1.8.7. None of the tests cover the filter bit though, which would be useful. Probably worth extracting this line into a method too.

test/templates/template_proxy_request_test.rb
```diff
@@ -0,0 +1,58 @@
...
+require 'webmock/test_unit'
+
+
+class TemplateProxyRequestTest < Test::Unit::TestCase
+  def setup
+    @foreman_url = 'https://foreman.example.com'
+    Proxy::SETTINGS.stubs(:foreman_url).returns(@foreman_url)
+    @template_url = 'http://proxy.lan:8443'
+    Proxy::Templates::Plugin.settings.stubs(:template_url).returns(@template_url)
+    @request = mock('request')
```
dmitri-d: This mocking wouldn't be necessary if you passed environment and ip parameters directly into `TemplateProxyRequest.get_template` method. Not only would this make it clear what the method operates on, but would also make tests less fragile. Don't use mocks/stubs unless absolutely necessary.

modules/templates/template_proxy_request.rb
lzap: Adding tests, actually passing just `env` hash is better in this case.

test/templates/template_proxy_request_test.rb
lzap: Yeah, fixed.

lzap:
Amended.

test/templates/template_proxy_request_test.rb
```diff
@@ -0,0 +1,73 @@
...
+    @foreman_url = 'https://foreman.example.com'
+    Proxy::SETTINGS.stubs(:foreman_url).returns(@foreman_url)
+    @template_url = 'http://proxy.lan:8443'
+    Proxy::Templates::Plugin.settings.stubs(:template_url).returns(@template_url)
+    @request_env = {
+      'REMOTE_ADDR' => '1.2.3.4'
+    }
+  end
+
+  def text_header_extraction_empty
```
dmitri-d: `test` is misspelled here and in two other tests below, which means they'll be skipped...

dmitri-d:
rubocop isn't happy either. Looks good other than that.

lzap:
Thanks for review!

lzap:
Fixed failed tests.

lzap:
@witlessbird any chance of merging that in before your PTO? Thanks!

dmitri-d:
@lzap: will test this in a bit.

dmitri-d:
merged in at 38a0935. Thanks, @lzap!

PR #301, opened 2015-07-13
fixes #11098: fixed ruby 1.8-specific issues around plugin and provider searches

dmitri-d:


dmitri-d:
Fixed another issue similar to reported in #11098; added tests.

dmitri-d:
Also, now if ssl is configured, but the configuration is incorrect (key/certificate cannot be read), smart-proxy will not load. This is a change in behaviour, but I think this is the desirable one?

domcleal:
Yeah, I'd say so.. think it originally used to work that way.

dmitri-d:
Also, it now prints a notification on the console if startup fails.

lib/smart_proxy.rb
```diff
@@ -150,6 +159,12 @@ def launch
       end
 
       (t1 || t2).join
+    rescue Exception => e
```
domcleal, line 165: This causes an issue when `daemon` is enabled...

<pre>
E, [2015-07-13T14:23:58.804946 #20218] ERROR -- : Error during startup, terminating. exit SystemExit
D, [2015-07-13T14:23:58.805110 #20218] DEBUG -- : exit:/usr/share/foreman-proxy/lib/daemon.rb:4:in `exit'
/usr/share/foreman-proxy/lib/daemon.rb:4:in `daemon'
/usr/share/foreman-proxy/lib/smart_proxy.rb:151:in `launch'
/usr/share/foreman-proxy/bin/smart-proxy:6
E, [2015-07-13T14:23:58.809238 #20219] ERROR -- : Error during startup, terminating. exit SystemExit
D, [2015-07-13T14:23:58.809417 #20219] DEBUG -- : exit:/usr/share/foreman-proxy/lib/daemon.rb:6:in `exit'
/usr/share/foreman-proxy/lib/daemon.rb:6:in `daemon'
/usr/share/foreman-proxy/lib/smart_proxy.rb:151:in `launch'
/usr/share/foreman-proxy/bin/smart-proxy:6
</pre>


I think what's happening is that the SystemExit inherits from Exception (rather than RuntimeError, which surprised me) and simply gets caught when the main thread legitimately exits.

dmitri-d:
@domcleal: fixed the issue with `SystemExit` being caught by the top-level exception handler.

domcleal:
> @domcleal: fixed the issue with SystemExit being caught by the top-level exception handler.

How'd you do that?  The only change I see is in the "Both http and https are disabled" conditional, which I don't think is applicable.

dmitri-d:
@domcleal: SystemExit's fixed now. For realz.

dmitri-d:
[test]

domcleal:
Merged as 0eb8179 for 1.10, thanks @witlessbird!

PR #302, opened 2015-07-16
fixes #11131 - use consistent timestamps in cache test expectations

domcleal:
Sorry, should've spotted this before closing #297.


dmitri-d:
I was wondering about that. Thanks for fixing this, @domcleal ! Merged in at f11f708.

PR #303, opened 2015-07-23
Fixes #11103 - removed unnecessary test

lzap:
This unit test has no sense and it always fail on systems which has different
TFTP root location defined in settings.


domcleal:
or call `Proxy::DhcpPlugin.load_test_settings({})`?

dmitri-d:
Looking into this; I don't think the test is unnecessary -- it attempts to verify that default setting is available, which is useful.

dmitri-d:
Closing in favour of https://github.com/theforeman/smart-proxy/pull/305.

PR #306, opened 2015-08-05
Fixes #11866: first pass at speeding up dhcp (both isc and ms).

dmitri-d:


dmitri-d:
This is the first set of changes to speed up dhcp. This PR replaces all but one (here: https://github.com/theforeman/smart-proxy/pull/306/files#diff-d045c300c8c8c9ae152c6c973bb2276eR31) linear searches to constant-time searches.

modules/dhcp/providers/server/native_ms.rb
```diff
@@ -186,7 +207,7 @@ def execute cmd, msg=nil, error_only=false
           response += std_err.readlines
         end
       rescue TimeoutError
-        raise Proxy::DHCP::Error.new("Netsh did not respond within #{tsecs} seconds")
+        raise Proxy::DHCP::Error.new("Netsh did nota respond within #{tsecs} seconds")
```
ekohl, line 215: Typo introduced here.

domcleal:
It's kind of difficult with the diff to see what the refactoring achieves.  Please correct me if necessary, but I understand it to be:

1) Move storage from the subnet to the subnet_service, move all data lookups queries through the server instead of dhcp_api using the subnet directly.
2) Keep interactions with the subnet_service largely in the server base class, extend methods in subclasses to add the "real" create/destroy actions etc.
3) The lifecycle of a server is a single request, and so is the subnet_service - so the cache is only used for one request and doesn't have any invalidation issues etc because of that.

dmitri-d:
@domcleal: Correct on all three items. I tried to decouple `Proxy::DHCP::Server`, `Proxy::DHCP::Subnet`, and `Proxy::DHCP::Record` from each other (to make the whole thing testable) and preserve existing behavior.

modules/dhcp/providers/server/isc.rb
```diff
@@ -104,38 +91,72 @@ def loadSubnetData subnet
...
       end
-      report "Enumerated hosts on #{subnet.network}"
     end
 
-    private
+    def loadSubnetData subnet
+      load_subnet_data(parse_config_and_leases)
+    end
+
+    def load_subnets
```
domcleal: This is incredibly similar naming to loadSubnets, can they be substantially different please?  Or incorporate this straight into loadSubnets, or rename loadSubnets to load_subnets.  (It would be nice to remove the camelCase names.)

modules/dhcp/dhcp_api.rb
```diff
@@ -16,17 +16,28 @@ class Proxy::DhcpApi < ::Sinatra::Base
         end
         @server = Proxy::DHCP::ISC.new(:name => "127.0.0.1",
                                        :config => Proxy::DhcpPlugin.settings.dhcp_config,
-                                       :leases => Proxy::DhcpPlugin.settings.dhcp_leases)
+                                       :leases => Proxy::DhcpPlugin.settings.dhcp_leases,
+                                       :service => Proxy::DHCP::SubnetService.new(Proxy::MemoryStore.new, Proxy::MemoryStore.new,
```
domcleal: Could these be optimised so it's not repeated for each provider?  Perhaps by initialising sane defaults inside the Proxy::DHCP::SubnetService.

modules/dhcp/subnet_service.rb
```diff
@@ -0,0 +1,98 @@
+class Proxy::DHCP::SubnetService
+  # rubocop:disable Metrics/ParameterLists
```
domcleal, line 10: missing an :enable?

domcleal:
The lifecycle of the store object was initially my biggest concern until I realised that it was only going to last for a single request.  That being the case, are you planning to use it for longer-term caching at any point as it has quite a few flaws if that was the plan?

As it stands, I'm not sure it's worth making alterations after adding/deleting records if it's not actually going to last for long.  I also wonder how much it's needed in other cases.

dmitri-d:
Nods, there's no caching happening atm, as all stores live only for the duration of a request. I think the biggest benefit of this PR is simplified testing, as speed improvements are marginal. The most expensive bit -- parsing of leases file is still present. 

I'd like to try replacing all operations on single records (view/add/delete) to omapi calls (in windows netsh should be able to provide similar functionality) -- we'll be trading lease file parsing for network latency, but the latter should be much cheaper for large networks. I think the only place where caching _might_ be useful is free IP address lookups.

I agree re: memory store updates on record additions/deletions, I'll remove those. Lemme know if you have questions/concerns.

modules/dhcp/dhcp_api.rb
dmitri-d: Nods, this is pretty unseemly.

dmitri-d:
@domcleal: renamed subnet- and record-loading methods in `Proxy::DHCP::Server` and subclasses, cleaned up provider instantiation. Memory store is no longer being updated on add/remove record operations.

modules/dhcp/providers/server/isc.rb
```diff
@@ -104,38 +96,72 @@ def loadSubnetData subnet
...
 
-        Proxy::DHCP::Subnet.new(self, match[0], match[1])
+        ret_val << Proxy::DHCP::Subnet.new(match[0], match[1])
       end
-      "Enumerated the scopes on #{@name}"
+
+      ret_val
+    end
+
+    def loadSubnets
```
shlomizadok, line 162: nitpick: rename this method name? (all over)

dmitri-d, line 162: It's a legacy name, I'd like to minimize renames in older parts of the code at this point.

modules/dhcp/subnet_service.rb
```diff
@@ -0,0 +1,105 @@
...
+    @leases_by_ip = leases_ip_store
+    @leases_by_mac = leases_mac_store
+    @reservations_by_ip = reservations_ip_store
+    @reservations_by_mac = reservations_mac_store
+    @reservations_by_name = reservations_name_store
+  end
+  # rubocop:enable Metrics/ParameterLists
+
+  def add_subnet(subnet)
+    raise Proxy::DHCP::Error, "Unable to Add Subnet #{subnet}" if find_subnet(subnet.network)
```
domcleal: Nitpick. lower case "add subnet"

modules/dhcp/subnet_service.rb
```diff
@@ -0,0 +1,105 @@
...
+  def all_subnets
+    @subnets.values
+  end
+
+  def add_lease(subnet_address, record)
+    @leases_by_ip[subnet_address, record.ip] = record
+    @leases_by_mac[subnet_address, record.mac] = record
+  end
+
+  def add_host(subnet_address, record)
```
domcleal, line 60: Could we have some debug or trace level logging in these methods?  I think it might be useful - we used to in add_record.

test/dhcp/subnet_service_test.rb
```diff
@@ -0,0 +1,187 @@
...
+    assert @service.all_subnets.include?(subnets.first)
+    assert @service.all_subnets.include?(subnets.last)
+  end
+
+  def test_find_subnet
+    subnets = [Proxy::DHCP::Subnet.new("192.168.0.0", "255.255.255.0"),
+               Proxy::DHCP::Subnet.new("192.168.1.0", "255.255.255.0")]
+    @service.add_subnets(*subnets)
+
+    assert_equal subnets.first, @service.find_subnet("192.168.0.0")
```
domcleal, line 53: This should also be tested for the latter part of the method, to ensure it can find a subnet by an IP within the net, not just the network address.

modules/dhcp/dhcp_api.rb
```diff
@@ -92,6 +105,10 @@ def load_subnet
   # create a new record in a network
   post "/:network" do
     begin
+
```
domcleal: Nitpick, remove blank lines at beginning of the block, ditto below.

domcleal:
Sorry for the slow trickle of review comments, I've been sidetracked.  I'd suggest waiting until I set the WoC label before trying to address them - thanks.

dmitri-d:
@domcleal: rebased and fixed things you pointed out (good catch re: missing test).

modules/dhcp/providers/server/isc.rb
```diff
@@ -25,7 +32,6 @@ def delRecord subnet, record
       omcmd "open"
       omcmd "remove"
       omcmd("disconnect", msg)
-      subnet.delete record
     end
```
domcleal, line 35: missing super() call in delRecord?

domcleal, line 35: Ignore, thought I saw it defined in the Server class.

dmitri-d, line 35: It used to be, but it was an empty method, didn't think there was any point to keep it around.

domcleal:
Okay, so I'm pretty happy with it and it tests well.  I'm not sure that we should use the associated ticket #6916 though, since I don't imagine that this fixes the issue - my understanding is that the native MS provider's as slow as it is because of the enumeration of subnets (which is still present), not the lookups.

dmitri-d:
@domcleal: there are still performance-related issues in both MS and ISC providers, this PR only addresses the "low-hanging fruit" -- various linear searches. If you don't think it's a sufficient justification I can open a new ticket instead.

domcleal:
@witlessbird if you could, I'd appreciate it.  I'm hesitant to close that one right now.

dmitri-d:
@domcleal: done and done: http://projects.theforeman.org/issues/11866

modules/dhcp/providers/server/native_ms.rb
domcleal, line 215: Fixed this on merge.

domcleal:
Merged as ce9ff4c, thanks @witlessbird.

PR #307, opened 2015-08-06
Fixes #10940: puppetrun works properly now on ruby > 1.8.7

dmitri-d:


domcleal:
That's working better, but there's another issue with customrun on 1.9+, up to you if you'd like me to file it separately as it's different.

Specifying customrun_args as a string containing multiple arguments (`:customrun_args: -a -b -c`) doesn't work as expected on 1.9+, because they'd be passed in as a single argument to the command rather than being parsed like a shell would do (and as our code does on 1.8, since it joins the array).

It's possible to specify an array of arguments in the YAML config file, though this isn't what's documented.  This would mean we need to flatten the resulting array.

dmitri-d:
@domcleal: this should do it.

domcleal:
Merged as b446e0c, thanks @witlessbird.

PR #308, opened 2015-08-07
Fixes #11310 - added kexec option to BMC shell provider

lzap:
This allows to pass in additional options via BODY. When these are present, the
host will be kexeced instead of rebooted. This is used on
foreman-discovery-image.

### DIFF
--- a/lib/proxy/http_downloads.rb
+++ b/lib/proxy/http_downloads.rb
@@ -7,9 +7,9 @@ module Proxy
         lock =  Proxy::FileLock.try_locking(dst)
         unless lock.nil?
           HttpDownload.new(src, dst, lock)
-          return true
+        else
+          false
         end
-        false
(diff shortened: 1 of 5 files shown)


domcleal:
I'm unsure that direct kernel loading should be part of the BMC API, particularly inside the power cycle action.  It ought to be in its own module as it's so specialist.  Remote code execution shouldn't be a core smart proxy feature!

lzap:
Absolutely, the same applies for the shell provider. I will remove that as well.

lzap:
Moved to new plugin: https://github.com/theforeman/smart_proxy_discovery_image/pull/1

Shell I initiate removal of unused shell provider rather now?

domcleal:
> Shell I initiate removal of unused shell provider rather now?

Fine by me.

PR #310, opened 2015-08-10
Fixes #11318 - improved HttpDownloads API

lzap:
Allows to `join` processes, useful for discovery image plugin.


domcleal:
Please note that tests are red.

dLobatog:
Minor comment, :+1: to the change, it's great. Can you change the commit name to make it highlight what it actually does? The current message is a bit vague.

lzap:
Tough task to do on 50 characters, but tried. Fixed tests.

domcleal:
A test should also be added to check the return value, not just non-nil, e.g. assert_instance_of.

test/http_downloads_test.rb
```diff
@@ -8,7 +8,7 @@ def command(src, dst)
 
 class HttpDownloadsTest < Test::Unit::TestCase
   def test_should_download_a_file
-    assert Proxy::HttpDownloads.start_download 'source', 'destination'
+    refute Proxy::HttpDownloads.start_download('source', 'destination').nil?
```
dmitri-d: use `assert_not_nil` instead. Better still, as @domcleal suggested verify that it returns an instance of expected class..

lzap:
Oh there was a missing `return` statement... Fixed.

test/http_downloads_test.rb
```diff
@@ -7,8 +7,14 @@ def command(src, dst)
 end
 
 class HttpDownloadsTest < Test::Unit::TestCase
+  def test_should_return_an_instance
+    assert_instance_of Proxy::HttpDownload, Proxy::HttpDownloads.start_download('source', 'destination')
+  ensure
+    File.delete('destination')
```
domcleal: Unless you join the resulting HttpDownload, could this be executing still in the background of the test?  Perhaps better to stub it like tftp_api_test?

dmitri-d: Most definitely. Joining a thread isn't great, as it will slow down tests. It might be worth it to break up HttpDownloads instantiation and starting and stub the latter in the test.

dmitri-d:
[test]

test/http_downloads_test.rb
lzap: > Most definitely. Joining a thread isn't great, as it will slow down tests. It would be worth it to break up HttpDownloads instantiation and starting, but is probably out of scope for this PR. Stubbing of HttpDownload constructor should work (ugh).

How about to mock Thread.new(acommand). This way I can fix all three.

Mocking popen3 is also possible, but challenging.

## 

Later,
 Lukas #lzap Zapletal

lzap:
Redesigned tests completely.

test/http_downloads_test.rb
```diff
@@ -1,23 +1,41 @@
...
-class HttpDownloadsTest < Test::Unit::TestCase
-  def test_should_download_a_file
-    assert Proxy::HttpDownloads.start_download 'source', 'destination'
-    assert File.exist?('destination')
+  def tmp(name)
+    File.join(Dir.tmpdir(), "http-download-#{name}.tmp")
+  end
+
+  def test_should_return_an_instance
+    src = tmp('source')
```
dmitri-d: source path doesn't matter here and in other tests, as it's evaluated by `wget` and no files are being created.

test/http_downloads_test.rb
```diff
@@ -1,23 +1,41 @@
...
+    dst = tmp('destination')
+    Thread.stubs(:new).returns(new_thread_mock)
+    assert_instance_of Proxy::HttpDownload, Proxy::HttpDownloads.start_download(src, dst)
+  ensure
+    File.delete(dst)
```
dmitri-d: Is this necessary? The block we pass to the thread at https://github.com/lzap/smart-proxy/blob/http-download-11318/lib/proxy/util.rb#L15 won't be executed, so no files will be created.

test/http_downloads_test.rb
```diff
@@ -1,23 +1,41 @@
...
+
+  def test_should_run_wget
+    src = tmp('source')
+    dst = tmp('destination')
+    cmd = "/usr/bin/wget --timeout=10 --tries=3 --no-check-certificate -nv -c \"#{src}\" -O \"#{dst}\""
+    Thread.expects(:new).with(cmd).returns(new_thread_mock)
+    assert_not_nil Proxy::HttpDownloads.start_download(src, dst)
   ensure
-    File.delete('destination')
+    File.delete(dst)
```
dmitri-d: ditto here.

test/http_downloads_test.rb
```diff
@@ -1,23 +1,41 @@
...
 
   def test_should_skip_download_if_one_is_in_progress
-    locked = Proxy::FileLock.try_locking("another_destination")
-    assert !(Proxy::HttpDownloads.start_download 'another_source', 'another_destination')
+    src = tmp('source')
+    locked = Proxy::FileLock.try_locking(tmp('other'))
+    refute Proxy::HttpDownloads.start_download(src, locked.path)
   ensure
-    File.delete(locked.path)
+     File.delete(locked.path)
```
dmitri-d: extra space in here.

test/http_downloads_test.rb
```diff
@@ -1,23 +1,41 @@
...
-    File.delete('destination')
+    File.delete(dst)
   end
 
   def test_should_skip_download_if_one_is_in_progress
-    locked = Proxy::FileLock.try_locking("another_destination")
-    assert !(Proxy::HttpDownloads.start_download 'another_source', 'another_destination')
+    src = tmp('source')
+    locked = Proxy::FileLock.try_locking(tmp('other'))
+    refute Proxy::HttpDownloads.start_download(src, locked.path)
```
dmitri-d: no `refute` on 1.8.7.

dmitri-d:
@lzap: sorry to be a pain, I think it would be much easier to test this thing if you broke up task instantiation and starting into two separate actions. You would then need to stub out (or override in a test-specific class) the method responsible for starting of the command, which isn't as invasive as stubbing out `Thread` constructor. As a bonus, this would avoid creation of temp files too.

test/http_downloads_test.rb
lzap: > Is this necessary? The block we pass to the thread at https://github.com/lzap/smart-proxy/blob/http-download-11318/lib/proxy/util.rb#L15 won't be executed, so no files will be created.

It is, I believe, since our `unlock` code only closes the file handle
(we do create the file to call `flock` on it).

Above comment fixed.

## 

Later,
 Lukas #lzap Zapletal

dmitri-d:
@lzap: rubocop is failing one of the checks, pls. also see my comment above.

lzap:
Ok this looks much more cleaner, also tests look nicer. I am actually only test the command construction, the command task object is well covered with true/false real cases. Should I rename the `http_downloads.rb` to `http_download.rb` now ?

lib/proxy/http_downloads.rb
```diff
@@ -1,29 +1,27 @@
...
-      end
-    end
-  end
-
   class HttpDownload < Proxy::Util::CommandTask
     include Util
 
-    def initialize(src, dst, lock)
-      super(command(src, dst)) {  Proxy::FileLock.unlock(lock) }
+    def initialize(src, dst, lock_dst = true)
```
dmitri-d: I can't think of any cases when we don't want to ensure that only a single download is in progress -- it results in corrupted files. What was the reason for adding of this flag?

dmitri-d:
@lzap: sure, renaming sounds good.

lzap:
> @lzap: sure, renaming sounds good. 

Done, removed the extra flag.

## 

Later,
 Lukas #lzap Zapletal

dmitri-d:
merged in at 5f7fd34. Thanks, @lzap!

PR #311, opened 2015-08-11
Fixes #11323 - fixed PID writing, interrupt trap and daemon logging

lzap:
Currently, our deamonizing works as follows:
- initialize configuration, logger and plugins
- create http/https rack servers
- daemonize the main process
- start rack servers in own threads (each writes own pid file into same file)
- wait five seconds
- override interrupt signal trap
- wait for both threads to finish

Problems:
- Rack writes main process pid twice (the main process should do this on its
  own)
- pid file is written way too long after the main process exited (this is
  causing irregular warnings with systemd "not readable (yet?) after start")
- we should get rid of the 5 seconds sleep (what is this needed for)?
- when logging is setup to STDOUT it does not work with daemon setting (these
  are actually exclusive, we fail out in that case)

This patch corrects PID writing, shortens the window between parent process
fork and PID creation, removes the useless 5 second sleep / install trap and
introduces new setting "SYSLOG" which can be used in daemonized mode instead of
STDOUT (used in Discovery Image with journald).

### DIFF
--- a/lib/proxy/log.rb
+++ b/lib/proxy/log.rb
@@ -1,6 +1,13 @@
 require 'logger'
+begin
+  require 'syslog/logger'
+  ::Syslog::Logger.class_eval { alias_method :write, :info }
+rescue LoadError
+  puts "Setting log_file=SYSLOG not supported on this platform, ignoring"
+end
 
-Logger.class_eval { alias_method :write, :'<<' } # ::Rack::CommonLogger expects loggers to implement 'write' method
+# ::Rack::CommonLogger expects loggers to implement 'write' method
+Logger.class_eval { alias_method :write, :info }
 
 module Proxy
   module Log
@@ -10,13 +17,27 @@ module Proxy
       @@logger ||= ::Proxy::Log.logger
     end
 
+    def self.default_logger(log_file)
+      # We keep the last 6 10MB log files
+      ::Logger.new(log_file, 6, 1024*1024*10)
+    end
+
     def self.logger
       log_file = ::Proxy::SETTINGS.log_file
       if log_file.upcase == 'STDOUT'
+        if SETTINGS.daemon
+          puts "Settings log_file=STDOUT and daemon=true are incompatible, exiting..."
+          exit 1
+        end
         logger = ::Logger.new(STDOUT)
+      elsif log_file.upcase == 'SYSLOG'
+        begin
+          logger = ::Syslog::Logger.new 'foreman-proxy'
+        rescue
+          logger = default_logger(log_file)
+        end
       else
-        # We keep the last 6 10MB log files
-        logger = ::Logger.new(log_file, 6, 1024*1024*10)
+        logger = default_logger(log_file)
       end
       logger.level = ::Logger.const_get(::Proxy::SETTINGS.log_level.upcase)
       logger
--- a/lib/smart_proxy.rb
+++ b/lib/smart_proxy.rb
@@ -1,12 +1,13 @@
 APP_ROOT = "#{File.dirname(__FILE__)}/.."
 
 require 'proxy'
+require 'launcher'
 
 require 'fileutils'
 require 'pathname'
 require 'checks'
 require 'webrick/https'
-require 'daemon' # FIXME: Do we still need this?
+require 'daemon'
 
 require 'checks'
 require 'proxy/log'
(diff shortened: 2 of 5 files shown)


domcleal:
Also rather red.

lzap:
Oh looks like Rubocop only. Fixed.

lzap:
Fixed load error.

lib/proxy/log.rb
```diff
@@ -1,6 +1,13 @@
...
+begin
+  require 'syslog/logger'
+rescue LoadError
+  puts "Setting log_file=SYSLOG not supported on this platform"
+end
 
-Logger.class_eval { alias_method :write, :'<<' } # ::Rack::CommonLogger expects loggers to implement 'write' method
+# ::Rack::CommonLogger expects loggers to implement 'write' method
+Logger.class_eval { alias_method :write, :info }
+::Syslog::Logger.class_eval { alias_method :write, :info }
```
dmitri-d: This isn't going to work on rubies < 2.0.0: there's no `Syslog::Logger` on these platforms, and what exists wildly differs from the interface used now. That's the reason behind tests breakage (also proxy won't start on older rubies now either).

lib/proxy/log.rb
```diff
@@ -13,7 +20,18 @@ def logger
...
+        if SETTINGS.daemon
+          puts "Settings log_file=STDOUT and daemon=true are incompatible, exiting..."
+          exit 1
+        end
         logger = ::Logger.new(STDOUT)
+      elsif log_file.upcase == 'SYSLOG'
+        begin
+          logger = ::Syslog::Logger.new 'foreman-proxy'
+        rescue
+          puts "Falling back to STDOUT logger"
```
dmitri-d: This is going to be printed out every time logger is used when Syslog::Logger isn't available.

lib/proxy/log.rb
```diff
@@ -13,7 +20,18 @@ def logger
...
+          logger = ::Logger.new(STDOUT)
+        end
       else
         # We keep the last 6 10MB log files
         logger = ::Logger.new(log_file, 6, 1024*1024*10)
```
dmitri-d: I find logger initialization logic is a bit convoluted now, in part because `Log#logger` deals with both initialization and instantiation.  I think it's worth breaking out initialization into a dedicated method that will be executed once, on startup (will have to rely on some dynamic programming or lambdas). This would allow to cleanly deal with config errors or fallbacks.

lib/proxy/log.rb
```diff
@@ -13,7 +20,18 @@ def logger
...
+          puts "Settings log_file=STDOUT and daemon=true are incompatible, exiting..."
+          exit 1
+        end
         logger = ::Logger.new(STDOUT)
+      elsif log_file.upcase == 'SYSLOG'
+        begin
+          logger = ::Syslog::Logger.new 'foreman-proxy'
+        rescue
+          puts "Falling back to STDOUT logger"
+          logger = ::Logger.new(STDOUT)
```
dmitri-d: Also, I think the fallback logger should to be to a file-based one, not STDOUT.

lib/smart_proxy.rb
```diff
@@ -136,24 +128,53 @@ def load_ssl_certificate(path)
...
+      when :dead
+        File.delete(pid_path)
+      end
+    end
+
+    def write_pid
+      FileUtils.mkdir_p(File.dirname(pid_path)) unless File.exist?(pid_path)
+      File.open(pid_path, ::File::CREAT | ::File::EXCL | ::File::WRONLY){|f| f.write("#{Process.pid}") }
+      at_exit { File.delete(pid_path) if File.exist?(pid_path) }
+    rescue Errno::EEXIST
```
dmitri-d: Why do we need this check?

lib/smart_proxy.rb
```diff
@@ -136,24 +128,53 @@ def load_ssl_certificate(path)
...
+        write_pid
+      end
 
       t1 = Thread.new { https_app.start } unless https_app.nil?
       t2 = Thread.new { http_app.start } unless http_app.nil?
 
-      sleep 5 # Rack installs its own trap; Sleeping for 5 secs insures we overwrite it with our own
-      trap(:INT) do
-        exit(0)
-      end
```
dmitri-d, line 155: This bit is needed as rack installs its own interrupt handler on startup. We are potentially running two instances of rack apps, in which case rack will install two interrupt handlers, one of them overwriting another. This leads to errors like this:

```
[2015-08-18 11:52:24] INFO  WEBrick::HTTPServer#start done.
^CE, [2015-08-18T11:52:24.830610 #11508] ERROR -- : Error during startup, terminating. undefined method `shutdown' for nil:NilClass
D, [2015-08-18T11:52:24.830832 #11508] DEBUG -- : undefined method `shutdown' for nil:NilClass:/home/wb/.rvm/gems/ruby-1.8.7-head@smart-proxy/gems/rack-1.5.5/lib/rack/handler/webrick.rb:25:in `shutdown'
/home/wb/.rvm/gems/ruby-1.8.7-head@smart-proxy/gems/rack-1.5.5/lib/rack/server.rb:258:in `start'
/home/wb/sandbox/smart-proxy/lib/smart_proxy.rb:178:in `call'
/home/wb/sandbox/smart-proxy/lib/smart_proxy.rb:178:in `join'
/home/wb/sandbox/smart-proxy/lib/smart_proxy.rb:178:in `launch'
bin/smart-proxy:6
Errors detected on startup, see log for details. Exiting: undefined method `shutdown' for nil:NilClass
```

dmitri-d, line 155: The delay is needed to ensure that rack finished its initialization before we install our own interrupt handler.

lib/proxy/log.rb
```diff
@@ -13,7 +20,18 @@ def logger
     def self.logger
       log_file = ::Proxy::SETTINGS.log_file
       if log_file.upcase == 'STDOUT'
+        if SETTINGS.daemon
+          puts "Settings log_file=STDOUT and daemon=true are incompatible, exiting..."
+          exit 1
+        end
         logger = ::Logger.new(STDOUT)
+      elsif log_file.upcase == 'SYSLOG'
```
dmitri-d, line 33: Should probably mention `SYSLOG` as a possible value in the example config file.

lib/proxy/log.rb
lzap: I agree, let me try to push fixed version (all your remarks above fixed) and we will see. I'd like to focus on the daemonizing part in this patch, touching this only because there is no way to log things while in daemon mode. That's the only reason why I include this work in the patch.

lib/smart_proxy.rb
lzap: We want to error out nicely when we try to spawn the proxy twice. In the past, we tried to run it and net stack errored out with "Port already bind" or something like that. This is definitely more sane.

lib/smart_proxy.rb
lzap, line 155: Ok, quite nasty. But as I understand this only happens on older version of Rack: https://github.com/rack/rack/blob/master/lib/rack/server.rb#L288-L294

I will put this back only when older version of Rack is found.

lzap, line 155: FYI looks like we need this code if rack_version < 1.6.4

https://github.com/rack/rack/commit/e516d89ffcdad1c9d58432aaaff4a382ed3997e7

lzap:
Pushed.

lib/smart_proxy.rb
dmitri-d, line 155: That seems to work, although I have no idea why (signal handler only stops webrick in one thread, somehow the other thread gracefully exits too). A concurrency issue maybe?

lib/smart_proxy.rb
dmitri-d: Ahh, so this is to handle a concurrent launch of multiple instances of smart-proxy? Could you add a comment here explaining this pls?

dmitri-d:
Looks good; I'd love to see tests for `Proxy#pid_status`, `Proxy#check_pid`, and `Proxy#write_pid` methods.

lib/smart_proxy.rb
lzap: Sure, adding tests as well.

lzap:
> Looks good; I'd love to see tests for `Proxy#pid_status`, `Proxy#check_pid`, and `Proxy#write_pid` methods.

Tests done. Added the environment variable for log_file.

## 

Later,
 Lukas #lzap Zapletal

lib/proxy/log.rb
```diff
@@ -10,13 +17,27 @@ def logger
...
     end
 
+    def self.default_logger(log_file)
+      # We keep the last 6 10MB log files
+      ::Logger.new(log_file, 6, 1024*1024*10)
+    end
+
     def self.logger
-      log_file = ::Proxy::SETTINGS.log_file
+      log_file = ENV['SMART_PROXY_LOGFILE'] || ::Proxy::SETTINGS.log_file
```
dmitri-d: Do we need environment override in production (I don't think so)? In which case, could you move this initialization to https://github.com/theforeman/smart-proxy/blob/develop/lib/smart_proxy_for_testing.rb#L20 ?

test/launcher_test.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  def test_pid_status_running
+    @launcher.write_pid
+    assert_equal :running, @launcher.pid_status
+  ensure
+    FileUtils.rm_f @launcher.pid_path
+  end
+
+  def test_write_pid_errj
```
dmitri-d: I think the name is misleading -- you are testing if check_pid forces exit if status is `:running`

test/launcher_test.rb
```diff
@@ -0,0 +1,36 @@
...
+    @launcher.write_pid
+    assert_raises SystemExit do
+      @launcher.check_pid
+    end
+  ensure
+    FileUtils.rm_f @launcher.pid_path
+  end
+end
```
dmitri-d, line 42: Could you add a test for `:dead` status -- both for `pid_status` and `check_status` methods?

lib/proxy/log.rb
lzap: > Do we need environment override in production (I don't think so)? In which case, could you move this initialization to https://github.com/theforeman/smart-proxy/blob/develop/lib/smart_proxy_for_testing.rb#L20 ?

Yeah, I will remove it. It's problematic, there are tests failing when
this is not log/test.log...

## 

Later,
 Lukas #lzap Zapletal

test/launcher_test.rb
lzap: > I think the name is misleading -- you are testing if check_pid forces exit if status is `:running`

This was a typo. Corrected to `test_check_pid_exits_program`

## 

Later,
 Lukas #lzap Zapletal

test/launcher_test.rb
lzap, line 42: > Could you add a test for `:dead` status -- both for `pid_status` and `check_status` methods?

Done!

## 

Later,
 Lukas #lzap Zapletal

test/launcher_test.rb
```diff
@@ -0,0 +1,42 @@
...
+    @launcher.write_pid
+    assert_equal :running, @launcher.pid_status
+  ensure
+    FileUtils.rm_f @launcher.pid_path
+  end
+
+  def test_check_pid_deletes_dead
+    Process.stubs(:kill).returns { raise Errno::ESRCH }
+    @launcher.check_pid
+    refute File.exist?(@launcher.pid_path)
```
dmitri-d: no `#refute` method on 1.8.7 (the reason for test failures).

lzap: > no `#refute` method on 1.8.7 (the reason for test failures).

Oh fixed, thanks for hint!

## 

Later,
 Lukas #lzap Zapletal

dmitri-d:
merged in at 335a161. Thanks, @lzap!

lzap:
Thanks for review, appreciated!

PR #313, opened 2015-08-28
RFC: some fixes required to work with Puppet 4.0 / Puppet Server

damoxc:
Despite finding http://projects.theforeman.org/issues/8447 I decided to try and get Foreman to work using Puppet 4, installing it manually.

So far these are the only changes I've needed to get the import to work, aside from updating paths in configuration files. Looking for feedback / thoughts more than anything else.

### DIFF
--- a/modules/puppet_proxy/class_scanner_eparser.rb
+++ b/modules/puppet_proxy/class_scanner_eparser.rb
@@ -44,7 +44,11 @@ if Puppet::PUPPETVERSION.to_f >= 3.2
       end
 
       def do_find ast
-        @@finder_visitor.visit_this(self, ast)
+        if Puppet::PUPPETVERSION.to_i < 4
+          @@finder_visitor.visit_this(self, ast)
+        else
+          @@finder_visitor.visit_this(self, ast, [])
+        end
       end
 
       def find_HostClassDefinition o
--- a/modules/puppet_proxy/environment.rb
+++ b/modules/puppet_proxy/environment.rb
@@ -146,7 +146,7 @@ class Proxy::Puppet::Environment
   def classes
     ::Proxy::Puppet::Initializer.load
     conf = ::Proxy::Puppet::ConfigReader.new(::Proxy::Puppet::Initializer.config).get
-    eparser = (conf[:main] && conf[:main][:parser] == 'future') || (conf[:master] && conf[:master][:parser] == 'future')
+    eparser = (conf[:main] && conf[:main][:parser] == 'future') || (conf[:master] && conf[:master][:parser] == 'future') || Puppet::PUPPETVERSION.to_i >= 4
     scanner = ::Proxy::Puppet::ClassScannerFactory.new(eparser).scanner
 
     paths.map {|path| scanner.scan_directory(path, name) }.flatten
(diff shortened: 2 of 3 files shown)


mmoll:
Very nice, just out of curiosity, is this on Fedora with native (non-AIO) packages or a similar setup? IIRC it has to be the same Ruby used by Puppet (agent in that case) and foreman-proxy.

dmitri-d:
[test]

damoxc:
This is on Ubuntu 14.04 using the packages available in the PuppetLabs PC1 repository. I have modified my environment so the ruby installed by the PL packages is being used (which other than requiring some `bundle install` or `gem install` to fix dependencies hasn't caused any issues).

mmoll:
I see. This is pretty great for me, because I don't want to use PC1 packages anyway, so Puppet will use the system Ruby like the proxy.

dmitri-d:
@damoxc: thanks for the PR. Changing the commit message to "Fixes #<issue_number>: short description" format should make the bot happy.

This PR breaks a whole bunch of tests when puppet > 4.0 is used (13 failures on my machine). You can run tests by executing `rake test` on your machine. I need to sort out puppet 4.0 install to test it too, I'm having some issues with that atm :(.

modules/puppet_proxy/class_scanner_eparser.rb
```diff
@@ -44,7 +44,11 @@ def initialize
       end
 
       def do_find ast
-        @@finder_visitor.visit_this(self, ast)
+        if Puppet::PUPPETVERSION.to_i < 4
+          @@finder_visitor.visit_this(self, ast)
+        else
+          @@finder_visitor.visit_this(self, ast, [])
+        end
```
dmitri-d, line 51: This conditional can be removed alltogether if you: change the constructor to something like:

```
      def initialize
        @klasses = []
        @finder_visitor = Puppet::Pops::Visitor.new(self,'find',0,0)
      end
```

and then update `do_find` method to:

```
      def do_find ast
        @finder_visitor.visit(ast)
      end
```

Note that I replaced class-variable `@@finder_visitor` with an instance variable `@finder_visitor`.

dmitri-d:
[test]

mmoll:
[test]

modules/puppet_proxy/class_scanner_eparser.rb
damoxc, line 51: Good call, will change that.

damoxc:
@witlessbird I've already modified the commit message to the bot's liking (I think), repushed as c2fde0c.

I'll take a look at the tests now. What issue are you having installing Puppet 4?

dmitri-d:
@damoxc: I think further changes are required to support puppet 4.0. Looks like both url and json for environments api changed starting with Puppet 4.0. I'm going to take a look into this tomorrow.

damoxc:
@witlessbird Ah yes, it has, the Puppet Server does implement a backwards compatible API for 3.X clients so that is probably why it is still working in my set up.

dmitri-d:
@damoxc: I'm going to close this PR in favour of https://github.com/theforeman/smart-proxy/pull/317. It's based on your work, but also intrduces changes to make various puppet configuration easier to test.

PR #314, opened 2015-09-02
fixes #11654: no longer rely on rubygems to determine rack version

dmitri-d:


dLobatog:
Merged as d00dc5860000839c21fdae3361ed230216e15243, thanks @witlessbird !

PR #315, opened 2015-09-02
fixes #11653: only display warning message about unavailability of 'SYSLOG' when it was configured

dmitri-d:


lzap:
Bringing issue discussion here:

> The smart proxy now warns about lack of syslog on some platforms on every startup, which is overly verbose, since #11323.
> 
> a) Formatting is messy with init scripts

We either go the more transprent way (which I personally prefer) or log the warning in the fallback file-based logger. This does not change this behavior if I read the code correctly.

> b) "Ignoring" implies that it was asked to use syslog, but it isn't configured for that (it has the default file).

Appears to be fixed.

> c) In a default configuration on Ruby 1.9.3, it's too verbose.

Not sure what you mean by that.

domcleal:
> > c) In a default configuration on Ruby 1.9.3, it's too verbose.
> 
> Not sure what you mean by that.

That out of the box with the default settings file (which does _not_ use syslog) under Ruby 1.9.3 it was issuing the warning on startup.  This looks to be fixed.

lzap:
Oh yeah, but the output can still break the service command output I guess.

lzap:
Merged as e0497ca9b3ab7d0be272847c2cbdae9850cc6e1d, thank you!

dmitri-d:
I wasn't sure re: showing the warning on STDOUT, but it's a trivial change to switch to use a logger instead.

PR #316, opened 2015-09-03
Fixes #11676 - added dhcp_server option

lzap:


dLobatog:
Looks fine, but I guess this needs an accompanying PR on puppet-foreman_proxy?

dmitri-d:
Could you update the test for the default configuration (https://github.com/theforeman/smart-proxy/blob/develop/test/dhcp/dhcp_config_test.rb) and add a test for `Proxy::DHCP::ISC#omcmd` method to verify that the `name` server attribute is being used.

lzap:
I don't see any value in testing every single possible default value for each plugin. This should be tested exactly once with exactly one default pair. Added anyway because its just one line, but I'd rather delete them all. Will add test for omcmd.

lzap:
Ok our `omshell` unit test coverage is none, I don't see a way how to verify the name without writing whole stubbing for that.

dmitri-d:
I think it's important to verify that the default values that we assume were defined are actually defined: this allows for us to get rid of nil-, default-value checks and setting initializations without getting various nil-value associated failures.

dmitri-d:
One approach would be to stuff individual lines of an omshell session into a list, and output them during the execution. A test could verify that the list contains stuff we expect.

lzap:
Done.

dmitri-d:
merged in at 77e0262. Thanks, @lzap!

PR #317, opened 2015-09-09
fixes #11599: introduced support for puppet 4.0

dmitri-d:
These changes are based on Damien Churchil's PR


dmitri-d:
This is based on https://github.com/theforeman/smart-proxy/pull/313. In addition to adding support for puppet 4.0, this PR:
- extracts all logic used to determine which Puppet interface to use into one place
- extracts various ways of interfacing with Puppet into dedicated classes
- correct Puppet interface is "wired" during module initialization now
- expands tests for around module configuration and puppet interfaces

dmitri-d:
For reviewers: I'm aware of failures on puppet 2.6, looking into this.

mmoll:
Maybe 2.6 support could be dropped?

dmitri-d:
Why use dependency injection?

The logic behind figuring out puppet configuration (old parser or future parser? config file environments or api?) got complicated enough and spread through a few methods/classes so that testing became harder than it should be. One of possible approaches in such a case is to create abstractions that handle specfic cases (we were mostly there), and then “wire” them up on startup. This dramatically simplifies testing: the dependency objects can now be tested in isolation, the rest can be tested using test doubles (mocks are dedicated test implemntations).

GregSutcliffe:
FYI that was my idea for some historical footnote on the use of injection, just because it might help next time we're debugging this :) - thanks @witlessbird!

lzap:
This also fixed issues with unit tests when puppet was not present on the system. Thanks.

modules/puppet_proxy/puppet_plugin.rb
```diff
@@ -1,10 +1,22 @@
...
-                     :salt_puppetrun_cmd => 'puppet.run'
+    default_settings :puppet_provider => 'puppetrun', :puppet_conf => '/etc/puppet/puppet.conf', :use_cache => true,
+                     :salt_puppetrun_cmd => 'puppet.run', :puppet_ssl_ca => '/var/lib/puppet/ssl/certs/ca.pem'
+
     plugin :puppet, ::Proxy::VERSION
+
+    validate_readable :puppet_conf
+
+    after_activation do
+      ::Proxy::Puppet::Initializer.reset_puppet
```
GregSutcliffe: Should be `.new.reset_puppet`, it's not a class method

modules/puppet_proxy/puppet_cache.rb
```diff
@@ -1,16 +1,17 @@
...
 module Proxy::Puppet
   class PuppetCache
     include Proxy::Log
+    extend Proxy::Puppet::DependencyInjection::Injectors
 
-    attr_reader :class_scanner
+    inject_attr :puppet_class_scanner_impl, :puppet_class_scanner
 
-    def initialize(class_scanner, classes_store, timestamps_store)
+    def initialize(classes_store = MemoryStore.new, timestamps_store = MemoryStore.new)
```
GregSutcliffe: This needs to be an explicit call to Proxy::MemoryStore, or you get "unitialized constant"

dmitri-d:
@GregSutcliffe: fixed the issues you pointed out.

GregSutcliffe:
Merged as 763d7ae, thanks @witlessbird, and @lzap for extra testing!

dmitri-d:
And thanks to @damoxc for the original work.

damoxc:
Thanks for your work everyone :smile:

lzap:
This patch stopped discovery image builds nightly from working, it errors out at https://github.com/theforeman/smart-proxy/blob/develop/modules/puppet_proxy/initializer.rb#L1 because puppet is not installed on the image. This line has not changed here, but is it expected to insist on puppet to be installed? It used to work, image only has "facter" installed.

dmitri-d:
@lzap: https://github.com/theforeman/smart-proxy/pull/321 fixes the issue.

PR #320, opened 2015-09-29
Fixes #11997 - DHCP option for NX-OS should be quoted

fernandocarolo:
This fixes the lack of quotes for DHCP option `tftp-server-name`, when a Cisco NX-OS switch is being provisioned via POAP.


dmitri-d:
merged in at f868365. Thanks, @fernandocarolo!

PR #322, opened 2015-09-30
fixes #12016 - this is for discussion

helge000:
Just to start discussion. IMHO we should use `STDOUT` as default, simply because it works everywhere.

### DIFF
--- a/lib/proxy/settings/global.rb
+++ b/lib/proxy/settings/global.rb
@@ -3,7 +3,7 @@ module ::Proxy::Settings
     DEFAULT_SETTINGS = {
       :settings_directory => Pathname.new(__FILE__).join("..","..","..","..","config","settings.d").expand_path.to_s,
       :https_port => 8443,
-      :log_file => "/var/log/foreman-proxy/proxy.log",
+      :log_file => ::PLATFORM =~ /mingw/ ? "STDOUT" : "/var/log/foreman-proxy/proxy.log",
       :log_level => "ERROR",
       :daemon => false,
       :daemon_pid => "/var/run/foreman-proxy/foreman-proxy.pid",


dmitri-d:
This is done via `log_file` setting (https://github.com/theforeman/smart-proxy/blob/develop/config/settings.yml.example#L55): use `STDOUT` to redirect output to the console. I haven't tested this under windows however.

helge000:
I know, `STDOUT` works well enough on windows (as anywhere; you are using logger). I created the issue about the defaults used. This one prevents startup on windows; so why use a special static path as default?

dmitri-d:
I don't think `STDOUT` is useful in production, the option is there for development environments only. I don't like conditional default values either, but I can be persuaded to use platform-specific sets, if there are enough differences.

Alternatively, we could point to a `log/` subdirectory inside smart-proxy source directory, but I'll have to check our packaging scripts to see if they would need changes.

@domcleal, @GregSutcliffe: thoughts?

helge000:
@witlessbird I only wanted to start a discussion about some defaults :)
@domcleal, @GregSutcliffe : I also think `:bind_host => "*"` is a non-optimal choice; we should use `:bind_host => "::"`?

dmitri-d:
`::` for `bind_host` parameter is IPv6 specific, I don't think we are quite there yet.

helge000:
@witlessbird This is not quite correct, because by default all ipv6 ports get mapped to ipv4 also (unless someone turns on `sysctl net.ipv6.bindv6only`). 

If IPv4 only is preferred, a better default would be `"0.0.0.0"`. `"*"` is error prone and will lead to `Exiting: getaddrinfo: No such host is known` (at least on windows, but I suspect many platforms do not like this as well).

mmoll:
`"::"` is IPv6 only on FreeBSD, although I have to state that `"*"` is also not working there as well.

dmitri-d:
@lzap: do you remember why we ended up using '*' instead of '0.0.0.0'?

domcleal:
https://github.com/theforeman/smart-proxy/pull/245#issuecomment-70090771 explains, but @helge000's point about using :: might be appropriate as long as a system always has an IPv6 stack available.  (You can disable the IPv6 module, I'm not sure if that will affect the behaviour of the bind to `::`.)

helge000:
@domcleal: sadly, the behavior varies by os. As by [rfc3493](http://tools.ietf.org/html/rfc3493#page-22) ipv6 only sockets can be defined and some operating systems treat this quite differently or various reasons (see below).

I cannot find the RFC requiring IPv6 addressing to be available (try to turn it off in Linux...) but I think we can safely assume at least `::1/128` is available in every supported os running smart-proxy.
This is why I think `::` a good default on the long run, as it at least works on any OS; even if it is only listening on `::1/128`. 
This is better then crashing on some operating systems as the current default `*` does. On most systems, `::` will listen on any available IPv4 and IPv6 address.

[Wikipedia](https://en.wikipedia.org/wiki/IPv6) sums it up quite nicely:

> Some common IPv6 stacks do not implement the IPv4-mapped address feature, either because the IPv6 and IPv4 stacks are separate implementations (e.g., Microsoft Windows 2000, XP, and Server 2003), or because of security concerns (OpenBSD). On these operating systems, a program must open a separate socket for each IP protocol it uses. On some systems, e.g., the Linux kernel, NetBSD, and FreeBSD, this feature is controlled by the socket option IPV6_V6ONLY, as specified in RFC 3493.

mmoll:
As I'm running again into the bind_host thing on FreeBSD, are there any objections against setting it to 0.0.0.0? If no, I'd open PRs to -proxy and its puppet module.

domcleal:
@mmoll is `::` OK for FreeBSD?  If so then @helge000's suggestion would be preferable as it'll listen on IPv6 too.

mmoll:
`::` is _only_ IPv6 there. As sad as it is, I think `0.0.0.0` is the best default at the moment :(

domcleal:
Perhaps we need per-OS defaults in the code then - use a special keyword in the file and change the value based on platform.

dmitri-d:
Closing this PR as it's been inactive for over a year.

PR #323, opened 2015-10-11
Fixes #12138 Add require statements

ekohl:
Adding these allows smart_proxy to be run from bundler.

### DIFF
--- a/lib/smart_proxy.rb
+++ b/lib/smart_proxy.rb
@@ -30,7 +30,10 @@ require 'proxy/request'
 require 'bundler_helper'
 Proxy::BundlerHelper.require_groups(:default)
 
+require 'json'
+require 'rack'
 require 'rack-patch' if Rack.release < "1.3"
+require 'sinatra'
 require 'sinatra-patch'
 require 'sinatra/authorization'
 require 'poodles-fix'
--- a/lib/smart_proxy_for_testing.rb
+++ b/lib/smart_proxy_for_testing.rb
@@ -1,5 +1,8 @@
 require 'proxy'
 require 'checks'
+require 'json'
+require 'rack'
+require 'sinatra'
 require 'proxy/log'
 require 'proxy/settings'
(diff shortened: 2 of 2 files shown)


ekohl:
I still have some issues. In particular the `require 'rake'` in the gemspec doesn't work on ruby 1.8.7. I also wonder why `s.default_executable = 'bin/smart_proxy.rb'` is there since that file doesn't exist.

domcleal:
> In particular the require 'rake' in the gemspec

I can't see why it'd be needed anyway, remove it?

dmitri-d:
@ekohl: could you file a ticket and explain the problem in there? At the moment  https://github.com/theforeman/smart-proxy/blob/develop/lib/smart_proxy.rb#L31 loads main dependencies which include rack and sinatra, and should be sufficient.

ekohl:
I created http://projects.theforeman.org/issues/12138 and will update the commit once I know what the correct solution is.

dmitri-d:
@ekohl: After thinking about this, I realised that your approach is a good way to resolve the issue. Could you add requires for 'json', 'sinatra', and 'rack' somewhere around here: https://github.com/theforeman/smart-proxy/pull/323/files#diff-283e99ae18637863a4569b30d7d62f7eR30 and https://github.com/theforeman/smart-proxy/blob/develop/lib/smart_proxy_for_testing.rb#L3. I'll clean up `Gemfile` and gemspec in a seprate PR.

dmitri-d:
@ekohl: or were you asking about https://github.com/theforeman/smart-proxy/pull/323/files#diff-283e99ae18637863a4569b30d7d62f7eR30?

dmitri-d:
[test]

dmitri-d:
@ekohl: ping?

ekohl:
@witlessbird something like this?

dmitri-d:
merged in at c9f0a07. Thanks, @ekohl!

PR #325, opened 2015-10-13
Fixes #12146: moved core dependencies into gemspec

dmitri-d:
depends on https://github.com/theforeman/smart-proxy/pull/323


smart_proxy.gemspec
```diff
@@ -9,19 +9,18 @@ Gem::Specification.new do |s|
...
   s.require_paths = ["lib", "modules"]
   s.test_files = FileList["{test}/**/*test.rb"].to_a
   s.license = 'GPLv3'
   s.has_rdoc = true
   s.extra_rdoc_files = ["README.md"]
-  s.add_dependency 'json', '~> 1.8'
-  s.add_dependency 'rack', '~> 1.5'
-  s.add_dependency 'sinatra', '~> 1.4'
+  s.add_dependency 'json', '~> 0'
+  s.add_dependency 'rack', '>= 1.1', '< 1.6'
```
mmoll: BTW, -proxy is working fine for me with rack 1.6.4 (default version on FreeBSD)

dmitri-d: Yeah, I think this can be relaxed.

dmitri-d:
@mmoll: fixed rack version spec.

ekohl:
Is the `require 'rake'` as the first line really needed?

dmitri-d:
@ekohl: this allows us to override rack version specified by sinatra (were you asking about https://github.com/theforeman/smart-proxy/pull/325/files#diff-350f81aa53246165f54f9fca18c224acR19 ?).

ekohl:
No, I was asking about https://github.com/theforeman/smart-proxy/blob/5f36f2e95761ca993fec96e6528c48caa5f8384d/smart_proxy.gemspec#L1

dmitri-d:
@ekohl: ahh, sorry -- we use `Rack::FileList` (https://github.com/theforeman/smart-proxy/pull/325/files#diff-350f81aa53246165f54f9fca18c224acR11) in `smart_proxy.gemspec`, which is defined in rake.

ekohl:
@witlessbird Ok. I did have an issue on ruby 1.8.7 that it couldn't load due to rake missing, but first I'll try to get some other failures fixed and then see if I can still reproduce it.

domcleal:
Having Gemfile use the gemspec, but the gemspec depend on rake is going to introduce a bit of a circular dependency as there's no guarantee that in a clean environment, rake will be available without having run bundler.  Can we replace the rake usage with Dir or something?

Also setting WoC until tests pass.

dmitri-d:
Removed gemspec's file dependency on rake. It looks like that sinatra (rack?) logger is being initialized differently when dependencies are defined in the gemspec (I'm seeing info-level logs on the console when executing tests), but it appears to be benign (if annoying). Will investigate this later.

domcleal:
Yes, I see the same.  Also interesting is that there are warnings for already-initialised constants just before the Sinatra logs:

<pre>
.........................../home/dcleal/code/foreman/smart-proxy/lib/smart_proxy.rb:42: warning: already initialized constant Proxy::SETTINGS
/home/dcleal/code/foreman/smart-proxy/lib/smart_proxy_for_testing.rb:26: warning: previous definition of SETTINGS was here
/home/dcleal/code/foreman/smart-proxy/lib/smart_proxy.rb:43: warning: already initialized constant Proxy::VERSION
/home/dcleal/code/foreman/smart-proxy/lib/smart_proxy_for_testing.rb:27: warning: previous definition of VERSION was here
</pre>

It appears that smart_proxy.rb is being loaded during rake test, while it usually doesn't.

<pre>
/home/dcleal/code/foreman/smart-proxy/lib/smart_proxy.rb:2:in `<top (required)>'
/home/dcleal/.rvm/gems/ruby-2.1.0@global/gems/bundler-1.5.2/lib/bundler/runtime.rb:76:in `require'
/home/dcleal/.rvm/gems/ruby-2.1.0@global/gems/bundler-1.5.2/lib/bundler/runtime.rb:76:in `block (2 levels) in require'
/home/dcleal/.rvm/gems/ruby-2.1.0@global/gems/bundler-1.5.2/lib/bundler/runtime.rb:72:in `each'
/home/dcleal/.rvm/gems/ruby-2.1.0@global/gems/bundler-1.5.2/lib/bundler/runtime.rb:72:in `block in require'
/home/dcleal/.rvm/gems/ruby-2.1.0@global/gems/bundler-1.5.2/lib/bundler/runtime.rb:61:in `each'
/home/dcleal/.rvm/gems/ruby-2.1.0@global/gems/bundler-1.5.2/lib/bundler/runtime.rb:61:in `require'
/home/dcleal/.rvm/gems/ruby-2.1.0@global/gems/bundler-1.5.2/lib/bundler.rb:131:in `require'
/home/dcleal/code/foreman/smart-proxy/lib/bundler_helper.rb:25:in `require_groups'
/home/dcleal/code/foreman/smart-proxy/lib/proxy/plugin.rb:136:in `configure_plugin'
/home/dcleal/code/foreman/smart-proxy/test/plugin_test.rb:32:in `test_enabled_plugins'
</pre>

dmitri-d:
Update: looks like when 'gemspec' line is used in `Gemfile`, bundler lists current project as a non-rubygems dependency in `Gemfile.lock` and adds it to 'default' group. Which is how `lib/smart_proxy` ends up being loaded (as we load all gems in 'default' group when initializing plugins). Renaming of `lib/smart_proxy.rb` stops automatic loading (naturally) of smart-proxy runtime. I'll do the renaming and push the update.

dmitri-d:
renamed `lib/smart_proxy.rb` to `lib/smart_proxy_main.rb` to avoid automatic loading by bundler.

smart_proxy.gemspec
```diff
@@ -8,20 +6,18 @@ Gem::Specification.new do |s|
...
+  s.test_files = Dir.glob("{test}/**/*test.rb")
   s.license = 'GPLv3'
   s.has_rdoc = true
   s.extra_rdoc_files = ["README.md"]
-  s.add_dependency 'json', '~> 1.8'
-  s.add_dependency 'rack', '~> 1.5'
-  s.add_dependency 'sinatra', '~> 1.4'
-  s.rubyforge_project = 'rake'
+  s.add_dependency 'json' 
+  s.add_dependency 'rack', '>= 1.5.4', '< 1.6' # ruby 1.8.7 support is broken in rack 1.6 versions < 1.6.4
```
domcleal: We still support 1.8.7 though, so I think this needs to permit all the way down to 1.1 as the Gemfile did before.  (Our packages use this for instance, though admittedly I don't think they'll care about the gemspec.)

smart_proxy.gemspec
```diff
@@ -8,20 +6,18 @@ Gem::Specification.new do |s|
...
+  s.add_dependency 'sinatra', '>= 1.4.6'
```
domcleal: Why the version spec here?  The Gemfile used to be >= 0.

domcleal:
This looks OK, but I see one plugin that mucks around with loading the smart proxy code externally to the smart proxy.  If you get the chance, you may want to send them a PR to fix this:

https://github.com/theforeman/smart_proxy_abrt/blob/master/bin/smart-proxy-abrt-send#L12

(Not a merge blocker.)

dmitri-d:
@domcleal: fixed the two things you pointed out. Will take a look at abrt plugin sometime today.

domcleal:
Merged as bf8936e, thanks @witlessbird.

PR #327, opened 2015-10-14
Fixes #12152: support for ruby 2.2.x

dmitri-d:


dmitri-d:
@domcleal: this should fix build issues on ruby 2.2.x

domcleal:
I'm running the tests manually via http://ci.theforeman.org/job/test_proxy_develop_pull_request/869/ with 2.2 in the matrix.

domcleal:
http://ci.theforeman.org/job/test_proxy_develop_pull_request/870/ tests 57f8dae.

bundler.d/test.rb
```diff
@@ -6,6 +6,7 @@
   gem 'ci_reporter', '>= 1.6.3', "< 2.0.0", :require => false
   gem 'rdoc'
   gem 'minitest', '~> 4.7', :platforms => :ruby_19
```
domcleal: Should we also replace minitest with test-unit on Ruby 1.9 for consistency?

As I understand it (http://projects.theforeman.org/issues/2261), we only have this in here for Ruby 1.9.2.  It might be that we can use test-unit on 1.9.

domcleal: And I suppose on Ruby 2.0 and 2.1, we're actually using minitest from Ruby core, with its test-unit compatibility.  Should it just be test-unit throughout?

dmitri-d: I'll try switching to test-unit for all versions of ruby.

dmitri-d:
@domcleal: seems to be happy with test-unit: http://ci.theforeman.org/job/test_proxy_develop_pull_request/871/

GregSutcliffe:
Merged as b617b3f to clear CI blocks, thanks @witlessbird and @domcleal for testing

PR #312, opened 2015-08-27
Fixes #11081 - broke out dhcp providers into modules

dmitri-d:

PR #329, opened 2015-10-19
Fixes #12209: introduced basic support for di in dns module

dmitri-d:


dmitri-d:
This is to support changes to powerdns module: https://github.com/theforeman/smart_proxy_dns_powerdns/pull/8

modules/dns/dns.rb
```diff
@@ -1,20 +1 @@
 require 'dns/dns_plugin'
-
-module Proxy::Dns
```
domcleal, line 3: The removal of these is going to cause a compatibility issue for the DNS plugins between 1.10 and 1.11.  Do they need to be moved into dns_common?

domcleal:
So is the intention for the plugin to extend the new Proxy::Dns::DependencyInjection::Dependencies class when loaded by declaring new dependencies which can be injected later?

Or would it create and load its own Dependencies class, which makes me wonder why the empty Proxy::Dns::DependencyInjection::Dependencies one is needed.

modules/dns/dns.rb
dmitri-d, line 3: I would very much like to do that, as otherwise we'll have weird dependency graph -- some things are going to depend on stuff in `dns_common`, others in `dns`. It wouldn't cause issues with dependency loading, but might cause confusion with locating of common dependencies (or require explanation why things are this way). I'll do a pr for route53, I assume @ekohl is ok doing the changes for powerdns?

dmitri-d:
@domcleal: It's neither. `Proxy::Dns::DependencyInjection::Container` is a singleton: `Container.instance` will return the same instance one it's been initialized. A shared instance is needed as the main module uses dependencies defined in providers. 
I also wanted to use a container per module -- to avoid specifying target container when resolving dependencies and to enforce isolation between modules. This coupled with how ruby parses classes and modules resulted in this bunch of boiler-plate code when defining a container. I'm not very fon dof it, but it's the best solution I could find.

modules/dns/dns.rb
domcleal, line 3: OK, fine by me if you want to change it - then my corresponding comment in the DHCP PR is all the more important, let's make sure we get it correct this second time.

ekohl, line 3: If we're going to break the API, I wonder if we shouldn't do some bigger refactoring. I'd very much like https://github.com/ekohl/smart_proxy_dns_powerdns/blob/bd5f16c189fab2f4f3e03c7fb08ac8f893484ff9/lib/smart_proxy_dns_powerdns/dns_powerdns_main.rb#L36-L48 to be moved into the core.

dmitri-d, line 3: Sry, which comment did you mean?

domcleal, line 3: https://github.com/theforeman/smart-proxy/pull/312#discussion_r44404640

domcleal:
> A shared instance is needed as the main module uses dependencies defined in providers.

I don't see where this is happening, can you show me please?

dmitri-d:
@domcleal: apologies, I forgot I decided to switch to di for provider resolution in dns module in a separate PR. Here's an example from dhcp module: https://github.com/theforeman/smart-proxy/pull/312/files#diff-03da59381a0cb4607cd2706060ae0310R9. `dhcp_provider` is wired up by the provider that was specified by the configuration.

domcleal:
Yeah, that's what I was beginning to suspect as well.. do you plan to include that in this PR?

dmitri-d:
Well, I wanted to keep it simple(er) and short(er) to make it easier to review. But personally I would prefer to bundle the disruptive changes in one PR. So sure, I can do that.

domcleal:
I don't mind either way really, but I agree with @ekohl that if we're going to make disruptive changes, let's try and get them done in one release cycle.  1.10's done and dusted by now, but we can avoid breaking plugins in both 1.11 and 1.12.

dmitri-d:
This is the bulk of changes. Dependency injection is now used for provider resolution. I also added a bunch of tests, but need to add more, tests for the api in particular. @ekohl I looked at https://github.com/ekohl/smart_proxy_dns_powerdns/blob/bd5f16c189fab2f4f3e03c7fb08ac8f893484ff9/lib/smart_proxy_dns_powerdns/dns_powerdns_main.rb#L36-L48 -- it looks pretty specific to power_dns, unless I missed something?

ekohl:
@witlessbird I'd say it's not. If you look at https://github.com/theforeman/smart-proxy/blob/f24be74d8f51f44aa8730769d2e468683804254c/modules/dns_nsupdate/dns_nsupdate_main.rb#L26-L44 then you see nsupdate normalizes directly in the commands.

modules/dns_nsupdate/dns_nsupdate_gss_plugin.rb
```diff
@@ -10,8 +9,11 @@ class Plugin < ::Proxy::Provider
 
     requires :dns, ::Proxy::VERSION
 
+    validate_readable :dns_key, :dns_tsig_keytab
```
domcleal: dns_key shouldn't be required for the GSS variant

modules/dns_virsh/dns_virsh_main.rb
```diff
@@ -7,45 +7,46 @@ class Record < ::Proxy::Dns::Record
...
 
-    def self.record(attrs = {})
-      new(attrs.merge(:virsh_network => ::Proxy::SETTINGS.virsh_network))
-    end
-
-
-    def initialize options = {}
-      @network = options[:virsh_network]
+    def initialize(network = nil)
+      @network = network ||  ::Proxy::SETTINGS.virsh_network
```
domcleal: Nitpick, double space after the || should be one space.

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -1,39 +1,35 @@
...
   class Record < ::Proxy::Dns::Record
     include Proxy::Log
     include Proxy::Util
-    attr_reader :resolver
 
-    def self.record(attrs = {})
-      new(attrs.merge(:server => ::Proxy::Dns::Dnscmd::Plugin.settings.dns_server,
-                      :ttl => ::Proxy::Dns::Plugin.settings.dns_ttl))
+    def initialize(server = nil, ttl = nil)
+      super(server || ::Proxy::Dns::Dnscmd::Plugin.settings.dns_server, ttl || ::Proxy::Dns::Plugin.settings.dns_ttl)
```
domcleal: Nitpick, can you split this into a few lines?  Probably `server ||= ::Proxy....settings..` and same for ttl, it'll be easier to read.

lib/proxy/plugin_validators.rb
```diff
@@ -7,6 +7,7 @@ def initialize(plugin, setting_name)
 
     def validate!
       setting_value = @plugin.settings.send(@setting_name)
+      return true if setting_value.nil?
```
domcleal: Is this a good idea?  If a user doesn't set a required file, it'll pass validation.

modules/dns_nsupdate/dns_nsupdate_gss_main.rb
```diff
@@ -6,17 +6,12 @@ class Record < Proxy::Dns::Nsupdate::Record
     include Proxy::Kerberos
     attr_reader :tsig_keytab, :tsig_principal
 
-    def initialize options = {}
+    def initialize(server = nil, ttl = nil)
       @tsig_keytab = ::Proxy::Dns::NsupdateGSS::Plugin.settings.dns_tsig_keytab
       @tsig_principal = ::Proxy::Dns::NsupdateGSS::Plugin.settings.dns_tsig_principal
-      raise "Keytab not configured via dns_tsig_keytab for DNS GSS-TSIG support" unless tsig_keytab
-      raise "Unable to read dns_tsig_keytab file at #{tsig_keytab}" unless File.exist?(tsig_keytab)
-      raise "Kerberos principal required - check dns_tsig_principal setting" unless tsig_principal
```
domcleal, line 14: I can't see where the tsig_principal check moved to, is it still present?

domcleal:
I'd still like to clarify whether the layout of `dns_common/errors` and `dns_common/record` is what we want to stick with, as the DHCP PR at https://github.com/theforeman/smart-proxy/pull/312#discussion_r44404640 puts both errors and some utility constants into `dhcp_common/dhcp`.

I wonder if we should simply have a `dns_common` that includes both errors and records, which plugins can require so they don't have to worry about the internal layout of the dns_common/ namespace?

lib/proxy/plugin_validators.rb
dmitri-d: It's a bit too clever, which I don't like: my thinking was that this is used together with default setting for the same parameter, and would require a deliberate effort to set the value to nil via config file.  I wonder if supporting a conditional here would be more clear?

domcleal: I think the default value is only present for the GSS keytab setting, but dns_key and probably stuff in other modules (Puppet?) etc may not have defaults, so might they have nil values too?

Remind me, what's it needed for?

modules/dns_nsupdate/dns_nsupdate_gss_main.rb
dmitri-d, line 14: There is a default value for it here: https://github.com/theforeman/smart-proxy/blob/develop/modules/dns_nsupdate/dns_nsupdate_gss_plugin.rb#L9. My reasoning here (and in other places where I removed similar nil checks) is that once default value is defined it's pretty unlikely it will ever be nil. It possible though to make a mistake in the config file (leave option without a value), I should filter out settings pairs without values before merging them with default values (here: https://github.com/theforeman/smart-proxy/blob/develop/lib/proxy/settings/plugin.rb#L10 and here: https://github.com/theforeman/smart-proxy/blob/develop/lib/proxy/settings/global.rb#L30)

domcleal, line 14: I'm not sure we should filter them out when there's no value, it could be unexpected, I think it's preferable to follow what the user entered but to raise an error.

A likely scenario is even in our own installer or config management systems - if a setting isn't initialised or set properly, a config file could be templated with a blank setting, e.g. `:foo: <%= @uninitialised_variable %>`.  It might be safer to let the proxy initialise the setting to nil and throw an error saying that the :foo setting's nil or not a file.

lib/proxy/plugin_validators.rb
dmitri-d: This is to support validation of optional parameters: the check is triggered only when the parameter was initialized.

domcleal: Sorry, I mean which setting or provider's triggering it?  It seems like we have a few settings in providers that could be nil by default.

dmitri-d: dns_key in dns_nsupdate provider: https://github.com/theforeman/smart-proxy/blob/develop/modules/dns_nsupdate/dns_nsupdate_plugin.rb#L6.

domcleal: Ah yes, thanks.  I think that should raise a validation error then if it's not set?

dmitri-d: It's an optional parameter: either we use an implementation similar to the one above (but that raises questions about handling of non-optional parameters), or add support for conditional validators...

domcleal: Or add another validator so we've got one for optional-file validation and one for required-file validation?

dmitri-d: Otoh, if all required (or optional -- don't think tagging both is needed) parameters were tagged in the Plugin class, validators could use that to switch logic accordingly...

domcleal: As an option to the validator, or separately?  It would be easy to pass options through validate_readable.

dmitri-d: I was thinking separately, something like:

```
default_settings :foo => ["value", :required], :bar => "another_value 
```

although I would rename default_settings to settings.

dmitri-d: Or, simply not define default values for optional parameters (which doesn't make sense anyway). In which case: I need an implicit presence validator for all default settings; For all other validators,if a parameter isn't among default settings, it should be treated as optional.

This would require documentation, but I think the logic is pretty clear/sensible? Also: backwards compatible!

dmitri-d:
@ekohl: don't think this makes sense for nsupdate -- the commands to create 'A' and 'PTR' records use different templates, and in this case makes normalization you rely on somewhat confusing. It's not applicable at all to dnscmd and virsh providers.

lib/proxy/plugin_validators.rb
domcleal: Yes, I like the presence/validation idea.  The two things it wouldn't permit are:
1. A required value without a default.  I guess in these cases we have to put _something_ just as an example, even if it isn't a useful default.
2. An optional value with a default.  This just becomes an example in the configuration file.

That seems acceptable, and I agree that it would be logical from a user and developer's perspective.

It might be worth a quick scout through the plugins (http://projects.theforeman.org/projects/foreman/wiki/List_of_Smart-Proxy_Plugins) to see if there are any default config files or settings that this might not work for.

ekohl:
@witlessbird the class (`IN`) is optional (default) and can be added or removed so `nsupdate` uses the same template for all types. It should also allow for supporting additional types, like CNAME and AAAA.

Maybe this is a v2 API thing, but it's a DNS API so why not standardize on what DNS does. Anything that requires different input (like `virsh`) is non-standard and should be exception rather than the rule.

That would mean:

``` json
{"type": "A", "name": "example.com.", "value": "192.168.1.1"}
{"type": "AAAA", "name": "example.com.", "value": "::1"}
{"type": "PTR", "name": "1.1.168.192.in-addr.arpa.", "value": "example.com."}
{"type": "CNAME", "name": "www.example.com.", "value": "www.example.org."}
```

With that the nsupdate command would become:

``` ruby
nsupdate "update add #{@name} #{@ttl} IN #{@type} #{@value}"
```

I think it's easy to transform the API v1 input to the v2 format. Then all providers could be simplified.

Another example: https://github.com/theforeman/smart_proxy_dns_route53/blob/2b2f68e59fab814f749514089034f383edd45c56/lib/smart_proxy_dns_route53/dns_route53_main.rb#L26-L49 does the same juggling with fqdn/value.

lib/proxy/plugin_validators.rb
dmitri-d: Only chef plugin is going to have an issue with this approach due to chef_ssl_pem_file setting (https://github.com/theforeman/smart_proxy_chef/blob/master/lib/smart_proxy_chef_plugin/chef_plugin.rb#L10). I can't see it being used anywhere though -- the fix is going to be trivial.

domcleal: Yeah, and again it looks backwards compatible - the setting remains "optional", especially as there's no use of the file validator.

dmitri-d: Hmm, not if the presence validator is automatically executed for all parameters listed under default_settings. By the looks of it, this is supposed to be an optional parameter to begin with, so removing it from default_settings looks ok to me.

domcleal: Ah yes, sounds good.

dmitri-d:
- Added `::Proxy::PluginValidators::Presence` validator which is automatically enabled for all required settings.
- FileReadable validator skips over optional parameters if they haven't been initialized.
- Merged common_dns/errors.rb and common_dns/record.rb
- Fixed a couple of issues in dns_nsupdate_gss: it was using wrong dns_server setting, and dns_key setting from dns_nsupdate plugin.
- A bunch of style-related fixes.

@ekohl: I'll get to normalization stuff first thing tomorrow.

lib/proxy/pluggable.rb
```diff
@@ -32,10 +32,11 @@ def log_used_default_settings
     def after_activation
       instance_eval(&self.class.after_activation_blk) if self.class.after_activation_blk
     end
-
+
```
domcleal: trailing whitespace

test/dns_common/record_test.rb
```diff
@@ -0,0 +1,19 @@
...
+  def test_dns_find_with_fqdn_parameter
+    Resolv::DNS.any_instance.expects(:getaddress).with('some.host').returns('127.13.0.2')
+    assert '127.13.0.2', Proxy::Dns::Record.new.dns_find('some.host')
+  end
+
+  def test_dns_find_key_not_found
+    Resolv::DNS.any_instance.expects(:getaddress).with('another.host').raises(Resolv::ResolvError.new('DNS result has no information'))
+    assert !Proxy::Dns::Record.new.dns_find('another.host')
+  end
+end
```
domcleal: Nitpick, missing \n

test/dns_nsupdate/dns_nsupdate_config_test.rb
```diff
@@ -2,38 +2,21 @@
...
   end
 
-  def test_initialize_nsupdate_gss_succeeds
-    Proxy::Dns::NsupdateGSS::Plugin.settings.expects(:dns_tsig_keytab).returns('./key')
-    Proxy::Dns::NsupdateGSS::Plugin.settings.expects(:dns_tsig_principal).returns('a@B')
-    File.expects(:exist?).with('./key').returns(true)
-    record = Proxy::Dns::NsupdateGSS::Record.new(:fqdn => 'example.com')
-    assert record
-    assert_equal './key', record.tsig_keytab
-    assert_equal 'a@B', record.tsig_principal
```
domcleal, line 28: Could we also keep tests that create an instance of the provider to test the initialiser?

For the GSS and virsh providers we should probably be testing that they retrieve and store the right keytab/principal/network settings.  It would probably still be useful for all providers to test that if the server's not passed in that they initialise it and the TTL from settings (usually the tests pass their own, right?).

modules/dns_virsh/dns_virsh_main.rb
```diff
@@ -7,45 +8,46 @@ class Record < ::Proxy::Dns::Record
...
-    def self.record(attrs = {})
-      new(attrs.merge(:virsh_network => ::Proxy::SETTINGS.virsh_network))
-    end
-
-
-    def initialize options = {}
-      @network = options[:virsh_network]
+    def initialize(network = nil)
+      @network = network || ::Proxy::SETTINGS.virsh_network
       raise "DNS virsh provider needs 'virsh_network' option" unless network
```
domcleal: This needs to test `@network`.

domcleal: Ah, I see what's happening - the local `network` variable is hiding the attr_reader defined by Proxy::Virsh.

dmitri-d:
@domcleal: fixed things you pointed out. @ekohl: I used your idea, but implemented it differently (https://github.com/theforeman/smart-proxy/pull/329/files#diff-cbba11df9dac1b9a044df6c1b0f8a6e9R17) -- I thought that re-assigning local vars could be hard to understand. This might be a bit more code, but is more readable.

ekohl:
@witlessbird I like the idea, but I wonder if we can pull it up a level higher to `Proxy::Dns::Record`. Plugins overriding create will continue working, but plugins can also override `do_create` to avoid duplication.

test/dns_dnscmd/dnscmd_test.rb
```diff
@@ -0,0 +1,75 @@
...
+    assert Proxy::Dns::Dnscmd::Record.new.remove('host.domain', '192.168.33.33', 'PTR')
+  end
+
+  def test_remove_nonexistent_ptr_record_raises_exception
+    Proxy::Dns::Dnscmd::Record.any_instance.expects(:dns_find).with('192.168.33.33').returns(false)
+    assert_raise Proxy::Dns::NotFound do
+      Proxy::Dns::Dnscmd::Record.new.remove('host.domain', '192.168.33.33', 'PTR')
+    end
+  end
+end
```
ekohl: Nitpick: Missing newline

dmitri-d:
@ekohl: done and done.

ekohl:
@witlessbird Looks good. Some bike shedding: I do wonder if `do_create` and `create` won't be a confusing API.

Other than that I'll try to do some testing with it this weekend.

dmitri-d:
Could be; I can add a comment explaining the reason this implementation, I don't think renaming this method would clear possible confusion.

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -1,39 +1,35 @@
...
+      super(a_server || ::Proxy::Dns::Dnscmd::Plugin.settings.dns_server,
+            a_ttl || ::Proxy::Dns::Plugin.settings.dns_ttl)
     end
 
-    def initialize options = {}
-      super(options)
-    end
+    # create("node01.lab", "192.168.100.2"}
+    # create("node01.lab", "3.100.168.192.in-addr.arpa","PTR"}
+    def create(fqdn, value, type = "A")
```
domcleal: Should this implement do_create to reuse the `raise` code?

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -43,24 +39,23 @@ def create
 
     # remove({ :fqdn => "node01.lab", :value => "192.168.100.2"}
     # remove({ :fqdn => "node01.lab", :value => "3.100.168.192.in-addr.arpa"}
-    def remove
-      @resolver = Resolv::DNS.new(:nameserver => @server)
-      case @type
+    def remove(fqdn, value, type = "A")
```
domcleal: ditto, do_remove?

domcleal:
> Some bike shedding: I do wonder if do_create and create won't be a confusing API.

The comment was helpful, thanks.  Would it be easier to split into `create_a_record` and `create_ptr_record`, or similar?

dmitri-d:
@domcleal: I think what you suggest is a better approach -- do the validation in the controller and let it call an appropriate method (both creation and removal).

ekohl:
@domcleal not sure a separate function per type is useful. I think most backends handle all types the same. The only difference I see is whether they want DNS names including a dot (`host.example.org.` like bind or without `host.example.org` like powerdns).

modules/dns_common/dns_common.rb
```diff
@@ -0,0 +1,48 @@
...
+    end
+
+    # Implement #do_create if checks here are useful; otherwise override #create method instead
+    # create("node01.lab", "192.168.100.2")
+    # create("node01.lab", "3.100.168.192.in-addr.arpa","PTR")
+    def create(fqdn, value, type = "A")
+      raise("'create' requires fqdn, value, and type parameters") if fqdn.nil? || value.nil? || type.nil?
+
+      do_create(fqdn, value, 'A') if type == 'A'
+      do_create(value, fqdn, 'PTR') if type == 'PTR'
```
ekohl: Raise an error if type is unsupported?

dmitri-d:
@domcleal. @ekohl: refactored creation and deletion of address and pointer records into dedicated methods. Moved parameter validation into the api controller. Added api controller tests.

ekohl:
@witlessbird for ease of review I wonder if we can split it into two commits. First the create refactor and then the DI refactor. I think we can keep the create refactor backwards compatible and thus easier to merge when we're happy. The DI refactor is much more likely to break things.

dmitri-d:
It's way more likely that I will fix an issue rather than rollback the commit. In this light, does it make sense to break up this PR (I _really_ want to be done with it at this point)?

As an aside, I think changes to create/delete methods are more likely to break stuff, even though I covered them with tests.

modules/dns_common/dns_common.rb
domcleal: I think this is resolved in the controller layer now.

dmitri-d: Nods: https://github.com/theforeman/smart-proxy/pull/329/files#diff-c603a2778d38b35e4c4c3f5a1edc6d35R18

domcleal:
> @domcleal not sure a separate function per type is useful. I think most backends handle all types the same. The only difference I see is whether they want DNS names including a dot (host.example.org. like bind or without host.example.org like powerdns).

Since the PR's been changed to the separate function anyway now, what do you think of how it looks?

You're right I think in that many of the providers do handle it the same - nsupdate, route53 and powerdns, but dnscmd and virsh don't, but it does give providers the option to use and call a shared method or not without them all needing to implement checks on `type`.  They can also easily add a dot if they need to.

> @witlessbird for ease of review I wonder if we can split it into two commits. First the create refactor and then the DI refactor. I think we can keep the create refactor backwards compatible and thus easier to merge when we're happy. The DI refactor is much more likely to break things.

The DI bit is the bit that's ready I think but is a breaking change, so I thought your original thought was to get the other breaking change in at the same time?

ekohl:
> > @domcleal not sure a separate function per type is useful. I think most backends handle all types the same. The only difference I see is whether they want DNS names including a dot (host.example.org. like bind or without host.example.org like powerdns).
> 
> Since the PR's been changed to the separate function anyway now, what do you think of how it looks?
> 
> You're right I think in that many of the providers do handle it the same - nsupdate, route53 and powerdns, but dnscmd and virsh don't, but it does give providers the option to use and call a shared method or not without them all needing to implement checks on type. They can also easily add a dot if they need to.

My initial thought was that this results in duplication between nsupdate, route53 and powerdns especially if you want to support new types. What you see in the nsupdate provider where `create_*_record` is all mapped to `do_create` with a type. I know I'll duplicate this in powerdns, but I also see the benefits of the hooking options it provides.

Maybe it makes sense to introduce a subclass of `::Proxy::Dns::Record` for unified providers, but we can always add that at a later point.

dmitri-d:
The amount of duplication introduced (in some cases) by splitting of create and delete methods is trivial and this split greatly improves overall readability -- I very much prefer this approach because of the latter.

ekohl:
@witlessbird I agree the current patch greatly improves readability and the duplication needed now is already greatly reduced to a trivial amount. Let's start with this and we can always see if we need further improvements later on.

domcleal:
> Maybe it makes sense to introduce a subclass of ::Proxy::Dns::Record for unified providers, but we can always add that at a later point.

Yeah, perhaps a mixin.

ekohl:
> > Maybe it makes sense to introduce a subclass of ::Proxy::Dns::Record for unified providers, but we can always add that at a later point.
> 
> Yeah, perhaps a mixin.

I like that idea.

domcleal:
Okay, I'm happy with this as-is then and will merge it shortly.  I'll try to follow up with another PR for a mixin that we can use between the providers that need it soon (unless one of you beats me to it).  I've submitted a PR for route53 that switches over to the current state.

ekohl:
Maybe a post to foreman-dev might be good so people know smart-proxy 1.10 is very different from develop with regards to DNS(-plugins).

domcleal:
Yes, I was going to suggest that, plus updates to the DNS plugin template and [wiki page](http://projects.theforeman.org/projects/foreman/wiki/How_to_Create_a_Smart-Proxy_Plugin) (perhaps with a pointer to the plugin PR as a guide).

domcleal:
Merged as 42b9581, thanks @witlessbird.  If you could do the three docs-related items above it would be appreciated.

dmitri-d:
Heh -- was going to suggest that I pick up the docs and examples, but you beat me to it, @domcleal.

PR #330, opened 2015-10-20
Fixes #12211 - Improve DHCP subnets parsing to get more informations …

bagasse:
…from DHCP server

Add ability to parse some subnet options and provides them to the foreman to enhance subnet imports.

This modify the dhcp smart-proxy API (adding options hash).

### DIFF
--- a/modules/dhcp/providers/server/isc.rb
+++ b/modules/dhcp/providers/server/isc.rb
@@ -54,7 +54,7 @@ module Proxy::DHCP
       statements += ztp_options_statements(options)
       statements += poap_options_statements(options)
 
-      omcmd "set statements = \"#{statements.join(" ")}\""      unless statements.empty?
+      omcmd "set statements = \"#{statements.join(' ')}\""      unless statements.empty?
       omcmd "create"
       omcmd("disconnect", "Added DHCP reservation for #{record}")
       record
@@ -108,7 +108,6 @@ module Proxy::DHCP
 
     def initialize_memory_store_with_dhcp_records(records)
       records.each do |record|
-
         case record
         when Proxy::DHCP::Reservation
           if record.options[:deleted]
@@ -149,11 +148,26 @@ module Proxy::DHCP
 
     def parse_config_for_subnets
       ret_val = []
-
-      @config.scan(/subnet\s+([\d\.]+)\s+netmask\s+([\d\.]+)/) do |match|
-        next unless managed_subnet? "#{match[0]}/#{match[1]}"
-
-        ret_val << Proxy::DHCP::Subnet.new(match[0], match[1])
+      # Extract subnets config block
+      @config.scan(/subnet\s+([\d\.]+)\s+netmask\s+([\d\.]+)\s*{([\w|;|\.|,|\-|\s]*)}/) do |match|
+        network, netmask, subnet_config_lines = match
+        options = {}
+        if subnet_config_lines
+          # Split each config lines and scan for some interesting options (dns, gateway, dhcp range) used in foreman subnets.
+          # Some other options can be added later, see manpages dhcpd.conf(5) or dhcpd-options(5) for available ISC DHCP subnet options.
+          subnet_config_lines.split(';').each do |option|
+            case option
+              when /option\s+routers\s+[\d\.]+/
+                options[:routers] = get_ip_list_from_config_line(option)
+              when /option\s+domain\-name\-servers\s+[\d\.]+/
+                options[:domain_name_servers] = get_ip_list_from_config_line(option)
+              when /range\s+[\d\.]+\s+[\d\.]+/
+                # get IP addr range used for this subnet
+                options[:range] = get_range_from_config_line(option)
+            end
+          end
+        end
+        ret_val << Proxy::DHCP::Subnet.new(network, netmask, options.reject{|key, value| value.nil? || value.empty? })
       end
 
       ret_val
@@ -288,9 +302,7 @@ module Proxy::DHCP
       logger.debug "Reading config file #{file}"
       config = []
       File.readlines(file).each do |line|
-
-        line.strip! # remove left and right whitespace
-        next if line.start_with?("#") # remove comments
+        line = line.split('#').first.strip # remove comments, left and right whitespace
         next if line.empty? # remove blank lines
 
         if /^include\s+"(.*)"\s*;/ =~ line
@@ -408,5 +420,15 @@ module Proxy::DHCP
       end
       statements
     end
+  end
+
+  # Get all IPv4 addresses provided by the ISC DHCP config line following the pattern "my-config-option-or-directive IPv4_ADDR[, IPv4_ADDR] [...];" and return an array.
+  def get_ip_list_from_config_line(option_line)
+    option_line.scan(/\s*((?:(?:\d{1,3}\.){3}\d{1,3})\s*(?:,\s*(?:(?:\d{1,3}\.){3}\d{1,3})\s*)*)/).first.first.gsub(/\s/,'').split(",").reject{|value| value.nil? || value.empty? }
+  end
+
+  # Get IPv4 range provided by the ISC DHCP config line following the pattern "range IPv4_ADDR IPv4_ADDR;" and return an array.
+  def get_range_from_config_line(range_line)
+    range_line.scan(/\s*((?:\d{1,3}\.){3}\d{1,3})\s*((?:\d{1,3}\.){3}\d{1,3})\s*/).first
   end
 end
--- a/modules/dhcp/subnet.rb
+++ b/modules/dhcp/subnet.rb
@@ -17,11 +17,19 @@ module Proxy::DHCP
     include Proxy::Log
     include Proxy::Validations
 
-    def initialize network, netmask
-      @network   = validate_ip network
-      @netmask   = validate_ip netmask
-      @options   = {}
-      @timestamp = Time.now
+    def initialize network, netmask, options = {}
+      @network = validate_ip network
+      @netmask = validate_ip netmask
+      @options = {}
+
+      @options[:routers] = options[:routers].each{|ip| validate_ip ip } if options[:routers]
+      @options[:domain_name] = options[:domain_name] if options[:domain_name]
+      @options[:domain_name_servers] = options[:domain_name_servers].each{|ip| validate_ip ip } if options[:domain_name_servers]
+      @options[:ntp_servers] = options[:ntp_servers].each{|ip| validate_ip ip } if options[:ntp_servers]
+      @options[:interface_mtu] = options[:interface_mtu].to_i if options[:interface_mtu]
+      @options[:range] = options[:range] if options[:range] && options[:range][0] && options[:range][1] && valid_range(:from => options[:range][0], :to => options[:range][1])
+
+      @timestamp     = Time.now
     end
 
     def include? ip
--- a/test/dhcp/dhcp_api_test.rb
+++ b/test/dhcp/dhcp_api_test.rb
@@ -37,8 +37,31 @@ class DhcpApiTest < Test::Unit::TestCase
     get "/"
     assert last_response.ok?, "Last response was not ok: #{last_response.status} #{last_response.body}"
     data = JSON.parse(last_response.body)
-    expected = [{"network"=>"192.168.122.0", "netmask"=>"255.255.255.0"}]
-    assert_equal expected, data
+    expected = [{
+        "network"=>"192.168.122.0",
+        "netmask"=>"255.255.255.0",
+        "options"=>{"routers"=>["192.168.122.250"]}
+      },{
+        "network"=>"192.168.123.0",
+        "netmask"=>"255.255.255.192",
+        "options"=>{
+          "routers"=>["192.168.123.1"],
+          "domain_name_servers"=>["192.168.123.1"],
+          "range"=>["192.168.123.2", "192.168.123.62"]
+        }
+      },{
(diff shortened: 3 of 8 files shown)


modules/dhcp/providers/server/isc.rb
```diff
@@ -149,11 +149,24 @@ def loadSubnetData subnet
...
-
-        ret_val << Proxy::DHCP::Subnet.new(match[0], match[1])
+      @config.scan(/subnet\s+([\d\.]+)\s+netmask\s+([\d\.]+)\s*{([\w|;|\.|,|\-|\s]*)}/) do |match|
+        options = {}
+        if match[2]
+          match[2].split(';').each do |option|
+            case option
+              when /option\s+routers\s+[\d\.]+/
+                # get the first router match and use it for gateway
+                options[:gateway] = option.scan(/option\s+routers\s+([\d\.]+)/).first.first
```
dmitri-d: why not keep the whole list?

modules/dhcp/providers/server/isc.rb
```diff
@@ -149,11 +149,24 @@ def loadSubnetData subnet
...
+              when /option\s+domain\-name\-servers\s+[\d\.]+/
+                # get #1 and #2 domain name servers
+                options[:dns_primary], options[:dns_secondary] = option.scan(/option\s+domain\-name\-servers\s+([\d\.]+)\s*[,]?\s*([\d\.]*)/).first
```
dmitri-d: similarly, why not keep the whole list? Also, this will leave `:dns_secondary` with an empty string if only one server has been specified.

dmitri-d:
@bagasse: cheers for the PR!

I feel that in the data we return we should keep option names and format of their values close to the original ones (if not the same): `gateway` would become `:routers` and contain the whole list, `:dns_primary` and `:dns_secondary` would become `:domain_name_servers` and keep all of the servers, etc.

Another issue that I see is that dhcp config file may contain global options applicable to all subnets (and potentially overridden in individual subnet configs). I we should support this if we are going to return these additional options.

modules/dhcp/providers/server/isc.rb
bagasse: I reject all nil and blank values just before return the array

dmitri-d: nods, see it now.

bagasse:
@witlessbird

I changed the options names to map names used on the foreman side. I will update the PR to keep names like in the dhcp server config, and keep all values too.
For the global options, i didn't take a look at it because you can have global options, shared networks blocks and group block (with some nested ones) that overrides some of the options. I wanted to keep this simple at this time and just take local subnet options. Parsing dhcpd config with regex with this kind of russian dolls config blocks was out of the scope of this PR (and testing all cases is not trivial), but if it's a requirement for you, i will take a look at it when i will have more time.

dmitri-d:
@bagasse: let's wait with the renaming until we hear something from the foreman side (although I think it makes sense to rename attribute names there and keep them close to original here).

Yeah, I agree that parsing global options makes things more complicated (and possibly warrants a real parser, not regexes). I'm ok with postponing this bit, as long as we can get the functionality in not too distant future.

dLobatog:
[test]

dLobatog:
@bagasse One of the tests you modified (get_root) is failing. see [Jenkins](http://ci.theforeman.org/job/test_proxy_develop_pull_request/888/)  or test locally with `rake test`. Rubocop is also throwing a few warnings, please fix them & update the PR.

bagasse:
@dLobatog: I've fixed rubocop warnings, sorry i forgot to run rubocop before on this PR. For the get_root  of dhcp_api_test, all test are green in my dev env (CentOS 7, ruby 1.9.3). I seen in jenkins that this test only fail with ruby 1.8.7. I don't know why, with this ruby version, networks presents in the list are not in the same order than in others ruby versions. I will take a look at this more deeply tomorrow.

dmitri-d:
[test]

modules/dhcp/providers/server/isc.rb
```diff
@@ -288,8 +298,7 @@ def read_config file, ignore_includes=false
       logger.debug "Reading config file #{file}"
       config = []
       File.readlines(file).each do |line|
-
-        line.strip! # remove left and right whitespace
+        line = line.split('#').first.strip # remove comments, left and right whitespace
```
dmitri-d, line 305: Why this change? The next line used to skip over comments (and isn't needed with this change).

test/dhcp/dhcp_api_test.rb
```diff
@@ -37,7 +37,26 @@ def test_api_01_get_root
...
+        }
+      },{
+        "network"=>"192.168.124.0",
+        "netmask"=>"255.255.255.0",
+        "options"=>{
+          "routers"=>["192.168.124.1", "192.168.124.2"],
+          "domain_name_servers"=>["192.168.123.1", "192.168.122.250"]
+        }
+    }]
     assert_equal expected, data
```
dmitri-d: Test failure here is due to the order of entries in `epxected` array being different from the order of entries in `data`. I would suggest using sets for equality comparison (we don't care about the order). You can find an example in other tests in this suite.

modules/dhcp/providers/server/isc.rb
bagasse, line 305: In my understanding, the next line remove lines that starts with comment, this remove comments at the end of lines. But if you want i can remove this.

bagasse:
@witlessbird: changed, arrays to sets, thanks for pointing this out to me.

dLobatog:
[test]

modules/dhcp/subnet.rb
```diff
@@ -17,11 +17,19 @@ class Subnet
     include Proxy::Log
     include Proxy::Validations
 
-    def initialize network, netmask
-      @network   = validate_ip network
-      @netmask   = validate_ip netmask
-      @options   = {}
-      @timestamp = Time.now
+    def initialize network, netmask, options = {}
+      @network       = validate_ip network
```
dmitri-d: nitpick: you can get rid of assignment formatting -- simple spaces around '=' would suffice.

modules/dhcp/providers/server/isc.rb
dmitri-d, line 305: once you call `String#split` all instances of the symbol you used to split the string will be removed,  which makes https://github.com/bagasse/smart-proxy/blob/Fixes_12211_-_Improve_DHCP_subnets_parsing_to_get_more_informations_from_DHCP_server/modules/dhcp/providers/server/isc.rb#L302 redundant. 

I'm ok with this change if you remove code on line 302.

test/dhcp/server_isc_test.rb
```diff
@@ -82,8 +82,24 @@ def test_poap_quirks
   def test_loadSubnets_loads_managed_subnets
     subnets = @dhcp.loadSubnets
```
dmitri-d: I know that this was here before, but it wasn't a very good test: `loadSubnets` returns subnets that were processed, not the data that was stored.

Another issue is that this tests too many things; which will make it harder to pinpoint an issue should code change. Could you break this up to test specific cases, for example handling of `domain_name_server` option when it's not present, when only one server is present, when both are present, etc. 

It is probably easier to work directly with subnet_service (see `@subnet_service` in setup -- you can query it for relevant subnets, as opposed to parsing the hash returned by `@dhcp.subnets`.

modules/dhcp/providers/server/isc.rb
```diff
@@ -409,4 +418,11 @@ def poap_options_statements(options)
       statements
     end
   end
+
+  private
+
+  def get_ip_list_from_config_line(option_line)
```
dmitri-d, line 426: can you add a test for this method (you'll need to drop `private` for this)?

test/dhcp/subnet_test.rb
```diff
@@ -24,6 +23,33 @@ def test_should_not_save_invalid_network_addresses
     assert_raise Proxy::Validations::Error do
       Proxy::DHCP::Subnet.new("1..1.1", @netmask)
     end
+    assert_raise Proxy::Validations::Error do
```
dmitri-d, line 29: Can you change this to test per option?

dmitri-d:
@bagasse: Looks pretty good, only needs some changes in tests around added functionality.

domcleal:
Thanks for your PRs and work on this feature @bagasse, and welcome to our GitHub organisation.  If you can set your membership to "public" on [this page](https://github.com/orgs/theforeman/people) then Jenkins will automatically test any updates you make.

modules/dhcp/providers/server/isc.rb
bagasse, line 305: line 302 removed.

modules/dhcp/subnet.rb
bagasse: done.

test/dhcp/server_isc_test.rb
bagasse: In my understanding, `loadSubnets` implementation of `Proxy::DHCP::Server::ISC` load subnets only from config files and `Proxy::DHCP::SubnetService` is only here to store loaded data, so i don't see the problem to use `loadSubnets` here. For the other point, i've separated tests in several methods.

modules/dhcp/providers/server/isc.rb
bagasse, line 426: done some basic test for this method

test/dhcp/subnet_test.rb
bagasse, line 29: moved per options test into other methods

domcleal:
[test] one last time :)

bagasse:
@domcleal Glad to join and thank you for the invitation :)

modules/dhcp/providers/server/isc.rb
```diff
@@ -149,11 +148,22 @@ def loadSubnetData subnet
...
     def parse_config_for_subnets
       ret_val = []
-
-      @config.scan(/subnet\s+([\d\.]+)\s+netmask\s+([\d\.]+)/) do |match|
-        next unless managed_subnet? "#{match[0]}/#{match[1]}"
-
-        ret_val << Proxy::DHCP::Subnet.new(match[0], match[1])
+      @config.scan(/subnet\s+([\d\.]+)\s+netmask\s+([\d\.]+)\s*{([\w|;|\.|,|\-|\s]*)}/) do |match|
+        options = {}
+        if match[2]
```
dLobatog: Mind to extract this logic to a method or something similar? Maybe write a comment / save `match[0] match[1] match[2]` to named variables? It's a bit difficult to understand without being well acquainted with the format of the conf file.

modules/dhcp/providers/server/isc.rb
```diff
@@ -149,11 +148,22 @@ def loadSubnetData subnet
...
+          match[2].split(';').each do |option|
+            case option
+              when /option\s+routers\s+[\d\.]+/
+                options[:routers] = get_ip_list_from_config_line(option)
+              when /option\s+domain\-name\-servers\s+[\d\.]+/
+                options[:domain_name_servers] = get_ip_list_from_config_line(option)
+              when /range\s+[\d\.]+\s+[\d\.]+/
+                # get IP addr range used for this subnet
+                options[:range] = option.scan(/\s*((?:\d{1,3}\.){3}\d{1,3})\s*((?:\d{1,3}\.){3}\d{1,3})\s*/).first
```
dLobatog: Perhaps you want to extract this regex to a method `ip_address_range`? Maybe move the regex to a constant?
 You use almost all of it below as well on `get_ip_list_from_config_line`, and it's hard to know what it is without a comment, maybe that would help making this easier to read.

dLobatog:
@bagasse I only have 2 comments about the regexes & maintainability, but overall great job :+1:. IMO if you could fix that that should be it. In any case @domcleal and @witlessbird feel free to merge if you think it's fine already, you are the ones that work on this code the most.

bagasse:
@dLobatog you are right, it's not really readable, i've extracted range regexp too and used named vars.

test/dhcp/server_isc_test.rb
dmitri-d: Nods,that's correct. However, currently these tests verify the behaviour of `parse_config_for_subnets` method: so either calls to `loadSubnets`should be replaced with calls to `parse_config_for_subnets`, or the assertions should be made against `Proxy::DHCP::SubnetService` (the main side-effect of `loadSubnets` is initialization of `SubnetService`).

modules/dhcp/providers/server/isc.rb
```diff
@@ -409,4 +421,14 @@ def poap_options_statements(options)
       statements
     end
   end
+
+  # Get all IPv4 addresses provided by the ISC DHCP config line following the pattern "my-config-option-or-directive IPv4_ADDR[, IPv4_ADDR] [...];" and return an array.
+  def get_ip_list_from_config_line(option_line)
```
dmitri-d, line 426: Could you add a test for this method?

bagasse, line 426: It was already there, but i've added a test for `get_range_from_config_line` that i've extracted too.  I changed `loadSubnets` calls to `parse_config_for_subnets` as suggested.

dmitri-d:
[test]

dmitri-d:
Waiting for tests to pass before merging.

dmitri-d:
merged in at e2a21ec. Thanks, @bagasse!

bagasse:
@witlessbird @dLobatog, thank you for pointers and re (re re re)-reviews :)

PR #331, opened 2015-10-20
Fixes #12217: support older versions of bundler

dmitri-d:
No longer rely on :platforms to install test-unit dependency


domcleal:
The rubocop job was aborted, so I've adjusted the timeout in the job configuration.  [test]

mmoll:
:+1:

domcleal:
Merged as 00875b3, thanks @witlessbird.

PR #332, opened 2015-10-28
Fixes #8946: Added a handler for SIGTERM

dmitri-d:


lib/launcher.rb
```diff
@@ -127,9 +127,8 @@ def launch
 
       (t1 || t2).join
     rescue SignalException => e
-      # This is to prevent the exception handler below from catching SignalException exceptions.
       logger.info("Caught #{e}. Exiting")
-      raise
+      e.signo == 15 ? exit(143) : raise # trap SIGTERM, used by systemd to stop/restart the service
```
lzap: Why 143? Time to create a table of exit codes at one place to get things in order?

Looks good.

dmitri-d: I wanted to retain the fact that it was terminated b/c of the signal. `143` is the exit code that shell will return for a process terminated by `SIGTERM` (that didn't handle the signal itself). 

Not sure how useful this behaviour is, I can return `0` instead.

lzap: > I wanted to retain the fact that it was terminated b/c of the signal. `143` is the exit code that shell will return for a process terminated by `SIGTERM` (that didn't handle the signal itself). 
> 
> Not sure how useful this behaviour is, I can return `0` instead.  

The usual practice is to return zero if that was a correct shutdown. But
143 seems misleading here (it's actually 128 + 15 which was number of
the signal in Bash). IMHO zero is more appropriate here.

## 

Later,
 Lukas #lzap Zapletal

dmitri-d:
Updated to use @lzap's `SignalHandler` class.

lib/proxy/signal_handler.rb
```diff
@@ -14,4 +23,22 @@ def self.install_ttin_trap
...
+        rescue Exception => e
+          logger.warn "Unable to overwrite interrupt trap: #{e}"
+        end
+      end
+    end
+  end
+
+  def install_term_trap
+    trap(:TERM) { exit(0) }
+  end
```
lzap, line 43: Shell we rather signal shutdown to our worker threads rather than hard exiting this? I was browsing through the Rack code and I don't know how to tell the stack to cleanly exit. Maybe it's not possible and we just need to do this, but it is worth some effort. Broken orchestrations are always problematic.

lzap, line 43: Looks like webrick is `.shutdown`-able but Rack does not seem to expose this API.

Another trick is to create some own API that will cause clean shutdown. We can generate and remember a security token and this will be passed along with the request.

dmitri-d, line 43: Rack relies on concrete web-server implementation of `#shutdown` (https://github.com/rack/rack/blob/master/lib/rack/server.rb#L288). You are right, this is a bit harsh, and we should shut down gracefully, but I'd implement that in a dedicated ticket.

dmitri-d:
Actually, now that I think of it, #shutdown handler is being triggered on exit call. Probably `at_exit` hook? Lemme check.

dmitri-d:
@lzap: webrick will always shutdown cleanly (starting with ruby 2.0), or at least attempt to (1.9.3, haven't checked 1.8.7): https://github.com/ruby/ruby/blob/trunk/lib/webrick/server.rb#L209. This is webrick-specific, however, and something more generic would be nice to have.

lib/proxy/signal_handler.rb
lzap, line 43: Ok it looks like this code is triggering correctly on TERM sig, so it does work as expected.

lzap:
Merged as 616d41a7841009a632e65f0b6360bdeec58df260, thank you!

PR #333, opened 2015-10-28
Fixes #12254: pinned addressable gem to ~> 2.3.8

dmitri-d:
This is the last version to support 1.8.7.

### DIFF
--- a/bundler.d/test.rb
+++ b/bundler.d/test.rb
@@ -6,6 +6,7 @@ group :test do
   gem 'ci_reporter', '>= 1.6.3', "< 2.0.0", :require => false
   gem 'rdoc'
   gem 'test-unit' if RUBY_VERSION > "1.8.7"
+  gem 'addressable', '~> 2.3.8' if RUBY_VERSION == '1.8.7' # 2.4.0 drops support for ruby 1.8.7
   gem 'webmock'
   gem 'rubocop-checkstyle_formatter' if RUBY_VERSION > "1.9.2"
 end


mmoll:
It would be nice to this only for 1.8, as 2.x.x should still be compatible to 2.3.8. The major problem is that we will have a hard time once a distro is updating its package of addressable (i.e. Ubuntu 16.04 or FreeBSD might hit this).

dmitri-d:
I'm very slightly concerned that this might cause strange test failures once the versions deviate enough (although I'm hoping that we'll drop 1.8.7 sooner rather than later). Having said that, this affects bundler-based installs only; for package-based installs we rely on bundler_ext that strips versioning information from gem requirements, forcing smart=proxy to use w/e is available on the system.

dmitri-d:
@mmoll: fixed

mmoll:
I do agree it's anything else than ideal, but in such a situation failing tests would be good IMHO so we know that there's a real action point.

bundler.d/test.rb
```diff
@@ -6,6 +6,8 @@
   gem 'ci_reporter', '>= 1.6.3', "< 2.0.0", :require => false
   gem 'rdoc'
   gem 'test-unit' if RUBY_VERSION > "1.8.7"
+  gem 'addressable', '~> 2.3.8' if RUBY_VERSION == '1.8.7' # 2.4.0 drops support for ruby 1.8.7
+  gem 'addressable' if RUBY_VERSION > '1.8.7'
```
domcleal: Is this needed?  It's brought in by webmock otherwise.

dmitri-d:
@domcleal: removed unnecessary gem specification.

domcleal:
Merged as 2487519, thanks @witlessbird.

PR #334, opened 2015-10-28
Fixes #12319: fixed race condition in templates module

dmitri-d:
Multiple simultaneous requests to /unattended/template_type no longer fail.


lzap:
Nice one, tested and it works fine!

lzap:
Merged as 26aba8520d3d32ca27b9708cb98169bc31d43d7b, thank you!

PR #336, opened 2015-11-02
Fixes #12370 - proxy logs stacktrace on SIGTTIN

lzap:
Good for investigation of 100 % CPU load.

### DIFF
--- a/lib/launcher.rb
+++ b/lib/launcher.rb
@@ -1,5 +1,6 @@
 require 'proxy/log'
 require 'proxy/settings'
+require 'proxy/signal_handler'
 
 module Proxy
   class Launcher
@@ -124,6 +125,8 @@ module Proxy
       rescue Exception => e
         logger.warn "Unable to overwrite interrupt trap: #{e}"
       end
+
+      ::Proxy::SignalHandler.install_ttin_trap
 
       (t1 || t2).join
     rescue SignalException => e
--- a/lib/proxy/signal_handler.rb
+++ b/lib/proxy/signal_handler.rb
@@ -1,0 +1,17 @@
+require 'proxy/log'
+
+class Proxy::SignalHandler
+  def self.install_ttin_trap
+    # logger can't be accessed from trap context
+    trap(:TTIN) do
+      puts "Starting thread dump for current Ruby process"
+      puts "============================================="
+      puts ""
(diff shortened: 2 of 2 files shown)


dmitri-d:
@lzap: I could live with one signal handler defined in `Proxy::Launcher.launch`, but if we are going to add this (and we could implement a handler for SIGTERM similarly), I think we should refactor signal handlers into a dedicated module/class. I can do that tomorrow if you don't have time...

lzap:
> @lzap: I could live with one signal handler defined in `Proxy::Launcher.launch`, but if we are going to add this (and we could implement a handler for SIGTERM similarly), I think we should refactor signal handlers into a dedicated module/class. I can do that tomorrow if you don't have time... 

Introduced new class, I am not going to work on SIGTERM in this ticket
tho. Tried to play with it for a while, but this deserves own ticket and
must be tested properly.

## 

Later,
 Lukas #lzap Zapletal

lzap:
Pushed the forgotten file.

Also for discussion: Loggingg the stackstrace via logger does not work, so I need to use STDOUT. But in case syslog is configured (and this is now the default on RHEL6), the output might end up nowhere. In systemd environments, the stdout is automatically acquired and added to journal, but I am not sure in non-systemd environments. Do we want to reopen STDOUT and redirect it into logs?

domcleal:
> But in case syslog is configured (and this is now the default on RHEL6)

What do you mean precisely about RHEL6?  A file logger is the default everywhere, AFAIK.

lzap:
> > But in case syslog is configured (and this is now the default on RHEL6)
> 
> What do you mean precisely about RHEL6?  A file logger is the default everywhere, AFAIK.

I was under impression we changed that to SYSLOG once it was introduced,
but that's not the case I see.

## 

Later,
 Lukas #lzap Zapletal

domcleal:
No, 335a1610 added the capability but didn't change the default behaviour.

lzap:
> No, 335a1610 added the capability but didn't change the default behaviour.

Created http://projects.theforeman.org/issues/12402 for the record. I
don't think the time has come yet.

## 

Later,
 Lukas #lzap Zapletal

dmitri-d:
merged in `2c40ea5`. Thanks, @lzap!

PR #337, opened 2015-11-12
fixes #12449 - load nsupdate_gss config from plugin settings

domcleal:


dmitri-d:
merged in at 839ca95. Thanks, @domcleal!

PR #338, opened 2015-11-13
Fixes #12465 -  handle ISC deleted entires correctly

brandonweeks:
~~Fixes #12465 with three changes: ~=
1. Don't catch Exception (the error it surfaced was opaque)
2. Added a test case for a record that doesn't have an IP
3. Skip records without an IP, otherwise they will error on trying to find the subnet~~

Alright, let's try this again correctly. Here's what's happening:
- The proxy creates a DHCP reservation in ISC via omshell. dhcpd.leases will contain an entry:

```
host deletedhost.example.com {
  dynamic;
  hardware ethernet aa:aa:aa:aa:aa:aa;
  fixed-address 192.168.0.10;
        supersede server.next-server = ff:ff:ff:ff;
        supersede host-name = "deletedhost.example.com";
}
```
- The proxy deletes the host. dhcpd.leases will now contain the original entry _and_ the following at the bottom of the file:

```
host deletedhost.example.com {
  dynamic;
  deleted;
}
```
- Foreman is trying to modify the DHCP reservation, so it calls `DELETE` on `/dhcp/192.168.0.0/aa:aa:aa:aa:aa:aa`.
- The proxy parses dhcpd.leases correctly parsing the first entry and adding it to the in memory database. It fails to parse the deleted entry because it doesn't have an address and since we're catching Exception it moves on.
- The proxy then tries to delete the record that it thinks exists and then #12469 is triggered.

### DIFF
--- a/modules/dhcp/providers/server/isc.rb
+++ b/modules/dhcp/providers/server/isc.rb
@@ -1,5 +1,6 @@
 require 'time'
 require 'dhcp/subnet'
+require 'dhcp/record/deleted_reservation'
 require 'dhcp/record/reservation'
 require 'dhcp/record/lease'
 require 'dhcp/server'
@@ -78,13 +79,13 @@ module Proxy::DHCP
             opts.merge!(parse_record_options(data))
           end
         end
-        begin
-          subnet = @service.find_subnet(opts[:ip])
-          next unless subnet
-          ret_val << Proxy::DHCP::Reservation.new(opts.merge(:subnet => subnet))
-        rescue Exception => e
-          logger.warn "skipped #{hostname} - #{e}"
+        if opts[:deleted]
+          ret_val << Proxy::DHCP::DeletedReservation.new(opts)
+          next
         end
+        subnet = @service.find_subnet(opts[:ip])
+        next unless subnet
+        ret_val << Proxy::DHCP::Reservation.new(opts.merge(:subnet => subnet))
       end
 
       conf.scan(/lease\s+(\S+\s*\{[^}]+\})/) do |lease|
@@ -109,12 +110,11 @@ module Proxy::DHCP
     def initialize_memory_store_with_dhcp_records(records)
       records.each do |record|
         case record
+        when Proxy::DHCP::DeletedReservation
+          record = @service.find_host_by_hostname(record.name)
+          @service.delete_host(record) if record
+          next
         when Proxy::DHCP::Reservation
-          if record.options[:deleted]
-            record = @service.find_host_by_hostname(record.subnet_address, record.name)
-            @service.delete_host(record) if record
-            next
-          end
           if dupe = @service.find_host_by_mac(record.subnet_address, record.mac)
             @service.delete_host(dupe)
           end
--- a/modules/dhcp/subnet_service.rb
+++ b/modules/dhcp/subnet_service.rb
@@ -60,7 +60,7 @@ class Proxy::DHCP::SubnetService
   def add_host(subnet_address, record)
     @reservations_by_ip[subnet_address, record.ip] = record
     @reservations_by_mac[subnet_address, record.mac] = record
-    @reservations_by_name[subnet_address, record.name] = record
+    @reservations_by_name[record.name] = record
     logger.debug("Added a reservation: #{record.ip}:#{record.mac}:#{record.name}")
   end
 
@@ -73,7 +73,7 @@ class Proxy::DHCP::SubnetService
   def delete_host(record)
     @reservations_by_ip.delete(record.subnet.network, record.ip)
     @reservations_by_mac.delete(record.subnet.network, record.mac)
-    @reservations_by_name.delete(record.subnet.network, record.name)
+    @reservations_by_name.delete(record.name)
     logger.debug("Deleted a reservation: #{record.ip}:#{record.mac}:#{record.name}")
   end
 
@@ -93,15 +93,15 @@ class Proxy::DHCP::SubnetService
     @reservations_by_ip[subnet_address, ip_address]
   end
 
-  def find_host_by_hostname(subnet_address, hostname)
-    @reservations_by_name[subnet_address, hostname]
+  def find_host_by_hostname(hostname)
+    return @reservations_by_name[hostname]
   end
 
   def all_hosts(subnet_address = nil)
     if subnet_address
-      return @reservations_by_name[subnet_address] ? @reservations_by_name.values(subnet_address) : []
+      return @reservations_by_ip[subnet_address] ? @reservations_by_ip.values(subnet_address) : []
     end
-    @reservations_by_name.values
+    @reservations_by_ip.values
   end
 
   def all_leases(subnet_address = nil)
--- a/test/fixtures/dhcp/dhcp.leases
+++ b/test/fixtures/dhcp/dhcp.leases
@@ -13,18 +13,32 @@ host ten.example.com {
 }
 
 # Reservation that is deleted
-host deleteme.example.com {
+host deleted.example.com {
   dynamic;
   hardware ethernet 10:10:10:10:20:20;
   fixed-address 192.168.122.20;
   supersede host-name = "deleteme.example.com";
 }
-host deleteme.example.com {
+host deleted.example.com {
   dynamic;
-  hardware ethernet 10:10:10:10:20:20;
-  fixed-address 192.168.122.20;
-  supersede host-name = "deleteme.example.com";
   deleted;
+}
+# Reservation that is deleted and then undeleted with same name
+host undeleted.example.com {
+  dynamic;
+  hardware ethernet 10:10:10:10:30:30;
+  fixed-address 192.168.122.30;
+  supersede host-name = "undeleted.example.com";
+}
+host undeleted.example.com {
+  dynamic;
+  deleted;
+}
+host undeleted.example.com {
+  dynamic;
+  hardware ethernet 10:10:10:10:30:30;
+  fixed-address 192.168.122.35;
+  supersede host-name = "undeleted.example.com";
 }
 
 # The following six leases are from booting a real-world DELL hardware with
--- a/modules/dhcp/record/deleted_reservation.rb
+++ b/modules/dhcp/record/deleted_reservation.rb
@@ -1,0 +1,13 @@
+require 'dhcp/record/reservation'
+
+module Proxy::DHCP
+  # represent a deleted DHCP Record
+  class DeletedReservation < Reservation
+    attr_reader :name
+
+    def initialize options = {}
+      @name = options[:name] || options[:hostname] || raise("Must define a name: #{options.inspect}")
+      @options = options
+    end
+  end
+end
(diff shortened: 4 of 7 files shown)


brandonweeks:
@domcleal this should most likely wait for 1.11, it doesn't repair any functionality.

domcleal:
Ah, because the exception only generates a warning message and skips it anyway?

brandonweeks:
Exactly. 

> W, [2015-11-13T05:09:15.403972 #26474]  WARN -- : skipped test.example.com - address family must be specified

This is the error that surfaces in the log, unfortunately for me test.example.com also happened to be the host that was suffering from theforeman/smart-proxy#339. So this had me barking up the wrong tree. :)

domcleal:
Right, I see.  It might still be important to get in soon, since it would ensure that the record is deleted in the memory store when marked as deleted in the leases file.  It looks like we could add the entry but not then call `@service.delete_host` since it's being skipped.

Certainly fixing the other PR will fix the handling if we did then try to delete a non-existent host.

domcleal:
> It looks like we could add the entry but not then call @service.delete_host since it's being skipped.

Oh, perhaps not.  It looks like we correctly "add" records with a deleted flag, but I was thinking of a case where we add a record with an IP, then see a deleted record without an IP.  I don't know if ISC does such a thing - certainly the example in the test has an IP in both the active and deleted leases...

modules/dhcp/providers/server/isc.rb
```diff
@@ -78,13 +78,10 @@ def parse_config_and_leases_for_records
...
           end
         end
-        begin
-          subnet = @service.find_subnet(opts[:ip])
-          next unless subnet
-          ret_val << Proxy::DHCP::Reservation.new(opts.merge(:subnet => subnet))
-        rescue Exception => e
-          logger.warn "skipped #{hostname} - #{e}"
-        end
+        next if !opts.has_key?(:ip)
```
domcleal: I don't think we should skip the entry here, instead we probably need to skip the find_subnet call and then create a reservation with no subnet assigned so it can be deleted.

Looking back at the behaviour of 5f7fd34, it would delete the entry from the in memory store: https://github.com/theforeman/smart-proxy/blob/5f7fd34c7b4dc9b373a5fc1faeef4f21bcbc7919/modules/dhcp/providers/server/isc.rb#L68-L77

If we apply this as-is, we'll get entries in the store (which show up on GET /dhcp/:subnet) and we'll also need your other PR to fix deletion of entries that are actually non-existent.

domcleal: Would you also mind adding a test (if you do the above) to check that the entry isn't showing up in the subnet data?

dmitri-d: Yeah, we shouldn't skip deleted entries, otherwise memory store may end up being initialized with stale records. However, It's not possible to create a reservation without a valid subnet at the moment: there a validator in the base class that will throw an exception if the subnet is missing (https://github.com/theforeman/smart-proxy/blob/develop/modules/dhcp/record.rb#L14). I think splitting Reservation class into Reservation and DeletedReservation (where subnet is optional) would provide a clean-ish solution.

brandonweeks:
@domcleal @witlessbird thanks for digging into this, your explanations have really helped me understand what the problem is. I'm updating the name/description to reflect the underlying issue.

brandonweeks:
This seems like the best way to solve the issue now that I understand exactly what's happening. This PR now fixes #12465 and #12469 without requiring f24be74d8f51f44aa8730769d2e468683804254c.
- Modified the test to reflect the deleted entry syntax I'm seeing with my production ISC dhcpd.
- Remove records from `ret_val` in `parse_config_and_leases_for_records` if we find them as deleted later in the config.

This seems like the simplest way to solve the problem, thoughts?

modules/dhcp/providers/server/isc.rb
```diff
@@ -78,13 +78,13 @@ def parse_config_and_leases_for_records
...
           end
         end
-        begin
-          subnet = @service.find_subnet(opts[:ip])
-          next unless subnet
-          ret_val << Proxy::DHCP::Reservation.new(opts.merge(:subnet => subnet))
-        rescue Exception => e
-          logger.warn "skipped #{hostname} - #{e}"
+        if opts[:deleted]
+          ret_val.each { |rec| ret_val.delete(rec) if rec.options[:hostname] == hostname }
```
domcleal: Use [#delete_if](http://ruby-doc.org/core-2.2.3/Array.html#method-i-delete_if) here to avoid the .each { .. delete } pattern.

domcleal:
> Modified the test to reflect the deleted entry syntax I'm seeing with my production ISC dhcpd.

Yep, seems the tests for subnet data and the API response are checking this is working.

> Remove records from ret_val in parse_config_and_leases_for_records if we find them as deleted later in the config.

This seems OK to me.  There's a chance it'll have a performance penalty relative to the number of records with big files, but the alternative is adding it as a DeletedReservation as @witlessbird mentioned, then removing it later (actually, it seems some of that code's still there, do we need it?  https://github.com/theforeman/smart-proxy/blob/1.10.0-RC2/modules/dhcp/providers/server/isc.rb#L113-L117).

I think that'll need a larger refactoring though to ensure the subnet service could actually deal with the entry without any subnet information, so I guess stick with this for now.

modules/dhcp/providers/server/isc.rb
brandonweeks:
```
next if ret_val.delete_if { |rec| rec.options[:hostname] == hostname }
```

Do you think this would be preferable so we raise an exception if we don't match anything to remove? Do you think adding a debug log makes sense here?

domcleal: Adding a debug log message if we fail to remove anything seems like a good idea.

modules/dhcp/providers/server/isc.rb
```diff
@@ -78,13 +78,13 @@ def parse_config_and_leases_for_records
...
-        begin
-          subnet = @service.find_subnet(opts[:ip])
-          next unless subnet
-          ret_val << Proxy::DHCP::Reservation.new(opts.merge(:subnet => subnet))
-        rescue Exception => e
-          logger.warn "skipped #{hostname} - #{e}"
+        if opts[:deleted]
+          ret_val.delete_if { |rec| rec.options[:hostname] == hostname }
+          next
         end
```
dmitri-d, line 85: This bit executes in polynomial time (and O(n*n) for the whole lease file), which isn't good, as lease file parsing is pretty slow already. You could replace the array here with a hashmap indexed by hostname, or introduce support for subnet-less reservations (pls. see my earlier comment).

dmitri-d:
@brandonweeks: ping?

brandonweeks:
@witlessbird ack. These past few days has been super busy for my team. I'm hoping to get a chance to rewrite this PR this weekend. :)

modules/dhcp/providers/server/isc.rb
sodabrew, line 85: The leases file has to be parsed in order; a deleted statement will apply to earlier entries, and then subsequent entries should be allowed through. so...yeah, actually a hash by hostname should be fine. I actually can't believe I never noticed that deleted entries weren't being parsed. This has been driving my team crazy for... two years?

sodabrew:
Corroborating @brandonweeks findings, this is what a created, then deleted, then updated entry  looks like in my ISC DHCPD (noting that the same dummy MAC address appears for foo and bar):

```
host foo {
  dynamic;
  hardware ethernet 00:11:22:33:44:55;
  fixed-address 172.2.3.4;
        supersede server.next-server = 1:2:3:4;
        supersede host-name = "foo";
}
host foo {
  dynamic;
  deleted;
}
host bar {
  dynamic;
  hardware ethernet 00:11:22:33:44:55;
  fixed-address 172.2.3.4;
        supersede server.next-server = 1:2:3:4;
        supersede host-name = "bar";
}
```

brandonweeks:
@sodabrew I'll update the test cases to include the created/deleted/created edge case as well. Thanks for pointing that out!

modules/dhcp/subnet_service.rb
```diff
@@ -93,13 +97,13 @@ def find_host_by_ip(subnet_address, ip_address)
     @reservations_by_ip[subnet_address, ip_address]
   end
 
-  def find_host_by_hostname(subnet_address, hostname)
-    @reservations_by_name[subnet_address, hostname]
+  def find_host_by_hostname(hostname, subnet_address = nil)
```
dmitri-d: The only place where find_host_by_hostname was used was https://github.com/theforeman/smart-proxy/blob/develop/modules/dhcp/providers/server/isc.rb#L114, which you changed. You can drop subnet_address from the method parameters now; all_hosts method can use either reservations_by_ip or reservations_by_mac to return host records for a subnet. After that you can get rid of reservations_by_subnet :).

dmitri-d:
looks good other than the changes I suggested around `SubnetService#find_host_by_hostname`.

brandonweeks:
@witlessbird well that's much cleaner. PR updated.

dmitri-d:
[test]

test/fixtures/dhcp/dhcp.leases
```diff
@@ -13,19 +13,33 @@ host ten.example.com {
...
   dynamic;
-  hardware ethernet 10:10:10:10:20:20;
-  fixed-address 192.168.122.20;
-  supersede host-name = "deleteme.example.com";
   deleted;
 }
+host undeleted.example.com {
+  dynamic;
+  hardware ethernet 10:10:10:10:30:30;
+  fixed-address 192.168.122.30;
```
sodabrew: Could you make this a different IP, and ensure in the tests that only the new IP is in use, and that the prior IP is available?

brandonweeks:
@sodabrew done.

sodabrew:
Rubocop is flagging this:
http://ci.theforeman.org/job/test_proxy_develop_pr_rubocop/484/checkstyleResult/source.23/#71

Says not to use a nested method definition here:

```
80   def test_delete_returns_error_if_value_is_missing
81     def test_create_returns_error_if_type_is_missing
82       delete '/'
83       assert_equal 400, last_response.status
84     end
85   end
```

sodabrew:
Actually that is in the DNS specs, not DHCP; unrelated failure to this PR.

brandonweeks:
Rubocop got upgraded in theforeman/smart-proxy#346, might be related. 

Edit: Well, that test doesn't seem to make much sense.

sodabrew:
I submitted a fix for the DNS API tests in #350

dmitri-d:
[test]

dmitri-d:
Made a tiny fix in test/dhcp/subnet_service_test.rb -- added 'test_' to test names that were missing it and merged in d981fe9. Thanks, @brandonweeks!

sodabrew:
Will this also get cherry-picked to the 1.10 branch?

domcleal:
@sodabrew yep, it will for the next 1.10 release (either .0 or RC4)

PR #340, opened 2015-11-16
Fixes #12466: isc dhcp provider handles subnet block declarations correctly now.

dmitri-d:


test/dhcp/server_isc_test.rb
```diff
@@ -79,6 +79,35 @@ def test_poap_quirks
...
+    assert "subnet 192.168.123.0 netmask 255.255.255.192 {filename \"filename\";}".match(Proxy::DHCP::ISC::SUBNET_BLOCK_REGEX)
+  end
+
+  def test_subnet_matching_with_declaration
+    assert "subnet 192.168.123.0 netmask 255.255.255.192 {pool{range 192.168.42.200 192.168.42.254;}}".match(Proxy::DHCP::ISC::SUBNET_BLOCK_REGEX)
+  end
+
+  def test_subnet_matching_with_declaration_and_parameter
+    assert "subnet 192.168.123.0 netmask 255.255.255.192 {pool{range 192.168.42.200 192.168.42.254;}option subnet-mask 255.255.255.192;}"
+             .match(Proxy::DHCP::ISC::SUBNET_BLOCK_REGEX)
```
brandonweeks: The tests fail on Ruby 1.8.7 here.

dmitri-d:
@brandonweeks: this should do the trick.

brandonweeks:
This resolves the issue in my environment. Thanks @witlessbird! :100:

modules/dhcp/providers/server/isc.rb
```diff
@@ -146,31 +146,34 @@ def loadSubnetData subnet
       initialize_memory_store_with_dhcp_records(parse_config_and_leases_for_records)
     end
 
+    SUBNET_BLOCK_REGEX = /subnet\s+([\d\.]+)\s+netmask\s+([\d\.]+)\s*\{([\w-]+\{[^{}]*\}|[\w-][^{}]*;)*\}/
```
domcleal: This removes the \s that was permitted between the two curly braces in the earlier version, so for example, this now fails:

```
assert "subnet 192.168.1.0 netmask 255.255.255.128 {pool\n{range abc;}}".match(Proxy::DHCP::ISC::SUBNET_BLOCK_REGEX)
```

It needs to be rather liberal in its accepting of whitespace.  Perhaps we should just match the start and end of the block with curly braces?  What's in the middle doesn't matter too much as it's parsed below.

domcleal: https://github.com/theforeman/puppet-dhcp/blob/master/templates/dhcpd.pool.erb is an example of just how messy a subnet definition can be.

dmitri-d: re comment #1: will fix the space issue; We can't completely ignore what's inside of curly braces, otherwise regex will either match everything between the first and the last curly braces (eager matching) or between the first and the second ones (lazy matching) (I did briefly consider rewriting this bit using a normal parser. I may return to this idea in the future if maintaining these regexes will turn out to be too much pain). Th would work with simpler subnet definitions, but will break with subnet definitions that contain pool and other similar definitions (I believe this is what happened in @brandonweeks setup).

re #2: Yeah, subnet definition can be quite messy; I went over the docs and (I think) covered possible use-cases with tests.

dmitri-d:
@domcleal: fixed the issue with spaces.

domcleal:
Merged as 2f513ec, thanks @witlessbird.

PR #341, opened 2015-11-17
Revert "Fixes #12469 - raise Proxy::DHCP::InvalidRecord"

brandonweeks:
This reverts commit f24be74d8f51f44aa8730769d2e468683804254c.


brandonweeks:
With the updated theforeman/smart-proxy#338 this change is no longer required. Since `parse_config_and_leases_for_records` removes the deleted records, `delete "/:network/:record"` will correctly return 404 as intended.

dmitri-d:
It is possible that a record gets deleted between lease file loading and omshell's "open" command execution. The original change handles this (pretty unlikely) scenario and I'd rather keep it.

brandonweeks:
I'm neutral if this should be merged or not. :)

domcleal:
Yeah, I liked this change as a fallback too, I think it has value even once the other bug's fixed.

brandonweeks:
The people have spoken!

PR #342, opened 2015-11-20
Fixes #12546: smart-proxy now loads under bundler_ext

dmitri-d:


dmitri-d:
@mmoll, @domcleal: ping

mmoll:
Tested with the Debian package that was broken before, works for me :+1:

domcleal:
Merged as 59809c6, thanks for the quick fix @witlessbird.

PR #345, opened 2015-11-24
Fixes #12573 - disable TLSv1

brandonweeks:
> [root@host smart-proxy]# cat /etc/redhat-release
> CentOS release 6.7 (Final)
> [root@host smart-proxy]# ruby -v
> ruby 1.8.7 (2013-06-27 patchlevel 374) [x86_64-linux]

Even on CentOS 6 (both Foreman and Smart-Proxy) with Ruby 1.8.7 we have TLSv1.1 and TLSv1.2 support.


domcleal:
We ought to check this works on our other older supported OSes (http://theforeman.org/manuals/1.10/index.html#3.1.1SupportedPlatforms), particularly 12.04 and Wheezy.

brandonweeks:
This is on hold, I'm working with our infosec team to validate these PRs.

PR #349, opened 2015-11-24
Fixes #12580 - configure WEBrick to use ::Proxy::Log.logger

brandonweeks:
Configures WEBrick to use the ::Proxy::Log.logger.


domcleal:
Ticket number should be #12580, not 12850.  While I'm here, should we also add the logger to the HTTP rack app?

brandonweeks:
@domcleal absolutely, good catch. :) Added and fixed commit message.

dmitri-d:
merged in at 63a513a. Thanks, @brandonweeks!

PR #350, opened 2015-11-24
Fixes #12597: remove extraneous method definition in the DNS API tests

sodabrew:
http://projects.theforeman.org/issues/12597

### DIFF
--- a/modules/dns/dns_api.rb
+++ b/modules/dns/dns_api.rb
@@ -28,7 +28,6 @@ module Proxy::Dns
     end
 
     delete "/:value" do
-      log_halt(400, "'remove' requires value parameter") if params[:value].nil?
       type = params[:value].match(/\.(in-addr|ip6)\.arpa$/) ? "PTR" : "A"
 
       begin
--- a/test/dns/dns_api_test.rb
+++ b/test/dns/dns_api_test.rb
@@ -78,9 +78,7 @@ class DnsApiTest < Test::Unit::TestCase
   end
 
   def test_delete_returns_error_if_value_is_missing
-    def test_create_returns_error_if_type_is_missing
-      delete '/'
-      assert_equal 400, last_response.status
-    end
+    delete '/'
+    assert_equal 404, last_response.status
   end
 end


sodabrew:
@witlessbird Could you take a look at the test failure? I removed the extra method definition, which seems to be blocking this test before, and now that it is executing it fails. Was originally added in 42b958156808eb2a16bd895863b196fca8f2977d

```
   def test_delete_returns_error_if_value_is_missing
     delete '/'
     assert_equal 400, last_response.status
   end
```

Smart Proxy is returning 404 in this case. That seems correct to me?

domcleal:
Oops, sorry about the failure, I should've noticed that in the review and also re-run tests with the new rubocop.

Either 404 or 400 seems reasonable to me.  Returning 400 as dns_api.rb intends would require changing the route to `/:value?`, but since the value _isn't_ actually optional (that's the point of the nil check) then returning 404 seems OK as the route `DELETE /` isn't valid.  I'd say removing the check and expecting 404 is probably OK.

dmitri-d:
I'm not sure why it's returning 404 (this implies parameter validator isn't catching that the required parameter is missing), let me look into this. I think I favour returning a 400, but 404 seems reasonable too.

domcleal:
> I'm not sure why it's returning 404 (this implies parameter validator isn't catching that the required parameter is missing)

The route doesn't match.  It's defined as `delete '/:value'` which implies :value must be present.  Adding a `?` would make the value optional, which would trigger the validator.

domcleal:
> Adding a ? would make the value optional, which would trigger the validator.

A thought - it looks like all other delete routes are the same, so we probably shouldn't change just this one.

dmitri-d:
Of course, always forget this bit. Agree that the behaviour in this case should  match other 'delete' calls.

sodabrew:
@witlessbird I merged in your further change - thank you! I will be offline a lot this week for the US Thanksgiving holiday. Please let me know if I need to squash on this branch, if it it'll get taken care of in a squash-cherry-pick to develop?

dmitri-d:
@sodabrew: could you squash the commits please?

sodabrew:
ok, got me while I'm still sitting at the desk!

dmitri-d:
merged in bef9bee. Thanks, @sodabrew!

PR #348, opened 2015-11-24
Fixes #12578 - set SSL_OP_CIPHER_SERVER_PREFERENCE

brandonweeks:

PR #347, opened 2015-11-24
Fixes #12576 - add SETTINGS.ssl_dhparam param for custom DH group

brandonweeks:
More details: http://projects.theforeman.org/issues/12576

This is a first stab at fixing the issue, the defaults are pretty poor still however.

PR #344, opened 2015-11-24
Fixes #12572 - update ciphersuite

brandonweeks:
https://wiki.mozilla.org/Security/Server_Side_TLS#Modern_compatibility

Updated the ciphersuite to Mozilla's "Modern compatibility" since we don't need legacy browser support for server-to-server communication.

PR #351, opened 2015-11-25
Fixes #12572 - refactor openssl settings

brandonweeks:
Obsoletes #344 #345 #347 #348 and fixes all of their corresponding security issues.
- Update cipher suites to sane defaults
  - Disable RC4 cipher suites  (RFC 7465)
  - Disable ECDHE  cipher suites since they don't work (filed as Ruby#11739)
  - Disable DHE cipher suites  since Ruby only ships with <1024-bit DH groups. #347 also solved the problem if the enduser bothered to generate a custom DH group, simply removing the cipher suites is a better solution to the problem.
- Sets `SSL_OP_CIPHER_SERVER_PREFERENCE`
- Sets `TLSv1_2_server_method`


dmitri-d:
[test]

lib/webrick-patch.rb
```diff
@@ -0,0 +1,36 @@
+require 'webrick/https'
+
+CIPHERS = 'AES128-GCM-SHA256:AES256-GCM-SHA384:AES128-SHA256:AES256-SHA256'
```
dmitri-d: I assume this is based on https://wiki.mozilla.org/Security/Server_Side_TLS minus ECDHE and DHE ciphers? Could you add a comment re: selection criteria here?

lib/launcher.rb
```diff
@@ -40,6 +40,9 @@ def https_app
           ::Proxy::Plugins.enabled_plugins.each {|p| instance_eval(p.https_rackup)}
         end
 
+        ssl_options = OpenSSL::SSL::SSLContext::DEFAULT_PARAMS[:options]
```
dmitri-d, line 43: Should we add `OpenSSL::SSL::OP_NO_COMPRESSION` option too? Looks like tls compression is on by default...

dmitri-d:
@brandonweeks: thanks for venturing into something pretty intricate -- ssl configuration! @domcleal: could you take a look when you have a chance (or perhaps you know someone knowledgeable in SSL ways who can take a look?). Looks ok to me, my only concern is that the list of ciphers is rather small, but then all of them seem to be widely available.

lib/webrick-patch.rb
```diff
@@ -0,0 +1,36 @@
...
+SSL_VERSION = 'TLSv1_2_server'
```
domcleal: On EL6 with Ruby 1.8.7 this fails: `E, [2015-12-01T15:06:25.054465 #22701] ERROR -- : Error during startup, terminating. unknown SSL method 'TLSv1_2_server'.`

dmitri-d: Could you post a traceback?

brandonweeks: > [weeks@foreman ~]$ irb
> irb(main):001:0> require 'openssl'
> => true
> irb(main):002:0> OpenSSL::SSL::SSLContext::METHODS
> => [:TLSv1, :TLSv1_server, :TLSv1_client, :SSLv2, :SSLv2_server, :SSLv2_client, :SSLv3, >:SSLv3_server, :SSLv3_client, :SSLv23, :SSLv23_server, :SSLv23_client]
> irb(main):003:0>

That makes sense actually, I must have missed testing my latest code against el6. I'll update the PR.

domcleal:
This may not exactly be our problem, but it's a slight concern.  I'm unable to connect using curl on EL6 or EL7.  On EL6 (where I've removed the SSL_VERSION spec) the error's vague:

<pre>
* About to connect() to 0.0.0.0 port 8443 (#0)
*   Trying 0.0.0.0... connected
* Connected to 0.0.0.0 (0.0.0.0) port 8443 (#0)
* Initializing NSS with certpath: sql:/etc/pki/nssdb
*   CAfile: /var/lib/puppet/ssl/certs/ca.pem
  CApath: none
* NSS error -12286
* Closing connection #0
* SSL connect error
curl: (35) SSL connect error
</pre>

While on EL7 (with this exact patch) it's failing to connect under TLS 1.2 and looks like it's attempting an SSLv3 fallback and failing:

<pre>
* About to connect() to 0.0.0.0 port 8443 (#0)
*   Trying 0.0.0.0...
* Connected to 0.0.0.0 (0.0.0.0) port 8443 (#0)
* Initializing NSS with certpath: sql:/etc/pki/nssdb
*   CAfile: /var/lib/puppet/ssl/certs/ca.pem
  CApath: none
* NSS error -12190 (SSL_ERROR_PROTOCOL_VERSION_ALERT)
* Peer reports incompatible or unsupported protocol version.
* Error in TLS handshake, trying SSLv3...
> GET /features HTTP/1.1
> User-Agent: curl/7.29.0
> Host: 0.0.0.0:8443
> Accept: */*
> 
* Connection died, retrying a fresh connect
* Closing connection 0
* Issue another request to this URL: 'https://0.0.0.0:8443/features'
* About to connect() to 0.0.0.0 port 8443 (#1)
*   Trying 0.0.0.0...
* Connected to 0.0.0.0 (0.0.0.0) port 8443 (#1)
* TLS disabled due to previous handshake failure
*   CAfile: /var/lib/puppet/ssl/certs/ca.pem
  CApath: none
* NSS error -12286 (SSL_ERROR_NO_CYPHER_OVERLAP)
* Cannot communicate securely with peer: no common encryption algorithm(s).
* Closing connection 1
curl: (35) Peer reports incompatible or unsupported protocol version.
</pre>

My feeling's that we should probably aim to have the cipher and protocol versions set as high as possible, and perhaps even consider defining it per OS or Ruby version (if applicable) so newer OSes get more secure settings.

It probably does need to be configurable though - there are going to be people with more mixed infrastructures who may need more relaxed configurations.  Do you agree?

I think we'll need to do much more testing on our other supported OSes, I've only done this quickly.

lib/webrick-patch.rb
domcleal: > Could you post a traceback?

<pre>
D, [2015-12-01T15:21:21.188543 #23426] DEBUG -- : unknown SSL method `TLSv1_2_server'.:/usr/share/foreman-proxy/lib/webrick-patch.rb:18:in `ssl_version='
/usr/share/foreman-proxy/lib/launcher.rb:125:in `join'
/usr/share/foreman-proxy/lib/launcher.rb:125:in `launch'
/usr/share/foreman-proxy/bin/smart-proxy:6
</pre>

dmitri-d:
I was wondering about that too; I don't think this is a ruby-version constraint, but rather an openssl one? We could define an order in which ssl versions should be tried, ie. TLSv1_2_server,  TLSv1_server, etc (I can't imagine a version of openssl not supporting TLSv1_server), and suites of ciphers for each (which I think is useful, as without it I ended up with an ssl connection with a weaker cipher). This would be less work than explicitly defining versions and ciphers per os...

brandonweeks:
@witlessbird I'm pretty sure it's Ruby, the OpenSSL version on el6 is rather up to date. From what I can tell for a protocol to appear in `OpenSSL::SSL::SSLContext::METHODS` it has to be defined in `ext/openssl/ossl_ssl.c`.

Ruby 1.8.7: https://github.com/ruby/ruby/blob/ruby_1_8_7/ext/openssl/ossl_ssl.c#L100
TLSv1.1 and TLSv1.2 support: https://github.com/ruby/ruby/commit/060184c347822b11dff3db6bef915c04a564c4e4

domcleal:
> I don't think this is a ruby-version constraint, but rather an openssl one?

I was imagining it was a bit of both, but I'm not very familiar with Ruby's SSL integration.  It might be that OpenSSL has certain capabilities, but a lack of equivalent constants in Ruby?

dmitri-d:
@brandonweeks: I stand corrected! I'll see what I can find out about ciphers...

brandonweeks:
I've updated the PR to enable TLSv1.2 implicitly instead of explicitly, which fixes the proxy not starting on el6/ruby 1.8.7. Both el6 and el7 have an OpenSSL version capable of connecting to the proxy with the current cipher suits, however curl apparently doesn't have all of the ciphers that OpenSSL does.

`openssl s_client -connect proxy:8443`

> SSL-Session:
>  Protocol  : TLSv1.2
>  Cipher    : AES128-GCM-SHA256

It looks like curl is compiled against NSS, not OpenSSL. Here's an article from RH (:100:) detailing how to see what ciphers NSS supports: https://access.redhat.com/solutions/1154393

https://gist.github.com/brandonweeks/03505e232ae24edb3c1b

Here's the `sslenum.c` from el6, `TLS_RSA_WITH_AES_128_GCM_SHA256` is present, so it should be working in theory.

Related bug: https://bugzilla.redhat.com/show_bug.cgi?id=1059682

So in summary it appears that TLS_RSA_WITH_AES_\* ciphers are broken in EL6 (and maybe EL7) Then TLS_DHE_RSA_WITH_AES\* and TLS_ECDHE_RSA_WITH_AES\* are broken in Ruby. Security is hard. :sob:

lib/webrick-patch.rb
brandonweeks: Well, not exactly. If you take Mozilla's "Modern" configuration and remove ECDHE+DHE you are left with nothing. This is because they're optimizing for PFS, which is less relevant to us (but still would be nice to have).

lib/launcher.rb
brandonweeks, line 43: Where are you seeing compression on by default? It was patched to be disabled as part of #9424 in 1.9.3 and newer. It also appears that RH has patched el6 to disable it as well.

brandonweeks:
Cipher suites tweaked a little bit to include two SSLv3 SHA1 based ciphers. It's not the _best_ option but it's better than the current state, at least until I can get some of the upstream issues addressed.

domcleal:
Seems to be OK to me, and I've tested it on EL6 and Ubuntu 12.04 without any issues.  The latter's curl is compiled against OpenSSL, so it doesn't have a problem finding a compatible cipher.

I'm OK with a smaller ciphers list by default, but I do think we should make it configurable, otherwise if somebody is relying on using curl+NSS etc. with the smart proxy then it'd require a code change.

dmitri-d:
I'm not sure re: configurable cipher list. In one of the examples I saw the list of ciphers was defined as: `OpenSSL::SSL::SSLContext::DEFAULT_PARAMS[:ciphers] = "TLSv1.2:!aNULL:!eNULL"`. We could do something similar, but include TLSv1 and TLSv1.1 suites (when needed, similarly to how we disable those when possible). This is more flexible, but also makes it possible to end up with a comparatively weak cipher.

brandonweeks:
@witlessbird unfortunately there are really only "SSLv3" and "TLS1.2" ciphers. TLSv1 and TLSv1.1 didn't bring support for modern cipher suites. With that cipher list I get `curl: (35) SSL connect error` when trying to connect from el6 and it is also vulnerable to Weak DH with Ruby. Given all of the upstream problems I enumerated above, we're not left with much to work with.

Do you think it's worth engaging the EL team (we're a customer, if it matters) in getting TLSv1.2 ciphers working with NSS in el6/el7? Fixing ECDHE needs to happen with Ruby upstream and then the distros, so I'm afraid that one is the farthest off. Alas, I'm assuming that the el6/el7 would still need to be updated to support those.

dmitri-d:
@brandonweeks: Hmm, yeah, in that case hand-picked list of ciphers is probably best. A possibility for an override might come handy too. 

Might be worth asking re: TLSv1.2 support -- it might happen, depending on complexity/support issues.

brandonweeks:
Never mind. "Known issue" doesn't begin to describe this.
- https://bugzilla.mozilla.org/show_bug.cgi?id=923089
- https://bugzilla.redhat.com//show_bug.cgi?id=1221946
- https://bugzilla.redhat.com/show_bug.cgi?id=1185708
- https://bugzilla.redhat.com/show_bug.cgi?id=1059670 (RHBA-2015:2121-1)
- https://bugzilla.redhat.com/show_bug.cgi?id=1202494

> Yes, there is progress. Enabling of ECC cipher-suites by default is currently under test by Red Hat. It's targeted for the next RHEL-7 y-stream release (RHEL-7.2) and we have submitted a subset of the patches upstream. I'll be bringing this to fedora once we have positive feedback from Red Hat QE. Hopefully, it won't take too long.

And a few bugs I can't see. So it seems like some improvement will be landing in 7.2 and possibly even longer for complete TLSv1.2 cipher support. I don't think there is any getting around AES+RSA+SHA1 for the time being.

dmitri-d:
I'm going to suggest to merge this as is and add a configuration option for ssl ciphers in a separate PR. Any objections?

domcleal:
Fine by me!

brandonweeks:
> ECDHE-RSA-AES128-GCM-SHA256:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-RSA-AES128-CBC-SHA:ECDHE-RSA-AES256-CBC-SHA:AES128-GCM-SHA256:AES256-GCM-SHA384:AES128-SHA256:AES256-SHA256:AES128-SHA:AES256-SHA

I propose tweaking the cipher suite to this one last time before merge. This is the same as the current diff except it is made explicit to prefer 128 bit over 256 bit for performance and readds the ECDHE ciphers so we don't have to make any changes if Ruby gets fixed.

dmitri-d:
@brandonweeks: sounds good to me; could you update the PR pls?

brandonweeks:
Updated.

dmitri-d:
merged in at 508a9e8. Thanks, @brandonweeks!

PR #353, opened 2015-12-09
Fixes #12555: switched to session-based freeipa api

dmitri-d:


dmitri-d:
[test]

modules/realm/freeipa.rb
```diff
@@ -39,17 +40,23 @@ def initialize
...
-        gssapi = GSSAPI::Simple.new(@ipa_server.host, "HTTP")
-        token = gssapi.init_context
+        @gssapi = GSSAPI::Simple.new(@ipa_server.host, "HTTP")
+        token = @gssapi.init_context
+
+        login = Net::HTTP.new(@ipa_server.host, 443)
+        login.use_ssl = true
+        login.verify_mode = OpenSSL::SSL::VERIFY_NONE
+
+        request = Net::HTTP::Post.new("/ipa/session/login_kerberos", 'Authorization'=>"Negotiate #{strict_encode64(token)}", 'Referer' => @ipa_server.to_s)
```
domcleal, line 50: Is this available in all versions, or just with the most recent update/change?

dmitri-d, line 50: Looks like support for sessions for xml api was introduced in freeipa version 3.0. That login end-point first appeared in version 2.2, I think.

domcleal:
@stbenjam did you want to look this over at all?

domcleal:
Merged as 07a6dae, thanks @witlessbird.

PR #354, opened 2015-12-11
Fixes #12775 - Handle nonexistent dhcp_vendor

phirince:
If the dhcp_vendor options is commented out, Proxy::DhcpPlugin.settings.dhcp_vendor is nil and downcase will not work on it. Detect early and error out.

-PP

### DIFF
--- a/modules/dhcp/dhcp_api.rb
+++ b/modules/dhcp/dhcp_api.rb
@@ -7,6 +7,7 @@ class Proxy::DhcpApi < ::Sinatra::Base
   before do
     begin
       raise "Smart Proxy is not configured to support DHCP" unless Proxy::DhcpPlugin.settings.enabled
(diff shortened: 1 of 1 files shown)


domcleal:
Thanks, this looks like a useful bug fix.  A big refactor of the DHCP module has just been merged which is why the code now conflicts.  The `use_provider` setting also now has a default value of `dhcp_isc`.

Do you want to reopen the PR against the 1.10-stable branch instead, and we can ship it in a patch release sooner?

phirince:
I've opened https://github.com/theforeman/smart-proxy/pull/355 again 1.10-stable branch

dmitri-d:
FYI: Starting with https://github.com/theforeman/smart-proxy/commit/42b958156808eb2a16bd895863b196fca8f2977d settings listed under 'default_settings' are checked during startup that their values are not nil. 

Looks good to me for 1.10.

PR #356, opened 2015-12-21
Fixes #12896 - Allow proxy to count classes

tbrisker:


dmitri-d:
Why introduce these end points? Class counts can be derived from already available api (I don't think we care that much about bandwidth use in this case). Do we need an endpoint for a total number of classes, seeing that we don't have a corresponding one for classes in all environments?

Also -- rubocop isn't happy.

domcleal:
House-keeping again: please don't use "refs" on an unshipped ticket in another project.  Global rule: use "fixes", with one ticket per commit.

tbrisker:
@witlessbird The logic is that right now (if I understand correctly) to get a class count we would need to iterate over all the environments, query each one separately and sum it all up, instead of getting a concise summery directly from the proxy (which might be possible to calculate more efficiently then it does right now? I'm still not familiar enough with the inner workings of the proxy). About the total count perhaps we can do without another endpoint and calculate it from the detailed count.

tbrisker:
Oh, and [test] failures seem to be unrelated.

modules/puppet_proxy/puppet_api.rb
```diff
@@ -70,4 +70,22 @@ def puppet_setup(opts = {})
       log_halt 406, "Failed to show puppet classes: #{e}"
     end
   end
+
+  get "/class_count" do
```
dmitri-d, line 74: Can we keep just the "/environments/class_count" end point? You could include a "total" key in it...

dmitri-d:
Could you add a test for the new end point?

tbrisker:
I'm closing this right now as preliminary testing indicates that pulling the counts from the existing end points should be fast enough. Will reopen if we run into a performance issue with this in the future and wish to push the processing to the capsule.

PR #357, opened 2016-01-12
fixes #13062 - handle missing or non-IP fixed-address ISC fields

domcleal:


domcleal:
The test failures in the native MS test appear to be intermittent failures on develop under 1.8.7.  Filed as http://projects.theforeman.org/issues/13129.

dmitri-d:
Nods, was looking into those; Looks good otherwise.

dmitri-d:
merged in at 31ab121. Thanks, @domcleal!

PR #358, opened 2016-01-12
Fixes #12715 - simple log buffer and API

lzap:
As per our discussion on the IRC, here is my attempt to implement Dmitri's
event system in a much more simplified way. Since we want something very simple
and efficient, this is essentially a ring buffer of structs (events), which is
synchronized.

There is only one buffer for all modules, there is a special "tail" buffer that
holds all the old values from the main buffer that has WARNING or ERROR level.
This way we won't loose important errors. The buffer can be large as this
implementation is fast and efficient.

The API is trivial, backtraces are not passed in the parameter, but you get the
idea. Our main goal is to get logs and important messages from proxy to
foreman. This accomplishes it easily and this buys us time to integrate with
other systems that does this thing well. Collectd comes to my mind, it actually
contains exactly same mechanism we need (it's called notifications and I
already have an implementation for smart-proxy).

I created this POC because I don't want us to reinvent the wheel building our
own messaging system for value readouts and notifications. Instead, let's
create a simple API and save time for other important things.

```
$ curl -sk http://localhost:8448/events | json_reformat 
{
    "events": [
        {
            "timestamp": 1452615190,
            "level": 300,
            "message": "'discovery' settings were initialized with default values: :node_port: 8443, :node_scheme: https",
            "backtrace": [

            ]
        },
        {
            "timestamp": 1452615190,
            "level": 300,
            "message": "Finished initialization of module 'discovery'",
            "backtrace": [

            ]
        },
        {
            "timestamp": 1452615190,
            "level": 300,
            "message": "Finished initialization of module 'discovery_image'",
            "backtrace": [

            ]
        },
        {
            "timestamp": 1452615190,
            "level": 300,
            "message": "Finished initialization of module 'salt'",
            "backtrace": [

            ]
        },
        {
            "timestamp": 1452615190,
            "level": 300,
            "message": "Finished initialization of module 'abrt'",
            "backtrace": [

            ]
        },
        {
            "timestamp": 1452615190,
            "level": 300,
            "message": "'colly' settings were initialized with default values: :socket: /var/run/collectd-unixsock, :timeout: 5",
            "backtrace": [

            ]
        },
        {
            "timestamp": 1452615190,
            "level": 100,
            "message": "Couldn't enable plugin colly: No such file or directory - \"/var/run/collectd-unixsock\"",
            "backtrace": [

            ]
        },
        {
            "timestamp": 1452615190,
            "level": 400,
            "message": "No such file or directory - \"/var/run/collectd-unixsock\":/home/lzap/work/collectd-uxsock/lib/uxsock/collectd_unix_socket.rb:12:in `initialize'\n/home/lzap/work/collectd-uxsock/lib/uxsock/collectd_unix_socket.rb:12:in `open'\n/home/lzap/work/collectd-uxsock/lib/uxsock/collectd_unix_socket.rb:12:in `initialize'\n/home/lzap/work/collectd-uxsock/lib/uxsock/collectd_unix_socket.rb:22:in `new'\n/home/lzap/work/collectd-uxsock/lib/uxsock/collectd_unix_socket.rb:22:in `open'\n/home/lzap/work/smart_proxy_colly/lib/smart_proxy_colly/writer.rb:31:in `notify'\n/home/lzap/work/smart_proxy_colly/lib/smart_proxy_colly/colly.rb:14:in `block in <class:Plugin>'\n/home/lzap/work/smart-proxy/lib/proxy/pluggable.rb:33:in `instance_eval'\n/home/lzap/work/smart-proxy/lib/proxy/pluggable.rb:33:in `after_activation'\n/home/lzap/work/smart-proxy/lib/proxy/plugin.rb:138:in `configure_plugin'\n/home/lzap/work/smart-proxy/lib/proxy/plugin.rb:29:in `block in configure_loaded_plugins'\n/home/lzap/work/smart-proxy/lib/proxy/plugin.rb:29:in `each'\n/home/lzap/work/smart-proxy/lib/proxy/plugin.rb:29:in `configure_loaded_plugins'\n/home/lzap/work/smart-proxy/lib/launcher.rb:111:in `launch'\nbin/smart-proxy:6:in `<main>'",
            "backtrace": [

            ]
        },
        {
            "timestamp": 1452615190,
            "level": 300,
            "message": "Finished initialization of module 'pulp'",
            "backtrace": [

            ]
        },
        {
            "timestamp": 1452615190,
            "level": 300,
            "message": "'pulpnode' module is disabled.",
            "backtrace": [

            ]
        },
        {
            "timestamp": 1452615190,
            "level": 300,
            "message": "'foreman_proxy' settings were initialized with default values: :enabled: true",
            "backtrace": [

            ]
        },
        {
            "timestamp": 1452615190,
            "level": 300,
            "message": "Finished initialization of module 'foreman_proxy'",
            "backtrace": [

            ]
        },
        {
            "timestamp": 1452615190,
            "level": 300,
            "message": "'facts' module is disabled.",
            "backtrace": [

            ]
        },
        {
            "timestamp": 1452615190,
            "level": 300,
            "message": "'dns' settings were initialized with default values: :dns_ttl: 86400",
            "backtrace": [

            ]
        },
        {
            "timestamp": 1452615190,
            "level": 300,
            "message": "Finished initialization of module 'dns'",
            "backtrace": [

            ]
        },
        {
            "timestamp": 1452615190,
            "level": 300,
            "message": "Finished initialization of module 'templates'",
            "backtrace": [

            ]
        },
        {
            "timestamp": 1452615190,
            "level": 300,
            "message": "Finished initialization of module 'tftp'",
            "backtrace": [

            ]
        },
        {
            "timestamp": 1452615190,
            "level": 300,
            "message": "'dhcp' settings were initialized with default values: :server: 127.0.0.1",
            "backtrace": [

            ]
        },
        {
            "timestamp": 1452615190,
            "level": 300,
            "message": "Finished initialization of module 'dhcp'",
            "backtrace": [

            ]
        },
        {
            "timestamp": 1452615191,
            "level": 300,
            "message": "Finished initialization of module 'puppetca'",
            "backtrace": [

            ]
        },
        {
            "timestamp": 1452615191,
            "level": 300,
            "message": "'puppet' settings were initialized with default values: :puppet_ssl_ca: /var/lib/puppet/ssl/certs/ca.pem, :salt_puppetrun_cmd: puppet.run, :use_cache: true",
            "backtrace": [

            ]
        },
        {
            "timestamp": 1452615191,
            "level": 100,
            "message": "Couldn't enable plugin puppet: Cannot set modulepath settings in puppet.conf",
            "backtrace": [

            ]
        },
        {
            "timestamp": 1452615191,
            "level": 400,
            "message": "Cannot set modulepath settings in puppet.conf:/home/lzap/.rbenv/versions/2.0.0-p353/lib/ruby/gems/2.0.0/gems/puppet-4.3.0/lib/puppet/settings.rb:1107:in `block in screen_non_puppet_conf_settings'\n/home/lzap/.rbenv/versions/2.0.0-p353/lib/ruby/gems/2.0.0/gems/puppet-4.3.0/lib/puppet/settings.rb:1105:in `each'\n/home/lzap/.rbenv/versions/2.0.0-p353/lib/ruby/gems/2.0.0/gems/puppet-4.3.0/lib/puppet/settings.rb:1105:in `screen_non_puppet_conf_settings'\n/home/lzap/.rbenv/versions/2.0.0-p353/lib/ruby/gems/2.0.0/gems/puppet-4.3.0/lib/puppet/settings.rb:538:in `parse_config'\n/home/lzap/.rbenv/versions/2.0.0-p353/lib/ruby/gems/2.0.0/gems/puppet-4.3.0/lib/puppet/settings.rb:595:in `parse_config_files'\n/home/lzap/.rbenv/versions/2.0.0-p353/lib/ruby/gems/2.0.0/gems/puppet-4.3.0/lib/puppet/settings.rb:262:in `initialize_global_settings'\n/home/lzap/.rbenv/versions/2.0.0-p353/lib/ruby/gems/2.0.0/gems/puppet-4.3.0/lib/puppet.rb:133:in `do_initialize_settings_for_run_mode'\n/home/lzap/.rbenv/versions/2.0.0-p353/lib/ruby/gems/2.0.0/gems/puppet-4.3.0/lib/puppet.rb:127:in `initialize_settings'\n/home/lzap/work/smart-proxy/modules/puppet_proxy/initializer.rb:20:in `reset_puppet'\n/home/lzap/work/smart-proxy/modules/puppet_proxy/puppet_plugin.rb:17:in `block in <class:Plugin>'\n/home/lzap/work/smart-proxy/lib/proxy/pluggable.rb:33:in `instance_eval'\n/home/lzap/work/smart-proxy/lib/proxy/pluggable.rb:33:in `after_activation'\n/home/lzap/work/smart-proxy/lib/proxy/plugin.rb:138:in `configure_plugin'\n/home/lzap/work/smart-proxy/lib/proxy/plugin.rb:29:in `block in configure_loaded_plugins'\n/home/lzap/work/smart-proxy/lib/proxy/plugin.rb:29:in `each'\n/home/lzap/work/smart-proxy/lib/proxy/plugin.rb:29:in `configure_loaded_plugins'\n/home/lzap/work/smart-proxy/lib/launcher.rb:111:in `launch'\nbin/smart-proxy:6:in `<main>'",
            "backtrace": [

            ]
        },
        {
            "timestamp": 1452615191,
            "level": 300,
            "message": "Finished initialization of module 'bmc'",
            "backtrace": [

            ]
        },
        {
            "timestamp": 1452615191,
            "level": 300,
            "message": "'realm' module is disabled.",
            "backtrace": [

            ]
        }
    ]
}
```

### DIFF
--- a/config/settings.yml.example
+++ b/config/settings.yml.example
@@ -56,3 +56,8 @@
 # Uncomment and modify if you want to change the log level
 # WARN, DEBUG, ERROR, FATAL, INFO, UNKNOWN
 #:log_level: ERROR
+
+# Log buffer size and extra buffer size (for errors). Defaults to 1500 messages in total,
+# which is about 250 kB request.
+#:log_buffer: 1000
+#:log_buffer_errors: 500
--- a/lib/launcher.rb
+++ b/lib/launcher.rb
@@ -132,9 +132,7 @@ module Proxy
       # do nothing. This is to prevent the exception handler below from catching SystemExit exceptions.
       raise
     rescue Exception => e
-      logger.error("Error during startup, terminating. #{e}")
-      logger.debug("#{e}:#{e.backtrace.join("\n")}")
-
+      logger.error("Error during startup, terminating. #{e}", e.backtrace)
       puts "Errors detected on startup, see log for details. Exiting: #{e}"
       exit(1)
     end
--- a/lib/proxy/helpers.rb
+++ b/lib/proxy/helpers.rb
@@ -19,8 +19,7 @@ module Proxy::Helpers
       code     = code || 400
     end
     content_type :json if request.accept?("application/json")
-    logger.error message
-    logger.debug exception.backtrace.join("\n") if exception.is_a?(Exception)
+    logger.error message, exception
     halt code, message
   end
 
--- a/lib/proxy/log_buffer/buffer.rb
+++ b/lib/proxy/log_buffer/buffer.rb
@@ -1,0 +1,97 @@
+require 'date'
+require 'proxy/log_buffer/ring_buffer'
+
+# Adopted from Celluloid library (ring_buffer.rb).
+# Copyright (c) 2011-2014 Tony Arcieri. Distributed under the MIT License.
+# https://github.com/celluloid/celluloid/blob/0-16-stable/lib/celluloid/logging/ring_buffer.rb
+module Proxy::LogBuffer
+
+  LogRecord = Struct.new(:timestamp, :level, :message, :backtrace) do
+    def to_h
+      h = {}
+      self.class.members.each{|m| h[m.to_sym] = self[m]}
+      h[:level] = case h[:level]
+                  when ::Logger::Severity::INFO
+                    :INFO
+                  when ::Logger::Severity::WARN
+                    :WARN
+                  when ::Logger::Severity::ERROR
+                    :ERROR
+                  when ::Logger::Severity::FATAL
+                    :FATAL
+                  when ::Logger::Severity::DEBUG
+                    :DEBUG
+                  else
+                    :UNKNOWN
+                  end
+      h.delete(:backtrace) unless h[:backtrace]
+      h
+    end
+  end
+
+  class Buffer
+    def self.instance
+      @@buffer ||= Buffer.new
+    end
+
+    def initialize(size = nil, size_tail = nil, level_tail = nil)
+      @mutex = Mutex.new
+      @failed_modules = {}
+      @main_buffer = RingBuffer.new(size || ::Proxy::SETTINGS.log_buffer.to_i)
+      @tail_buffer = RingBuffer.new(size_tail || ::Proxy::SETTINGS.log_buffer_errors.to_i)
+      @level_tail = level_tail || ::Logger::Severity::ERROR
+    end
+
+    def push(rec)
+      @mutex.synchronize do
+        rec.timestamp = Time.now.utc.to_f
+        old_value = @main_buffer.push(rec)
+        @tail_buffer.push(old_value) if old_value && old_value.level >= @level_tail
+      end
+    end
+
+    def iterate_ascending
+      @mutex.synchronize do
+        @tail_buffer.iterate_ascending { |x| yield x }
+        @main_buffer.iterate_ascending { |x| yield x }
+      end
+    end
+
+    def iterate_descending
+      @mutex.synchronize do
+        @main_buffer.iterate_descending { |x| yield x }
+        @tail_buffer.iterate_descending { |x| yield x }
+      end
+    end
+
+    def to_a(from_timestamp = 0)
+      result = []
+      if from_timestamp == 0
+        iterate_ascending do |x|
+          result << x if x
+        end
+      else
+        iterate_ascending do |x|
+          result << x if x && x.timestamp >= from_timestamp
+        end
+      end
+      result
+    end
+
+    # Singleton logger does not allow per-module logging, until this is fixed
+    # initialization errors are kept in this explicit hash.
+    def failed_module(a_module, message)
+      @failed_modules[a_module] = message
+    end
+
+    def info
+      {
+        :size => @main_buffer.size,
+        :tail_size => @tail_buffer.size,
+        :level => @level,
+        :level_tail => @level_tail,
+        :failed_modules => @failed_modules
+      }
+    end
+  end
+end
--- a/lib/proxy/log_buffer/decorator.rb
+++ b/lib/proxy/log_buffer/decorator.rb
@@ -1,0 +1,61 @@
+require 'proxy/log_buffer/decorator'
+require 'proxy/log_buffer/buffer'
+
+module Proxy::LogBuffer
+  class Decorator
+    def initialize(logger, buffer = Proxy::LogBuffer::Buffer.instance)
+      @logger = logger
+      @buffer = buffer
+    end
+
+    def add(severity, message = nil, progname = nil, backtrace = nil, a_module = 'core', &block)
+      severity ||= UNKNOWN
+      progname ||= @logger.progname
+      if message.nil?
+        if block_given?
+          message = yield
+        else
+          message = progname
+          progname = @logger.progname
+        end
+      end
+      # add to the logger first
+      @logger.add(severity, message, progname)
+      @logger.add(::Logger::Severity::DEBUG, backtrace) if backtrace
+      # add add to the buffer
+      if severity >= @logger.level
+        # we accept backtrace, exception and simple string
+        backtrace = backtrace.is_a?(Exception) ? backtrace.backtrace : backtrace
+        backtrace = backtrace.respond_to?(:join) ? backtrace.join("\n") : backtrace
+        rec = Proxy::LogBuffer::LogRecord.new(nil, severity, message, backtrace)
+        @buffer.push(rec)
+      end
+    end
+
+    def debug(msg_or_progname, exception_or_backtrace = nil, &block)
+      add(::Logger::Severity::DEBUG, nil, msg_or_progname, exception_or_backtrace, caller, &block)
+    end
+
+    def info(msg_or_progname, exception_or_backtrace = nil, &block)
+      add(::Logger::Severity::INFO, nil, msg_or_progname, exception_or_backtrace, caller, &block)
+    end
+    alias_method :write, :info
+
+    def warn(msg_or_progname, exception_or_backtrace = nil, &block)
+      add(::Logger::Severity::WARN, nil, msg_or_progname, exception_or_backtrace, caller, &block)
+    end
+    alias_method :warning, :warn
+
+    def error(msg_or_progname, exception_or_backtrace = nil, &block)
+      add(::Logger::Severity::ERROR, nil, msg_or_progname, exception_or_backtrace, caller, &block)
+    end
+
+    def fatal(msg_or_progname, exception_or_backtrace = nil, &block)
+      add(::Logger::Severity::FATAL, nil, msg_or_progname, exception_or_backtrace, caller, &block)
+    end
+
+    def method_missing(symbol, *args);
+      @logger.send(symbol, *args)
+    end
+  end
+end
--- a/lib/proxy/log_buffer/ring_buffer.rb
+++ b/lib/proxy/log_buffer/ring_buffer.rb
@@ -1,0 +1,63 @@
+module Proxy::LogBuffer
+  class RingBuffer
+    attr_reader :size, :count
+
+    def initialize(size)
+      raise "size must be > 1" if size <= 1
+      @size = size
+      @start = 0
+      @count = 0
+      @buffer = Array.new(size)
+    end
+
+    def full?
+      @count == @size
+    end
+
+    def push(value)
+      stop = (@start + @count) % @size
+      old_value = @buffer[stop]
+      @buffer[stop] = value
+      if full?
+        @start = (@start + 1) % @size
+        old_value
+      else
+        @count += 1
+        nil
+      end
+    end
+
+    def clear
+      @buffer = Array.new(@size)
+      @start = 0
+      @count = 0
+    end
+
+    def iterate_ascending
+      0.step(@size - 1, 1) do |size_iterator|
+        element = @buffer[(@start + size_iterator) % @size]
+        yield element if element
+      end
+    end
+
+    def iterate_descending
+      (@size - 1).step(0, -1) do |size_iterator|
+        element = @buffer[(@start + size_iterator) % @size]
+        yield element if element
+      end
+    end
+
+    # shallow copy
+    def copy(new_size)
+      new_buffer = RingBuffer.new(new_size)
+      iterate_ascending { |x| new_buffer.push(x) }
+      new_buffer
+    end
+
+    def to_a
+      result = []
+      iterate_ascending { |x| result << x }
+      result
+    end
+  end
+end
--- a/lib/proxy/plugin.rb
+++ b/lib/proxy/plugin.rb
@@ -140,8 +140,8 @@ class ::Proxy::Plugin
       logger.info("'#{plugin_name}' module is disabled.")
     end
   rescue Exception => e
-    logger.error("Couldn't enable plugin #{plugin_name}: #{e}")
-    logger.debug("#{e}:#{e.backtrace.join("\n")}")
+    logger.error("Couldn't enable plugin #{plugin_name}: #{e}", e.backtrace)
     ::Proxy::Plugins.disable_plugin(plugin_name)
+    ::Proxy::LogBuffer::Buffer.instance.failed_module(plugin_name, e.message)
   end
 end
--- a/modules/root/http_config.ru
+++ b/modules/root/http_config.ru
@@ -1,5 +1,10 @@
 require 'root/root_api'
+require 'root/logs_api'
 
 map "/" do
   run Proxy::RootApi
+end
+
+map "/logs" do
+  run Proxy::LogsApi
 end
--- a/test/bmc/bmc_api_test.rb
+++ b/test/bmc/bmc_api_test.rb
@@ -245,7 +245,6 @@ class BmcApiTest < Test::Unit::TestCase
     Proxy::BMC::IPMI.any_instance.stubs(:poweron?).returns(true)
     get "/#{host}/chassis/power/on", args
     assert_not_equal 'Rubyipmi', Proxy::BMC::IPMI.logger.progname
-    assert_equal "./logs/test.log", Proxy::BMC::IPMI.logger.instance_variable_get("@logdev").filename
   end
 
   def test_api_can_put_power_action
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -43,3 +43,7 @@ Style/SymbolProc:
 # Workaround for bbatsov/rubocop#2451 (found in v0.35.1)
 Style/StabbyLambdaParentheses:
   Enabled: false
+
+# Some logging methods need more params than usual
+Metrics/ParameterLists:
+  Max: 8
--- a/LICENSE
+++ b/LICENSE
@@ -1,3 +1,39 @@
+The Foreman repository/package is licensed under the GNU GPL v3 or newer,
+with some exceptions.
+
+Copyright (c) 2009-2015 to Ohad Levy, Paul Kelly and their respective owners
+
+All copyright holders for the Foreman project are in the separate file called
+Contributors.
+
+Except where specified below, this program and entire repository is free
+software: you can redistribute it and/or modify it under the terms of the GNU
+General Public License as published by the Free Software Foundation, either
+version 3 of the License, or any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program.
+If not, see [GNU licenses](http://www.gnu.org/licenses/).
+
+The following files and directories are exceptions:
+
+* proxy/log_buffer/ring_buffer.rb is under MIT license
+
+Holders for code above are:
+
+  (c) 2011-2014 Tony Arcieri
+
+All rights reserved.
+
+All the licenses mentioned above now follows:
+
+-----------------------------------------------------------------------------
+
                     GNU GENERAL PUBLIC LICENSE
                        Version 3, 29 June 2007
 
@@ -672,3 +708,25 @@ may consider it more useful to permit linking proprietary applications with
 the library.  If this is what you want to do, use the GNU Lesser General
 Public License instead of this License.  But first, please read
 <http://www.gnu.org/philosophy/why-not-lgpl.html>.
+
+-----------------------------------------------------------------------------
+
+MIT license
+
+Permission is hereby granted, free of charge, to any person obtaining a copy of
+this software and associated documentation files (the "Software"), to deal in
+the Software without restriction, including without limitation the rights to
+use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies
+of the Software, and to permit persons to whom the Software is furnished to do
+so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
--- a/lib/proxy/log.rb
+++ b/lib/proxy/log.rb
@@ -1,4 +1,6 @@
 require 'logger'
+require 'proxy/log_buffer/decorator'
+
 begin
   require 'syslog/logger'
   ::Syslog::Logger.class_eval { alias_method :write, :info }
@@ -42,7 +44,7 @@ module Proxy
         logger = default_logger(log_file)
       end
       logger.level = ::Logger.const_get(::Proxy::SETTINGS.log_level.upcase)
-      logger
+      ::Proxy::LogBuffer::Decorator.new(logger)
     end
   end
 
--- a/lib/proxy/settings/global.rb
+++ b/lib/proxy/settings/global.rb
@@ -9,7 +9,10 @@ module ::Proxy::Settings
       :daemon_pid => "/var/run/foreman-proxy/foreman-proxy.pid",
       :forward_verify => true,
       :bind_host => "*",
-      :virsh_network => 'default'
+      :virsh_network => 'default',
+      :log_buffer => 2000,
+      :log_buffer_errors => 1000,
+      :log_buffer_level => "INFO"
     }
 
     HOW_TO_NORMALIZE = {
--- a/modules/root/logs_api.rb
+++ b/modules/root/logs_api.rb
@@ -1,0 +1,20 @@
+require 'proxy/log_buffer/decorator'
+require 'proxy/log_buffer/buffer'
+
+class Proxy::LogsApi < Sinatra::Base
+  helpers ::Proxy::Helpers
+  authorize_with_trusted_hosts
+  authorize_with_ssl_client
+
+  get "/" do
+    begin
+      content_type :json
+      buffer = ::Proxy::LogBuffer::Buffer.instance
+      from_timestamp = params[:from_timestamp].to_f rescue 0
+      records = buffer.to_a(from_timestamp).collect(&:to_h)
+      { :info => buffer.info, :logs => records }.to_json
+    rescue => e
+      log_halt 400, e
+    end
+  end
+end
--- a/test/log_buffer/buffer_test.rb
+++ b/test/log_buffer/buffer_test.rb
@@ -1,0 +1,48 @@
+require 'test_helper'
+
+class BufferTest < Test::Unit::TestCase
+  SIZE = 3
+  SIZE_TAIL = 2
+
+  DEBUG = ::Logger::Severity::DEBUG
+  INFO = ::Logger::Severity::INFO
+  ERR = ::Logger::Severity::ERROR
+
+  def setup
+    @buffer = Proxy::LogBuffer::Buffer.new(SIZE, SIZE_TAIL, ERR)
+  end
+
+  def log_event(*args)
+    Proxy::LogBuffer::LogRecord.new(*args)
+  end
+
+  def test_should_be_empty
+    @buffer.iterate_ascending { |x| fail }
+    @buffer.iterate_descending { |x| fail }
+  end
+
+  def test_should_contain_one_value
+    @buffer.push(log_event(Time.now.utc.to_f, ERR, "msg"))
+    @buffer.iterate_ascending { |x| assert_equal "msg", x.message }
+  end
+
+  def test_should_store_size
+    1.step(SIZE) { |i| @buffer.push(log_event(Time.now.utc.to_f, ERR, i)) }
+    assert_equal [1, 2, 3], @buffer.to_a.collect(&:message)
+  end
+
+  def test_should_store_full_size_errors
+    1.step(SIZE + SIZE_TAIL) { |i| @buffer.push(log_event(Time.now.utc.to_f, ERR, i)) }
+    assert_equal [1, 2, 3, 4, 5], @buffer.to_a.collect(&:message)
+  end
+
+  def test_should_store_full_size_plus_one_errors
+    1.step(SIZE + SIZE_TAIL + 1) { |i| @buffer.push(log_event(Time.now.utc.to_f, ERR, i)) }
+    assert_equal [2, 3, 4, 5, 6], @buffer.to_a.collect(&:message)
+  end
+
+  def test_should_store_full_size_plus_one_info
+    1.step(SIZE + SIZE_TAIL + 1) { |i| @buffer.push(log_event(Time.now.utc.to_f, INFO, i)) }
+    assert_equal [4, 5, 6], @buffer.to_a.collect(&:message)
+  end
+end
--- a/test/log_buffer/decorator_test.rb
+++ b/test/log_buffer/decorator_test.rb
@@ -1,0 +1,33 @@
+require 'test_helper'
+
+class DecoratorTest < Test::Unit::TestCase
+  SIZE = 10
+  SIZE_TAIL = 5
+
+  DEBUG = ::Logger::Severity::DEBUG
+  INFO = ::Logger::Severity::INFO
+  ERR = ::Logger::Severity::ERROR
+
+  def setup
+    @buffer = Proxy::LogBuffer::Buffer.new(SIZE, SIZE_TAIL, ERR)
+    @logger = ::Logger.new("/dev/null")
+    @logger.level = INFO
+    @decorator = ::Proxy::LogBuffer::Decorator.new(@logger, @buffer)
+  end
+
+  def test_should_not_ignore_infos
+    1.step(5) { |i| @decorator.info(i) }
+    assert_equal [1, 2, 3, 4, 5], @buffer.to_a.collect(&:message)
+  end
+
+  def test_should_ignore_debugs
+    1.step(5) { |i| @decorator.debug(i) }
+    assert_equal [], @buffer.to_a.collect(&:message)
+  end
+
+  def test_should_not_ignore_debugs
+    @logger.level = DEBUG
+    1.step(5) { |i| @decorator.debug(i) }
+    assert_equal [1, 2, 3, 4, 5], @buffer.to_a.collect(&:message)
+  end
+end
--- a/test/log_buffer/ring_buffer_test.rb
+++ b/test/log_buffer/ring_buffer_test.rb
@@ -1,0 +1,53 @@
+require 'test_helper'
+
+class RingBufferTest < Test::Unit::TestCase
+  SIZE = 3
+
+  def setup
+    @buffer = Proxy::LogBuffer::RingBuffer.new(SIZE)
+  end
+
+  def test_should_be_empty
+    assert_equal false, @buffer.full?
+    @buffer.iterate_ascending { |x| fail }
+    @buffer.iterate_descending { |x| fail }
+  end
+
+  def test_should_store_one_value
+    @buffer.push(:a)
+    assert_equal 1, @buffer.count
+  end
+
+  def test_should_contain_one_value
+    @buffer.push(:a)
+    @buffer.iterate_ascending { |x| assert_equal :a, x }
+  end
+
+  def test_should_store_size
+    1.step(SIZE) { |i| @buffer.push(i) }
+    assert_equal [1, 2, 3], @buffer.to_a
+  end
+
+  def test_copy_should_store_values
+    @buffer.push(:a)
+    @buffer.push(:b)
+    new_buffer = @buffer.copy(SIZE)
+    assert_equal [:a, :b], new_buffer.to_a
+  end
+
+  def test_full_buffer
+    1.step(SIZE) { |i| @buffer.push(i) }
+    assert @buffer.full?
+  end
+
+  def test_should_store_size_plus_one
+    1.step(SIZE + 1) { |i| @buffer.push(i) }
+    assert_equal [2, 3, 4], @buffer.to_a
+  end
+
+  def test_clear
+    1.step(SIZE) { |i| @buffer.push(i) }
+    @buffer.clear
+    assert_equal [], @buffer.to_a
+  end
+end
--- a/test/root/logs_api_test.rb
+++ b/test/root/logs_api_test.rb
@@ -1,0 +1,25 @@
+require 'test_helper'
+require 'json'
+require 'root/logs_api'
+
+ENV['RACK_ENV'] = 'test'
+
+class LogsApiTest < Test::Unit::TestCase
+  include Rack::Test::Methods
+
+  def app
+    Proxy::LogsApi.new
+  end
+
+  def test_log_buffer
+    get "/"
+    assert_equal 3, JSON.parse(last_response.body)["info"]["level_tail"]
+  end
+
+  def test_log_buffer_timestamp
+    get "/"
+    t1 = JSON.parse(last_response.body)["logs"][0]["timestamp"]
+    t2 = JSON.parse(last_response.body)["logs"][1]["timestamp"]
+    assert t1 < t2
+  end
+end


dmitri-d:
This is rather different from what I had in mind and been working on; It's much simpler, not really an events collection system: essentially a drop-in replacement for a logger that also buffers messages. 

My approach is more complicated, and introduces a system for emitting and collecting events for individual modules. Logger and /events api end points are just two consumers and aggregators of these events. See https://github.com/witlessbird/smart-proxy/tree/events, but you'll probably need a walk-through.

As we now have two competing approaches, we need to figure out which one to use. @ohadlevy, @domcleal, not sure whom else to ping.

lzap:
Thanks for the explanation, can you please put your work together into a PR so we can discuss? You are right, it is not an event system actually and I think we need more input from @ohadlevy and @shlomizadok about what the future plans are for this feature. Maybe what you are building is overkill, maybe this is too simple and won't fit our needs.

dmitri-d:
I have another few modules to convert to use rx; I'd rather wait until we decide on requirements/architecture as not to turn this into busy-work; The PR I referred to in my comment should be sufficient for discussion and demonstration purposes.

domcleal:
> As we now have two competing approaches, we need to figure out which one to use. @ohadlevy, @domcleal, not sure whom else to ping.

I've no idea what this is about.  Link to the original discussion and ticket?

lzap:
Yeah, the only problem with github branch comments is they disappear on branch deletion I think while PR can be closed forever.

I am pushing one more change - I created simple Logger Decorator therefore no changes are necessary in individual modules/plugins - the logging buffer system now simply works across whole proxy. Also added increasing serial number, so the API can now skip already loaded items. I expect a JavaScript to serve continuous log "stream" with filters:

```
$ curl -sk http://localhost:8448/events | json_reformat 
{
    "events": [
        {
            "serial": 2,
            "timestamp": 1452676733,
            "level": 1,
            "message": "'discovery' settings were initialized with default values: :node_port: 8443, :node_scheme: https"
        },
        {
            "serial": 3,
            "timestamp": 1452676733,
            "level": 1,
            "message": "Finished initialization of module 'discovery'"
        },
        {
            "serial": 4,
            "timestamp": 1452676733,
            "level": 1,
            "message": "Finished initialization of module 'discovery_image'"
        },
        {
            "serial": 5,
            "timestamp": 1452676733,
            "level": 1,
            "message": "Finished initialization of module 'salt'"
        },
        {
            "serial": 6,
            "timestamp": 1452676733,
            "level": 1,
            "message": "Finished initialization of module 'abrt'"
        },
        {
            "serial": 7,
            "timestamp": 1452676733,
            "level": 1,
            "message": "'colly' settings were initialized with default values: :socket: /var/run/collectd-unixsock, :timeout: 5"
        },
        {
            "serial": 8,
            "timestamp": 1452676733,
            "level": 3,
            "message": "Couldn't enable plugin colly: No such file or directory - \"/var/run/collectd-unixsock\"",
            "backtrace": "/home/lzap/work/collectd-uxsock/lib/uxsock/collectd_unix_socket.rb:12:in `initialize'\n/home/lzap/work/collectd-uxsock/lib/uxsock/collectd_unix_socket.rb:12:in `open'\n/home/lzap/work/collectd-uxsock/lib/uxsock/collectd_unix_socket.rb:12:in `initialize'\n/home/lzap/work/collectd-uxsock/lib/uxsock/collectd_unix_socket.rb:22:in `new'\n/home/lzap/work/collectd-uxsock/lib/uxsock/collectd_unix_socket.rb:22:in `open'\n/home/lzap/work/smart_proxy_colly/lib/smart_proxy_colly/writer.rb:31:in `notify'\n/home/lzap/work/smart_proxy_colly/lib/smart_proxy_colly/colly.rb:14:in `block in <class:Plugin>'\n/home/lzap/work/smart-proxy/lib/proxy/pluggable.rb:33:in `instance_eval'\n/home/lzap/work/smart-proxy/lib/proxy/pluggable.rb:33:in `after_activation'\n/home/lzap/work/smart-proxy/lib/proxy/plugin.rb:137:in `configure_plugin'\n/home/lzap/work/smart-proxy/lib/proxy/plugin.rb:28:in `block in configure_loaded_plugins'\n/home/lzap/work/smart-proxy/lib/proxy/plugin.rb:28:in `each'\n/home/lzap/work/smart-proxy/lib/proxy/plugin.rb:28:in `configure_loaded_plugins'\n/home/lzap/work/smart-proxy/lib/launcher.rb:110:in `launch'\nbin/smart-proxy:6:in `<main>'"
        },
        {
            "serial": 9,
            "timestamp": 1452676733,
            "level": 1,
            "message": "Finished initialization of module 'pulp'"
        },
        {
            "serial": 10,
            "timestamp": 1452676733,
            "level": 2,
            "message": "Couldn't find settings file /home/lzap/work/smart-proxy/config/settings.d/pulpnode.yml. Using default settings."
        },
        {
            "serial": 11,
            "timestamp": 1452676733,
            "level": 1,
            "message": "'pulpnode' module is disabled."
        },
        {
            "serial": 12,
            "timestamp": 1452676733,
            "level": 1,
            "message": "'foreman_proxy' settings were initialized with default values: :enabled: true"
        },
        {
            "serial": 13,
            "timestamp": 1452676733,
            "level": 1,
            "message": "Finished initialization of module 'foreman_proxy'"
        },
        {
            "serial": 14,
            "timestamp": 1452676733,
            "level": 2,
            "message": "Couldn't find settings file /home/lzap/work/smart-proxy/config/settings.d/facts.yml. Using default settings."
        },
        {
            "serial": 15,
            "timestamp": 1452676733,
            "level": 1,
            "message": "'facts' module is disabled."
        },
        {
            "serial": 16,
            "timestamp": 1452676733,
            "level": 1,
            "message": "'dns_virsh' settings were initialized with default values: :enabled: false"
        },
        {
            "serial": 17,
            "timestamp": 1452676733,
            "level": 1,
            "message": "'dns' settings were initialized with default values: :dns_ttl: 86400"
        },
        {
            "serial": 18,
            "timestamp": 1452676733,
            "level": 1,
            "message": "Finished initialization of module 'dns'"
        },
        {
            "serial": 19,
            "timestamp": 1452676733,
            "level": 1,
            "message": "Finished initialization of module 'templates'"
        },
        {
            "serial": 20,
            "timestamp": 1452676733,
            "level": 1,
            "message": "Finished initialization of module 'tftp'"
        },
        {
            "serial": 21,
            "timestamp": 1452676733,
            "level": 2,
            "message": "Couldn't find settings file /home/lzap/work/smart-proxy/config/settings.d/dhcp_virsh.yml. Using default settings."
        },
        {
            "serial": 22,
            "timestamp": 1452676733,
            "level": 1,
            "message": "'dhcp_virsh' settings were initialized with default values: :enabled: false"
        },
        {
            "serial": 23,
            "timestamp": 1452676733,
            "level": 1,
            "message": "'dhcp' settings were initialized with default values: :server: 127.0.0.1"
        },
        {
            "serial": 24,
            "timestamp": 1452676733,
            "level": 1,
            "message": "Finished initialization of module 'dhcp'"
        },
        {
            "serial": 25,
            "timestamp": 1452676734,
            "level": 1,
            "message": "Finished initialization of module 'puppetca'"
        },
        {
            "serial": 26,
            "timestamp": 1452676734,
            "level": 1,
            "message": "'puppet' settings were initialized with default values: :puppet_ssl_ca: /var/lib/puppet/ssl/certs/ca.pem, :salt_puppetrun_cmd: puppet.run, :use_cache: true"
        },
        {
            "serial": 27,
            "timestamp": 1452676734,
            "level": 1,
            "message": "Initializing from Puppet config file: /home/lzap/.puppet/puppet.conf"
        },
        {
            "serial": 28,
            "timestamp": 1452676734,
            "level": 3,
            "message": "Couldn't enable plugin puppet: Cannot set modulepath settings in puppet.conf",
            "backtrace": "/home/lzap/.rbenv/versions/2.0.0-p353/lib/ruby/gems/2.0.0/gems/puppet-4.3.0/lib/puppet/settings.rb:1107:in `block in screen_non_puppet_conf_settings'\n/home/lzap/.rbenv/versions/2.0.0-p353/lib/ruby/gems/2.0.0/gems/puppet-4.3.0/lib/puppet/settings.rb:1105:in `each'\n/home/lzap/.rbenv/versions/2.0.0-p353/lib/ruby/gems/2.0.0/gems/puppet-4.3.0/lib/puppet/settings.rb:1105:in `screen_non_puppet_conf_settings'\n/home/lzap/.rbenv/versions/2.0.0-p353/lib/ruby/gems/2.0.0/gems/puppet-4.3.0/lib/puppet/settings.rb:538:in `parse_config'\n/home/lzap/.rbenv/versions/2.0.0-p353/lib/ruby/gems/2.0.0/gems/puppet-4.3.0/lib/puppet/settings.rb:595:in `parse_config_files'\n/home/lzap/.rbenv/versions/2.0.0-p353/lib/ruby/gems/2.0.0/gems/puppet-4.3.0/lib/puppet/settings.rb:262:in `initialize_global_settings'\n/home/lzap/.rbenv/versions/2.0.0-p353/lib/ruby/gems/2.0.0/gems/puppet-4.3.0/lib/puppet.rb:133:in `do_initialize_settings_for_run_mode'\n/home/lzap/.rbenv/versions/2.0.0-p353/lib/ruby/gems/2.0.0/gems/puppet-4.3.0/lib/puppet.rb:127:in `initialize_settings'\n/home/lzap/work/smart-proxy/modules/puppet_proxy/initializer.rb:20:in `reset_puppet'\n/home/lzap/work/smart-proxy/modules/puppet_proxy/puppet_plugin.rb:17:in `block in <class:Plugin>'\n/home/lzap/work/smart-proxy/lib/proxy/pluggable.rb:33:in `instance_eval'\n/home/lzap/work/smart-proxy/lib/proxy/pluggable.rb:33:in `after_activation'\n/home/lzap/work/smart-proxy/lib/proxy/plugin.rb:137:in `configure_plugin'\n/home/lzap/work/smart-proxy/lib/proxy/plugin.rb:28:in `block in configure_loaded_plugins'\n/home/lzap/work/smart-proxy/lib/proxy/plugin.rb:28:in `each'\n/home/lzap/work/smart-proxy/lib/proxy/plugin.rb:28:in `configure_loaded_plugins'\n/home/lzap/work/smart-proxy/lib/launcher.rb:110:in `launch'\nbin/smart-proxy:6:in `<main>'"
        },
        {
            "serial": 29,
            "timestamp": 1452676734,
            "level": 1,
            "message": "Finished initialization of module 'bmc'"
        },
        {
            "serial": 30,
            "timestamp": 1452676734,
            "level": 1,
            "message": "'realm' module is disabled."
        },
        {
            "serial": 31,
            "timestamp": 1452676735,
            "level": 1,
            "message": "WEBrick 1.3.1"
        },
        {
            "serial": 32,
            "timestamp": 1452676735,
            "level": 1,
            "message": "WEBrick 1.3.1"
        },
        {
            "serial": 33,
            "timestamp": 1452676735,
            "level": 1,
            "message": "ruby 2.0.0 (2013-11-22) [x86_64-linux]"
        },
        {
            "serial": 34,
            "timestamp": 1452676735,
            "level": 1,
            "message": "ruby 2.0.0 (2013-11-22) [x86_64-linux]"
        },
        {
            "serial": 35,
            "timestamp": 1452676735,
            "level": 2,
            "message": "TCPServer Error: Address already in use - bind(2)"
        },
        {
            "serial": 36,
            "timestamp": 1452676735,
            "level": 1,
            "message": "\nCertificate:\n    Data:\n        Version: 3 (0x2)\n        Serial Number: 2 (0x2)\n    Signature Algorithm: sha256WithRSAEncryption\n        Issuer: CN=Puppet CA: lzapx.brq.redhat.com\n        Validity\n            Not Before: Jul 10 13:21:45 2014 GMT\n            Not After : Jul 10 13:21:45 2019 GMT\n        Subject: CN=lzapx.brq.redhat.com\n        Subject Public Key Info:\n            Public Key Algorithm: rsaEncryption\n                Public-Key: (4096 bit)\n                Modulus:\n                    00:d7:84:99:fe:fc:5b:82:f9:5b:0c:e8:83:a1:65:\n                    5e:b8:55:60:28:ab:49:51:6f:36:f5:32:eb:79:7f:\n                    03:74:bf:01:70:70:80:ea:93:71:83:85:19:3f:ef:\n                    bc:de:0f:e6:84:52:35:d3:53:ac:ce:c5:45:41:3d:\n                    72:60:37:09:af:95:d1:0f:e7:db:1f:96:ef:a4:98:\n                    16:d7:d7:90:5a:4d:76:21:00:b7:65:02:75:d2:e1:\n                    1c:64:ec:2e:44:6f:22:64:c6:38:73:06:35:07:cc:\n                    1d:39:14:3b:72:29:53:b2:14:77:39:62:42:e3:21:\n                    c1:78:62:b5:3f:80:02:57:32:53:cb:84:d0:99:7e:\n                    4a:7c:07:79:f7:52:b4:35:59:40:a1:bf:7f:b0:14:\n                    d5:8b:ab:7b:91:3a:de:4a:8e:16:ed:8d:7b:db:2f:\n                    eb:a4:15:6a:84:1f:8c:e5:db:1d:a4:79:e3:c6:77:\n                    1d:62:79:28:e3:43:54:1e:d9:29:3c:f5:28:62:17:\n                    0d:a7:03:ff:97:4e:01:e9:74:36:32:dd:7a:c2:a3:\n                    99:53:9c:a1:c7:e2:ef:db:42:65:a6:59:1d:f7:71:\n                    c8:07:7c:03:02:80:e8:ea:85:34:0b:f1:b8:2e:5c:\n                    90:88:34:13:b8:36:de:83:c9:ef:b7:a9:e8:65:57:\n                    2d:99:a9:b5:d2:3e:0d:b5:0d:e4:82:e8:3a:fd:45:\n                    c4:92:08:88:ca:b1:58:6a:8c:b3:87:76:e7:27:60:\n                    da:bc:c9:d0:12:40:67:b0:bf:8b:87:db:54:b0:c0:\n                    31:ae:25:bb:cb:07:31:be:54:f2:93:96:15:78:b3:\n                    dc:7c:c9:e2:93:e6:a8:91:48:ca:d3:1b:da:10:b9:\n                    5c:57:d2:64:2d:3a:29:44:bf:d7:66:79:3b:e1:f9:\n                    8a:58:61:ba:8f:94:38:c8:e6:e8:28:96:a9:c9:1f:\n                    30:f1:c9:af:59:2a:0a:60:1e:89:e1:74:6f:4e:52:\n                    93:23:04:38:aa:81:8d:d7:e9:0f:02:f4:9d:04:d7:\n                    54:39:48:3e:4a:7d:9a:5e:aa:92:1b:cb:94:36:31:\n                    eb:65:f3:a4:6c:de:a0:ed:19:9f:e7:8c:2f:a3:b1:\n                    19:93:ad:ca:20:45:0e:e0:8f:8f:ec:7e:ff:d2:d0:\n                    9b:f4:19:20:f1:59:e2:c7:95:e3:4b:45:51:b1:3c:\n                    eb:84:2f:60:66:ab:5a:ef:df:52:6a:1d:94:bd:54:\n                    6b:ec:e9:69:7b:44:7b:3c:90:1a:af:bb:b0:7f:31:\n                    b8:fe:73:83:b1:97:ff:4f:8f:70:03:78:68:fc:d6:\n                    73:46:60:f3:4b:a1:f8:61:45:d4:ae:f1:bc:3b:6f:\n                    1e:fb:23\n                Exponent: 65537 (0x10001)\n        X509v3 extensions:\n            Netscape Comment: \n                Puppet Ruby/OpenSSL Internal Certificate\n            X509v3 Subject Alternative Name: \n                DNS:lzapx.brq.redhat.com, DNS:puppet, DNS:puppet.brq.redhat.com\n            X509v3 Key Usage: critical\n                Digital Signature, Key Encipherment\n            X509v3 Extended Key Usage: critical\n                TLS Web Server Authentication, TLS Web Client Authentication\n            X509v3 Basic Constraints: critical\n                CA:FALSE\n            X509v3 Subject Key Identifier: \n                58:FB:B9:7F:ED:AF:F8:6F:45:9D:78:BA:A6:F9:91:EE:7C:16:1A:35\n            X509v3 Authority Key Identifier: \n                keyid:11:35:7C:0A:23:4D:3C:26:16:49:39:9F:4D:57:CE:58:C9:3C:4F:F8\n\n    Signature Algorithm: sha256WithRSAEncryption\n         59:3b:36:3a:7b:7e:a7:c8:fb:06:89:d7:25:5f:a9:dc:f6:68:\n         96:a8:f2:81:fb:4f:aa:38:9f:26:67:75:f3:4d:c6:db:44:4d:\n         e8:58:9d:23:b0:73:c4:dc:d3:04:db:e4:20:f9:a2:65:14:38:\n         2d:fa:77:ef:63:4f:56:1e:1c:e7:61:db:dc:f5:9b:68:2f:3a:\n         9f:b7:a8:b5:66:b2:84:79:db:ac:ae:89:7c:d0:86:bd:41:cd:\n         2e:17:02:29:d2:73:95:75:29:fb:39:c4:f2:d5:82:2a:3e:6a:\n         19:e8:ee:e0:b3:4b:fe:af:03:44:28:bf:f5:01:b8:ac:22:ec:\n         c4:8e:55:81:14:1d:e0:c4:45:c5:5a:e3:42:f0:fe:12:1d:2b:\n         77:0d:f7:35:26:91:c0:b5:3d:7b:b7:3a:dd:df:9a:24:1c:ef:\n         e6:05:c4:4b:e9:cf:c6:73:b9:43:a1:22:5c:a5:8a:7e:fe:64:\n         35:f0:77:94:2f:73:76:27:08:db:35:d0:05:28:9d:47:8b:9d:\n         dd:dc:7c:e3:b8:8d:5b:d8:04:1c:c1:f6:26:b5:76:f2:85:1b:\n         23:51:1c:40:0c:cf:2b:82:3f:a9:94:35:98:06:6a:35:5b:54:\n         76:3d:df:f8:f0:be:06:a7:61:7e:c4:15:bb:95:7d:75:bd:68:\n         c4:25:1f:93:f0:85:7c:ae:ff:44:6c:3c:a7:b6:30:d8:d5:28:\n         18:c2:a2:c6:91:db:b8:25:ea:7f:56:76:3d:10:bd:ea:3a:66:\n         24:39:09:03:01:1c:85:73:c7:cb:b7:66:1c:5e:59:9a:f1:a8:\n         86:7c:d4:fb:12:58:a0:1e:89:c4:2f:c6:e9:70:81:9f:d5:cf:\n         ae:0d:e8:5d:48:f9:7c:d7:7a:bd:74:5d:4c:64:0a:0f:be:18:\n         2d:c3:87:42:bb:05:95:0e:e3:1c:3b:d2:cf:ca:99:d3:16:e5:\n         33:aa:06:07:52:b0:22:7f:a7:11:1e:6b:11:05:b7:ff:c6:04:\n         c8:91:7b:44:78:10:cd:76:31:20:84:9d:f8:23:1a:37:52:8a:\n         54:32:ec:a0:d4:15:a5:5e:38:9c:67:d5:41:bf:d1:72:16:6d:\n         eb:ff:7e:f4:14:01:b8:1a:2c:95:6a:5e:68:6f:07:1c:eb:8d:\n         ce:bc:09:bf:6a:de:67:a6:1a:73:ef:dd:64:3a:47:7d:e1:4b:\n         ec:a9:ea:97:9d:b5:d4:38:a4:0b:1a:79:b1:74:a7:59:d1:43:\n         13:77:bc:e6:66:c6:e2:70:4a:c1:29:4d:fe:a4:69:4b:34:ae:\n         45:d2:d6:ce:01:f4:3a:51:86:81:70:03:f7:0b:ee:e6:6b:f7:\n         2e:cc:b0:0b:80:90:8f:d1\n"
        },
        {
            "serial": 37,
            "timestamp": 1452676735,
            "level": 1,
            "message": "WEBrick::HTTPServer#start: pid=32408 port=8443"
        },
        {
            "serial": 38,
            "timestamp": 1452676735,
            "level": 2,
            "message": "TCPServer Error: Address already in use - bind(2)"
        },
        {
            "serial": 39,
            "timestamp": 1452676735,
            "level": 1,
            "message": "WEBrick::HTTPServer#start: pid=32408 port=8448"
        },
        {
            "serial": 40,
            "timestamp": 1452676737,
            "level": 1,
            "message": "127.0.0.1 - - [13/Jan/2016 10:18:57] \"GET /events HTTP/1.1\" 200 12964 0.0398\n"
        },
        {
            "serial": 41,
            "timestamp": 1452676751,
            "level": 1,
            "message": "127.0.0.1 - - [13/Jan/2016 10:19:11] \"GET /dhcp/192.168.122.0/unused_ip?from=192.168.122.10&to=192.168.122.250 HTTP/1.1\" 200 23 2.0602\n"
        },
        {
            "serial": 42,
            "timestamp": 1452676768,
            "level": 1,
            "message": "127.0.0.1 - - [13/Jan/2016 10:19:28] \"GET /tftp/serverName HTTP/1.1\" 200 30 0.0005\n"
        },
        {
            "serial": 43,
            "timestamp": 1452676768,
            "level": 3,
            "message": "Record 192.168.122.0/192.168.122.19 not found",
            "backtrace": "Record 192.168.122.0/192.168.122.19 not found"
        },
        {
            "serial": 44,
            "timestamp": 1452676768,
            "level": 1,
            "message": "127.0.0.1 - - [13/Jan/2016 10:19:28] \"GET /dhcp/192.168.122.0/192.168.122.19 HTTP/1.1\" 404 45 0.0406\n"
        },
        {
            "serial": 45,
            "timestamp": 1452676769,
            "level": 1,
            "message": "127.0.0.1 - - [13/Jan/2016 10:19:29] \"POST /dhcp/192.168.122.0 HTTP/1.1\" 200 - 0.2271\n"
        },
        {
            "serial": 46,
            "timestamp": 1452676769,
            "level": 1,
            "message": "127.0.0.1 - - [13/Jan/2016 10:19:29] \"POST /dns/ HTTP/1.1\" 200 - 0.0391\n"
        },
        {
            "serial": 47,
            "timestamp": 1452676769,
            "level": 2,
            "message": "not creating PTR record for teoerou.local.lan (unsupported)"
        },
        {
            "serial": 48,
            "timestamp": 1452676769,
            "level": 1,
            "message": "127.0.0.1 - - [13/Jan/2016 10:19:29] \"POST /dns/ HTTP/1.1\" 200 - 0.0005\n"
        },
        {
            "serial": 49,
            "timestamp": 1452676769,
            "level": 1,
            "message": "127.0.0.1 - - [13/Jan/2016 10:19:29] \"GET /unattended/templateServer HTTP/1.1\" 200 46 0.0003\n"
        },
        {
            "serial": 50,
            "timestamp": 1452676769,
            "level": 1,
            "message": "TFTP: entry for 52:54:00:dc:10:21 created successfully"
        },
        {
            "serial": 51,
            "timestamp": 1452676769,
            "level": 1,
            "message": "127.0.0.1 - - [13/Jan/2016 10:19:29] \"POST /tftp/syslinux/52:54:00:dc:10:21 HTTP/1.1\" 200 - 0.0022\n"
        },
        {
            "serial": 52,
            "timestamp": 1452676769,
            "level": 1,
            "message": "127.0.0.1 - - [13/Jan/2016 10:19:29] \"POST /tftp/fetch_boot_file HTTP/1.1\" 200 - 0.0051\n"
        },
        {
            "serial": 53,
            "timestamp": 1452676769,
            "level": 3,
            "message": "[32598] failed: Connection refused.\n"
        },
        {
            "serial": 54,
            "timestamp": 1452676769,
            "level": 1,
            "message": "127.0.0.1 - - [13/Jan/2016 10:19:29] \"POST /tftp/fetch_boot_file HTTP/1.1\" 200 - 0.0011\n"
        },
        {
            "serial": 55,
            "timestamp": 1452676769,
            "level": 3,
            "message": "[32605] failed: Connection refused.\n"
        }
    ]
}
[lzap@lzapx smart-proxy]$ curl -sk http://localhost:8448/events?from_serial=54 | json_reformat 
{
    "events": [
        {
            "serial": 54,
            "timestamp": 1452676769,
            "level": 1,
            "message": "127.0.0.1 - - [13/Jan/2016 10:19:29] \"POST /tftp/fetch_boot_file HTTP/1.1\" 200 - 0.0011\n"
        },
        {
            "serial": 55,
            "timestamp": 1452676769,
            "level": 3,
            "message": "[32605] failed: Connection refused.\n"
        },
        {
            "serial": 56,
            "timestamp": 1452676773,
            "level": 1,
            "message": "127.0.0.1 - - [13/Jan/2016 10:19:33] \"GET /events HTTP/1.1\" 200 15243 0.0008\n"
        }
    ]
}
```

lzap:
I don't know about the Smart Proxy feature ticket, but I am working on the UI part of this feature and this is http://projects.theforeman.org/issues/12718 @shlomizadok right?

lzap:
@shlomizadok @ohadlevy here is my UI prototype, started today, not much but the idea is to build the scrolling table with javascript and then to add some filtering: https://github.com/lzap/foreman/commit/6ace5699e73bb110033128a30bb887db28c47b8d

I think I shall rename "events" in this patch to prevent future collision with the real eventing system. This could be named `LogBuffer` and the API endpoing could be `/latest_logs`.

dmitri-d:
w/e api end-point name you pick, pls. don't treat as a temporary: we don't have api versioning atm, so this is pretty much going to be a permanent change.

lib/proxy/events/buffer.rb
```diff
@@ -0,0 +1,90 @@
...
+        yield @buffer[(@start + size_iterator) % @size]
+      end
+    end
+
+    def iterate_descending
+      (@size - 1).step(0, -1) do |size_iterator|
+        yield @buffer[(@start + size_iterator) % @size]
+      end
+    end
+  end
```
lzap: For the record, this class is simplified and slightly enhanced version of https://github.com/celluloid/celluloid/blob/0-16-stable/lib/celluloid/logging/ring_buffer.rb (I added iterators and removed unused methods).

dmitri-d: You should leave a note that this contains code from celluloid and the copyright on parts of this code belongs to Tony Arcieri and others (MIT license thing).

lzap:
@ohadlevy @shlomizadok @witlessbird I added settings support, late initialization (to support that) and additional info returned through the API. This is now ready for review.

Since I am not available next two days, this patch can be easily used for developing/testing the UI. Simply checkout, start the proxy and then access the endpoint. There are about 50 log entries after start already:

```
curl -sk http://localhost:8448/events
```

lzap:
There is a trivial unit test failure on 1.8.7 because of `assert_false`. Will fix that later today.

config/settings.yml.example
```diff
@@ -56,3 +56,10 @@
 # Uncomment and modify if you want to change the log level
 # WARN, DEBUG, ERROR, FATAL, INFO, UNKNOWN
 #:log_level: ERROR
+
+# Log buffer size and extra buffer size (for errors)
+#:log_buffer: 2000
+#:log_buffer_errors: 1000
+# Log buffer threshold (debug messages are ignored by default)
+#:log_buffer_level: INFO
+
```
ohadlevy: nitpick: extra line

lib/proxy/events/buffer.rb
```diff
@@ -0,0 +1,129 @@
...
+      {
+        "size" => @main_buffer.size,
+        "tail_size" => @tail_buffer.size,
+        "level" => @level,
+        "level_tail" => @level_tail
+      }
+    end
+  end
+
+  class RingBuffer
```
ohadlevy: why two classes in the same file?

modules/root/root_api.rb
```diff
@@ -25,4 +27,19 @@ class Proxy::RootApi < Sinatra::Base
       log_halt 400, e
     end
   end
+
+  get "/events" do
+    begin
+      content_type :json
+
+      events = []
+      from_serial = params[:from_serial].to_i rescue 0
```
ohadlevy: nitpick: probably rescue is not required as nil.to_i and ''.to_i in both cases will be 0.

modules/root/root_api.rb
```diff
@@ -25,4 +27,19 @@ class Proxy::RootApi < Sinatra::Base
...
+      ::Proxy::Events::Buffer.instance.iterate_ascending do |event|
+        events << event.to_h if event && from_serial == 0 || (event && from_serial > 0 && event.serial >= from_serial)
```
ohadlevy: nitpick: maybe extract this logic outside here?
regardless, if from_serial is zero, you probably don't need to compare it for every event?

test/bmc/bmc_api_test.rb
```diff
@@ -245,7 +245,6 @@ def test_api_uses_default_smart_proxy_logger
     Proxy::BMC::IPMI.any_instance.stubs(:poweron?).returns(true)
     get "/#{host}/chassis/power/on", args
     assert_not_equal 'Rubyipmi', Proxy::BMC::IPMI.logger.progname
-    assert_equal "./logs/test.log", Proxy::BMC::IPMI.logger.instance_variable_get("@logdev").filename
```
ohadlevy, line 248: related?

lib/proxy/events/buffer.rb
```diff
@@ -0,0 +1,129 @@
...
+      end
+    end
+
+    def to_a
+      result = []
+      iterate_ascending { |x| result << x }
+      result
+    end
+
+    def get_info
```
ohadlevy: lib/proxy/events/buffer.rb:64:9: C: Do not prefix reader method names with get_.

lib/proxy/events/buffer.rb
```diff
@@ -0,0 +1,129 @@
...
+      @buffer[stop] = value
+      if full?
+        @start = (@start + 1) % @size
+        old_value
+      else
+        @count += 1
+        nil
+      end
+    end
+    alias :<< :push
```
ohadlevy: lib/proxy/events/buffer.rb:101:5: C: Use alias_method instead of alias.

ohadlevy:
one thing that i am missing is which part of the app provided the event, for example, if i want to see dhcp related issues, i cant tell by the reported event?

modules/root/root_api.rb
ohadlevy: also, why is it called from_serial? its a timestamp isnt it? so it should be called from_timestamp?

lib/proxy/events/log_decorator.rb
```diff
@@ -0,0 +1,59 @@
+require 'proxy/events/buffer'
+
+module Proxy::Events
+  class LogDecorator
+    def initialize(logger)
+      @logger = logger
+    end
+
+    def add(severity, message = nil, progname = nil, backtrace = nil, &block)
```
ohadlevy: &block is unused?

lib/proxy/events/buffer.rb
```diff
@@ -0,0 +1,129 @@
...
+      result = []
+      iterate_ascending { |x| result << x }
+      result
+    end
+
+    def get_info
+      {
+        "size" => @main_buffer.size,
+        "tail_size" => @tail_buffer.size,
+        "level" => @level,
```
ohadlevy: why string as hash keys?

modules/root/root_api.rb
ohadlevy: ah, its not a timestamp, nm :)

lib/proxy/events/buffer.rb
lzap: Fixed, plus all above.

modules/root/root_api.rb
lzap: Nah, every record in the buffer has a serial (increasing integer), which can be used to skip N records when doing additional request (e.g. in one user UI session). From timestamp is not supported, I don't see a use case for that. 

Removed the extra rescue, good point.

ohadlevy:
@lzap not sure if you were still around, i've created #359

modules/root/root_api.rb
lzap: Extracted, good point with the comparison. Optimized.

test/bmc/bmc_api_test.rb
lzap, line 248: Yes, the test is actually getting an internal from the logger instance. It is not available anymore. I don't think this is a good test. Anyway, line 247 already test that the instance is not `rubyipmi`, which is the main purpose. The extra test for the internal was useless IMHO.

lib/proxy/events/buffer.rb
lzap: Rubocop? Fixed.

lib/proxy/events/buffer.rb
lzap: We don't actually use that, removed.

lzap:
@ohadlevy I am around, I fixed all your remarks myself already :-(

Since I renamed events/ directory to log_buffer to avoid confision for the future, I cannot apply your patch, please close. But I fixed everything. Pushing in a bit.

lzap:
In regard to the plugin/module which generated a log entry - we don't have that, but we could dig this out of the callers (stacktrace) perhaps. Probably an enhancement - only if we have time?

ohadlevy:
@lzap i guess we could use a different logger per feature? or just wait for @witlessbird new / improved event system?

lib/proxy/events/log_decorator.rb
lzap: What do you mean? It is not used in smart-proxy (the block syntax) but it is the logger contract and I don't want to break it. Logging should be working "as is".

lib/proxy/events/buffer.rb
lzap: It's being serialized to JSON, but no big deal, changed to symbols.

lib/proxy/events/log_decorator.rb
ohadlevy: &block isn't necessary here as an argument and makes it slower, yield is
enough

lzap:
I'd rather keep this simple, let's proceed without module being reported and I can take a look in getting this from stacktrace later on.

Pushed fixes, develop tests are broken due to different issue:

https://github.com/theforeman/smart-proxy/pull/357#issuecomment-170902186

lzap:
For the record, I changed the endpoint, so to test this patch, use the following request:

```
$ curl -sk http://localhost:8448/log_buffer?from_serial=40 | json_reformat 
{
    "info": {
        "size": 100,
        "tail_size": 50,
        "level": 1,
        "level_tail": 3
    },
    "log": [
        {
            "serial": 40,
            "timestamp": 1452717320,
            "level": 1,
            "message": "127.0.0.1 - - [13/Jan/2016 21:35:20] \"GET /log_buffer?from_serial=54 HTTP/1.1\" 200 73 0.0007\n"
        },
        {
            "serial": 41,
            "timestamp": 1452717323,
            "level": 1,
            "message": "127.0.0.1 - - [13/Jan/2016 21:35:23] \"GET /log_buffer?from_serial=40 HTTP/1.1\" 200 228 0.0008\n"
        }
    ]
}
```

ohadlevy:
nitpick: you ignored all of my whitespaces and nitpick fixes :)

lzap:
Yeah, I was fixing it as I was reading. The patch is now free of "events" term, I used "records" now (e.g. `LogRecord`) which is much better. This is no rocket science indeed.

lzap:
And added a functional test. I call it a day, and a week as well. See the above `curl` command for an example use. This can be used for UI development straight away!

domcleal:
The linked Redmine ticket is in the Foreman project, but it also links to one in the Smart Proxy project.  Please either move it or use the other if that's the one you intend to implement.

modules/root/root_api.rb
```diff
@@ -25,4 +28,16 @@ class Proxy::RootApi < Sinatra::Base
       log_halt 400, e
     end
   end
+
+  get "/log_buffer" do
+    begin
+      content_type :json
+      buffer = ::Proxy::LogBuffer::Buffer.instance
+      from_serial = params[:from_serial].to_i
+      records = buffer.to_a(from_serial).collect(&:to_h)
```
ohadlevy: whats to_h in this context? its undefined for rubies older than 2.0

ohadlevy: ah, i see struct doesnt have a to_hash method.

ohadlevy: changing this line to:

``` ruby
records = buffer.to_a(from_serial).collect{|r| Hash[r.each_pair.to_a]}
```

not sure about performance

config/settings.yml.example
```diff
@@ -56,3 +56,9 @@
 # Uncomment and modify if you want to change the log level
 # WARN, DEBUG, ERROR, FATAL, INFO, UNKNOWN
 #:log_level: ERROR
+
+# Log buffer size and extra buffer size (for errors)
+#:log_buffer: 2000
+#:log_buffer_errors: 1000
+# Log buffer threshold (debug messages are ignored by default)
+#:log_buffer_level: INFO
```
dmitri-d: Why do we need separate levels for the logger and the event buffer? The latter shouldn't filter any events, it's up to final consumers; otherwise it's possible to configure logger for debug, buffer for info, and in the end not see any debug-level messages in the log.

lib/launcher.rb
```diff
@@ -132,9 +132,7 @@ def launch
       # do nothing. This is to prevent the exception handler below from catching SystemExit exceptions.
       raise
     rescue Exception => e
-      logger.error("Error during startup, terminating. #{e}")
-      logger.debug("#{e}:#{e.backtrace.join("\n")}")
-
+      logger.error("Error during startup, terminating. #{e}", e.backtrace)
```
dmitri-d, line 135: this was split on purpose, in order not to pollute error-level log with backtraces.

lib/proxy/helpers.rb
```diff
@@ -19,8 +19,7 @@ def log_halt code=nil, exception=nil
       code     = code || 400
     end
     content_type :json if request.accept?("application/json")
-    logger.error message
-    logger.debug exception.backtrace.join("\n") if exception.is_a?(Exception)
+    logger.error message, exception
```
dmitri-d, line 22: ditto.

lib/proxy/plugin.rb
```diff
@@ -140,8 +140,7 @@ def configure_plugin
       logger.info("'#{plugin_name}' module is disabled.")
     end
   rescue Exception => e
-    logger.error("Couldn't enable plugin #{plugin_name}: #{e}")
-    logger.debug("#{e}:#{e.backtrace.join("\n")}")
+    logger.error("Couldn't enable plugin #{plugin_name}: #{e}", e.backtrace)
```
dmitri-d, line 143: pls. don't join error and debug messages.

modules/root/root_api.rb
dmitri-d: I don't like the name of this end point, as it refers to the underlying implementation; I'd suggest log, logs. events, etc. Keep in mind we cannot rename this end-point after the api has been made public.

lib/proxy/log_buffer/buffer.rb
```diff
@@ -0,0 +1,83 @@
+require 'proxy/log_buffer/ring_buffer'
+
+module Proxy::LogBuffer
+
+  LogRecord = Struct.new(:serial, :timestamp, :level, :message)
```
dmitri-d: Why do we need a serial? Why not use a string/constant for the level of the message, instead of a cryptic integer. Not gonna insist on message classes, but having two structs probably means that you should consider that.

GregSutcliffe:
Might I make a few process comments here?

> lzap: the only problem with github branch comments is they disappear on branch deletion I think while PR can be closed forever.

Could we discuss on the -dev list then? I'm not sure a PR is the place to decide the relative merits of two approaches, and the discussion would be better in a place likely to be seen by the whole team (rather than just those who subscribe to the smart-proxy repo). I haven't seen any mails discussing this feature at all, despite that being our usual place to discuss direction (other than Kyle's wireframes a long time back, which is only tangentially related to this problem).

> lzap: this patch can be easily used for developing/testing the UI. 

Should we be asking others to work on something that hasn't been agreed on? If we decide not to go this way, then a lot of people have wasted their time.

> ohad: i guess we could use a different logger per feature? or just wait for @witlessbird new / improved event system?

As Dmitri says, the API is unversioned. We have only once chance to get this right. If we're _ever_ going to use the eventing system he proposes, then it must be done now.

I'm struggling to see what the user stories / requirements for this feature are, because the tickets have _zero_ description in them. Can someone elaborate on this? Working from my own opinions though, I would regard the ability to filter events to a specific module as fairly important - requesting the whole JSON stack of everything every time seems wasteful. Also, from a community/plugin perspective, giving plugin authors the option to _only_ consume the events from their proxy plugins is a nice thing to do.

I'm going to stop here as I feel this debate should be had on the dev list. I suggest we get a list of requirements for this feature and debate the merits of each approach there.

lib/proxy/log_buffer/buffer.rb
dmitri-d: Additionally, you are not persisting the serial counter between restarts, dropping it's monotonicity property, which makes it less useful. Maybe consider using a Time.to_f as a timestamp, which will pretty much be unique per message.

lib/proxy/log_buffer/ring_buffer.rb
```diff
@@ -0,0 +1,56 @@
+module Proxy::LogBuffer
```
dmitri-d, line 1: No copyright notice that this uses code from the celluloid project; MIT license under which celluloid is distributed requires attribution.

lib/proxy/log_buffer/buffer.rb
```diff
@@ -0,0 +1,83 @@
...
+      @serial = 0
+      @buffers_set_up = false
+    end
+
+    def new_serial
+      @serial += 1
+    end
+
+    # late initialization
+    def setup_buffers(size = nil, size_tail = nil, level = nil, level_tail = nil)
```
dmitri-d: why delay the initialization? Our startup times aren't slow, this does delay potential issues, making them harder to troubleshoot.

lib/launcher.rb
dmitri-d, line 135: Apologies, I see now that you debug-level messages are being automatically created. Pls. disregard this and other comments about that.

test/buffer_test.rb
```diff
@@ -0,0 +1,99 @@
+require 'test_helper'
```
dmitri-d: I assume this is from celluloid too? Pls. add a copyright notice then.

lzap:
@ohadlevy said on IRC modules are blocker for this, I added modules support, this took me 15 minutes, I am getting this from the caller stack and it works pretty well: https://gist.github.com/lzap/1eec112f7fd2993a5bc3

I am going to incorporate remarks anyway.

modules/root/root_api.rb
lzap: Fixed `to_h` with a backport, renamed endpoint to `logs`.

config/settings.yml.example
lzap: The idea is simple - there are two buffers and the "tail" one only stores errors and fatals after they expire. This is not configurable. What you can configure is the input level filtering, by default it ignores debug messages.

lib/proxy/log_buffer/buffer.rb
lzap: Changed to epoch-miliseconds serial.

lib/proxy/log_buffer/ring_buffer.rb
lzap, line 1: Added attribution.

lib/proxy/log_buffer/buffer.rb
lzap: I was getting chicken and egg problem. If there is a better way, tell me.

test/buffer_test.rb
lzap: Nah I wrote the tests.

lzap:
@ohadlevy added modules support, fixed comments. This is how records now look like:

```
        {
            "module": "puppet_proxy",
            "timestamp": 1453107943.7815278,
            "level": 1,
            "message": "Initializing from Puppet config file: /home/lzap/.puppet/puppet.conf"
        }
```

No serial anymore, timestamp is acccurate enough to use it for filtering. Module is determined from the caller stack, I search for `modules/XYZ` in the first three elements, it looks like 99% of all matches is the first element. Yes, it's a hack, but it gives us what we need without rearchitecting the smart proxy.

config/settings.yml.example
dmitri-d: Still don't understand why we'd need filters on the buffer: right now it's possible to configure logger for debug, buffer for info, and in the end not see any debug-level messages in the log.

dmitri-d:
Why not use a string representation of the event type, instead of an integer? Deriving module name based on callstack won't always work: there's no requirement for modules and providers to share the same namespace (and a few plugins do just that).

lzap:
Changed to string representation.

It's not ideal, it works with most core modules and when we have time, I can add per-module logging. But that is a bigger change we don't have time for I think. This should be out of the table for now IMHO.

lib/proxy/log_buffer/buffer.rb
```diff
@@ -0,0 +1,102 @@
+require 'date'
+require 'proxy/log_buffer/ring_buffer'
+
+# Adopted from Celluloid library (ring_buffer.rb).
+# Copyright (c) 2011-2014 Tony Arcieri. Distributed under the MIT License.
+# https://github.com/celluloid/celluloid/blob/0-16-stable/lib/celluloid/logging/ring_buffer.rb
+module Proxy::LogBuffer
+
+  LogRecord = Struct.new(:module, :timestamp, :level, :message, :backtrace) do
+    def to_h
```
ohadlevy, line 10: IMHO put in a separate file

lib/proxy/log_buffer/decorator.rb
```diff
@@ -0,0 +1,70 @@
+require 'proxy/log_buffer/decorator'
+require 'proxy/log_buffer/buffer'
+
+module Proxy::LogBuffer
+  class Decorator
+    def initialize(logger)
+      @logger = logger
+    end
+
+    def add(severity, message = nil, progname = nil, backtrace = nil, a_module = 'core', &block)
```
ohadlevy, line 11: nitpick: &block is not needed. you can still pass blocks
@dLobatog told me that passing &block makes it slower, but i have no data to back this up :)

lib/proxy/log_buffer/ring_buffer.rb
ohadlevy, line 1: should you update https://github.com/theforeman/smart-proxy/blob/develop/LICENSE too ?

lzap:
I will change the code to per-module loggers and add error counters so we are able to easily tell which modules failed during initialization. This should solve our problem while not changing any existing API. My initial idea of extending the features call does not work as the API does not follow the golden rule of a root element:

```
$ curl -sk http://localhost:8448/features | json_reformat 
[
    "abrt",
    "bmc",
    "dhcp",
    "discovery",
    "discovery_image",
    "dns",
    "pulp",
    "puppetca",
    "salt",
    "templates",
    "tftp"
]
```

dmitri-d:
How are you planning on doing this? The changes I made to the DI container to support something like this aren't in the core yet. I'd like to avoid per-module singletons too: this will affect all plugins (both in and outside of the core), would suck to force other developers to go keep changing their designs.

lzap:
Yeah, I just realized this too, there is little time to do this, therefore I am removing all module-related code including the hack [1] which will not work properly and I am only adding explicit hash of initialization failures. This only needs one extra call in the plugin initialization rescue block to give us list of failed plugins with their latest error messages:

```
$ curl -sk http://localhost:8448/logs?from_timestamp=1 | json_reformat 
{
    "info": {
        "size": 100,
        "tail_size": 50,
        "level": 1,
        "level_tail": 3,
        "failed_modules": {
            "colly": "No such file or directory - \"/var/run/collectd-unixsock\"",
            "puppet": "Cannot set modulepath settings in puppet.conf"
        }
    },
    "buffer": ...
```

Therefore the "logs" API gets us what Ohad wants:
- list of failed modules (only initialization phase - we can't do better at this point)
- the initialization error messages themselves
- log buffer for futher investigation

[1]:

``` ruby
    MODULE_REGEXP = %r!modules/([^/]+)!
    def detect_module(the_caller)
      # look four calls deep in the stack
      match = MODULE_REGEXP.match(the_caller[0])
      1.upto(3) do |i|
        match ||= MODULE_REGEXP.match(the_caller[i]) if the_caller[i]
      end
      if match
        match[1]
      else
        'core'
      end
    end
```

lzap:
When logging is refactored per-module, then we can easily refresh this API and provide also runtime-error messages per-module and also add module field for buffered log messages.

lzap:
Pushed, example output: 

```
curl -sk http://localhost:8448/logs?from_timestamp=1 | json_reformat | paste
http://sprunge.us/XiUH
```

config/settings.yml.example
dmitri-d: @lzap: ping?

lib/proxy/log_buffer/buffer.rb
```diff
@@ -0,0 +1,110 @@
...
+    end
+
+    def initialize
+      @mutex = Mutex.new
+      @buffers_set_up = false
+      @failed_modules = {}
+    end
+
+    # late initialization
+    def setup_buffers(size = nil, size_tail = nil, level = nil, level_tail = nil)
```
dmitri-d: Why delay initialization? Doesn't look like anything can raise an exception here, other than requiring of proxy/settings (shouldn't need that, it's one of the first things to load).

lib/proxy/log_buffer/ring_buffer.rb
dmitri-d, line 1: You should probably add attribution in every file that has celluloid code in it...

modules/root/root_api.rb
```diff
@@ -25,4 +28,16 @@ class Proxy::RootApi < Sinatra::Base
...
     end
   end
+
+  get "/logs" do
+    begin
+      content_type :json
+      buffer = ::Proxy::LogBuffer::Buffer.instance
+      from_timestamp = params[:from_timestamp].to_i
+      records = buffer.to_a(from_timestamp).collect(&:to_h)
+      { :info => buffer.info, :buffer => records }.to_json
```
dmitri-d: Wouldn't :log or :events be more descriptive than :buffer here?

lzap: Yes, future implementations might not use buffer at all. Changing to logs.

lib/proxy/log_buffer/buffer.rb
lzap: Because of settings. Logging is initialized before settings are loaded, apparently we want to log loading of settings. Therefore I am not able to initialize buffers in the constructor and I need late initialization after settings are loaded.

The workaround would be to create temporary buffer to store all messages at that early stage and then re-initialize and copy the records over to the final destination. I though it's not worth it. I can do it this way if you insist.

lzap: Oh you were right, the require was not needed at all. Initialization is now as simple as the contructor call!

lzap:
Pushed the changes.

lzap:
Rubocop failed for no apparent reason, [test].

lzap:
Hostname problem this time, networking is not in a good shape there. [test]

lzap:
Fixed Rubocop dependency issue: https://groups.google.com/forum/#!topic/foreman-dev/pgRYATnT-qo

lzap:
[test]

lzap:
[test]

lzap:
I don't get it, when I execute tests manually, it's all green. When github hook starts it, I see some MS DHCP failures. _deleted yet another re-test_ Any idee @domcleal ?

domcleal:
There's a PR open about that, please stop re-testing for an intermittent failure.

lzap:
Hmmm I missed that because it looked develop is all green. Thanks.

dmitri-d:
@lzap: https://github.com/theforeman/smart-proxy/pull/362

lzap:
@witlessbird please highlight your concerns here, audio was bad and I did not understand clearly, thanks.

config/settings.yml.example
```diff
@@ -56,3 +56,10 @@
 # Uncomment and modify if you want to change the log level
 # WARN, DEBUG, ERROR, FATAL, INFO, UNKNOWN
 #:log_level: ERROR
+
+# Log buffer size and extra buffer size (for errors). Defaults to 1500 messages in total,
+# which is about 250 kB request.
+#:log_buffer: 1000
+#:log_buffer_errors: 500
+# Log buffer threshold (debug messages are ignored by default)
+#:log_buffer_level: INFO
```
dmitri-d: @lzap: I still do not understand the need for specifying log level in both the buffer and the logger. This could lead to a situation when 'debug' was specified for the logger, 'info' for the buffer, resulting in no debug-level information making it to the logger.

dmitri-d:
Another note I had was in relation to the attribution notices: we either need to include them in all files where celluloid code has been included, or have a generic notice that says something like "this code contains code from the celluloid project". Possibly both (don't know, probably worth asking the lawyers).

ohadlevy:
I would suggest to do the same as foreman, in a license file?

dmitri-d:
Sure, if it works for foreman, it works for smart-proxy.

config/settings.yml.example
lzap: I understand, I can simply inherit it from the logging setting. Good point.

lzap:
Important remark - I need to enable

```
  authorize_with_trusted_hosts
  authorize_with_ssl_client
```

for this endpoint for security.

lzap:
Ok done:
- Removed the level setting, it respects what's set in the decorated logger instance.
- Improved tests (split into separate files in `logs_buffer` subdir)
- Changed LICENSE in the same form as in Foreman Core
- Enabled security, the endpoint is now available only via HTTPS with cert or trusted host
- Rebased on top of the jenkins fix PR (all tests are green now)

modules/root/http_config.ru
```diff
@@ -1,5 +1,10 @@
 require 'root/root_api'
+require 'root/logs_api'
 
 map "/" do
   run Proxy::RootApi
 end
+
+map "/logs" do
```
dmitri-d, line 8: Should probably go into a dedicated module then; I'm ok with delaying this though...

dmitri-d:
merged in at 91d4aa6. Thanks, @lzap!

lzap:
Thanks for comments.

domcleal:
Please add this to the smart proxy API docs: http://projects.theforeman.org/projects/smart-proxy/wiki/API

domcleal:
@lzap could you document this please?  See above.

lzap:
> @lzap could you document this please?  See above.

Sure, I am just little late on github notifications.

## 

Later,
 Lukas #lzap Zapletal

PR #360, opened 2016-01-15
Fixes #13219: force utf-8 external encoding on puppet classes

dmitri-d:


dmitri-d:
Argh, dhcp tests strike again. [test]

domcleal:
Is this only on particular versions of Ruby, or certain locales?  I tried running under WEBrick on Ruby 1.9.3 and 2.1.0 with LANG set to UTF-8 and ISO-88591 but couldn't trigger the original bug.

dmitri-d:
I think only earlier 1.9.3s had external encoding set to US-ASCII, any recent build should use UTF-8 instead. I could only trigger the issue when I modified the code to read puppet classes using US-ASCII external encoding.

domcleal:
Okay, merged as 0a70eea, thanks @witlessbird.

PR #361, opened 2016-01-15
[DO NOT MERGE] investigating intermittent test failures

dmitri-d:


dmitri-d:
[test]

witlessbird:
[test]

domcleal:
You'll probably find it quicker to just rebuild the job on Jenkins (http://ci.theforeman.org/job/test_proxy_develop_pr_core/526/rebuild).  The keyword is only checked for every 30 minutes.

dmitri-d:
Found the issue: https://github.com/theforeman/smart-proxy/blob/develop/modules/dhcp_native_ms/dhcp_native_ms_main.rb#L265

Apparently on older rubies the order of keys after merge/during select is non-deterministic, hence intermittent errors. Unfortunately, this bit is quite broken (there are several reports about broken behavour: http://projects.theforeman.org/issues/10556, http://projects.theforeman.org/issues/5481, http://projects.theforeman.org/issues/4148) and the options parser needs to be re-written. Good news is this affects windows only.

dmitri-d:
closed in favour of https://github.com/theforeman/smart-proxy/pull/362

PR #362, opened 2016-01-19
Fixes #13129: fixed parsing of dhcp options for ms dhcp server

dmitri-d:


domcleal:
Could you add some info to the commit message body please about what this is fixing exactly?  I'm struggling to see the link between the ticket and where exactly the fix is.  Did you fix the intermittent behaviour somewhere and then noticed options parsing was broken?

dmitri-d:
In retrospect I should've moved my earlier comment into this PR:

Found the issue: https://github.com/theforeman/smart-proxy/blob/develop/modules/dhcp_native_ms/dhcp_native_ms_main.rb#L265

Apparently on older rubies the order of keys after merge/during select is non-deterministic, hence intermittent errors. Unfortunately, this bit is quite broken (there are several reports about broken behavour: http://projects.theforeman.org/issues/10556, http://projects.theforeman.org/issues/5481, http://projects.theforeman.org/issues/4148) and the options parser needs to be re-written. Good news is this affects windows only.

I'll expand the commit message too.

dmitri-d:
@domcleal: updated the commit message.

domcleal:
Thanks @witlessbird , merged as b9309bf.

PR #363, opened 2016-01-22
Fixes #13326 - Sending feature list to Foreman on startup

xprazak2:
On startup, Smart proxy prods Foreman to ask for a feature list.

### DIFF
--- a/lib/proxy/startup_info.rb
+++ b/lib/proxy/startup_info.rb
@@ -1,0 +1,16 @@
+require 'proxy/request'
+require 'socket'
+
+module Proxy
+  class StartupInfo < Proxy::HttpRequest::ForemanRequest
+    include Proxy::Log
+
+    def put_features
+      params = { :proxy_name => Socket.gethostname, :startup_refresh => true }.to_json
+      req = request_factory.create_put("/api/v2/smart_proxies/startup_refresh", params)
+      response = send_request(req)
+      logger.warn "Failed to notify Foreman on startup. Received response: #{response.code} #{response.msg}" unless response.code == "200"
+      response
+    end
+  end
+end
--- a/lib/launcher.rb
+++ b/lib/launcher.rb
@@ -1,6 +1,7 @@
 require 'proxy/log'
 require 'proxy/settings'
 require 'proxy/signal_handler'
+require 'proxy/startup_info'
 
 module Proxy
   class Launcher
@@ -122,6 +123,7 @@ module Proxy
       t1 = Thread.new { https_app.start } unless https_app.nil?
       t2 = Thread.new { http_app.start } unless http_app.nil?
 
+      Proxy::StartupInfo.new.put_features
       Proxy::SignalHandler.install_traps
 
       (t1 || t2).join
--- a/lib/proxy/request.rb
+++ b/lib/proxy/request.rb
@@ -39,6 +39,13 @@ module Proxy::HttpRequest
       req.body = body
       req
     end
+
+    def create_put(path, body, headers = {})
+      req = Net::HTTP::Put.new(uri(path).path)
+      req = add_headers(req, headers)
+      req.body = body
+      req
+    end
   end
 
   class ForemanRequest
--- a/test/request_test.rb
+++ b/test/request_test.rb
@@ -37,4 +37,11 @@ class RequestTest < Test::Unit::TestCase
     result = @request.send_request(proxy_req)
     assert_equal("body", result.body)
   end
+
+  def test_put
+    stub_request(:put, @foreman_url+'/path').with(:body => "body").to_return(:status => [200, 'OK'], :body => "body")
+    proxy_req = @request.request_factory.create_put("/path", "body")
+    result = @request.send_request(proxy_req)
+    assert_equal("body", result.body)
+  end
 end
--- a/test/startup_info_test.rb
+++ b/test/startup_info_test.rb
@@ -1,0 +1,21 @@
+require 'test_helper'
+require 'net/http'
+require 'mocha'
+require 'webmock/test_unit'
+require 'proxy/startup_info'
+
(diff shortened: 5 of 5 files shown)


domcleal:
This ticket is assigned to the Foreman project (again?), please use a different one.

dmitri-d:
I'm still not convinced this is a good way of doing this. What will happen if proxy cannot reach foreman (a firewall between foreman and proxy comes first to mind)? Atm we don't have such a requirement (@domcleal, @GregSutcliffe, thoughts?), and the only plugin that reaches back to forman is templates, which doesn't perform any updates.

Implementation-wise: this needs to handle situations when foreman isn't available or slow to respond. I'd also suggest performing this call in a separate thread, in order not to slow down the boot up.

domcleal:
Indeed, looks like a terrible idea to me.

lib/proxy/startup_info.rb
```diff
@@ -0,0 +1,16 @@
+require 'proxy/request'
+require 'socket'
+
+module Proxy
+  class StartupInfo < Proxy::HttpRequest::ForemanRequest
+    include Proxy::Log
+
+    def put_features
+      params = { :proxy_name => Socket.gethostname, :startup_refresh => true }.to_json
+      req = request_factory.create_put("/api/v2/smart_proxies/startup_refresh", params)
```
dmitri-d, line 10: Why using put here? This doesn't update a specific resource (no resource id in the url), a 'post' would be more appropriate, I think.

lib/proxy/startup_info.rb
```diff
@@ -0,0 +1,16 @@
+require 'proxy/request'
+require 'socket'
+
+module Proxy
+  class StartupInfo < Proxy::HttpRequest::ForemanRequest
+    include Proxy::Log
+
+    def put_features
+      params = { :proxy_name => Socket.gethostname, :startup_refresh => true }.to_json
```
dmitri-d, line 9: This may not return return expected (by foreman) hostname on a host with multiple interfaces. A slightly better approach would be to check 'bind_host' global setting and use the address there to determine the hostname. But: this would require reverse address resolution to work, and won't help when wildcard (or similar) was used.

GregSutcliffe:
-1 it's not a good idea. My concerns:
- What are the use case(s) we're trying to solve? I see very little description of a problem on 13290
- As @witlessbird says, what if Foreman is down? Or unreachable due to network topology
- If more than one Foreman host (eg a prod and test instance) are using the same proxy, how will you update both?

However my main concern is taking action to update Foreman without specific consent of the user. I would _not_ want my proxies to auto-update Foreman. What if I'm testing a new proxy plugin? What if I've accidentally removed the Puppet feature and now Foreman has dropped the puppetmaster from all my hosts?

Can we _please_ have the usecase/design discussion somewhere sensible (Redmine or the dev list) _before_ writing implementations? Otherwise, we are writing code that may never be merged.

xprazak2:
I am afraid I cannot answer the questions regarding all the use cases we are trying to solve here since this got to me from @ohadlevy and @shlomizadok. The exact description for this feature I received is: 
`Proxy post to Foreman's "/refresh" when booting (to let know of new changes)` 
so that is what I wrote into Redmine. 

I apologize for the wrong ticket, I will be more careful in the future.

GregSutcliffe:
@xprazak2 my comments weren't aimed at you in particular - I was raising that there has been no record of the user stories or requirements. If there _had_ been, then you and I would _both_ know why you're implementing this feature!

ohadlevy:
IMHO the reasons to implement a callback from the proxy to foreman are the following:
1. a way to invalidate the smart proxy related cache.
2. a way to detect if new feature are enabled on the proxy.

the stories i would like to address are:
- as a user, I've restarted my proxy and I see out of date data in the UI/API. (covered with 1)
- as a user, I've ran the installer (e.g. added the rex plugin) yet I have no idea i need to go to the smart proxies and click refresh. (covered partly with 2, more below)
  as a user i might want to see older errors from the proxy yet know they are from pre-restart (related to 1 and @lzap work on logs).

lets break each on a bit down:
for 1. i could see an alternative implementation, by simply keeping a timestamp of when the proxy was started, and provide that timestamp in one of the api calls (for example in one of the ping/status calls).
for 2. i think foreman needs to know a restart happened, what it does with it can be handled differently, for example:
if a new feature was added, i think its fair to simply add it and report it in the ui.
if a feature was removed, we can probably detect if it was due to configuration error (via the plugin init loggin) or not, in either case we don't remove it for the user, rather report it (e.g. in the status icon).

I think the only "allowed" automatic update in this case would be add new features to the proxy, and I would consider it an improvement, as its the default process a user usually take.

I do think that over time, the proxy will need to send an event stream back to foreman, in order to enrich the visibility / management in the application and enable a rich provisioning state machine.

regarding the implementation details that @witlessbird and @GregSutcliffe mentioned I fully agree it should not block the proxy in any level etc.
I do not think that the usage case of mutlitple foreman hosts using the same proxy is that common, if we however wish to support it we can simply provide an array of foreman urls (which we dont afaik support today anyway).

GregSutcliffe:
@ohadlevy that's a start, however, this is still the wrong place for it - it should be in the Redmine ticket or on the dev list (and, ideally we should decide which and _write it down_ in the Handbook). Can you update the ticket please? Having this discussion on this PR reinforces the idea that we've decided _this implementation_ is the right way forward, and that's clearly still under discussion.

I'll point out that @xprazak2 says he actually did not know the context of what he was trying to implement, and that's entirely because we have _no record_ of the stories/goals. This is avoided if we do things properly - anyone should be able to pick up the task based on what's in Redmine, with further context from the dev-list discussions.

lzap:
As I stated on the discussion list, I also don't think this is something we want in our codebase.

dmitri-d:
Closing this PR as it's been inactive for more than six months.

PR #364, opened 2016-01-22
Fixes #13028: provider will not be configured if the main plugin is disabled

dmitri-d:
- deprecated ::Proxy::Pluggable.initialize_after: this method predates providers and was intended for ordering of plugins during initialization. 
- made plugin initialization and configuration more testable.
- improved test coverage around plugin loading and initialization.


lib/proxy/pluggable.rb
```diff
@@ -100,6 +86,7 @@ def default_settings(a_hash = {})
     end
 
     def initialize_after(*module_names)
+      logger.warn("#{plugin_name}: 'initialize_after' has been deprecated and will be removed in future versions.")
```
domcleal: It appears that this isn't actual functional any more, since `validate_dependencies!` is no longer called with `@initialize_after`.  A few points follow:
1. Was there any use for feature this in existing plugins?  It looks like we now lose the ability of a provider or module to order itself after another module, and I don't know if we even take the `require` into account for ordering.  It still seems useful to me, despite the general ordering improvements in this PR.
2. It's not so much deprecated as disabled, so the code below should be removed as I don't think it's functional.
3. If being removed, it would be better to raise an exception as it's going to behave very differently.

lib/proxy/plugin.rb
```diff
@@ -116,10 +112,32 @@ def plugin(plugin_name, aversion)
...
+      end
+      if !(all_plugins.find {|plugin| plugin[:name] == p.to_sym && plugin[:enabled] == true})
+        raise ::Proxy::PluginMisconfigured, "Dependency '#{p}' of '#{plugin_name}' has not been enabled."
+      end
     end
   end
 
+  def validate!(all_plugins)
+    validate_dependencies!(all_plugins, self.class.dependencies)
+    validate_prerequisites_enabled!(all_plugins, [settings.use_provider]) if uses_provider?
```
domcleal, line 131: This method could probably stay in `Pluggable` as 3/4 of it's the same as the provider, and then call `validate_prerequisites_enabled!` & `super`.

modules/templates/templates_plugin.rb
```diff
@@ -6,15 +6,8 @@ class Plugin < ::Proxy::Plugin
...
-        logger.warn "missing :foreman_url: from configurations. 'templates' module is disabled"
-        ::Proxy::Plugins.disable_plugin(:templates)
-      end
-
-      unless Proxy::Templates::Plugin.settings.template_url
-        logger.warn "missing :templates_url: from configurations. 'templates' module is disabled"
-        ::Proxy::Plugins.disable_plugin(:templates)
-      end
+      loading_failed "missing :foreman_url: from configuration." unless Proxy::SETTINGS.foreman_url
+      loading_failed "missing :templates_url: from configuration." unless Proxy::Templates::Plugin.settings.template_url
```
domcleal: Looks like this should use `validate_presence` as it's in the plugin settings - it'll also check for empty values too then.  I don't think foreman_url can be checked like that, as it's a proxy/global setting.

lib/proxy/plugin.rb
```diff
@@ -15,59 +16,53 @@ def initialize(aname, aversion)
...
 
     def plugin_enabled?(plugin_name)
-      !!@@enabled[plugin_name.to_sym]
+      plugin = loaded.find {|p| p[:name] == plugin_name.to_sym}
+      plugin.nil? ? false : !!plugin[:enabled]
     end
 
     def disable_plugin(plugin_name)
-      @@enabled.delete(plugin_name.to_sym)
+      logger.error("::Proxy::Plugins.disable_plugin has been deprecated and will be removed from future versions of smart-proxy")
```
domcleal: 1. warn, not error
2. suggest raising an exception as the preferred alternative (Edit: or `loading_failed`)

lib/proxy/pluggable.rb
dmitri-d: > Was there any use for feature this in existing plugins?

This feature is not being used In plugins I checked (all under foreman org in github). 

> It looks like we now lose the ability of a provider or module to order itself after another module, and I don't know if we even take the require into account for ordering. It still seems useful to me, despite the general ordering improvements in this PR.

The goal is to remove guarantees about the order in which modules are loaded (modules are supposed to be independent of each other), other than providers before main modules. The functionality was introduced before providers, with idea was that plugins will be used to implement provider logic. Lack of use in real world in combination with not being able to handle dependency loops makes me want to remove the functionality rather than fix it.

Please note that the less strict module dependency functionality  is present and working (probably should be fixed to take into account enabled modules, as opposed to loaded ones).

> It's not so much deprecated as disabled, so the code below should be removed as I don't think it's functional.
> If being removed, it would be better to raise an exception as it's going to behave very differently.

True, will remove the code and raise an exception instead.

modules/templates/templates_plugin.rb
dmitri-d: good call re: `validate_presence` for `templates_url` setting.

dmitri-d:
Made changes suggested in the feedback.

modules/templates/templates_plugin.rb
```diff
@@ -5,16 +5,10 @@ class Plugin < ::Proxy::Plugin
 
     plugin :templates, ::Proxy::VERSION
 
-    after_activation do
-      unless Proxy::SETTINGS.foreman_url
-        logger.warn "missing :foreman_url: from configurations. 'templates' module is disabled"
-        ::Proxy::Plugins.disable_plugin(:templates)
-      end
+    validate_readable :template_url
```
domcleal: `validate_presence`?  It isn't a file.

lib/proxy/pluggable.rb
domcleal: That's all fine, thanks.  I just wanted to be clear about the new behaviour of the method, raising the exception's good.

lib/proxy/pluggable.rb
```diff
@@ -33,37 +33,30 @@ def after_activation
...
-    def validate_prerequisites_enabled!(prerequisites)
-      prerequisites.each do |p|
-        if !(::Proxy::Plugins.find_plugin(p))
-          raise ::Proxy::PluginMisconfigured, "Unable to find dependency '#{p}' of '#{plugin_name}'."
-        end
-        if !(::Proxy::Plugins.plugin_enabled?(p))
-          raise ::Proxy::PluginMisconfigured, "Dependency '#{p}' of '#{plugin_name}' has not been enabled."
-        end
-      end
+    def validate!(all_plugins)
```
domcleal, line 36: Nitpick, this has moved making the diff larger than it should be, it used to be above `execute_validators`.

dmitri-d:
- templates module uses 'validate_presence' instead of 'validate_readable' now
- moved `Pluggable#validate!` back to its original spot in the file

domcleal:
Merged as 53e560f, thanks @witlessbird.

PR #365, opened 2016-01-24
Fixes #13346 - revert 'each_line'

helge000:
**Works for me**: Simply reverts `.each_line` to `.each`. 
Maybe we could come up with something more stable if the use of [.readlines](https://github.com/theforeman/smart-proxy/blob/1.10-stable/modules/dhcp/providers/server/native_ms.rb#L211) becomes a sting in ruby =< 2.0.0

Ugly; but should work:

``` ruby
execute(cmd, msg).to_s.split("\n").each do |line|
 # ...
end
```

eg, [here](https://github.com/theforeman/smart-proxy/pull/365/files?diff=split#diff-08c4251c1dbcdc5c20755615254654eaL101)

### DIFF
--- a/test/dhcp/server_ms_test.rb
+++ b/test/dhcp/server_ms_test.rb
@@ -27,7 +27,7 @@ class DHCPServerMicrosoftTest < Test::Unit::TestCase
  172.29.216.0   - 255.255.254.0  -Active        -DC BRS               -
 
  Total No. of Scopes = 6
-Command completed successfully.')
+Command completed successfully.'.split("\n"))
     @server.stubs(:execute).with("scope 172.29.205.0 show reservedip", "Enumerated hosts on 172.29.205.0").returns('
 Changed the current scope context to 172.29.205.0 scope.
 
@@ -97,7 +97,7 @@ Changed the current scope context to 172.29.205.0 scope.
 No of ReservedIPs : 57 in the Scope : 172.29.205.0.
 
 Command completed successfully.
-')
+'.split("\n"))
     @server.stubs(:execute).with("scope 172.29.205.0 Show OptionValue", "Queried 172.29.205.0 options").returns('
 Changed the current scope context to 172.29.205.0 scope.
 
@@ -121,7 +121,7 @@ Options for Scope 172.29.205.0:
                 Option Element Type = IPADDRESS
                 Option Element Value = 172.29.205.1
 Command completed successfully.
-    ')
+    '.split("\n"))
     @server.stubs(:execute).with(
         regexp_matches(/^scope 172.29.205.0 Show ReservedOptionValue 172.29.205.25/),
         regexp_matches(/^Queried .+ options/)).returns('
@@ -147,7 +147,7 @@ Options for the Reservation Address 172.29.205.25 in the Scope 172.29.205.0 :
                 Option Element Type = STRING
                 Option Element Value = brslcs25
 Command completed successfully.
-')
+'.split("\n"))
     @server.loadSubnets
   end
 
--- a/modules/dhcp/providers/server/native_ms.rb
+++ b/modules/dhcp/providers/server/native_ms.rb
@@ -98,7 +98,7 @@ module Proxy::DHCP
 
       to_return = []
       # Extract the data
-      execute(cmd, msg).each_line do |line|
+      execute(cmd, msg).each do |line|
         #     172.29.216.6      -    00-a0-e7-21-41-00-
         if line =~ /^\s+([\w\.]+)\s+-\s+([-a-f\d]+)/
           ip  = $1
@@ -181,7 +181,7 @@ module Proxy::DHCP
       msg = "Enumerated the scopes on #{@name}"
 
       ret_val = []
-      execute(cmd, msg).each_line do |line|
+      execute(cmd, msg).each do |line|
         # 172.29.216.0   - 255.255.254.0  -Active        -DC BRS               -
         if match = line.match(/^\s*([\d\.]+)\s*-\s*([\d\.]+)\s*-\s*(Active|Disabled)/)
           next unless managed_subnet? "#{match[1]}/#{match[2]}"
@@ -252,7 +252,7 @@ module Proxy::DHCP
     def parse_options response
       optionId = nil
       options  = {}
-      response.each_line do |line|
+      response.each do |line|
         line.chomp!
         break if line.match(/^Command completed/)
 


dLobatog:
@helge000 Mind to write a test for this? Thanks!

helge000:
@dLobatog: certainly, though I think I'll need some assistance ;)

I've looked [here](https://github.com/theforeman/smart-proxy/blob/develop/test/dhcp_ms_native/server_ms_test.rb#L17) for starters. Your test case returns a string - though ruby doc (even for 1.9) [states readlines will be of type array](http://ruby-doc.org/core-1.9.3/IO.html#method-c-readlines)...

Where should I start putting the test case?

My ruby returns an array; split by lines, like this:

``` ruby
["\n", "==============================================================================\n"," Scope Address  - Subnet Mask    - State        - Scope Name          -  Comment   \n", "==============================================================================\n", "\n", " 192.168.0.0      - 255.255.255.0  -Active        -management           -vlan 1        \n", " 192.168.2.0      - 255.255.255.0  -Active        -stora                -vlan 2        \n", " 192.168.3.0      - 255.255.255.0  -Active        -storb            -vlan3         \n", " 192.168.10.0     - 255.255.255.0  -Active        -servers              -vlan 10       \n", "\n", " Total No. of Scopes = 9 \n", "Command completed successfully.\n"]
```

dmitri-d:
@helge000: I'd suggest you update existing tests -- you'll need to break up strings returned by stubs into arrays in this case.

helge000:
@witlessbird I hope this is about right

test/dhcp/server_ms_test.rb
```diff
@@ -14,140 +14,12 @@ def setup
...
-        OptionId : 3
-        Option Value:
-                Number of Option Elements = 1
-                Option Element Type = IPADDRESS
-                Option Element Value = 172.29.205.1
-Command completed successfully.
-    ')
+    @server.stubs(:execute).with("show scope", "Enumerated the scopes on 1.2.3.4").returns(["", "==============================================================================", " Scope Address  - Subnet Mask    - State        - Scope Name          -  Comment", "==============================================================================", " 172.24.166.0   - 255.255.255.128-Active        -WLAN SMW Clients BRS -WLAN range SMW", " 172.24.230.0   - 255.255.255.128-Active        -WLAN SMW Clients BRS -WLAN range SMW", " 172.29.204.0   - 255.255.255.0  -Active        -Client VLAN Scope    -DC Standardiza", " 172.29.205.0   - 255.255.255.128-Active        -Server VLAN Scope    -DC Standardiza", " 172.29.205.128 - 255.255.255.128-Active        -Management VLAN Scope-DC Standardiza", " 172.29.216.0   - 255.255.254.0  -Active        -DC BRS               -", " Total No. of Scopes = 6", "Command completed successfully."])
```
dmitri-d: This is pretty hard to read. What I had in mind was to keep existing multi-line strings around, but to split them on new-lines...

test/dhcp/server_ms_test.rb
```diff
@@ -14,121 +14,116 @@ def setup
...
+      "    172.29.205.27     -    00-03-ba-45-79-31-",
+      "    172.29.205.23     -    00-0c-29-f6-f9-ce-",
+      "    172.29.205.25     -    00-0c-29-64-a1-f1-",
+      "    172.29.205.26     -    00-0c-29-46-8f-10-",
+      "",
+      "",
+      "No of ReservedIPs : 57 in the Scope : 172.29.205.0.",
+      "",
+      "Command completed successfully."])
+    @server.stubs(:execute).with("scope 172.29.205.0 Show OptionValue", "Queried 172.29.205.0 options").returns(["",
```
helge000: @witlessbird I am not sure on this one. Currently I think GitHub shows it wrong it seems; but I think this needs also be an array; since it uses `server.stubs(:execute)`?

dmitri-d:
@helge000: My apologies, I should've been more clear -- you can use String#split("\n") to split a string on newlines; the result is going to be an array of strings. A multi-line string is a bit easier to deal with than a manually constructed array of strings...

helge000:
@witlessbird Hm, ok. I'll still would argue passing the "real life" array is better for a test case then using split on the string.

test/dhcp/server_ms_test.rb
dmitri-d: Not sure what you mean. #execute method will always behave the same way, and should be stubbed/mocked similarly across the board.

dmitri-d:
@helge000: not sure it makes much of a difference -- we are stubbing a call here; A multi-line string is easier to deal with both reading- and maintenance-wise though.

helge000:
@witlessbird no worries; your project, your test case! 
The provider is much more important to me anyway ;)

(Though I'm still unconvinced of the maintenance-wise argument; at least in this particular `netsh.exe` - case)

dmitri-d:
[test]

dmitri-d:
@helge000: could you squash your commits pls?

helge000:
@witlessbird I see no conflicts; but suppose you mean rebase on _develop_?

dmitri-d:
@helge000: your PR consists of four commits, I was wondering if you could merge all these commits into one (a bit less work for me during merging).

helge000:
@witlessbird sorry, noob here. Forgot about --amend and --force. Hopefully better now

dmitri-d:
When testing your changes I noticed that your PR is based on 1.10-stable branch. Could you rebase it against develop? Otherwise looks good!

domcleal:
The PR would need to be created against develop too for the tests to run, which means re-opening it (branch can't be changed on a PR).

helge000:
@witlessbird , @domcleal: I thought of this as a hotfix; so I opened it against 1.10-stable. Just tell me how you like it, opening a new PR is simple enough.

domcleal:
Unless the bug has actually been fixed in develop (by some other means, which I've not checked), then always open a PR to develop and then we'll handle the patching of the stable release after.  If we fix it only in a stable branch then future versions will be broken.

helge000:
@domcleal I'll mind this for the future. As long as it gets backported to 1.10-stable... Expect a new PR; I hope foreman-bot allows it for the same issue

helge000:
For the record, opened #365 for `develop`. Mind you, it will be easier to merge this PR in 1.10-stable

dmitri-d:
I'm going to close this PR, as @domcleal mentioned we'll do backporting once this has been fixed in the  develop branch.

PR #367, opened 2016-01-27
Refs #12715 - decoupled logs API from core

lzap:
This decouples the log buffer API from core. It has two main purposes. First,
it's a cleaner design. Second, clients can now use the standard Features API to
detect, if Logs API is enabled. Therefore integration with older version of
Smart Proxies is easier.


lzap:
@witlessbird could you give it a quick read?

.gitignore
```diff
@@ -5,7 +5,7 @@ config/settings.d/*.yml
 config/migration_state
 config_*/*
 .idea
-logs
+/logs
```
ohadlevy, line 8: why /?

lzap, line 8: Because it ignores all logs subdirectories, we want only _the_ proxy logs.

config/settings.d/logs.yml.example
```diff
@@ -0,0 +1,5 @@
+---
+# Can be true, false, or http/https to enable just one of the protocols
+:enabled: false
+
+# Log buffer configuration is in core file (settings.yml)
```
dLobatog, line 5: If log buffer is only used to show it in the Foreman smart proxy page - can we move the log_buffer and log_buffer errors to this file instead?

dmitri-d, line 5: This config file deals with log api only (exposing events over http). This module may or may not be enabled, but internal logging is always available.

I think the main settings file is a good spot for these.

modules/logs/logs_plugin.rb
```diff
@@ -0,0 +1,14 @@
...
+  http_rackup_path File.expand_path("http_config.ru", File.expand_path("../", __FILE__))
+  https_rackup_path File.expand_path("http_config.ru", File.expand_path("../", __FILE__))
+
+  plugin :logs, ::Proxy::VERSION
+  default_settings :enabled => true
+
+  after_activation do
+    begin
+      buffer = Proxy::LogBuffer::Buffer.instance
+      logger.info "Log buffer plugin activated with #{buffer.size}/#{buffer.size_tail} capacity"
```
dmitri-d: I think this is misleading, as the logging subsystem is initialized elsewhere.

lzap: Will `Log buffer API initialized, available capacity: #{buffer.size}/#{buffer.size_tail}` be better? Or do you prefer not to print anything?

lzap: I reworded and also changed to DEBUG level.

dmitri-d:
merged in at e26eb4f. Thanks, @lzap.

dLobatog:
I get a very strange stack trace after this change, and I can't start the proxy after this change. Not sure what does this module have to do with Puppet initialization though...

https://gist.github.com/dLobatog/0865f01841381a2f3088

dmitri-d:
Don't think this is related to that change -- this appears to be a conflict between two versions of puppet: one from /usr/share/ruby/vendor_ruby and another /usr/local/share/gems/gems/puppet-3.6.2. I suspect this is happening when puppet_proxy or puppet_ca module is being loaded...

PR #369, opened 2016-01-27
Fixes #13421 - dnscmd now adds/removes records in sub zones

helge000:
I could only test this on my production env running 1.10-stable, but this seems to work now running the same patch. 

Please advice reading from setting `::Proxy::Dns::Plugin.settings.dns_domain` directly...
Also, I never ever ran Windows DNS without AD, hence I cannot really tell if the old regex method works in this case. 
But in case of AD, for sure the `zonename` parameter has to match the active directory zone and must not contain any sub zone.

Also, as I stated in the ticket, I can think of no reliable way to determine this zone name apart from a setting - witch means better ideas welcome!
(Ok, not true, we could use `dnscmd /EnumZones`, parse it and select the best match... but I really think this would overdue it, aka I do not like to be the one implementing this ;)

That said a better fallback could be the second level domain?

```
D, [2016-01-27T23:36:54.516411 #3676] DEBUG -- : executing: c:\windows\system32\cmd.exe /c c:\Windows\System32\dnscmd.exe localhost /RecordAdd domain.com testdns2.sec.domain.com. A 192.168.0.155
I, [2016-01-27T23:36:54.563119 #3676]  INFO -- : Added DNS entry testdns2.sec.domain.com => 192.168.0.155
I, [2016-01-27T23:36:54.563119 #3676]  INFO -- : 2001:dbff:2c2c::15 - - [27/Jan/2016 23:36:54] "POST /dns/ HTTP/1.1" 200 - 0.0936
```

### DIFF
--- a/modules/dns_dnscmd/dns_dnscmd_main.rb
+++ b/modules/dns_dnscmd/dns_dnscmd_main.rb
@@ -16,7 +16,7 @@ module Proxy::Dns::Dnscmd
       if found = dns_find(fqdn)
         raise(Proxy::Dns::Collision, "#{fqdn} is already used by #{found}") if found != ip
       else
-        zone = fqdn.sub(/[^.]+./,'')
+        zone = match_zone(fqdn)
         msg = "Added DNS entry #{fqdn} => #{ip}"
         cmd = "/RecordAdd #{zone} #{fqdn}. A #{ip}"
         execute(cmd, msg)
@@ -33,7 +33,7 @@ module Proxy::Dns::Dnscmd
     def remove_a_record(fqdn)
       ip = dns_find(fqdn)
       raise Proxy::Dns::NotFound.new("Cannot find DNS entry for #{fqdn}") unless ip
-      zone = fqdn.sub(/[^.]+./,'')
+      zone = match_zone(fqdn)
       msg = "Removed DNS entry #{fqdn} => #{ip}"
       cmd = "/RecordDelete #{zone} #{fqdn}. A /f"
       execute(cmd, msg)
@@ -67,6 +67,7 @@ module Proxy::Dns::Dnscmd
         std_err.close unless std_in.nil?
       end
       report msg, response, error_only
+      response
     end
 
     def report msg, response, error_only
@@ -84,6 +85,38 @@ module Proxy::Dns::Dnscmd
     rescue
       logger.error "Dnscmd failed:\n" + (response.is_a?(Array) ? response.join("\n") : "Response was not an array! #{response}")
       raise Proxy::Dns::Error.new("Unknown error while processing '#{msg}'")
+    end
+
+    def match_zone(fqdn)
+      dns_zones = enum_zones
+      weight = 0 # sub zones might be independent from similar named parent zones; use weight for longest suffix match
+      matched_zone = nil
+
+      dns_zones.each do |zone|
+        zone_labels = zone.split(".").reverse
+        zone_weight = zone_labels.length
+        fqdn_labels = fqdn.split(".")
+        fqdn_labels.shift
+        is_match = zone_labels.all? { |zone_label| zone_label == fqdn_labels.pop }
+        # match only the longest zone suffix
+        if is_match && zone_weight >= weight
+          matched_zone = zone
+          weight = zone_weight
+        end
+      end
+      raise Proxy::Dns::NotFound.new("The DNS server has no authoritative zone for #{fqdn}") unless matched_zone
+      matched_zone
+    end
+
+    def enum_zones
+      zones = []
+      response = execute('/EnumZones')
+      response.each do |line|
+        next unless line =~  / Primary /
+        zones << line.sub(/^ +/, '').sub(/ +.*$/, '').chomp("\n")
+      end
+      logger.debug "Enumerated dns zones: #{zones}"
+      zones
     end
   end
 end
--- a/test/dns_dnscmd/dnscmd_test.rb
+++ b/test/dns_dnscmd/dnscmd_test.rb
@@ -2,7 +2,25 @@ require 'test_helper'
 require 'dns_dnscmd/dns_dnscmd'
 require 'dns_dnscmd/dns_dnscmd_main'
 
+class DnscmdForTesting < Proxy::Dns::Dnscmd::Record
+  def initialize(dns_zones)
+    @enum_zones = dns_zones
+  end
+  attr_accessor :enum_zones
+end
+
 class DnsCmdTest < Test::Unit::TestCase
+  def setup
+    @server = DnscmdForTesting.new([
+      "_msdcs.bar.domain.local",
+      "168.192.in-addr.arpa",
+      "33.168.192.in-addr.arpa",
+      "e.e.d.8.b.d.0.1.0.0.2.ip6.arpa",
+      "bar.domain.local",
+      "domain.local",
+      "TrustAnchors"])
+  end
+
   def test_dnscmd_provider_initialization
     Proxy::Dns::Dnscmd::Plugin.load_test_settings(:dns_server => 'a_server')
     Proxy::Dns::Plugin.load_test_settings(:dns_ttl => 999)
@@ -12,64 +30,110 @@ class DnsCmdTest < Test::Unit::TestCase
     assert_equal 999, server.ttl
   end
 
-  def test_create_address_record
-    Proxy::Dns::Dnscmd::Record.any_instance.expects(:dns_find).with('host.domain').returns(false)
-    Proxy::Dns::Dnscmd::Record.any_instance.expects(:execute).with('/RecordAdd domain host.domain. A 192.168.33.33', anything).returns(true)
-    assert Proxy::Dns::Dnscmd::Record.new.create_a_record('host.domain', '192.168.33.33')
+  def test_create_address_record_with_longest_zone_match
+    Proxy::Dns::Dnscmd::Record.any_instance.expects(:dns_find).with('host.foo.bar.domain.local').returns(false)
+    Proxy::Dns::Dnscmd::Record.any_instance.expects(:execute).with('/RecordAdd bar.domain.local host.foo.bar.domain.local. A 192.168.33.33', anything).returns(true)
+    assert @server.create_a_record('host.foo.bar.domain.local', '192.168.33.33')
   end
 
-  def test_overwrite_address_record
-    Proxy::Dns::Dnscmd::Record.any_instance.expects(:dns_find).with('host.domain').returns('192.168.33.33')
-    Proxy::Dns::Dnscmd::Record.new.create_a_record('host.domain', '192.168.33.33')
+  def test_overwrite_address_record_with_longest_zone_match
+    Proxy::Dns::Dnscmd::Record.any_instance.expects(:dns_find).with('host.foo.bar.domain.local').returns('192.168.33.33')
+    @server.create_a_record('host.foo.bar.domain.local', '192.168.33.33')
   end
 
   def test_create_duplicate_address_record_fails
-    Proxy::Dns::Dnscmd::Record.any_instance.expects(:dns_find).with('host.domain').returns('192.168.33.34')
+    Proxy::Dns::Dnscmd::Record.any_instance.expects(:dns_find).with('host.foo.bar.domain.local').returns('192.168.33.34')
 
     assert_raise Proxy::Dns::Collision do
-      Proxy::Dns::Dnscmd::Record.new.create_a_record('host.domain', '192.168.33.33')
+      @server.create_a_record('host.foo.bar.domain.local', '192.168.33.33')
     end
   end
 
   def test_create_ptr_record
     Proxy::Dns::Dnscmd::Record.any_instance.expects(:dns_find).with('192.168.33.33').returns(false)
-    assert Proxy::Dns::Dnscmd::Record.new.create_ptr_record('host.domain', '192.168.33.33')
+    assert @server.create_ptr_record('host.foo.bar.domain.local', '192.168.33.33')
   end
 
   def test_overwrite_ptr_record
-    Proxy::Dns::Dnscmd::Record.any_instance.expects(:dns_find).with('192.168.33.33').returns('host.domain')
-    Proxy::Dns::Dnscmd::Record.new.create_ptr_record('host.domain', '192.168.33.33')
+    Proxy::Dns::Dnscmd::Record.any_instance.expects(:dns_find).with('192.168.33.33').returns('host.foo.bar.domain.local')
+    @server.create_ptr_record('host.foo.bar.domain.local', '192.168.33.33')
   end
 
   def test_create_duplicate_ptr_record_fails
-    Proxy::Dns::Dnscmd::Record.any_instance.expects(:dns_find).with('192.168.33.33').returns('another.host.domain')
+    Proxy::Dns::Dnscmd::Record.any_instance.expects(:dns_find).with('192.168.33.33').returns('another.host.foo.bar.domain.local')
     assert_raise Proxy::Dns::Collision do
-      Proxy::Dns::Dnscmd::Record.new.create_ptr_record('host.domain', '192.168.33.33')
+      @server.create_ptr_record('host.foo.bar.domain.local', '192.168.33.33')
     end
   end
 
-  def test_remove_address_record
-    Proxy::Dns::Dnscmd::Record.any_instance.expects(:dns_find).with('host.domain').returns(true)
-    Proxy::Dns::Dnscmd::Record.any_instance.expects(:execute).with('/RecordDelete domain host.domain. A /f', anything).returns(true)
-    assert Proxy::Dns::Dnscmd::Record.new.remove_a_record('host.domain')
+  def test_remove_address_record_with_longest_zone_match
+    Proxy::Dns::Dnscmd::Record.any_instance.expects(:dns_find).with('host.foo.bar.domain.local').returns(true)
+    Proxy::Dns::Dnscmd::Record.any_instance.expects(:execute).with('/RecordDelete bar.domain.local host.foo.bar.domain.local. A /f', anything).returns(true)
+    assert @server.remove_a_record('host.foo.bar.domain.local')
   end
 
   def test_remove_non_existent_address_record_raises_exception
-    Proxy::Dns::Dnscmd::Record.any_instance.expects(:dns_find).with('host.domain').returns(false)
+    Proxy::Dns::Dnscmd::Record.any_instance.expects(:dns_find).with('host.foo.bar.domain.local').returns(false)
     assert_raise Proxy::Dns::NotFound do
-      Proxy::Dns::Dnscmd::Record.new.remove_a_record('host.domain')
+      @server.remove_a_record('host.foo.bar.domain.local')
     end
   end
 
   def test_remove_ptr_record
     Proxy::Dns::Dnscmd::Record.any_instance.expects(:dns_find).with('192.168.33.33').returns(true)
-    assert Proxy::Dns::Dnscmd::Record.new.remove_ptr_record('192.168.33.33')
+    assert @server.remove_ptr_record('192.168.33.33')
   end
 
   def test_remove_nonexistent_ptr_record_raises_exception
     Proxy::Dns::Dnscmd::Record.any_instance.expects(:dns_find).with('192.168.33.33').returns(false)
     assert_raise Proxy::Dns::NotFound do
-      Proxy::Dns::Dnscmd::Record.new.remove_ptr_record('192.168.33.33')
+      @server.remove_ptr_record('192.168.33.33')
     end
+  end
+
+  def test_dns_zone_matches_second_best_match_if_zone_name_equals_host_name
+    assert_equal('33.168.192.in-addr.arpa', @server.match_zone('33.33.168.192.in-addr.arpa'))
+    assert_equal('domain.local', @server.match_zone('bar.domain.local'))
+  end
+
+  def test_dns_zone_matches_sole_available_zone
+    server = DnscmdForTesting.new(["sole.domain"])
+    assert_equal('sole.domain', server.match_zone('host.foo.bar.sole.domain'))
+  end
+
+  def test_dns_zone_no_match_raises_exception
+    assert_raise Proxy::Dns::NotFound do
+      @server.match_zone('host.foo.bar.domain.com')
+    end
+  end
+
+  def test_dnscmd_enum_zones_parses_primary_zones_only
+    to_parse = '
+Enumerated zone list:
+        Zone count = 8
+
+ Zone name                      Type       Storage         Properties
+
+ .                              Cache      AD-Domain
+ _msdcs.bar.domain.local        Primary    AD-Forest       Secure
+ 168.192.in-addr.arpa           Primary    AD-Domain       Rev
+ 33.168.192.in-addr.arpa        Primary    AD-Domain       Secure Rev Aging
+ e.e.d.8.b.d.0.1.0.0.2.ip6.arpa Primary    AD-Domain       Secure Rev
+ bar.domain.local               Primary    AD-Domain       Secure Aging
+ domain.local                   Primary    AD-Domain       Secure
+ domain.com                     Secondary    File
+ TrustAnchors                   Primary    AD-Forest
+
+
+Command completed successfully.'.split("\n")
+    Proxy::Dns::Dnscmd::Record.any_instance.expects(:execute).with('/EnumZones').returns(to_parse)
+    assert_equal [
+      "_msdcs.bar.domain.local",
+      "168.192.in-addr.arpa",
+      "33.168.192.in-addr.arpa",
+      "e.e.d.8.b.d.0.1.0.0.2.ip6.arpa",
+      "bar.domain.local",
+      "domain.local",
+      "TrustAnchors"], Proxy::Dns::Dnscmd::Record.new.enum_zones
   end
 end


domcleal:
Administrivia: I invited you to our GitHub org to get your PRs tested automatically, thanks for your contributions @helge000.  If you accept, please set membership to public on https://github.com/orgs/theforeman/people so Jenkins can see it.

This PR contains a couple of commits - you might need to rebase the latest one on top of develop (`git rebase --onto theforeman/develop HEAD^` or similar).

[test]

dmitri-d:
Good catch.

I think I'd prefer the second approach you suggested -- querying available primary zones via /enumzones. Main reason for this preference is that it would be annoying to be forced to update smart-proxy config and then restart after creating a new network/domain in foreman (I can also see people forgetting about this bit and looking for assistance).

The second issue is that it should be not one primary zone, but a list of those, as both foreman and ms dns support this. 

@helge000, @domcleal: thoughts?

helge000:
@witlessbird @domcleal What do you think of this? I added a new function `mach_zone(fqdn)` witch reads an array of managed zones from settings and selects the closest match. I run this patch already on my 1.10-stable smart proxy

helge000:
@domcleal Thanks for making me part of the team!

helge000:
@witlessbird 

> The second issue is that it should be not one primary zone, but a list of those, as both foreman and ms dns support this.
> I think the current patch fixes this issue?

Further, lazy me, argues that users have this way more control over the AD DNS zones smart proxy is 'messing' with.  
Since a ad top zone also covers all zones below it a new zone is a fairly rare thing in real life.

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -85,5 +85,31 @@ def report msg, response, error_only
...
       raise Proxy::Dns::Error.new("Unknown error while processing '#{msg}'")
     end
+
+    def match_zone(fqdn)
+      dns_zones = ::Proxy::Dns::Plugin.settings.dns_zones
+      found_match = false
+      wight = 0
+      matched_zone = fqdn.sub(/[^.]+./,'')
+
+      if dns_zones.is_a?(Array)
```
witlessbird: You can do this check without a regex: 
```
zone_labels = zone.split(".").reverse
fqdn_labels = fqdn.split(".")
found_match = zone_labels.all? { |zone_label| zone_label == fqdn_labels.pop } 
```

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -85,5 +85,31 @@ def report msg, response, error_only
...
+        dns_zones.each do |zone|
+          zone.sub!(/^\./, '')
```
dmitri-d: You can do this check without a regex:

```
zone_labels = zone.split(".").reverse
fqdn_labels = fqdn.split(".")
found_match = zone_labels.all? { |zone_label| zone_label == fqdn_labels.pop } 
```

dmitri-d:
> Since a ad top zone also covers all zones below it a new zone is a fairly rare thing in real life.

In an AD deployment, sure. A standalone dns server can have multiple top-level independent zones. As foreman supports this, I feel we need to support this on the smart-proxy side.

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -85,5 +85,31 @@ def report msg, response, error_only
       logger.error "Dnscmd failed:\n" + (response.is_a?(Array) ? response.join("\n") : "Response was not an array! #{response}")
       raise Proxy::Dns::Error.new("Unknown error while processing '#{msg}'")
     end
+
+    def match_zone(fqdn)
+      dns_zones = ::Proxy::Dns::Plugin.settings.dns_zones
```
dmitri-d: I'd prefer if this was initialized in the constructor; and an accessor was used instead of referring to `::Proxy::Dns::Plugin.settings` directly. 

This property is a required one, you should validate that it's present in dns_dnscmd_plugin.rb: `validate_presence :dns_zones`. I'll update the validator to handle arrays (right now validation will pass even if the array is empty).

modules/dns_dnscmd/dns_dnscmd_main.rb
helge000: That's pretty cool. Adding it ASAP

dmitri-d:
I still think that fetching the zone list from the server on launch is a much better option...

modules/dns_dnscmd/dns_dnscmd_main.rb
helge000: > This property is a required one

Actually not. The function falls back to the old method and always returns a zone

dmitri-d: Let me rephrase this.  I think this should be a required property -- otherwise it's possible to skip this config setting and end up with the issue that you found and fixed.

helge000:
@witlessbird 

> I still think that fetching the zone list from the server on launch is a much better option...

True; though we either need to run every time we create/delete a record or have to restart smart proxy. In my environment the first method might come cheap; but might as well take time.

dmitri-d:
I was thinking about initializing the zone list once, and then keeping it for all instances. A restart would be required when zones are added/removed.

modules/dns_dnscmd/dns_dnscmd_main.rb
dmitri-d: Either need to complain on startup that dns_zones parameter isn't an array (the idea is to report config errors on startup, as it makes troubleshooting easier), or handle this parameter being a string.

helge000:
@witlessbird Another option could be a combination: Running with default settings or not finding a match invokes `execute(enum_zones)`. This way it would neither require a restart and would be fast in the configured state. 
As a nice side effect the changes would be more contained to the provider (though a test case needs to be added, but I think I know now how to deal with this... :)

dmitri-d:
By default you mean an empty list of zones? Dunno, I think this would complicate things, including troubleshooting: misspelled domain name in settings results in an additional lookup on every call, resulting in an overall slowdown.

If you think restarts are an issue, I'd suggest asynchronously refreshing the list every n minutes (15?).

helge000:
@witlessbird I added your suggestion and scrubbed the regex. Also, fixed an edge case if a host is named like a sub zone.

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -85,5 +85,32 @@ def report msg, response, error_only
...
+      found_match = false
+      weight = 0
+      matched_zone = fqdn.sub(/[^.]+./,'')
+
+      if dns_zones.is_a?(Array)
+        dns_zones.each do |zone|
+          zone_labels = zone.split(".").reverse
+          zone_weight = zone_labels.length
+          fqdn_labels = fqdn.split(".")
+          is_match = zone_labels.all? { |zone_label| zone_label == fqdn_labels.pop } || false
```
dmitri-d: no need for `|| false` -- #all? returns true if condition holds for elements of the array or false.

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -85,5 +85,32 @@ def report msg, response, error_only
...
+          if is_match && zone_weight >= weight && fqdn != zone # match longest zone suffix unless zone == fqdn
```
dmitri-d: Will ms dns allow to create a record if subdomain doesn't exist already (if the zone is a.b. and we mistakenly are trying to add host a.b. -- it would require a pre-existing subdomain b.a.b, right?)?

modules/dns_dnscmd/dns_dnscmd_main.rb
helge000: Right .. hehe, already removed locally.

modules/dns_dnscmd/dns_dnscmd_main.rb
helge000: SInce record != zone, the record will become the parent zone's record (`(same as parent)`). I do think we never want this?

dmitri-d: No we most definitely do not, was checking on how ms dns would handle such a situation.

I don't understand the point of zone_weight, however. It's always going to greater than 0, and `zone_weight >= weight` check will always evaluate to true.

helge000: Hm; I use zone_weight to determine the _longest match_, since we most certianly have a zone config like this:

```
["ad.company.com", "company.com", "test.ad.company.com", "0.168.192.in-addr.arpa"]
```

I think you have something better in mind?

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -85,5 +85,32 @@ def report msg, response, error_only
...
+      dns_zones = ::Proxy::Dns::Plugin.settings.dns_zones || nil
+      found_match = false
+      weight = 0
+      matched_zone = fqdn.sub(/[^.]+./,'')
+
+      if dns_zones.is_a?(Array)
+        dns_zones.each do |zone|
+          zone_labels = zone.split(".").reverse
+          zone_weight = zone_labels.length
+          fqdn_labels = fqdn.split(".")
```
dmitri-d: Just noticed the issue -- this shouldn't use fqdn, just the domain part of the fqdn: 

```
fqdn_labels = fqdn.split(".")
fqdn_labels.shift
```

is_match will then only be true for the longest available domain. Also, the `fqdn != zone` check on line 101 won't be needed either.

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -85,5 +85,30 @@ def report msg, response, error_only
...
+          fqdn_labels.shift
+          is_match = zone_labels.all? { |zone_label| zone_label == fqdn_labels.pop }
+          # match only the longest zone suffix
+          (matched_zone, weight, found_match = zone, zone_weight, true) if is_match && zone_weight >= weight
```
helge000: @witlessbird Great! I hope this is what you had in mind?

dmitri-d: not quite:

```
    def match_zone(fqdn)
      dns_zones = ::Proxy::Dns::Plugin.settings.dns_zones || []

      zone_list = dns_zones.is_a?(String) ? [dns_zones] : dns_zones
      zone_list.each do |zone|
        zone_labels = zone.split(".").reverse
        fqdn_labels = fqdn.split(".")
        fqdn_labels.shift
        is_matched = zone_labels.all? { |zone_label| zone_label == fqdn_labels.pop }
        return zone if is_matched
      end

      matched_zone = fqdn.sub(/[^.]+./,'')
      logger.warn "Unable to determine dns zone for dnscmd.exe, falling back to #{matched_zone}"
      matched_zone
    end
```

helge000: The problem with the above: it returns the _first_ match; witch might not good.
But I suspect reverse sorting the hole **zone_list** might be just as good

dmitri-d: Ahh, right:

```
     def match_zone(fqdn)
      dns_zones = ::Proxy::Dns::Plugin.settings.dns_zones || []

      zone_list = dns_zones.is_a?(String) ? [dns_zones] : dns_zones
      host_domain = fqdn.sub(/[^.]+./,'')

      found = zone_list.find {|zone| zone == host_domain}
      return found if found

      logger.warn "Unable to determine dns zone for dnscmd.exe, falling back to #{host_domain}"
      host_domain
    end
```

helge000: @witlessbird This does not work either. 
I just run a series of tests with real life data from our zones; my version of `match_zone()` might be less elegant, but reliably returns forward and reverse zone machtes.

Yours works with reverse zones only. It never finds a forward match; I fail to see why though.

dmitri-d: @helge000: Please add tests for this (for w/e version you pick). Then I might be able to see why.

helge000: @witlessbird Certainly. Added `test_dns_zone_matching`. I've opted for a separate test rather than extending the existing tests; tell me what you like this better.

Right now I have trouble with settings. They get ignored. So far I tried changing [this](https://github.com/theforeman/smart-proxy/blob/develop/test/dns_dnscmd/dnscmd_test.rb#L7):

``` ruby
Proxy::Dns::Dnscmd::Plugin.load_test_settings(
        :dns_server => 'a_server',
        :dns_zones => ['ad.domain.com', 'domain.com', '33.168.192.in-addr.arpa'])
```

helge000:
@witlessbird @domcleal - dnscmd _PTR_'s are currently noop. Since this patch solves the underlying issue it I could handle it here, too.. Do you prefer a new issue or should I extent this patch?

dmitri-d:
@helge000: I don't see the test among the changes. Did you forget to commit the  new file? I think it'd be be better to have PTR-related changes in a separate PR.

helge000:
@witlessbird I pushed it now, but mind you: I use a dirty trick to get the settings working, see question above

test/dns_dnscmd/dnscmd_test.rb
```diff
@@ -13,51 +15,51 @@ def test_dnscmd_provider_initialization
   end
 
   def test_create_address_record
-    Proxy::Dns::Dnscmd::Record.any_instance.expects(:dns_find).with('host.domain').returns(false)
-    Proxy::Dns::Dnscmd::Record.any_instance.expects(:execute).with('/RecordAdd domain host.domain. A 192.168.33.33', anything).returns(true)
-    assert Proxy::Dns::Dnscmd::Record.new.create_a_record('host.domain', '192.168.33.33')
+    Proxy::Dns::Dnscmd::Record.any_instance.expects(:dns_find).with('host.foo.bar.domain.local').returns(false)
+    Proxy::Dns::Dnscmd::Record.any_instance.expects(:execute).with('/RecordAdd bar.domain.local host.foo.bar.domain.local. A 192.168.33.33', anything).returns(true)
```
dmitri-d, line 35: I don't understand this bit: we are adding a host 'host.foo.bar.domain.local' to 'bar.domain.local'. 'host.foo' isn't a legal host name, how would ms dns handle this -- is it going to automatically create a subdomain 'foo.bar.domain.local'? I don't think this is a desired behaviour...

helge000, line 35: Indeed, dnscmd just adds the zone if it does not exist; hence `host.foo` is a perfectly valid hostname. Since we are are explicitly defining a FQDN with  `host.foo.bar.domain.local.` our A record would be created with the expected result (*). 
Since users are going to create their sub domains and zones in foreman, I think we can safely leverage the resulting domain to users: We should get  `host.foo` if we have actually created a zone named `foo` in foreman, resulting in the desired behavior?

(*) Using a relative domain name with dnscmd would create the host in the defined domain; eg 
`/RecordAdd bar.domain.local host.foo.bar.domain.local A 192.168.33.33` would result in the FQDN: `host.foo.bar.domain.local.bar.domain.local`

test/dns_dnscmd/dnscmd_test.rb
```diff
@@ -4,7 +4,9 @@
 
 class DnsCmdTest < Test::Unit::TestCase
   def test_dnscmd_provider_initialization
-    Proxy::Dns::Dnscmd::Plugin.load_test_settings(:dns_server => 'a_server')
+    Proxy::Dns::Dnscmd::Plugin.load_test_settings(
+        :dns_server => 'a_server',
+        :dns_zones => ['bar.domain.local', 'domain.local', '33.168.192.in-addr.arpa', '168.192.in-addr.arpa'])
```
helge000: How can I get this working?

dmitri-d: Not sure what you mean, this passes:

```
  def test_dnscmd_provider_initialization
    Proxy::Dns::Dnscmd::Plugin.load_test_settings(
        :dns_server => 'a_server',
        :dns_zones => ['bar.domain.local', 'domain.local', '33.168.192.in-addr.arpa', '168.192.in-addr.arpa'])
    Proxy::Dns::Plugin.load_test_settings(:dns_ttl => 999)
    server = Proxy::Dns::Dnscmd::Record.new

    assert_equal ['bar.domain.local', 'domain.local', '33.168.192.in-addr.arpa', '168.192.in-addr.arpa'], Proxy::Dns::Dnscmd::Plugin.settings.dns_zones
    assert_equal "a_server", server.server
    assert_equal 999, server.ttl
  end
```

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -85,5 +85,34 @@ def report msg, response, error_only
       logger.error "Dnscmd failed:\n" + (response.is_a?(Array) ? response.join("\n") : "Response was not an array! #{response}")
       raise Proxy::Dns::Error.new("Unknown error while processing '#{msg}'")
     end
+
+    def match_zone(fqdn)
+      dns_zones = ::Proxy::Dns::Plugin.settings.dns_zones || ['bar.domain.local', 'domain.local', '33.168.192.in-addr.arpa', '168.192.in-addr.arpa']
```
helge000: @witlessbird This is actually why the test unit is passing and what I meant with dirty trick

dmitri-d: You'll need to use `setup` method which is called before every test to define plugin settings in tests:

```
def setup
  Proxy::Dns::Plugin.load_test_settings(
        :dns_server => 'a_server',
        :dns_zones => ['bar.domain.local', 'domain.local', '33.168.192.in-addr.arpa', '168.192.in-addr.arpa'])
end
```

test/dns_dnscmd/dnscmd_test.rb
```diff
@@ -3,61 +3,70 @@
 require 'dns_dnscmd/dns_dnscmd_main'
 
 class DnsCmdTest < Test::Unit::TestCase
+  def setup
+    Proxy::Dns::Plugin.load_test_settings(
+        :dns_server => 'a_server',
+        :dns_zones => ['bar.domain.local', 'domain.local', '33.168.192.in-addr.arpa', '168.192.in-addr.arpa'])
+  end
+
   def test_dnscmd_provider_initialization
```
helge000, line 24: @witlessbird Thanks! setup works fine. What about `test_dnscmd_provider_initialization` test, is it redundant?

dmitri-d, line 24: The test is still usefull -- it verifies that dnscmd respects parameters defined in the configuration file. However, plugin initialization code is redundant, could you remove it?

I see that the dns_zones setting is defined in the dnscmd config file, but is retrieved from the main plugin settings in the code. Test setup method is wrong too.

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -85,5 +85,34 @@ def report msg, response, error_only
       logger.error "Dnscmd failed:\n" + (response.is_a?(Array) ? response.join("\n") : "Response was not an array! #{response}")
       raise Proxy::Dns::Error.new("Unknown error while processing '#{msg}'")
     end
+
+    def match_zone(fqdn)
+      dns_zones = ::Proxy::Dns::Plugin.settings.dns_zones || nil
```
dmitri-d: Can you move this into the constructor? No need to set it to nil, nil is returned for missing settings.

dmitri-d:
One of the issues that has yet to be resolved is retrieval of the domain list from the server. Will you be able to tackle this?

helge000:
You think about scrubbing the settings method altogether and running / parsing `dnscmd /EnumZones`? While I think I can do that I certainly need help *where* to put and where to start. since we only wanted to run it on server startup once. Also, I would really like to be able to use the #execute method of the provider.

test/dns_dnscmd/dnscmd_test.rb
helge000, line 24: > However, plugin initialization code is redundant, could you remove it?

Already tried this, though the unit fails if I do so. I'll need to loo at it.

Can you help me out what method I should use best?

helge000:
This simple function I could come up with for `/EnumZones`, it returns an array with Primary zones:
```ruby
    def enum_dns_zones
      zones = []
      cmd = "/EnumZones"
      output = execute(cmd)
      output.each do |line|
        if line =~ /Primary/
          zones << line.sub(/^ +/, '').sub(/ +.*$/, '')
        end
      end
      zones
    end
```

Currently, I cannot figure out the test unit and where to best put it in the provider. A sample return from dnscmd:
```
Enumerated zone list:
        Zone count = 8

 Zone name                      Type       Storage         Properties

 .                              Cache      AD-Domain
 _msdcs.bar.domain.local        Primary    AD-Forest       Secure
 168.192.in-addr.arpa           Primary    AD-Domain       Rev
 33.168.192.in-addr.arpa        Primary    AD-Domain       Secure Rev Aging
 e.e.d.8.b.d.0.1.0.0.2.ip6.arpa Primary    AD-Domain       Secure Rev
 bar.domain.local               Primary    AD-Domain       Secure Aging
 domain.local                   Primary    AD-Domain       Secure
 domain.com                     Primary    File
 TrustAnchors                   Primary    AD-Forest


Command completed successfully.
```

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -85,5 +86,45 @@ def report msg, response, error_only
       logger.error "Dnscmd failed:\n" + (response.is_a?(Array) ? response.join("\n") : "Response was not an array! #{response}")
       raise Proxy::Dns::Error.new("Unknown error while processing '#{msg}'")
     end
+
+    def match_zone(fqdn)
+      dns_zones = ::Proxy::Dns::Plugin.settings.dns_zones || enum_dns_zones
```
helge000: This allows the test units to succeed. However, I think it is a good idea: Either use settings or enumerate dns zones

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -85,5 +86,45 @@ def report msg, response, error_only
...
+      end
+      unless found_match
+        logger.warn "Unable to determine dns zone for dnscmd.exe, falling back to #{matched_zone}"
+      else
+        logger.debug "Matched dns zone for dnscmd.exe, #{matched_zone}"
+      end
+      matched_zone
+    end
+
+    def enum_dns_zones
```
helge000: I patched my production env with this function; does its job. Maybe we _should_ run it on every match zone call as it would be the most transparent for the user as the execution time is negligible. 
I just timed it with 30 Zones on `localhost`. We can also almost half the execution time by not calling `cmd.exe`:

``` powershell
PS C:\Users\Administrator> Measure-Command {c:\windows\system32\cmd.exe /c c:\Windows\System32\dnscmd.exe localhost /EnumZones}

TotalMilliseconds : 42,897


PS C:\Users\Administrator> Measure-Command {c:\Windows\System32\dnscmd.exe localhost /EnumZones}

TotalMilliseconds : 27,3026
```

test/dns_dnscmd/dnscmd_test.rb
```diff
@@ -3,61 +3,90 @@
...
+ 33.168.192.in-addr.arpa        Primary    AD-Domain       Secure Rev Aging
+ e.e.d.8.b.d.0.1.0.0.2.ip6.arpa Primary    AD-Domain       Secure Rev
+ bar.domain.local               Primary    AD-Domain       Secure Aging
+ domain.local                   Primary    AD-Domain       Secure
+ domain.com                     Secondary    File
+ TrustAnchors                   Primary    AD-Forest
+
+
+Command completed successfully.'.split("\n")
+    Proxy::Dns::Dnscmd::Record.stubs(:execute).with('/EnumZones').returns(to_parse)
```
helge000: I obviously need help setting up this test unit...

modules/dns_dnscmd/dns_dnscmd_main.rb
dmitri-d: I don't think this is a good idea -- why have such an override at all? Esp. considering potential issues -- the list getting out of sync, wrong entries in the list, etc. 

I'm ok with retrieving of dns zones on every call for now however.

modules/dns_dnscmd/dns_dnscmd_main.rb
dmitri-d: Can you extract the logic for parsing of the command output into a separate method and add tests for it?

test/dns_dnscmd/dnscmd_test.rb
dmitri-d: You actually don't need to stub Proxy::Dns::Dnscmd::Record#execute. Instead:

```
class DnscmdForTesting < Proxy::Dns::Dnscmd::Record
  def initialize(dns_zones)
    @enum_dns_zones = dns_zones
  end
  attr_accessor : enum_dns_zones
end
```

then for tests that are validating logic in methods _other_ than Proxy::Dns::Dnscmd::Record.#enum_dns_zones:

```
server = DnscmdForTesting.new(['domain.com', 'domain.local' ...])
```

Alternatively, you can move the line above into `setup` method.

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -67,6 +67,7 @@ def execute cmd, msg=nil, error_only=false
         std_err.close unless std_in.nil?
       end
       report msg, response, error_only
+      response
```
helge000, line 70: @witlessbird: Please mind this change I need to make in order to be able to use the `execute()` I can't see any side effects, but maybe you see a better way.

dmitri-d, line 70: looks ok to me.

helge000:
@witlessbird: Implement your suggestion for output parsing and added the test case. Also, since agreed on running `execute` each time I removed all settings - related things; further returning `nil` and _error_ if the zone could not be matched (in theory, the DNS server lacks a primary zone for the fqdn)

Still todo: 
- The test case setup. Hope you are patient with me with me: I really do not understand how you want me to init the zones; currently I am using the old method. 
- With my limited unterstanding I would rather stub `execute('/EnumZones')` like in the DHCP test unit setup, though I don't know if/how this is posible

helge000:
Any clue why the build is failing? My precommit hook checked out...

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -85,5 +86,40 @@ def report msg, response, error_only
       logger.error "Dnscmd failed:\n" + (response.is_a?(Array) ? response.join("\n") : "Response was not an array! #{response}")
       raise Proxy::Dns::Error.new("Unknown error while processing '#{msg}'")
     end
+
+    def match_zone(fqdn)
+      dns_zones = ::Proxy::Dns::Plugin.settings.dns_zones || parse_dns_zones(execute('/EnumZones'))
```
dmitri-d: This still uses zones from the config file...

helge000: Right: Only temporary until the test unit works

dmitri-d:
> - The test case setup. Hope you are patient with me with me: I really do not understand how you want me to init the zones; currently I am using the old method.
> - With my limited unterstanding I would rather stub execute('/EnumZones') like in the DHCP test unit setup, though I don't know if/how this is posible

These two are related. 

In general you should mock/stub calls as close as possible (call-chain wise) to the method/class you are testing: as you move farther away from the method you are testing, the chances that the code will change independently of the current system under test (Proxy::Dns::Dnscmd::Record class, in our case) increase. This results in fragile tests (on top of the fact that mocking/stubbing is fragile to begin with) and more maintenance. 

(If you are curious about how to use mocks/stubs better I can suggest an excellent (but long) article on mocking vs. traditional state-based testing: http://martinfowler.com/articles/mocksArentStubs.html)

In this case the code we are testing is under our control, and it's easier (maintenance-wise) to create a test-specific subclass of `Proxy::Dns::Dnscmd::Record` that lets you define desired output of `Proxy::Dns::Dnscmd::Record#parse_dns_zones` method. `test_zone_parse` should use the original implementation, however. 

Hope this makes sense.

dmitri-d:
re: test failures -- looks like connectivity issues. I'll schedule another run.

[test]

helge000:
> Hope this makes sense.

Thanks, and I hope I slowly get it. The test unit checks out now (*) using the class you suggested. All traces of config files should be gone now. Note, I had to revert to old `enum_zones`, but the test should cover it.

(*) at least it should, even if Jenkins currently thinks diffrent

dmitri-d:
[test]

test/dns_dnscmd/dnscmd_test.rb
```diff
@@ -13,63 +31,98 @@ def test_dnscmd_provider_initialization
...
+ 33.168.192.in-addr.arpa        Primary    AD-Domain       Secure Rev Aging
+ e.e.d.8.b.d.0.1.0.0.2.ip6.arpa Primary    AD-Domain       Secure Rev
+ bar.domain.local               Primary    AD-Domain       Secure Aging
+ domain.local                   Primary    AD-Domain       Secure
+ domain.com                     Secondary    File
+ TrustAnchors                   Primary    AD-Forest
+
+
+Command completed successfully.'.split("\n")
+    Proxy::Dns::Dnscmd::Record.any_instance.expects(:execute).with('/EnumZones').returns(to_parse)
```
dmitri-d, line 129: Good call re: expects (vs. stubs)!

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -85,5 +86,41 @@ def report msg, response, error_only
...
+        fqdn_labels.shift
+        is_match = zone_labels.all? { |zone_label| zone_label == fqdn_labels.pop }
+        # match only the longest zone suffix
+        if is_match && zone_weight >= weight
+          matched_zone = zone
+          weight = zone_weight
+        end
+      end
+      unless matched_zone
+        logger.error "The DNS server has no authoritative zone for #{fqdn}!"
```
dmitri-d: This should raise a Proxy::Dns::NotFound exception instead. No need for logging it either, the main plugin will do that instead.

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -85,5 +86,41 @@ def report msg, response, error_only
...
+      else
+        logger.debug "Matched dns zone to #{matched_zone}"
```
dmitri-d: Pls. remove this debug: this isn't needed in production.

test/dns_dnscmd/dnscmd_test.rb
```diff
@@ -13,63 +31,98 @@ def test_dnscmd_provider_initialization
...
     Proxy::Dns::Dnscmd::Record.any_instance.expects(:dns_find).with('192.168.33.33').returns(false)
     assert_raise Proxy::Dns::NotFound do
-      Proxy::Dns::Dnscmd::Record.new.remove_ptr_record('192.168.33.33')
+      @server.remove_ptr_record('192.168.33.33')
     end
   end
+
+  def test_dns_zone_matching_edge_cases
+    assert_equal('33.168.192.in-addr.arpa', @server.match_zone('33.33.168.192.in-addr.arpa'))
+    assert_equal('domain.local', @server.match_zone('bar.domain.local'))
```
dmitri-d, line 96: Shouldn't this match 'bar.domain.local' zone?

test/dns_dnscmd/dnscmd_test.rb
```diff
@@ -13,63 +31,98 @@ def test_dnscmd_provider_initialization
...
 
   def test_remove_nonexistent_ptr_record_raises_exception
     Proxy::Dns::Dnscmd::Record.any_instance.expects(:dns_find).with('192.168.33.33').returns(false)
     assert_raise Proxy::Dns::NotFound do
-      Proxy::Dns::Dnscmd::Record.new.remove_ptr_record('192.168.33.33')
+      @server.remove_ptr_record('192.168.33.33')
     end
   end
+
+  def test_dns_zone_matching_edge_cases
```
dmitri-d: can you break up this test and add a couple more tests that verify:
- it can match the only available zone (fqdn is 'a.b.c' and we have a zone 'a.b'. Probably should verify with fqdn 'a.b.c.d' and available zone 'c.d')
- it matches the longest available zone
- returns nil if no zone has been matched

test/dns_dnscmd/dnscmd_test.rb
helge000, line 96: `bar.domain.local` is the host name in this case. As `bar.domain.local` is already a zone, I am guessing the user's intention  is to rather add it to the next best match (we talked about it, dnscmd adds a `(same as parent)` A record in this case).

The alternative would be to raise `Proxy::Dns::Collision`. But it is still an edge case when we have  both zones available; I'd rather not fail. What do you think?

dmitri-d, line 96: Let me think about it...

test/dns_dnscmd/dnscmd_test.rb
helge000: Certainly. The longest available zone test gets done in the above A record tests. Should I do this again?

dmitri-d: Can you rename the test so it's clear that it verifies that the longest available zone has been matched?

helge000:
@witlessbird Removed logging in favor for raising an exception and added the following tests:
- mach sole available zone
- no match raises exception (thus not checking for nil)
- renamed tests witch already matching the longest available zone

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -85,5 +86,37 @@ def report msg, response, error_only
...
+    end
+
+    def enum_zones
+      zones = []
+      response = execute('/EnumZones')
+      response.each do |line|
+        next unless line =~  / Primary /
+        zones << line.sub(/^ +/, '').sub(/ +.*$/, '').chomp("\n")
+      end
+      logger.debug "Enumerated dns zones: #{zones}"
```
helge000, line 118: What about this logging? I found it quite useful.

helge000:
May I push this a little, @witlessbird? I need this PR merged to get working on [#13536](http://projects.theforeman.org/issues/13536)

dmitri-d:
I need to test how bind and ms dns (two main dns providers supported by smart-proxy) react to an attempt to add an 'A' record like 'a.b' when 'b' isn't a pre-existing zone. My guess is that bind isn't going to like this. I need to figure out then how important is to have a uniform behaviour for record creation across different providers.

You can create a branch off the branch you used to create this PR and start/continue with new changes there...

helge000:
Good to know! 
IMHO the above applies to nsupdate as well; though I think something like `mach_zone` is [already build in:](http://linux.die.net/man/8/nsupdate)

> **zone zonename**
> Specifies that all updates are to be made to the zone zonename. If no zone statement is provided, nsupdate will attempt determine the correct zone to update based on the rest of the input.

dmitri-d:
Merged in at 310433f. Thanks, @helge000!

PR #370, opened 2016-01-28
Fixes #13427 - fixed log buffer with SYSLOG

lzap:
Proxy was not even starting when configured with SYSLOG. The handling of
progname was not necessary, we pass nil all the time. We just need to make sure
the message itself is correctly retrieved according to the API docs:

http://ruby-doc.org/stdlib-2.1.0/libdoc/logger/rdoc/Logger.html#method-i-add

### DIFF
--- a/lib/proxy/log_buffer/decorator.rb
+++ b/lib/proxy/log_buffer/decorator.rb
@@ -8,19 +8,17 @@ module Proxy::LogBuffer
       @buffer = buffer
     end
 
-    def add(severity, message = nil, progname = nil, backtrace = nil, a_module = 'core', &block)
+    def add(severity, message = nil, progname = nil, backtrace = nil)
       severity ||= UNKNOWN
-      progname ||= @logger.progname
       if message.nil?
         if block_given?
           message = yield
         else
           message = progname
-          progname = @logger.progname
         end
       end
       # add to the logger first
-      @logger.add(severity, message, progname)
+      @logger.add(severity, message)
       @logger.add(::Logger::Severity::DEBUG, backtrace) if backtrace
       # add add to the buffer
       if severity >= @logger.level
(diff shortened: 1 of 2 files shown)


lib/proxy/log_buffer/decorator.rb
```diff
@@ -10,17 +10,15 @@ def initialize(logger, buffer = Proxy::LogBuffer::Buffer.instance)
 
     def add(severity, message = nil, progname = nil, backtrace = nil, a_module = 'core', &block)
```
dmitri-d: You don't need to pass block as a parameter -- you are using yield.

lzap: But that's the way it's defined in Ruby Core, I want to keep the exact same interface in the decorator.

lzap: Also I use the last parameter in all the helper methods here.

dLobatog: I think @witlessbird made a good point here - it even incurs a serious performance penalty. If you're not calling block explicitely, just remove it IMO - http://mudge.name/2011/01/26/passing-blocks-in-ruby-without-block.html

Where is this defined in Ruby core? If `block` is defined but doesn't need to be passed because `yield` is used, submit a patch to fix the performance issue :smile:

dLobatog:
:+1: to this change, it fixes the problem for me - however a test to ensure we don't get this bug again would be great @lzap . Thanks!

lib/proxy/log_buffer/decorator.rb
dmitri-d: What @dLobatog said. Unless you explicitly call `block.call`, no need to include block as a parameter in method's argument list.

dLobatog: Ah I saw this logger is based upon Ruby's logger, I didn't know that - https://github.com/ruby/ruby/pull/1240

lzap: Heh, you should leave a comment so I would not bother reading about how to contribute to Ruby :-)

https://bugs.ruby-lang.org/issues/12054

It's been fixed in trunk already. Interesting article, the speedup was not so obvious. I will remove this than.

lzap:
Done, added few tests which actually helped to find another extra parameter that was just a leftover. Removed as well.

dmitri-d:
@lzap: pls. take a look at test failures.

lzap:
Thanks for the heads up, rebased.

lzap:
Geez, Rubocop! Rebased.

test/log_buffer/decorator_test.rb
```diff
@@ -15,6 +16,29 @@ def setup
     @decorator = ::Proxy::LogBuffer::Decorator.new(@logger, @buffer)
   end
 
+  def test_should_pass_logs
+    @logger.expects(:add).with(DEBUG, "message")
+    @decorator.debug("message")
+  end
+
+  def test_should_pass_logs_to_syslog
```
dmitri-d, line 24: I don't see why this test is needed. You verified the interface in test_should_pass_logs (probably worth renaming to test_should_pass_logs_to_logger).

lzap, line 24: I test the `add` method in case we will change it in future sending some methods to loggers which are not available in syslog implementation. This implementation is a bit unusual and missing some features (for obvious reasons). Therefore this test helps to avoid these mistakes (the patch actually fixes one of these - I was assuming that progname is a method that is implemented for all logger implementations).

dmitri-d:
merged in at a838204. Thanks, @lzap.

PR #371, opened 2016-02-03
Fixes #13544: Support IPv6 in DNS

ekohl:


ekohl:
@domcleal @witlessbird this is very much a work in progress, but I'm mostly wondering about the API part. If you think that's the way to go, I'll work on adding the support to providers.

modules/dns/dns_api.rb
```diff
@@ -15,11 +16,22 @@ class Api < ::Sinatra::Base
...
       log_halt(400, "'create' requires fqdn, value, and type parameters") if fqdn.nil? || value.nil? || type.nil?
-      log_halt(400, "unrecognized 'type' parameter: #{type}") unless type == 'A' || type == 'PTR'
+      validate_dns_name(fqdn)
 
       begin
-        server.create_a_record(fqdn, value) if type == 'A'
-        server.create_ptr_record(fqdn, value) if type == 'PTR'
+        case type
+        when 'A'
+          ip = IPAddr.new(value, Socket::AF_INET).to_s
```
ekohl, line 25: I didn't want to break the API, but what do you think about removing the to_s?

dmitri-d, line 25: I'm ok with changing the internal API. I'm not aware of any non-core dns providers other than power-dns.

modules/dns_common/dns_common.rb
```diff
@@ -15,12 +17,20 @@ def resolver
       Resolv::DNS.new(:nameserver => @server)
     end
 
-    def dns_find key
-      if match = key.match(/(\d{1,3})\.(\d{1,3})\.(\d{1,3})\.(\d{1,3})/)
-        resolver.getname(match[1..4].reverse.join(".")).to_s
-      else
-        resolver.getaddress(key).to_s
+    def dns_find key, type='auto'
+      if type == 'auto'
```
dmitri-d: can you replace this with something like:

```
def dns_find key, type='auto'
  key, type = case key
  when Resolv::IPv4::Regex
    [Resolv::IPv4.create(key).to_name.to_s, 'PTR']
  when Resolv::IPv6::Regex
    [Resolv::IPv6.create(key).to_name.to_s, 'PTR']
  else
    [key, 'A']
  end if key == 'auto'

  resolver.getresource(key, type).to_s
end  
```

test/dns/dns_api_test.rb
```diff
@@ -32,53 +39,173 @@ def app
...
     assert_equal 400, last_response.status
   end
 
   def test_create_returns_error_if_type_is_unrecognized
-    post '/', :fqdn => 'test.com', :value => '192.168.33.33', :type => "FFF"
+    post '/', :fqdn => get_fqdn, :value => get_ip, :type => "FFF"
+    assert_equal 400, last_response.status
+  end
+
+  def test_create_returns_error_if_invalid_name_for_a
```
dmitri-d, line 68: Could you create unit-type tests for the name validator, and keep one integration-level test for a valid fqdn and one for an invalid one instead?

modules/dns_common/dns_common.rb
dmitri-d: Also, looking at dns_find-related tests in dns_common/record_test.rb, it's worth breaking logic used for the type == 'auto' case into a separate class: corresponding tests won't require mocking then.

modules/dns/dns_api.rb
ekohl, line 25: And Dominics route53 provider.

modules/dns_common/dns_common.rb
ekohl: Do you mean write two test classes? One for auto and one for explicit?

dmitri-d:
https://github.com/ekohl/smart-proxy/blob/aaaa-records/modules/dns_nsupdate/dns_nsupdate_main.rb#L30 isn't going to work for shortened ipv6 addresses (which is why tests are failing): need to either create IPAddr in here and compare that, or switch internal api to use IPAddr as you suggested.

modules/dns_common/dns_common.rb
dmitri-d: I'm suggesting breaking up https://github.com/ekohl/smart-proxy/blob/aaaa-records/modules/dns_common/dns_common.rb#L20 into two methods, something along these lines:

```
def calculate_key_type(key, type)
  key, type = case key
  when Resolv::IPv4::Regex
    [Resolv::IPv4.create(key).to_name.to_s, 'PTR']
  when Resolv::IPv6::Regex
    [Resolv::IPv6.create(key).to_name.to_s, 'PTR']
  else
    [key, 'A']
  end if key == 'auto'

  [key, type]
end

def dns_find(key, type='auto')
  resolver.getresource(*calculate_key_type(key, type)).to_s
end
```

Then you can test the "interesting" logic in calculate_key_type without any mocking. Probably worth adding one integration-type test to verify that dns_find calls calculate_key_type and resolver.getresource though.

dmitri-d:
Looks pretty good, although I haven't tried running this against bind...

modules/dns/dns_api.rb
dmitri-d, line 25: Ahh, right, I can do a PR for that one.

modules/dns_common/dns_common.rb
ekohl: That sounds like a good idea.

ekohl:
I think the shortened part should work because the dns_api shortens IPv6. So any provider will always get the shortened IP. Other than that I still have to run integration testing. So far it's just code that should work.

modules/dns_common/dns_common.rb
ekohl: I chose a slightly different API, but the general idea is the same.

ekohl:
Haven't addressed all comments yet.

ekohl:
Updated. It includes https://github.com/theforeman/smart-proxy/pull/384.

ekohl:
@helge000 I also added code for dnscmd but I have no way to test it. Mind having a look as well?

modules/dns/dns_api.rb
```diff
@@ -27,17 +39,41 @@ class Api < ::Sinatra::Base
...
+        end
       rescue Proxy::Dns::NotFound => e
         log_halt 404, e
       rescue => e
         log_halt 400, e
       end
     end
+
+    def validate_dns_name name
+      log_halt(400, "Invalid DNS name #{name}") unless name =~ /^([a-zA-Z0-9]([-a-zA-Z0-9]+)?\.?)+$/
```
dmitri-d: pls. don't perform validation and handling of failed validation in the same method.

modules/dns/dns_api.rb
```diff
@@ -27,17 +39,41 @@ class Api < ::Sinatra::Base
...
+    end
+
+    def validate_reverse_dns_name name
+      validate_dns_name name
+      log_halt(400, "Invalid reverse DNS #{name}") unless name =~ /\.(in-addr|ip6)\.arpa$/
```
dmitri-d: ditto here.

modules/dns/dns_api.rb
ekohl: Should I just throw a Proxy::Dns::Error?

dmitri-d:
> Should I just throw a Proxy::Dns::Error?

Either return a boolean, or raise an exception. In the latter case, please rename the method to `validate_dns_name!` to show a presence of a side-effect (the exception).

ekohl:
Updated using the exception method.

ekohl:
Updated, now with passing tests.

modules/dns_nsupdate/dns_nsupdate_main.rb
```diff
@@ -13,22 +13,41 @@ def initialize(a_server = nil, a_ttl = nil)
     end
 
     def create_a_record(fqdn, ip)
-      do_create(fqdn, ip, "A")
+      case a_record_conflicts(fqdn, ip) #returns -1, 0, 1
+      when 1
+        raise(Proxy::Dns::Collision, "'#{fqdn} 'is already in use")
+      when 0 then
+        return nil
```
dmitri-d, line 20: No need to call 'return' here and in other similar spots -- w/e expression evaluates to will be returned from the method.

dmitri-d:
This looks pretty good to me, just a couple of pretty minor nitpicks. Could you squash the commits? I'll give this a try then...

modules/dns_nsupdate/dns_nsupdate_main.rb
ekohl, line 20: I copied this from dns_dnscmd_main so that should be modified as well then.

dmitri-d, line 20: It's not a mistake per se, just redundant.

helge000, line 20: It expecitly terminates the method, though

modules/dns_nsupdate/dns_nsupdate_main.rb
```diff
@@ -13,22 +13,41 @@ def initialize(a_server = nil, a_ttl = nil)
...
-      do_create(ip, fqdn, "PTR")
+    def create_aaaa_record(fqdn, ip)
+      case aaaa_record_conflicts(fqdn, ip) #returns -1, 0, 1
+      when 1
+        raise(Proxy::Dns::Collision, "'#{fqdn} 'is already in use")
+      when 0 then
+        return nil
+      else
+        do_create(fqdn, ip, "AAAA")
+      end
```
helge000, line 34: You might also want to explicitly return `nil` and alter the tests to use `asstert_nil` make sure not to break foreman

ekohl:
Includes https://github.com/theforeman/smart-proxy/pull/399 and https://github.com/theforeman/smart-proxy/pull/400.

modules/dns_nsupdate/dns_nsupdate_main.rb
ekohl, line 34: Done (though in do_create and do_remove).

ekohl:
Rebased

modules/dns_nsupdate/dns_nsupdate_main.rb
ekohl, line 20: Because it terminates the method, I decided to keep it in.

dmitri-d:
merged as b007039. Thanks, @ekohl! 

PS: I took a liberty of removing a duplicate test: https://github.com/theforeman/smart-proxy/pull/371/files#diff-1bad6d4be1e2837af6415295fec4412fR74

ekohl:
Thanks for merging and reviewing,  but that wasn't duplicate. It checks if an uppercase IPv6 address won't give a conflict when the client submits a lowercase address (``2001:DB8:DEEF::1` vs `2001:dB8:deef::1`).

dmitri-d:
My apologies. I did notice the difference in case, but didn't think it was important. Could you do a PR with just that test (maybe rename it to make it clearer what it's testing?).

ekohl:
https://github.com/theforeman/smart-proxy/pull/407

PR #372, opened 2016-02-05
fixes #13575 - Don't fail on missing PXEClient dhcp opt

agx:
Due to #2870 the dhcp_native_ms provider always tries to set the
PXEClient opt to empty. However this does not exist in all DHCP servers
(tested on 2008R2) and in this case we currently fail with:

```
  INFO -- : Added DHCP reservation for foo.example.com(10.0.2.20 / 52:54:00:84:64:03)
 DEBUG -- : key: "PXEClient"
 DEBUG -- : executing: c:\windows\system32\cmd.exe /c c:\Windows\System32\netsh.exe -c dhcp server 127.0.0.1 scope 10.0.0.0 set reservedoptionvalue 10.0.2.20 60 String ""
 INFO -- : Vendor class not found
ERROR -- : Netsh failed:

Changed the current scope context to 10.0.0.0 scope.

The specified option does not exist.

ERROR -- : Unknown error while processing ''
DEBUG -- : Unknown error while processing '' (Proxy::DHCP::Error)
C:/Users/Administrator/Downloads/smart-proxy/modules/dhcp_native_ms/dhcp_native_ms_main.rb:246:in `rescue in report'
C:/Users/Administrator/Downloads/smart-proxy/modules/dhcp_native_ms/dhcp_native_ms_main.rb:223:in `report'
C:/Users/Administrator/Downloads/smart-proxy/modules/dhcp_native_ms/dhcp_native_ms_main.rb:218:in `execute'
C:/Users/Administrator/Downloads/smart-proxy/modules/dhcp_native_ms/dhcp_native_ms_main.rb:59:in `block in add_record'
C:/Users/Administrator/Downloads/smart-proxy/modules/dhcp_native_ms/dhcp_native_ms_main.rb:46:in `each'
C:/Users/Administrator/Downloads/smart-proxy/modules/dhcp_native_ms/dhcp_native_ms_main.rb:46:in `add_record'
C:/Users/Administrator/Downloads/smart-proxy/modules/dhcp/dhcp_api.rb:97:in `block in <class:DhcpApi>'
C:/Users/Administrator/Downloads/smart-proxy/vendor/ruby/2.1.0/gems/sinatra-1.4.7/lib/sinatra/base.rb:1611:in `call'
```

so ignore errors to set that option.

### DIFF
--- a/modules/dhcp_native_ms/dhcp_native_ms_main.rb
+++ b/modules/dhcp_native_ms/dhcp_native_ms_main.rb
@@ -57,7 +57,11 @@ module Proxy::DHCP::NativeMS
         else
           logger.debug "key: " + key.inspect
           k = Standard[key] || Standard[key.to_sym]
-          execute "scope #{record.subnet.network} set reservedoptionvalue #{record.ip} #{k[:code]} #{k[:kind]} \"#{value}\"", msg, true
+          begin
+            execute "scope #{record.subnet.network} set reservedoptionvalue #{record.ip} #{k[:code]} #{k[:kind]} \"#{value}\"", msg, true
+          rescue Proxy::DHCP::Error => e
+            raise(e) unless key.to_s == "PXEClient"
+          end
         end
       end
       execute("scope #{record.subnet.network} set dnsconfig #{record.ip} 0 0 0 0", 'disable Dynamic DNS', true) if Proxy::DHCP::NativeMS::Plugin.settings.disable_ddns
--- a/test/dhcp_ms_native/server_ms_test.rb
+++ b/test/dhcp_ms_native/server_ms_test.rb
@@ -13,7 +13,6 @@ class DHCPServerMicrosoftTest < Test::Unit::TestCase
     @subnet_service = Proxy::DHCP::SubnetService.new
     @server = Proxy::DHCP::NativeMS::Provider.new.initialize_for_testing(:name => "1.2.3.4",
                                                                          :service => @subnet_service)
-
     @server.stubs(:execute).with("show scope", "Enumerated the scopes on 1.2.3.4").returns('
 ==============================================================================
  Scope Address  - Subnet Mask    - State        - Scope Name          -  Comment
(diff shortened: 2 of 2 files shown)


dLobatog:
[test]

dLobatog:
@agx You need to make your Foreman membership public so that our bot recognizes it and launches tests - it should show up as a 'badge' when you go to http://github.com/agx - see https://help.github.com/articles/publicizing-or-hiding-organization-membership/ for more info.

dmitri-d:
@agx: can you add a test that verifies this behaviour (no exception is re-thrown if the option was PXEClient, and is re-thrown otherwise)?

agx:
@witlessbird I added some spec to for add_record but how would I mock :execute to raise on certain conditions. My attempts using @server.stubs(:execute).with("....").raises(...) did not work and looking at mocha's docs I don't see an example for it neither do I find precedence in the smart-proxy tests.

dmitri-d:
Not sure what the issue is. This:

```
@server.stubs(:execute).with('scope 172.24.166.0 set reservedoptionvalue 172.24.166.11 60 String "pxeclientval"', nil, true).raises(Proxy::DHCP::Error)
```

works for me.

dmitri-d:
[test]

agx:
On Thu, Feb 18, 2016 at 05:23:03AM -0800, Dmitri Dolguikh wrote:

> Not sure what the issue is. This:
> 
> ```
> @server.stubs(:execute).with('scope 172.24.166.0 set reservedoptionvalue 172.24.166.11 60 String "pxeclientval"', nil, true).raises(Proxy::DHCP::Error)
> ```
> 
> works for me.

Hmm...if I use:

```
diff --git a/test/dhcp_ms_native/server_ms_test.rb b/test/dhcp_ms_native/server_ms_test.rb
index 151b6ce..1548083 100644
--- a/test/dhcp_ms_native/server_ms_test.rb
+++ b/test/dhcp_ms_native/server_ms_test.rb
@@ -293,7 +293,7 @@ Command completed successfully.

     @server.expects(:execute).with('scope 172.24.166.0 add reservedip 172.24.166.11 0011bbccddee test.example.com', 'Added DHCP reservation for test.example.com (172.24.166.11 / 00:11:bb:cc:dd:ee)')
     @server.expects(:execute).with('scope 172.24.166.0 set reservedoptionvalue 172.24.166.11 12 String "test.example.com"', nil, true)
-    @server.expects(:execute).with('scope 172.24.166.0 set reservedoptionvalue 172.24.166.11 60 String "pxeclientval"', nil, true)
+    @server.stubs(:execute).with('scope 172.24.166.0 set reservedoptionvalue 172.24.166.11 60 String "pxeclientval"', nil, true).raises(Proxy::DHCP::Error)
     @server.add_record(to_add)
   end
 end
```

the test should fail since we would raise an exception but this doesn't
happen if I run:

```
bundle exec rspec -I"lib:.:lib:test" test/dhcp_ms_native/server_ms_test.rb
```

so while the stub is added the .raises() is not used.

dmitri-d:
@agx: try using a regular test runner instead of rspec (there are no rspec tests in smart-proxy anyway):

`bundle exec rake test TEST=test/dhcp_ms_native/server_ms_test`

agx:
On Fri, Feb 26, 2016 at 12:31:32AM -0800, Dmitri Dolguikh wrote:

> @agx: try using a regular test runner instead of rspec (there are no rspec tests in smart-proxy anyway):
> 
> `bundle exec rake test TEST=test/dhcp_ms_native/server_ms_test`

Doesn't make a difference. I sure tested that things fail properly with
the line above. Running just one test is just faster.

dmitri-d:
The test isn't failing b/c the exception is raised when "PXEClient" option is being added, which results in the exception being silenced, which is what we wanted.

Please add a dedicated test for this case though, as the one you changed tests a slightly different scenario -- when adding a "PXEClient" option doesn't result in an exception.

dmitri-d:
Also, could you rebase your branch against develop please?

alexjfisher:
I have a single subnet with DHCP coming from a Win 2008 server, (other subnets are all on a Win 2012 server).  This patch worked for me.  Any chance of it being included in 1.13?

Thanks

dmitri-d:
@alexjfisher: there are issues with tests in this PR, I can't merge it before they have been resolved.

agx:
On Tue, Jul 19, 2016 at 03:29:11AM -0700, Dmitri Dolguikh wrote:

> @alexjfisher: there are issues with tests in this PR, I can't merge it before they have been resolved.

I've updated the tests, hopefully they now meet your expectations.

dmitri-d:
[test]

dmitri-d:
Merged as e5aa234. Thanks, @agx.

PR #373, opened 2016-02-10
Fixes #13637 - Upgrade rubocop to 0.37.1

brandonweeks:

PR #374, opened 2016-02-10
fixes #13638 - Enable rubocop 0.37.1 cops

brandonweeks:
Depends on #373.


lib/proxy/log.rb
```diff
@@ -27,13 +27,13 @@ def self.default_logger(log_file)
 
     def self.logger
       log_file = ::Proxy::SETTINGS.log_file
-      if log_file.upcase == 'STDOUT'
+      if log_file.casecmp('STDOUT')
```
domcleal: This returns an int, it needs comparing to zero I think. You could add `.zero?`

lib/proxy/log.rb
```diff
@@ -27,13 +27,13 @@ def self.default_logger(log_file)
...
         if SETTINGS.daemon
           puts "Settings log_file=STDOUT and daemon=true are incompatible, exiting..."
           exit 1
         end
         logger = ::Logger.new(STDOUT)
-      elsif log_file.upcase == 'SYSLOG'
+      elsif log_file.casecmp('SYSLOG')
```
domcleal: ditto, `.zero?`

tasks/pkg.rake
```diff
@@ -6,7 +6,7 @@ namespace :pkg do
     File.exist?('pkg') || FileUtils.mkdir('pkg')
     ref = ENV['ref'] || 'HEAD'
     version = `git show #{ref}:VERSION`.chomp.chomp('-develop')
-    raise "can't find VERSION from #{ref}" if version.length == 0
+    raise "can't find VERSION from #{ref}" if !version.lengt.empty?
```
domcleal: typo, `!version.empty?`

test/dhcp_isc/server_isc_test.rb
```diff
@@ -64,15 +64,15 @@ def test_sparc_host_quirks
...
-      %q{option SUNW.root-server-hostname \"itgsyddev807.macbank\";},
-      %q{option SUNW.root-server-ip-address 192.168.122.24;},
-      %q{option SUNW.sysid-config-file-server \"192.168.122.24:/Solaris/jumpstart/sysidcfg/sysidcfg_primary\";},
-      %q{vendor-option-space SUNW;}
+      'option SUNW.JumpStart-server \"192.168.122.24:/Solaris/jumpstart\";',
+      'option SUNW.install-path \"/Solaris/install/Solaris_5.10_sparc_hw0811\";',
+      'option SUNW.install-server-hostname \"itgsyddev807.macbank\";',
+      'option SUNW.install-server-ip-address 192.168.122.24;',
+      'option SUNW.root-path-name \"/Solaris/install/Solaris_5.10_sparc_hw0811/Solaris_10/Tools/Boot\";',
+      'option SUNW.root-server-hostname \"itgsyddev807.macbank\";',
```
domcleal: I think the meaning of the escapes may change here, as `'\"'` is the value `\"` while `%q{\"}` is `"`.

domcleal:
Sorry for the duplicate comments, I incorrectly commented on the commit rather than the PR, so my comments would get lost on GitHub.

dmitri-d:
I'm still thinking about this, but I'm not sure how much I like enforcing coding style to such a degree...

domcleal:
The cops I'm particularly supportive of are the ones with a performance impact - those under Performance/, but also some stylistic ones that have a small impact.  Some of the whitespace type ones can end up in excessive bikeshedding, but otherwise I think the consistency it brings to the project - both internally, and with the wider community - is worth it.

dLobatog:
I'm fine with nearly of all of it. The ABC complexity, class size and other cops are a bit hard to enforce but overall I like how the code looks the same no matter where you look with rubocop.

brandonweeks:
[test]

test/dhcp_isc/server_isc_test.rb
domcleal: Unfixed?

domcleal:
Would you mind rebasing this?  There's also an inline comment about escaping that I'm concerned about.

test/dhcp_isc/server_isc_test.rb
dmitri-d: This is an issue indeed.

brandonweeks:
Done. :)

domcleal:
Merged as 5f5c216, thanks @brandonweeks.

PR #375, opened 2016-02-10
fixes #13522 add correct path to puppet binary

choonming:
Puppet 4 AIO uses /opt/puppetlabs/bin to store the puppet executable. This allows the smart-proxy to list all the certs that are signed/unsigned/revoked


domcleal:
Thanks for sending the patch in @choonming.  I'll kick off the tests on Jenkins: [test]

Don't worry too much about the commit message comment for now, it's just missing punctuation after the ticket number which I can fix on merge.

domcleal:
Merged as 9bd4470, thanks @choonming!

PR #376, opened 2016-02-11
Fixes #13667: puppet gem is no longer being loaded in puppet version >= 4.0 environments

dmitri-d:


dmitri-d:
puppet_proxy will no longer load puppet gem when running in puppet >= 4.0 environments. Similar change was introduced for puppet_ca module: it doesn't load puppet gem independent of puppet version used.

This change introduces a new required parameter: 'puppet_version'. I'm not sure if this is a good name: should it be 'puppet_server_version' instead? Is there a good default value(s) for puppet_proxy and puppet_ca?

Depending  on 'puppet_version',  'puppet_use_environment_api', and puppet configuration used, 'puppet_url', 'puppet_ssl_ca', 'puppet_ssl_cert', and 'puppet_ssl_key' may be required to used puppet_proxy module (these are mandatory for puppet versions 4.0 and above).

config/settings.d/puppet.yml.example
```diff
@@ -41,16 +41,20 @@
 # If you want to override the puppet_user above just for mco commands
 #:mcollective_user: peadmin
 
-# URL of the puppet master itself for API requests
+# Override use of Puppet's API to list environments, by default it will use only if
+# environmentpath is given in puppet.conf, else will look for environments in puppet.conf
+# (Puppet versions prior to 4.0 only)
+#:puppet_use_environment_api: true
+
+# Puppet master version used
```
GregSutcliffe: nitpick: puppet server is more correct going forwards (the master still exists, but not for long)

GregSutcliffe:
Just tried it out (against a puppet server on another host, nifty :P) and it seems solid. Two small things:
- Had to delete the puppet line in `bundler.d/puppet.rb` to ensure puppet wasn't being loaded - we'll need to figure out the right way to ensure the presence/absence of that.
- Same file has augeas in it which I'm not sure we need any more - it's used to load the `puppet.conf` but I don't think we actually use any values from it now. Can we drop it? Simplifies packaging.

dmitri-d:
Augeas is no longer required when smart-proxy is used with puppet versions > 4.0. 
I wonder if using `bundle install --without puppet puppetca` is good enough for source installs in puppet 4.0 environments?

dmitri-d:
rebased.

dmitri-d:
A note to myself: facts module has a dependency on facter, which will need to be added to the appropriate bundler group.

config/settings.d/puppet.yml.example
domcleal: Even nitpickier: Puppet Server has its own version scheme, which contains a specific version of Puppet (like Puppet Agent w.x contains Puppet y.z).  I think the current wording is slightly safer, or just remove "master".

config/settings.d/puppetca.yml.example
```diff
@@ -2,6 +2,10 @@
 # Can be true, false, or http/https to enable just one of the protocols
 :enabled: false
 
+# Puppet master version used
+#:puppet_version: 4.1
+
+:puppet_version:
```
domcleal: should this be left out, and remove validate_presence?

modules/puppet_proxy/dependency_injection/dependencies.rb
```diff
@@ -1,45 +1,46 @@
-require 'puppet_proxy/runtime_configuration'
-
+# rubocop:disable Style/Lambda
```
domcleal: Please keep this enabled globally, disabling it per-file makes the codebase inconsistent.

config/settings.d/puppet.yml.example
```diff
@@ -41,16 +41,20 @@
...
+#:puppet_version: 4.1
```
domcleal, line 50: This probably contradicts what I wrote on the puppetca.yml file, but this should probably have a default value, or at least a config migration to add one.  At the moment it doesn't work out of the box, and fails to startup after upgrade.

config/settings.d/puppet.yml.example
dmitri-d: Yeah, I wasn't sure what to use, I figured most people would know what version of puppet they are running straight away, and we know what version of puppet server (or more importantly version of api) that corresponds too. Will drop 'master' from the comment.

modules/puppetca/puppetca_main.rb
```diff
@@ -106,7 +105,7 @@ def find_puppetca
       default_path = ["/opt/puppet/bin", "/opt/puppet/sbin", "/opt/puppetlabs/bin"]
 
       # puppetca is the old method of using puppet cert which is new in puppet 2.6
-      if Puppet::PUPPETVERSION.to_i < 3
+      if Proxy::PuppetCa::Plugin.settings.puppet_version.to_s < '3'
         @puppetca = which("puppetca", default_path) || which("puppet", default_path)
```
dmitri-d: I guess the conditional can be dropped with only `@puppetca = which("puppetca", default_path) || which("puppet", default_path)` bit left. In that case puppet_version property can go.

config/settings.d/puppet.yml.example
dmitri-d, line 50: I like the idea of a default value, but not sure what it should be?

modules/puppet_proxy/dependency_injection/dependencies.rb
dmitri-d: Not sure what you mean?

modules/puppetca/puppetca_main.rb
domcleal: Yes, that's probably true in most cases.  It might have been written more defensively to begin with (when binary names were switched in 2 to 3) as having multiple versions might have picked the wrong binary.

modules/puppet_proxy/dependency_injection/dependencies.rb
domcleal: Please delete the line and avoid disabling cops per-file.

dmitri-d: I can disable this cop globally. Can't switch to stabby lambdas yet (which this cop insists on), because 1.8.7.

config/settings.d/puppet.yml.example
domcleal, line 50: There are three common scenarios I think we should handle:
1. Upgrades on a regular, existing Puppet 2/3 installation - could a migration detect the current Puppet version and fill that in?  Puppet should be require-able.
2. New Foreman installer installations, the installer will be templating this file and can supply the right value.
3. New installations, the config file could have a default value of the latest version of Puppet.  Perhaps we don't even need this one?

test/dependency_injection_test.rb
```diff
@@ -1,7 +1,15 @@
+# rubocop:disable Style/Lambda
```
domcleal: Also delete this please, keep it enabled.

test/puppet/configuration_validator_test.rb
```diff
@@ -0,0 +1,89 @@
+# rubocop:disable  Style/BlockDelimiters
```
domcleal: Definitely keep this enabled, the line it fixes can be trivially improved.

test/puppet/configuration_validator_test.rb
```diff
@@ -0,0 +1,89 @@
...
+  end
+
+  def test_validate_puppet_conf_returns_true_if_file_is_readable
+    @ssl_config_validator.classes_retriever = :future_parser
+    assert @ssl_config_validator.validate_puppet_conf!(certs_path('puppet_ca.pem'))
+  end
+
+  def test_validate_puppet_conf_raises_exception_if_file_does_not_exist
+    @ssl_config_validator.environments_retriever = :future_parser
+    assert_raises Proxy::Error::ConfigurationError do @ssl_config_validator.validate_puppet_conf!("non-existent") end
```
domcleal: `{ .. }`

modules/puppet_proxy/dependency_injection/dependencies.rb
domcleal: Ah yes, definitely disable it globally then as it's applicable to the whole codebase.

test/puppet/puppet_config_environments_retriever_test.rb
```diff
@@ -1,5 +1,6 @@
+# rubocop:disable  Style/BlockDelimiters
```
domcleal: ditto, use `{ .. }` instead of `do .. end`

domcleal:
> Augeas is no longer required when smart-proxy is used with puppet versions > 4.0.
> I wonder if using bundle install --without puppet puppetca is good enough for source installs in puppet 4.0 environments?

I tried this, but the Bundler group is still going to be required by the plugin initialisation.  Is this a case for multiple providers with different sets of dependencies?

dmitri-d:
- removed puppet_version config setting from puppetca module
- added migration for puppet_version setting for puppet_proxy which uses Puppet::PUPPETVERSION if available or "4.3.1"
- fixed rubocop issues

Still need to figure out what to do with augeas and loading of bundler groups.

dmitri-d:
fixed a style issue that rubocop wasn't happy with.

jyaworski:
LGTM.

alexjfisher:
I've been testing this PR today.

It looks like there's an issue importing environments if an empty environment is encountered.

```
Oops, we're sorry but something went wrong

ERF12-4115 [ProxyAPI::ProxyException]: Unable to get classes from Puppet for common ([RestClient::NotAcceptable]: 406 Not Acceptable) for proxy
```

Not too sure how I ended up with an empty 'common' environment.  Looks like https://github.com/theforeman/puppet-puppet/blob/master/manifests/server/config.pp#L156 might have been responsible.  In any case, gracefully ignoring the environment would be better.

alexjfisher:
There's a new API in puppetserver 2.3 that replaces the 'Resource Type' API. https://docs.puppetlabs.com/puppetserver/2.3/release_notes.html#new-feature-environmentclasses-api

mmoll:
@witlessbird what's the status here?

dmitri-d:
I'm going through the tests broken during splitting up of the module into providers. Another two days max.

dmitri-d:
closing in favour of https://github.com/theforeman/smart-proxy/pull/406

PR #379, opened 2016-02-16
Fixes #13536 - working ptr's for dnscmd

helge000:
Makes PTR's for dnscmd working again.

~~Sadly~~ I had to introduce a ~~rather complicated~~ method `#get_ptr_record` ~~because dnscmd's rather strange implementation of adding pointers~~. It should also handle IPv6 reverse zones, see tests.

It executes dnscmd in the following manner:

Eg for host `host.foo.bar.domain.local.` with `192.168.33.2` and a given /16 reverse zone `168.192.in-addr.arpa`:
#### IPv4

```
dnscmd /RecordAdd 168.192.in-addr.arpa 2.33.168.192.in-addr.arpa. PTR host.foo.bar.domain.local.
```
#### IPv6

```
dnscmd /RecordAdd f.e.e.d.8.b.d.0.1.0.0.2.ip6.arpa 1.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.f.e.e.d.8.b.d.0.1.0.0.2.ip6.arpa. PTR host.foo.bar.domain.local.
```

~~**Note:** This further fixes a rubocop offense introduced with overlapping #369 and #373 merges (it seems)~~

### DIFF
--- a/modules/dns_common/dns_common.rb
+++ b/modules/dns_common/dns_common.rb
@@ -1,3 +1,6 @@
+require 'resolv'
+require 'ipaddr'
+
 module Proxy::Dns
   class Error < RuntimeError; end
   class NotFound < RuntimeError; end
@@ -23,6 +26,44 @@ module Proxy::Dns
       end
     rescue Resolv::ResolvError
       false
+    end
+
+    def ptr_to_ip ptr
+     if ptr =~ /\.in-addr\.arpa$/
+       ptr.split('.')[0..-3].reverse.join('.')
+     elsif ptr =~ /\.ip6\.arpa$/
+       ptr.split('.')[0..-3].reverse.each_slice(4).inject([]) {|address, word| address << word.join}.join(":")
+     else
+       raise Proxy::Dns::Error.new("Not a PTR record: '#{ptr}'")
+     end
+    end
+
+    # conflict methods return values:
+    # no conflict: -1; conflict: 1, conflict but record / ip matches: 0
+    def a_record_conflicts(fqdn, ip)
+      if ip_addr = to_ipaddress(ip)
+        addresses = resolver.getaddresses(fqdn)
+        return -1 if addresses.empty?
+        return 0 if addresses.any? {|a| IPAddr.new(a) == ip_addr}
+        1
+      else
+        raise Proxy::Dns::Error.new("Not an IP Address: '#{ip}'")
+      end
+    end
+
+    def ptr_record_conflicts(fqdn, ip)
+      if ip_addr = to_ipaddress(ip)
+        names = resolver.getnames(ip_addr.to_s)
+        return -1 if names.empty?
+        return 0 if names.any? {|n| n.to_s.casecmp(fqdn) == 0}
+        1
+      else
+        raise Proxy::Dns::Error.new("Not an IP Address: '#{ip}'")
+      end
+    end
+
+    def to_ipaddress ip
+      IPAddr.new(ip) rescue false
     end
   end
 end
--- a/modules/dns_dnscmd/dns_dnscmd_main.rb
+++ b/modules/dns_dnscmd/dns_dnscmd_main.rb
@@ -1,4 +1,3 @@
-require 'resolv'
 require 'open3'
 require 'dns_common/dns_common'
 
@@ -13,36 +12,49 @@ module Proxy::Dns::Dnscmd
     end
 
     def create_a_record(fqdn, ip)
-      if found = dns_find(fqdn)
-        raise(Proxy::Dns::Collision, "#{fqdn} is already used by #{found}") if found != ip
-      else
-        zone = match_zone(fqdn)
-        msg = "Added DNS entry #{fqdn} => #{ip}"
-        cmd = "/RecordAdd #{zone} #{fqdn}. A #{ip}"
-        execute(cmd, msg)
+      case a_record_conflicts(fqdn, ip) #returns -1, 0, 1
+        when 1 then
+          raise(Proxy::Dns::Collision, "'#{fqdn} 'is already in use")
+        when 0 then
+          return nil
+        else
+          zone = match_zone(fqdn, enum_zones)
+          msg = "Added DNS entry #{fqdn} => #{ip}"
+          cmd = "/RecordAdd #{zone} #{fqdn}. A #{ip}"
+          execute(cmd, msg)
+          nil
       end
     end
 
-    # noop
-    def create_ptr_record(fqdn, ip)
-      found = dns_find(ip)
-      raise(Proxy::Dns::Collision, "#{ip} is already used by #{found}") if found && found != fqdn
-      true
+    def create_ptr_record(fqdn, ptr)
+      case ptr_record_conflicts(fqdn, ptr_to_ip(ptr)) #returns -1, 0, 1
+        when 1 then
+          raise(Proxy::Dns::Collision, "'#{ptr}' is already in use")
+        when 0
+          return nil
+        else
+          zone = match_zone(ptr, enum_zones)
+          msg = "Added PTR entry #{ptr} => #{fqdn}"
+          cmd = "/RecordAdd #{zone} #{ptr}. PTR #{fqdn}."
+          execute(cmd, msg)
+          nil
+      end
     end
 
     def remove_a_record(fqdn)
-      ip = dns_find(fqdn)
-      raise Proxy::Dns::NotFound.new("Cannot find DNS entry for #{fqdn}") unless ip
-      zone = match_zone(fqdn)
-      msg = "Removed DNS entry #{fqdn} => #{ip}"
+      zone = match_zone(fqdn, enum_zones)
+      msg = "Removed DNS entry #{fqdn}"
       cmd = "/RecordDelete #{zone} #{fqdn}. A /f"
       execute(cmd, msg)
+      nil
     end
 
-    # noop
-    def remove_ptr_record(ip)
-      raise Proxy::Dns::NotFound.new("Cannot find DNS entry for #{ip}") unless dns_find(ip)
-      true
+    def remove_ptr_record(ptr)
+      zone = match_zone(ptr, enum_zones)
+      msg = "Removed PTR entry #{ptr}"
+      cmd = "/RecordDelete #{zone} #{ptr}. PTR /f"
+      execute(cmd, msg)
+      nil
     end
 
     def execute cmd, msg=nil, error_only=false
@@ -87,15 +99,13 @@ module Proxy::Dns::Dnscmd
       raise Proxy::Dns::Error.new("Unknown error while processing '#{msg}'")
     end
 
-    def match_zone(fqdn)
-      dns_zones = enum_zones
+    def match_zone(record, zone_list)
       weight = 0 # sub zones might be independent from similar named parent zones; use weight for longest suffix match
       matched_zone = nil
-
-      dns_zones.each do |zone|
+      zone_list.each do |zone|
         zone_labels = zone.split(".").reverse
         zone_weight = zone_labels.length
-        fqdn_labels = fqdn.split(".")
+        fqdn_labels = record.split(".")
         fqdn_labels.shift
         is_match = zone_labels.all? { |zone_label| zone_label == fqdn_labels.pop }
         # match only the longest zone suffix
@@ -104,18 +114,18 @@ module Proxy::Dns::Dnscmd
           weight = zone_weight
         end
       end
-      raise Proxy::Dns::NotFound.new("The DNS server has no authoritative zone for #{fqdn}") unless matched_zone
+      raise Proxy::Dns::NotFound.new("The DNS server has no authoritative zone for #{record}") unless matched_zone
       matched_zone
     end
 
     def enum_zones
       zones = []
-      response = execute('/EnumZones')
+      response = execute '/EnumZones', nil, true
       response.each do |line|
         next unless line =~  / Primary /
         zones << line.sub(/^ +/, '').sub(/ +.*$/, '').chomp("\n")
       end
-      logger.debug "Enumerated dns zones: #{zones}"
+      logger.debug "Enumerated authoritative dns zones: #{zones}"
       zones
     end
   end
--- a/test/dns_dnscmd/dnscmd_test.rb
+++ b/test/dns_dnscmd/dnscmd_test.rb
@@ -14,7 +14,7 @@ class DnsCmdTest < Test::Unit::TestCase
     @server = DnscmdForTesting.new(["_msdcs.bar.domain.local",
                                     "168.192.in-addr.arpa",
                                     "33.168.192.in-addr.arpa",
-                                    "e.e.d.8.b.d.0.1.0.0.2.ip6.arpa",
+                                    "f.e.e.d.8.b.d.0.1.0.0.2.ip6.arpa",
                                     "bar.domain.local",
                                     "domain.local",
                                     "TrustAnchors"])
@@ -30,79 +30,68 @@ class DnsCmdTest < Test::Unit::TestCase
   end
 
   def test_create_address_record_with_longest_zone_match
-    Proxy::Dns::Dnscmd::Record.any_instance.expects(:dns_find).with('host.foo.bar.domain.local').returns(false)
+    Proxy::Dns::Dnscmd::Record.any_instance.expects(:a_record_conflicts).with('host.foo.bar.domain.local', '192.168.33.33').returns(-1)
     Proxy::Dns::Dnscmd::Record.any_instance.expects(:execute).with('/RecordAdd bar.domain.local host.foo.bar.domain.local. A 192.168.33.33', anything).returns(true)
-    assert @server.create_a_record('host.foo.bar.domain.local', '192.168.33.33')
+    assert_nil @server.create_a_record('host.foo.bar.domain.local', '192.168.33.33')
   end
 
   def test_overwrite_address_record_with_longest_zone_match
-    Proxy::Dns::Dnscmd::Record.any_instance.expects(:dns_find).with('host.foo.bar.domain.local').returns('192.168.33.33')
-    @server.create_a_record('host.foo.bar.domain.local', '192.168.33.33')
+    Proxy::Dns::Dnscmd::Record.any_instance.expects(:a_record_conflicts).with('host.foo.bar.domain.local', '192.168.33.33').returns(0)
+    assert_nil @server.create_a_record('host.foo.bar.domain.local', '192.168.33.33')
   end
 
   def test_create_duplicate_address_record_fails
-    Proxy::Dns::Dnscmd::Record.any_instance.expects(:dns_find).with('host.foo.bar.domain.local').returns('192.168.33.34')
-
+    Proxy::Dns::Dnscmd::Record.any_instance.expects(:a_record_conflicts).with('host.foo.bar.domain.local', '192.168.33.33').returns(1)
     assert_raise Proxy::Dns::Collision do
       @server.create_a_record('host.foo.bar.domain.local', '192.168.33.33')
     end
   end
 
   def test_create_ptr_record
-    Proxy::Dns::Dnscmd::Record.any_instance.expects(:dns_find).with('192.168.33.33').returns(false)
-    assert @server.create_ptr_record('host.foo.bar.domain.local', '192.168.33.33')
+    Proxy::Dns::Dnscmd::Record.any_instance.expects(:ptr_record_conflicts).with('host.foo.bar.domain.local', '192.168.33.33').returns(-1)
+    Proxy::Dns::Dnscmd::Record.any_instance.expects(:execute).with('/RecordAdd 33.168.192.in-addr.arpa 33.33.168.192.in-addr.arpa. PTR host.foo.bar.domain.local.', anything).returns(true)
+    assert_nil @server.create_ptr_record('host.foo.bar.domain.local', '33.33.168.192.in-addr.arpa')
   end
 
   def test_overwrite_ptr_record
-    Proxy::Dns::Dnscmd::Record.any_instance.expects(:dns_find).with('192.168.33.33').returns('host.foo.bar.domain.local')
-    @server.create_ptr_record('host.foo.bar.domain.local', '192.168.33.33')
+    Proxy::Dns::Dnscmd::Record.any_instance.expects(:ptr_record_conflicts).with('host.foo.bar.domain.local', '192.168.33.33').returns(0)
+    assert_nil @server.create_ptr_record('host.foo.bar.domain.local', '33.33.168.192.in-addr.arpa')
   end
 
   def test_create_duplicate_ptr_record_fails
-    Proxy::Dns::Dnscmd::Record.any_instance.expects(:dns_find).with('192.168.33.33').returns('another.host.foo.bar.domain.local')
+    Proxy::Dns::Dnscmd::Record.any_instance.expects(:ptr_record_conflicts).with('host.foo.bar.domain.local', '192.168.33.33').returns(1)
     assert_raise Proxy::Dns::Collision do
-      @server.create_ptr_record('host.foo.bar.domain.local', '192.168.33.33')
+      @server.create_ptr_record('host.foo.bar.domain.local', '33.33.168.192.in-addr.arpa')
     end
   end
 
   def test_remove_address_record_with_longest_zone_match
-    Proxy::Dns::Dnscmd::Record.any_instance.expects(:dns_find).with('host.foo.bar.domain.local').returns(true)
     Proxy::Dns::Dnscmd::Record.any_instance.expects(:execute).with('/RecordDelete bar.domain.local host.foo.bar.domain.local. A /f', anything).returns(true)
-    assert @server.remove_a_record('host.foo.bar.domain.local')
-  end
-
-  def test_remove_non_existent_address_record_raises_exception
-    Proxy::Dns::Dnscmd::Record.any_instance.expects(:dns_find).with('host.foo.bar.domain.local').returns(false)
-    assert_raise Proxy::Dns::NotFound do
-      @server.remove_a_record('host.foo.bar.domain.local')
-    end
+    assert_nil @server.remove_a_record('host.foo.bar.domain.local')
   end
 
   def test_remove_ptr_record
-    Proxy::Dns::Dnscmd::Record.any_instance.expects(:dns_find).with('192.168.33.33').returns(true)
-    assert @server.remove_ptr_record('192.168.33.33')
-  end
-
-  def test_remove_nonexistent_ptr_record_raises_exception
-    Proxy::Dns::Dnscmd::Record.any_instance.expects(:dns_find).with('192.168.33.33').returns(false)
-    assert_raise Proxy::Dns::NotFound do
-      @server.remove_ptr_record('192.168.33.33')
-    end
+    Proxy::Dns::Dnscmd::Record.any_instance.expects(:execute).with('/RecordDelete 33.168.192.in-addr.arpa 33.33.168.192.in-addr.arpa. PTR /f', anything).returns(true)
+    assert_nil @server.remove_ptr_record('33.33.168.192.in-addr.arpa')
   end
 
   def test_dns_zone_matches_second_best_match_if_zone_name_equals_host_name
-    assert_equal('33.168.192.in-addr.arpa', @server.match_zone('33.33.168.192.in-addr.arpa'))
-    assert_equal('domain.local', @server.match_zone('bar.domain.local'))
+    assert_equal('domain.local', @server.match_zone('bar.domain.local', @server.enum_zones))
   end
 
   def test_dns_zone_matches_sole_available_zone
-    server = DnscmdForTesting.new(["sole.domain"])
-    assert_equal('sole.domain', server.match_zone('host.foo.bar.sole.domain'))
+    assert_equal('sole.domain', Proxy::Dns::Dnscmd::Record.new.match_zone('host.foo.bar.sole.domain', ["sole.domain"]))
   end
 
-  def test_dns_zone_no_match_raises_exception
+  def test_dns_non_authoritative_zone_raises_exception
     assert_raise Proxy::Dns::NotFound do
-      @server.match_zone('host.foo.bar.domain.com')
+      @server.match_zone('host.foo.bar.domain.com', ['domain.local'])
+    end
+    assert_raise Proxy::Dns::NotFound do
+      @server.match_zone('33.33.16.192.in-addr.arpa', ['168.192.in-addr.arpa'])
+    end
+    assert_raise Proxy::Dns::NotFound do
+      @server.match_zone('1.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.f.e.e.d.8.b.d.1.1.0.0.2.ip6.arpa', ['f.e.e.d.8.b.d.0.1.0.0.2.ip6.arpa'])
     end
   end
 
@@ -117,7 +106,7 @@ Enumerated zone list:
  _msdcs.bar.domain.local        Primary    AD-Forest       Secure
  168.192.in-addr.arpa           Primary    AD-Domain       Rev
  33.168.192.in-addr.arpa        Primary    AD-Domain       Secure Rev Aging
- e.e.d.8.b.d.0.1.0.0.2.ip6.arpa Primary    AD-Domain       Secure Rev
+ f.e.e.d.8.b.d.0.1.0.0.2.ip6.arpa  Primary    AD-Domain       Secure Rev
  bar.domain.local               Primary    AD-Domain       Secure Aging
  domain.local                   Primary    AD-Domain       Secure
  domain.com                     Secondary    File
@@ -125,12 +114,12 @@ Enumerated zone list:
 
 
 Command completed successfully.'.split("\n")
-    Proxy::Dns::Dnscmd::Record.any_instance.expects(:execute).with('/EnumZones').returns(to_parse)
+    Proxy::Dns::Dnscmd::Record.any_instance.expects(:execute).with('/EnumZones', nil, true).returns(to_parse)
     assert_equal [
       "_msdcs.bar.domain.local",
       "168.192.in-addr.arpa",
       "33.168.192.in-addr.arpa",
-      "e.e.d.8.b.d.0.1.0.0.2.ip6.arpa",
+      "f.e.e.d.8.b.d.0.1.0.0.2.ip6.arpa",
       "bar.domain.local",
       "domain.local",
       "TrustAnchors"], Proxy::Dns::Dnscmd::Record.new.enum_zones
--- a/test/dns_common/record_test.rb
+++ b/test/dns_common/record_test.rb
@@ -16,4 +16,54 @@ class DnsRecordTest < Test::Unit::TestCase
     Resolv::DNS.any_instance.expects(:getaddress).with('another.host').raises(Resolv::ResolvError.new('DNS result has no information'))
     assert !Proxy::Dns::Record.new.dns_find('another.host')
   end
+
+  def test_ptr_to_ip_ipv4
+    assert_equal('192.168.33.30', Proxy::Dns::Record.new.ptr_to_ip('30.33.168.192.in-addr.arpa'))
+  end
+
+  def test_ptr_to_ip_ipv6
+    assert_equal('2001:0db8:deef:0000:0000:0000:0000:0001', Proxy::Dns::Record.new.ptr_to_ip('1.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.f.e.e.d.8.b.d.0.1.0.0.2.ip6.arpa'))
+  end
+
+  def test_ptr_to_ip_without_record_exception
+    assert_raise Proxy::Dns::Error do
+      Proxy::Dns::Record.new.ptr_to_ip('host.example.com')
+    end
+  end
+
+  def test_a_record_conflicts_no_conflict
+    Resolv::DNS.any_instance.expects(:getaddresses).with('some.host').returns([])
+    assert_equal -1, Proxy::Dns::Record.new.a_record_conflicts('some.host', '192.168.33.33')
+  end
+
+  def test_a_record_conflicts_has_conflict
+    Resolv::DNS.any_instance.expects(:getaddresses).with('some.host').returns(['192.168.33.33', '2001:DB8:DEEF::1'])
+    assert_equal 1, Proxy::Dns::Record.new.a_record_conflicts('some.host', '2001:db8:deef:1000::1')
+  end
+
+  def test_a_record_conflicts_but_nothing_todo
+    Resolv::DNS.any_instance.expects(:getaddresses).with('some.host').returns(['192.168.33.33', '2001:DB8:DEEF::1'])
+    assert_equal 0, Proxy::Dns::Record.new.a_record_conflicts('some.host', '192.168.33.33')
+  end
+
+  def test_ptr_record_conflicts_no_conflict
+    Resolv::DNS.any_instance.expects(:getnames).with('192.168.33.33').returns([])
+    assert_equal -1, Proxy::Dns::Record.new.ptr_record_conflicts('some.host', '192.168.33.33')
+  end
+
+  def test_ptr_record_conflicts_has_conflict
+    Resolv::DNS.any_instance.expects(:getnames).with('2001:db8:deef::1').returns(['some.host'])
+    assert_equal 1, Proxy::Dns::Record.new.ptr_record_conflicts('another.host', '2001:db8:deef::1')
+  end
+
+  def test_ptr_record_conflicts_but_nothing_todo
+    Resolv::DNS.any_instance.expects(:getnames).with('192.168.33.33').returns(['some.host'])
+    assert_equal 0, Proxy::Dns::Record.new.ptr_record_conflicts('some.host', '192.168.33.33')
+  end
+
+  def test_validate_ip
+    assert_equal '192.168.33.33', Proxy::Dns::Record.new.to_ipaddress('192.168.33.33').to_s
+    assert_equal '2001:db8:deef::1', Proxy::Dns::Record.new.to_ipaddress('2001:db8:deef::1').to_s
+    assert_equal false, Proxy::Dns::Record.new.to_ipaddress('some.host')
+  end
 end


modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -118,5 +124,34 @@ def enum_zones
       logger.debug "Enumerated dns zones: #{zones}"
       zones
     end
+
+    def get_ptr_ip(zone, ip)
```
helge000: I am counting on you to optimitze this, @witlessbird

test/dns_dnscmd/dnscmd_test.rb
```diff
@@ -51,12 +44,13 @@ def test_create_duplicate_address_record_fails
 
   def test_create_ptr_record
     Proxy::Dns::Dnscmd::Record.any_instance.expects(:dns_find).with('192.168.33.33').returns(false)
+    Proxy::Dns::Dnscmd::Record.any_instance.expects(:execute).with('/RecordAdd 33.168.192.in-addr.arpa 33 PTR host.foo.bar.domain.local.', anything).returns(true)
     assert @server.create_ptr_record('host.foo.bar.domain.local', '192.168.33.33')
   end
 
   def test_overwrite_ptr_record
```
helge000, line 56: I cannot get this test working...

test/dns_dnscmd/dnscmd_test.rb
```diff
@@ -11,14 +11,7 @@ def initialize(dns_zones)
...
   def setup
-    @server = DnscmdForTesting.new([
-      "_msdcs.bar.domain.local",
-      "168.192.in-addr.arpa",
-      "33.168.192.in-addr.arpa",
-      "e.e.d.8.b.d.0.1.0.0.2.ip6.arpa",
-      "bar.domain.local",
-      "domain.local",
-      "TrustAnchors"])
+    @server = DnscmdForTesting.new(["_msdcs.bar.domain.local", "168.192.in-addr.arpa", "33.168.192.in-addr.arpa", "f.e.e.d.8.b.d.0.1.0.0.2.ip6.arpa", "bar.domain.local", "domain.local", "TrustAnchors"])
```
helge000: Changed it back to one line. Rubocup seems to like it better after #373

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -118,5 +127,30 @@ def enum_zones
...
+      ip6 = ip.include?(':')
```
dmitri-d: Use `IPAddr#ipv6?` instead.

dmitri-d: You are assuming that ipv6 ptr zone boundary is on each four bytes, while it's on a nibble (each four bits/a single hex digit). This is complicated by ipv6 that can be passed in in a shortened format. 

I think what needs to be done is:
1.  incoming address converted to an IPAddr
2.  reverse zone converted to an address with mask and converted to an IPAddr 
3.  check that passed in address belongs to the subnet (IPAddr#include?)
4.  calculate address bit mask: ptr zone address mask XOR 0xffffffff for ipv4 (0xffffffffffffffffffffffffffffffff for ipv6)
5.  using address from 1) and bit mask from 4): IPAddr.new(ip).to_i & mask
6.  split address from 5) into octets for ipv4, nibbles for ipv6, reverse the order, and return

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -118,5 +127,30 @@ def enum_zones
...
+          if partial_ip6.length == 4
+            ip6_zone << partial_ip6
+            partial_ip6 = ''
+          end
+        end
+        ip6_zone + [partial_ip6] unless partial_ip6.empty?
+      end
+      if ip6
+        ip.slice!("#{ip6_zone.join(':')}:")
+        result = "#{ip.split(':').reverse.join(':')}"
```
dmitri-d: shouldn't this be nibbles separated by '.'?

helge000: Right! I could not find any documentation. Turns out we can use the full reverse output of the IP to add the record. However, I cannot point to any documentation on this.

modules/dns_dnscmd/dns_dnscmd_main.rb
helge000: > Use IPAddr#ipv6? instead.

I am working on all that right now; much easier now that I know I can require `ipaddr`

modules/dns_dnscmd/dns_dnscmd_main.rb
dmitri-d: What about removing? If so, IPAddr.reverse and IPAddr#ip6_arpa is all you need (as opposed to logic I described above).

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -118,5 +127,30 @@ def enum_zones
...
+      else
+        result = "#{ip.split('.').reverse.join('.')}.in-addr.arpa"
+        result.slice!(".#{zone}")
```
helge000: You dont't like this simple slice @witlessbird? Do have an example when this would not work?

dmitri-d: I don't like the idea of treating an address as a string for purposes of separation of ptr and zone parts of a record. 
I _suspect_ however, you can pass in the whole address (including ip6.arpa/in-addr.arpa bit), probably ending with a '.' to add/remove record command.

modules/dns_dnscmd/dns_dnscmd_main.rb
helge000: > 1. check that passed in address belongs to the subnet (IPAddr#include?)

This one is pretty hard. Can we levarage this on dnscmd? It will fail `DNS_ERROR_NOT_ALLOWED_UNDER_DELEGATION` if we add an IP out of range...

modules/dns_dnscmd/dns_dnscmd_main.rb
helge000: Yep, new commit with a much shorter method comming right up

modules/dns_dnscmd/dns_dnscmd_main.rb
dmitri-d: This works.

modules/dns_dnscmd/dns_dnscmd_main.rb
helge000: > I suspect however, you can pass in the whole address (including ip6.arpa/in-addr.arpa bit), probably ending with a '.' to add/remove record command.

This is only true in case of IPv6. IPv4 PTRs need to be zone relative; my tests adding `.` raised `DNS_ERROR_NOT_ALLOWED_UNDER_DELEGATION`

helge000:
Using `IPAddr::` I could simplify the `#get_ptr_ip` alot. Further I added the functionality to use `#match_zone` directly with an IP _or_ FQDN. Both methods should fully support IPv4 and IPv6.

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -104,7 +123,7 @@ def match_zone(fqdn)
           weight = zone_weight
         end
       end
-      raise Proxy::Dns::NotFound.new("The DNS server has no authoritative zone for #{fqdn}") unless matched_zone
+      raise Proxy::Dns::NotFound.new("The DNS server has no authoritative zone for #{fqdn_or_ip}") unless matched_zone
```
helge000: Using IP's with mach_zone should also solve the the problem of IP not in subnet since we would not have mached_zone here, @witlessbird

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -87,15 +97,24 @@ def report msg, response, error_only
       raise Proxy::Dns::Error.new("Unknown error while processing '#{msg}'")
     end
 
-    def match_zone(fqdn)
+    def match_zone(fqdn_or_ip)
       dns_zones = enum_zones
       weight = 0 # sub zones might be independent from similar named parent zones; use weight for longest suffix match
       matched_zone = nil
+      ip_check = validate_ip6(fqdn_or_ip)
```
dmitri-d: what happens when an address is ipv4?

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -87,15 +97,24 @@ def report msg, response, error_only
       raise Proxy::Dns::Error.new("Unknown error while processing '#{msg}'")
     end
 
-    def match_zone(fqdn)
+    def match_zone(fqdn_or_ip)
```
dmitri-d: Pass in zone_name as a parameter. You can remove conditional logic used to determine whether zone is forward or ptr one.

modules/dns_dnscmd/dns_dnscmd_main.rb
helge000: `true` if Ipv6, `false` if IPv4, `nil` if anything else

modules/dns_dnscmd/dns_dnscmd_main.rb
helge000: ~~Sorry, not getting that. How would I pass the zone name if it is the method's job to determine exactly that?~~
I think you mean something like `match_zone(get_ptr_record(ip))` in case of PTR's?
I rather like the idea to be able to pass an IP or FQDN.

modules/dns_dnscmd/dns_dnscmd_main.rb
dmitri-d: Please don't do this: nil is also evaluated to false, anything but nil evaluates as true, this is going to cause issues. As I suggested, remove this conditional altogether and pass in the zone name as a parameter.

modules/dns_dnscmd/dns_dnscmd_main.rb
dmitri-d: You just made this method responsible for determining the zone name. This logic is implicit in the context of the call to `match_zone`: create_a/remove_a methods use fqdn, create_ptr/remove_ptr need a ptr. I'd also pass in zone list as a parameter while I'm at it -- makes testing easier. So, for the former:

```
match_zone(fqdn, enum_zones)
```

for ptr-related methods:

```
match_zone(get_ptr_record(fqdn_or_ip), enum_zones)
```

dmitri-d: You increased `match_zone`'s cyclomatic complexity, made it responsible for more than one thing, relied on tricky call semantics, made testing harder -- not a good change.

helge000: Yea, you're right.. ;) 
I'm at it

helge000:
Removed the validate_ip method and passing zone_list as parameter

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -115,8 +123,23 @@ def enum_zones
...
+      begin
+        ipaddr = IPAddr.new(ip)
+      rescue
+        raise Proxy::Dns::Error.new("Not an IPv4 or IPv4 address: '#{ip}'")
+      end
+
+      if ipaddr.ipv6?
+        result = ipaddr.reverse
+      else
+        result = "#{ip.split('.').reverse.join('.')}.in-addr.arpa"
```
dmitri-d: Why not use `IPAddr#reverse` for ipv4 too?

dmitri-d:
Looks good, other than one small issue above!

helge000:
> Why not use IPAddr#reverse for ipv4 too?

because I misunderstood the [method description](http://ruby-doc.org/stdlib-1.9.3/libdoc/ipaddr/rdoc/IPAddr.html#method-i-reverse) in a way it only works for IPv6 addresses ;) 

Should be better now.
Are we already too late for 1.11 with this?

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -115,8 +123,17 @@ def enum_zones
         next unless line =~  / Primary /
         zones << line.sub(/^ +/, '').sub(/ +.*$/, '').chomp("\n")
       end
-      logger.debug "Enumerated dns zones: #{zones}"
+      logger.debug "Enumerated dns authoritative dns zones: #{zones}"
       zones
     end
+
+    def get_ptr_record(ip)
+      begin
```
dmitri-d: You can remove explicit try-rescue boundaries in the context of a method definition:

```
def get_ptr_record(ip)
  ipaddr = IPAddr.new(ip)
  ipaddr.reverse
rescue
  raise Proxy::Dns::Error.new("Not an IP address: '#{ip}'")
end
```

helge000: Right, forgot that. Stange though: rubocop did offend this in the past

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -16,33 +17,42 @@ def create_a_record(fqdn, ip)
...
       msg = "Removed DNS entry #{fqdn} => #{ip}"
       cmd = "/RecordDelete #{zone} #{fqdn}. A /f"
       execute(cmd, msg)
     end
 
-    # noop
     def remove_ptr_record(ip)
-      raise Proxy::Dns::NotFound.new("Cannot find DNS entry for #{ip}") unless dns_find(ip)
-      true
+      fqdn = dns_find(ip)
```
dmitri-d: This fails when ipv6 is used (dns_find is ipv4-specific). I think it's ok to drop this check altogether: we either successfully delete the record, or it's not there in the first place, end result is the same.

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -16,33 +17,42 @@ def create_a_record(fqdn, ip)
...
+      raise Proxy::Dns::NotFound.new("Cannot find DNS entry for #{ip}") unless fqdn
+      ptr_record = get_ptr_record(ip)
```
dmitri-d: This is redundant, as ip address parameter to PTR delete operation (https://github.com/theforeman/smart-proxy/blob/develop/modules/dns/dns_api.rb#L30) is assumed to be in the reversed order.

modules/dns_dnscmd/dns_dnscmd_main.rb
helge000: Removing a non is exiting ptr is [probably not ok](https://github.com/theforeman/smart-proxy/pull/379/files#diff-aede4453fda32f153191fffeb8d474dcR89)

dmitri-d: MS dns server doesn't return an error when deleting a non-existent record (comes back with a success).

modules/dns_dnscmd/dns_dnscmd_main.rb
helge000: Really... the parameter name [and the tests](https://github.com/theforeman/smart-proxy/blob/develop/test/dns_dnscmd/dnscmd_test.rb#L53) are quite misleading in this case.

modules/dns_dnscmd/dns_dnscmd_main.rb
helge000: You know your dnscmd much better than me :+1:

dmitri-d: Hah. Not really, have a vm with ms dns on it. Spent the whole afternoon testing smart-proxy against it.

modules/dns_dnscmd/dns_dnscmd_main.rb
dmitri-d: https://github.com/theforeman/smart-proxy/blob/develop/modules/dns/dns_api.rb#L31 is the line in question. Also checked code on the foreman side too. And you are indeed correct, the tests are wrong.

test/dns_dnscmd/dnscmd_test.rb
```diff
@@ -79,30 +80,41 @@ def test_remove_non_existent_address_record_raises_exception
   end
 
   def test_remove_ptr_record
-    Proxy::Dns::Dnscmd::Record.any_instance.expects(:dns_find).with('192.168.33.33').returns(true)
-    assert @server.remove_ptr_record('192.168.33.33')
+    Proxy::Dns::Dnscmd::Record.any_instance.expects(:execute).with('/RecordDelete 33.168.192.in-addr.arpa 33.33.168.192.in-addr.arpa. PTR /f', anything).returns(true)
+    assert @server.remove_ptr_record('33.33.168.192.in-addr.arpa')
```
helge000: This should be be more realistic?

modules/dns_dnscmd/dns_dnscmd_main.rb
helge000: > Spent the whole afternoon testing smart-proxy against it.

Lucky you!

If you allow me one off topic question: Did smart-proxy successfully remove DHCP reservations? This is really bugging me since the time I started out with foremen.

dimatha:
Hi all,

I just applied this patch and I'm getting: "DEBUG -- : Not an IP address: '100.10.10.10.in-addr.arpa' (Proxy::Dns::Error) while trying to create PTR". ip is already in a reverse form ?

More details here: http://pastebin.com/5KM4r0qJ

Thanks
Dmitry

helge000:
>  ip is already in a reverse form ?

Looks that way;  found this in [foreman's orchestration](https://github.com/theforeman/foreman/blob/develop/app/models/concerns/orchestration/dns.rb#L46) leading to this [DNS module](https://github.com/theforeman/foreman/blob/8bbfa2c36cb07d6c9d48631384b051c2532c4e67/lib/net/dns/ptr_record.rb#L41).

Mind you, I am far from an expert; can you confirm this behavior @witlessbird , @domcleal ? If so, our tests need further tweaking. Strangely [nsupdate uses the same tests with IP adresses?](https://github.com/theforeman/smart-proxy/blob/develop/test/dns_nsupdate/dns_nsupdate_test.rb#L87)

dmitri-d:
This is the relevant bit on the foreman side: https://github.com/theforeman/foreman/blob/develop/lib/net/dns/ptr_record.rb#L13. 

As I mentioned before, foreman does send ptr address to 'delete' call reversed, so the tests here need to be updated.

helge000:
@dimatha, can you confirm you are running 1.11 foreman and smart proxy, and did not just patch the one file?

dmitri-d:
The latest changes look good though, will test today. I realised the question from @dimatha was about record creation, not deletion: ptr 'create' call uses regular ip address and host name. This is due to 'create' being the same for 'A' and 'PTR' (and 'AAAA' in the future) types of records. Here's the corresponding bit from the foreman side: https://github.com/theforeman/foreman/blob/develop/lib/net/dns/ptr_record.rb#L18

helge000:
> Here's the corresponding bit from the foreman side:

While that lead me to look into `def attr` - which confirm that forman sends us  PTR records on create as well, see links above (always given I found the right part)

dmitri-d:
@helge000: you are indeed correct -- PTR create uses reversed ip address as a value.

helge000:
Since I question the the reasoning on sending PTR-records altogether, I would workaround this by figuring out whatever we are dealing with an IP address or PTR record, @witlessbird?

dmitri-d:
Not sure what you mean, @helge000. https://github.com/theforeman/smart-proxy/blob/develop/modules/dns/dns_api.rb#L34 and https://github.com/theforeman/smart-proxy/blob/develop/modules/dns/dns_api.rb#L21 are responsible for calling an appropriate method. API itself isn't going to change either: regular ip addresses for operations on 'A' records and reversed ip's for 'PTR' records...

helge000:
Never mind, I am just composing a message to the ML about this.

The more pressing issue: If I get a PTR record from foreman, I am not able to check for a duplicate record without applying some questionable string voodoo. While this would fix the immediate issue, there's no way of converting IPv6 PTR's.

OTOH, we can just let `#execute` on creating the duplicate ptr, which is a bad thing in my opinion (apprart from delete, dnscmd fails on creating duplicate PTRs).

dmitri-d:
Here: https://github.com/helge000/smart-proxy/blob/13536_fix_dnscmd_ptr/modules/dns_dnscmd/dns_dnscmd_main.rb#L28

instead of calling `dns_find(ip)` (which is broken anyway, as Resolv::DNS uses normal, not reversed ip addresses for #getname call, which leads to it always returning `false` here), you can call `Resolv::DNS#getaddresses(fqdn)`, and reverse each and compare to the one that was passed in. If none match, it's a collision. If one matches, you can return without adding a record (already there), otherwise (empty list came back) create a new PTR.

helge000:
> If none match, it's a collision. If one matches, you can return without adding a record (already there), otherwise (empty list came back) create a new PTR.

Good call, @witlessbird. I will have time for this in the evening.

dmitri-d:
Hmm, I realised that substituting a reverse lookup with a forward lookup isn't great when creating a ptr record -- address should take precedence over fqdn here (fqdn to ip mapping doesn't necessarily match ip to fqdn mapping). 

In this light, to convert a reversed ipv6:

```
"1.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.f.e.e.d.8.b.d.0.1.0.0.2.ip6.int"
    .slice(0..-9) # remove 'ip6.int'
    .split('.') 
    .reverse
    .each_slice(4) # break into groups of four nibbles each. Works, b/c reverse ipv6 is never shortened
    .each_with_object([]) {|word, address| address << word.join } # convert groups of four to strings 
    .join(":")
```

dimatha:
@helge000, sorry for my delayed reply. Indeed I'm running Foreman 1.10.1 and development foreman-proxy version with this patch applied.

helge000:
> substituting a reverse lookup with a forward lookup isn't great when creating a ptr record 

I was about to reply exactly that before I read your hole reply... For good reason I think you talked me out of using what you call a lightwight method ;). Anyway; sine foreman's side is not likely to change maybe the best we can do

helge000:
@dimatha: Once I've implemented @witlessbird's method, would you give it another try? I think I can push in the evening (CET)

dimatha:
@helge000 Sure, just drop a short note pls.

ekohl:
> instead of calling dns_find(ip) (which is broken anyway, as Resolv::DNS uses normal, not reversed ip addresses for #getname call, which leads to it always returning false here), you can call Resolv::DNS#getaddresses(fqdn), and reverse each and compare to the one that was passed in. If none match, it's a collision. If one matches, you can return without adding a record (already there), otherwise (empty list came back) create a new PTR.

https://github.com/ekohl/smart-proxy/commit/00e182a487180560a78858f0d52aaaff0fdfbc4e changes it to make it consistent.

dmitri-d:
@ekohl, your changes to `dns_find` aren't going to work for PTR record creation, see my comment.

helge000:
@witlessbird, I added a new method to convert PTRs back to IPs; though I slidely diverted from your example. 
Also, I went ahead and put them right into `dns_common.rb`, I reasoned they might be useful for other providers - hope that's alright with you? Mind compat 1.8 (see below)
Further, I did re-enabled the the check for existing records if we delete a ptr.

Build is expected to fail for 1.8 rubies because `#each_with_object` is apparently too new.
@dimatha, can you test this again? Please mind the above! It checks out with rubies >= 1.9

modules/dns_common/dns_common.rb
```diff
@@ -24,5 +26,19 @@ def dns_find key
...
+
+    def ip_to_ptr ip
+      IPAddr.new(ip).reverse rescue raise Proxy::Dns::Error.new("Not an IP address: '#{ip}'")
+    end
+
+    def ptr_to_ip ptr
+     if ptr =~ /\.in-addr\.arpa$/
+       ptr.split('.')[0..-3].reverse.join('.')
+     elsif ptr =~ /\.ip6\.arpa$/
+       ptr.split('.')[0..-3].reverse.each_slice(4).each_with_object([]) {|word, address| address << word.join }.join(":")
```
dmitri-d: You can use `inject([]) {|address, word| address << word.join}` instead of `each_with_object`, should work on 1.8.7

helge000: Alright! I could not get RVM with 1.8.7 working, have to leave this to CI

helge000:
The one test failure seems to be unrelated? Can I do the [test] also?

dmitri-d:
> The one test failure seems to be unrelated? Can I do the [test] also?

Does look like a fluke. You should be able to trigger tests -- you are a member of the foreman org. Keep in mind the check is every 30 mins or so. I'll leave the honours to you :P.

helge000:
> I'll leave the honours to you :P.

Hehe, my inline string already did the trick it seems. Good to know!

dimatha:
@helge000 hmm when I'm trying to provision a new host now - it fails on DNS creation and tries to rollback. Proxy log (DNS part only looks ok 200): http://pastebin.com/5iqtVqvv It returns back to the host creation page with no errors. At least it looks that both records are initially created, later it kills A record and keeps the PTR only. 

Foreman 1.10.1
Proxy development version + your patch
Foreman proxy node: win2008R2 Ruby 1.9
DNS/DHCP servers: W2k12R2

dimatha:
Update: Now I found why it fails (from the foreman log):

```
---
2016-02-23 00:42:59 [app] [I] Add DNS A record for rheltmpt08.example.com/10.10.10.115
2016-02-23 00:42:59 [app] [I] Add DNS PTR record for 10.10.10.115/rheltmpt08.example.com
2016-02-23 00:43:00 [app] [W] Failed to parse response:
 | Add PTR Record for 115.10.10.10.in-addr.arpa. at 10.in-addr.arpa
 | Command completed successfully.
 |
 |  -> 746: unexpected token at 'Add PTR Record for 115.10.10.10.in-addr.arpa. at 10.in-addr.arpa
 | Command completed successfully.
 |
 | '
2016-02-23 00:43:00 [app] [I] Delete the DNS A record for rheltmpt08.example.com/10.10.10.115
2016-02-23 00:43:00 [app] [I]
----
```

helge000:
Seems to me like a side effect to [execute return](https://github.com/theforeman/smart-proxy/blob/develop/modules/dns_dnscmd/dns_dnscmd_main.rb#L70), @witlessbird ?
If so, I would add another parameter to it `return_response=false`

Note to self: Rubies < 2.0 do not support [keyword arguments](https://github.com/ruby/ruby/blob/2adb9fd4a51b054543cc40555718bd09901a324f/ChangeLog#L9623)

dmitri-d:
https://github.com/helge000/smart-proxy/blob/13536_fix_dnscmd_ptr/modules/dns_dnscmd/dns_dnscmd_main.rb#L78 is the reason for errors on the foreman side. I missed that when it was introduced here: https://github.com/helge000/smart-proxy/commit/310433f729c5743b0c9f489e664470c0959857b8

Removing it will fix the issue.

modules/dns_common/dns_common.rb
```diff
@@ -24,5 +26,19 @@ def dns_find key
     rescue Resolv::ResolvError
       false
     end
+
+    def ip_to_ptr ip
```
dmitri-d: Don't see where this is being used?

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -16,33 +16,41 @@ def create_a_record(fqdn, ip)
...
+        zone = match_zone(ptr, enum_zones)
+        msg = "Added PTR entry #{ptr} => #{fqdn}"
+        cmd = "/RecordAdd #{zone} #{ptr}. PTR #{fqdn}."
+        execute(cmd, msg)
+      end
     end
 
     def remove_a_record(fqdn)
       ip = dns_find(fqdn)
       raise Proxy::Dns::NotFound.new("Cannot find DNS entry for #{fqdn}") unless ip
```
dmitri-d: No point in doing this check: ms dns returns a success when deleting a non-existing record, foreman ignores 'not found' errors on delete. This only slowes down the response.

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -16,33 +16,41 @@ def create_a_record(fqdn, ip)
...
       cmd = "/RecordDelete #{zone} #{fqdn}. A /f"
       execute(cmd, msg)
     end
 
-    # noop
-    def remove_ptr_record(ip)
-      raise Proxy::Dns::NotFound.new("Cannot find DNS entry for #{ip}") unless dns_find(ip)
-      true
+    def remove_ptr_record(ptr)
+      ip = ptr_to_ip ptr
```
dmitri-d: ditto.

modules/dns_dnscmd/dns_dnscmd_main.rb
helge000: We save one unnecessary execute?

modules/dns_common/dns_common.rb
helge000: Right, I thought it might prove useful in the future - I'll remove it

modules/dns_dnscmd/dns_dnscmd_main.rb
dmitri-d: dns_find uses a Resolv::DNS call, which is roughly as slow as the delete operation itself.

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -67,7 +73,7 @@ def execute cmd, msg=nil, error_only=false
         std_err.close unless std_in.nil?
       end
       report msg, response, error_only
-      response
+      response if return_result
```
helge000: hope this counts as fix, @witlessbird

modules/dns_dnscmd/dns_dnscmd_main.rb
ekohl: Returning a 404 on a delete is IMHO useless and I chose not to do it in smart_proxy_dns_powerdns either.

helge000: Sorry, I missed we where talking about A's here in the current commit

modules/dns_dnscmd/dns_dnscmd_main.rb
dmitri-d: nowhere in the code where it's used the result is needed. No need for conditionals (and additional tests), just don't return anything from the method.

modules/dns_dnscmd/dns_dnscmd_main.rb
dmitri-d: Nods, same issue in creation of both 'A' and 'PTR' records.

modules/dns_dnscmd/dns_dnscmd_main.rb
helge000: Not quite, I'll highly depend on it in [#enum_zones](https://github.com/theforeman/smart-proxy/blob/develop/modules/dns_dnscmd/dns_dnscmd_main.rb#L113)

modules/dns_dnscmd/dns_dnscmd_main.rb
helge000: > same issue in creation of both 'A' and 'PTR' records.

Hm? I'm quite certain creating a duplicate A / PTR fails. We can only remove it in the remove methods

dmitri-d: Apologies, I meant deletion of 'A' and 'PTR' records.

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -16,36 +16,42 @@ def create_a_record(fqdn, ip)
...
+        raise(Proxy::Dns::Collision, "#{ptr} is already used by #{found}") unless fqdn == found
+      else
+        zone = match_zone(ptr, enum_zones)
+        msg = "Added PTR entry #{ptr} => #{fqdn}"
+        cmd = "/RecordAdd #{zone} #{ptr}. PTR #{fqdn}."
+        execute(cmd, msg)
+      end
     end
 
     def remove_a_record(fqdn)
```
helge000, line 44: While I'm at it: I think this method is highly questionable; depending on dns_find() to get us exactly _the_ one IP... Anything I should do about it?

modules/dns_dnscmd/dns_dnscmd_main.rb
dmitri-d: Oops, of course. Still, I don't like the idea of a conditional return; let's explicitly return nils from  create_\* and remove_\* methods instead.

helge000: Good to know, Foreman expects `nil`. I think our test unit is getting better by the minute:  use `assert_nil`.

modules/dns_dnscmd/dns_dnscmd_main.rb
dmitri-d, line 44: You mean `dns_find`, right? There is indeed an issue with it -- if fqdn or ip has multiple mappings, `Resolv::DNS` will round-robin (I believe) return them. Changes in `dns_find` will affect all modules though, I'd rather do that in a dedicated PR.

helge000, line 44: Ok,  I'll keep this in mind.

modules/dns_dnscmd/dns_dnscmd_main.rb
dmitri-d: Foreman assumes that all calls return json. Empty body is ok though. And yep, good tests are good.

dmitri-d:
@helge000: looks good, will test the changes in a bit.

helge000:
@witlessbird How's the procedure for #369 - should I open a ref for it, since it is already in 1.11?

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -16,33 +16,42 @@ def create_a_record(fqdn, ip)
...
     end
 
-    # noop
-    def create_ptr_record(fqdn, ip)
-      found = dns_find(ip)
-      raise(Proxy::Dns::Collision, "#{ip} is already used by #{found}") if found && found != fqdn
-      true
+    def create_ptr_record(fqdn, ptr)
+      ip = ptr_to_ip ptr
+      if found = dns_find(ip)
```
dmitri-d: Sigh. This is broken, because `dns_find` reverses ip address before calling `Resolv::DNS.getname`, which then returns false and creation of a ptr record fails, event though it should've succeeded.

I would suggest:
- copy dns_find code into this module
- break it up into find_ptr and find_fqdn
- fix issues with multiple mappings

We'll move the changes into dns_common once the dust settles here.

helge000: > fix issues with multiple mappings

How should I best proceed with his one in case if A records?
- return first / last mapping to #create
- fail altogether

I rather fail in this case...

dmitri-d: The main logic should stay the same -- it's a conflict unless the ip that came back maps to fqdn. 

I think instead of dns_find, it should be something like: 

```
def a_record_conflicts?(ip, fqdn)
  addresses = resolv.getaddresses(fqdn)
  return false if addresses.empty?
  return false if addresses.size == 1 && addresses.first.to_s == ip
  true
end
```

modules/dns_common/dns_common.rb
```diff
@@ -24,5 +24,29 @@ def dns_find key
...
+     elsif ptr =~ /\.ip6\.arpa$/
+       ptr.split('.')[0..-3].reverse.each_slice(4).inject([]) {|address, word| address << word.join}.join(":")
+     else
+       raise Proxy::Dns::Error.new("Not a PTR record: '#{ptr}'")
+     end
+    end
+
+    def a_record_conflicts?(fqdn, ip)
+      addresses = resolver.getaddresses(fqdn)
+      addresses.each { |address| return address.to_s if address.to_s.casecmp(ip) == 0 }
```
helge000: I diverted form your example @witlessbird, because we have actually three cases to handle:
- no collision - `false`
- a collision - `true`
- overwrite - `name, ip` = nothing to do

I reason here that any found match is a valid overwrite case, this is later compared in create_a|ptr

Note, the `#casecmp` is needed because IPv6 addresses are valid written with upper and lower case. I this method will certainly need some tweaking with IPv6 address comparison. I will need to add the validate_ip method again... ;)

helge000:
@dimatha  - thanks btw for finding the above issue with foreman, saved us a bad regression. This should be fixed by now. You're welcome to try this again!

dimatha:
Thanks a lot for your efforts ! Looks good to me. Both records are in place once you build a machine. Removal process also works fine. No errors when you have DNS entries already created before you start building a machine.

modules/dns_common/dns_common.rb
dmitri-d: Trying to fit three outcomes into a boolean method results in weird semantics. You can return symbols or (-1, 0, 1) (pls. add a comment about the meaning of values in this case) instead. Probably should rename the method and drop '?' from it (denotes boolean methods) too.

Ditto for ptr_record_conflicts?.

helge000: Dejavu: I had to lookup casecmp yesterday because I did not know about it's returns; should have known you would suggest the same thing ;)

modules/dns_common/dns_common.rb
```diff
@@ -24,5 +27,43 @@ def dns_find key
...
+     end
+    end
+
+    # conflict methods return values:
+    # no conflict: -1; conflict: 1, conflict but record / ip matches: 0
+    def a_record_conflicts(fqdn, ip)
+      if ip_addr = validate_ip(ip)
+        addresses = resolver.getaddresses(fqdn)
+        return -1 if addresses.empty?
+        addresses.each { |address| return 0 if IPAddr.new(address.to_s).to_s.casecmp(ip_addr.to_s) == 0 }
```
helge000: This is manly for IPv6. I hope this is not overly complicated; can you think of a better way, @witlessbird ?

dmitri-d: You can compare IPAddr instances directly, no need to convert them to strings. Also, use #any? instead of each {... predicate_here }:

```
return 0 if addresses.any? {|a| IPAddr.new(a) == ip_addr}
```

modules/dns_common/dns_common.rb
```diff
@@ -24,5 +27,43 @@ def dns_find key
...
+       ptr.split('.')[0..-3].reverse.each_slice(4).inject([]) {|address, word| address << word.join}.join(":")
+     else
+       raise Proxy::Dns::Error.new("Not a PTR record: '#{ptr}'")
+     end
+    end
+
+    # conflict methods return values:
+    # no conflict: -1; conflict: 1, conflict but record / ip matches: 0
+    def a_record_conflicts(fqdn, ip)
+      if ip_addr = validate_ip(ip)
```
dmitri-d: Not a good name for this method: I would expect validate method to return a boolean, while it returns an instance of IPAddr if the parameter is valid. Maybe rename it to to_ipaddr or something similar?

modules/dns_common/dns_common.rb
```diff
@@ -24,5 +27,43 @@ def dns_find key
...
+      else
+        raise Proxy::Dns::Error.new("Not an IP Address: '#{ip}'")
+      end
+    end
+
+    def ptr_record_conflicts(fqdn, ip)
+      if ip_addr = to_ipaddress(ip)
+        names = resolver.getnames(ip_addr.to_s)
+        return -1 if names.empty?
+        return 0 if names.any? {|n| n.to_s.casecmp(fqdn) == 0}
```
dmitri-d, line 58: hostnames are case-insensitive ('HOSTname' == 'hostname'), better to lowercase both sides and compare the results.

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -13,36 +12,45 @@ def initialize(a_server = nil, a_ttl = nil)
     end
 
     def create_a_record(fqdn, ip)
-      if found = dns_find(fqdn)
-        raise(Proxy::Dns::Collision, "#{fqdn} is already used by #{found}") if found != ip
+      conflict = a_record_conflicts(fqdn, ip) #returns -1, 0, 1
+      if conflict >= 0
```
dmitri-d: Split this into two checks, you are doing another one anyway when raising `Proxy::Dns::Collision`

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -13,36 +12,45 @@ def initialize(a_server = nil, a_ttl = nil)
...
     end
 
-    # noop
-    def create_ptr_record(fqdn, ip)
-      found = dns_find(ip)
-      raise(Proxy::Dns::Collision, "#{ip} is already used by #{found}") if found && found != fqdn
-      true
+    def create_ptr_record(fqdn, ptr)
+      conflict = ptr_record_conflicts(fqdn, ptr_to_ip(ptr)) #returns -1, 0, 1
+      if conflict >= 0
```
dmitri-d: ditto.

dmitri-d:
looks good other than a couple of minor things. Testing it now.

modules/dns_common/dns_common.rb
helge000, line 58: Was my first thought as well. But here comes rubocop, the reason I was using #casecmp in the fist place:

``` ruby
Offenses:

modules/dns_common/dns_common.rb:58:37: C: Use casecmp instead of downcase ==.
        return 0 if names.any? {|n| n.to_s.downcase == fqdn.to_s.downcase}
                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

modules/dns_dnscmd/dns_dnscmd_main.rb
helge000: would case be alright:

``` ruby
case collision
  when 1 then raise 
  when 0 then return nil
  else
    #..
```

modules/dns_common/dns_common.rb
dmitri-d, line 58: Oops. My apologies, should check docs before making a comment. `casecmp` is cool, you and rubocop win!

modules/dns_dnscmd/dns_dnscmd_main.rb
dmitri-d: sounds good to me.

dmitri-d: I'd get rid of the local var and use the result of `a_record_conflicts(fqdn, ip)` expression in case directly.

helge000: Yea, this was just for the example above, to save me some writing. Commit coming right up!

dmitri-d:
Hit a strange issue when testing creation of IPv6 PTR records -- ruby resolver wouldn't return host names, even though DNS server was sending replies. Resolver worked from ruby running on linux too.

Merged in at 450e0b6, as this very much looks like an issue with my setup. Thanks a bunch, @helge000!

helge000:
> Hit a strange issue when testing creation of IPv6 PTR records -- ruby resolver wouldn't return host names, even though DNS server was sending replies. Resolver worked from ruby running on linux too.

@witlessbird, I am hitting the same issue. nslookup returns name, `#getnames` does come up empty. IPv4 reverse works. Tested with rubies 1.9 .. 2.2.1. This looks more like an issue with rubies stub resolver. Could not find anything mention that so far...

helge000:
Btw, how do we handle the regression? I assume you keep track of it?

PR #380, opened 2016-02-17
Fixes #13747 - Allow configuration of allowed SSL cipher suites

tbrisker:


shlomizadok:
:+1: 
rubocop failure seems unrelated

dmitri-d:
Any information on which suites need to be removed to pass audits? We could make this list tighter (drop AES128-SHA and AES256-SHA for example)... I would like to avoid exposing the list of ciphers to users as much as possible -- figuring out the cipher list is pretty involved (see discussion in https://github.com/theforeman/smart-proxy/pull/351).

@brandonweeks: any thoughts on this?

brandonweeks:
From an industry perspective, these types of configurations are very helpful. Our audit requirements are often arbitrary and fast changing. As an example when Logjam/CVE-2015-4000 was released our security team started flagging services that were using improperly configured DH groups. Either I could have waited for someone to make the change upstream and for it to be released, or I could have just disabled cipher suites that used DH key exchange using the option this PR implements.

Now with all of that said, configuring the particular stack smart-proxy uses is non-trivial and I expect many admins/engineers will not do it correctly. There are a number of caveats and upstream issues I discovered during #351 (some of which our security team have gotten fixed in Ruby core!). 

A better long term solution would be to move away from using WEBrick as the web server for smart-proxy. But I might be getting ahead of myself here...

tbrisker:
@witlessbird @brandonweeks How do you suggest we proceed here? I think giving users the ability to change the suites without having to change the source/wait for us to do it is important - as Brandon mentioned, often a vulnerability is discovered that requires immediate changes to the allowed cyphers. I think defaulting to the hard coded list provides a good solution for those who don't know what they are doing. Maybe we should add a warning to the config file, or make it a "hidden" option (i.e. allow it to be set but not provide an example in the example yml)?

dmitri-d:
I have a few concerns with the proposed approach:
- How do we handle updates? This override will always take precedence, even if the ciphers have been updated.
- How do we support this? Even the names of ciphers in ruby differ from those in OpenSSL. Different combinations of ruby interpreter and OS will have different available ciphers.

And finally: If someone has enough understanding of ruby + openssl,  they can change the source code directly to suit their needs...

tbrisker:
@witlessbird Ok, what do you think about this option: change the setting to disabled ciphers - this way if a certain cipher needs disabling due to a vulnerability or security policy, it can easily be done in the setting without touching the hard coded values. This way, any updates will still be effective while the disabled suites will remain disabled.

brandonweeks:
@tbrisker I'm fully in support of that idea. It might be worth waiting until the outcome of #396 is decided though.

tbrisker:
@witlessbird @brandonweeks changed to disabled ciphers.

lib/webrick-patch.rb
```diff
@@ -14,7 +14,7 @@ def setup_ssl_context(config) # :nodoc:
       end
       ctx = OpenSSL::SSL::SSLContext.new
       ctx.set_params
-      ctx.ciphers = CIPHERS
+      ctx.ciphers = (CIPHERS.split(':') - config.fetch(:SSLDisabledCiphers, '').split(':')).join(':')
```
dmitri-d: Have you committed all the changes? I don't see how values from `ssl_disabled_ciphers` get into `SSLDisabledCiphers`. I would like to avoid introducing custom webrick config options however; you can access global settings directly from here. Additionally, you can set a default value for `ssl_disabled_ciphers` to an empty array, which will allow you to dispense with defensive code around fetching it.

dmitri-d:
I think the docs need to be updated with the list of enabled ciphers (from here: https://github.com/tbrisker/smart-proxy/blob/13747/lib/webrick-patch.rb#L3). I also wonder if the docs should contain a link to a mapping of tls/ssl cipher names to openssl cipher names (this one? https://www.openssl.org/docs/manmaster/apps/ciphers.html. Look for "CIPHER SUITE NAMES").

tbrisker:
@witlessbird Thanks for your comment, still not quite used to how the proxy works :smile: 
Now changed to use the `SETTINGS` directly and add default for empty array. I agree we should update the docs to describe this, will do once it's merged. I don't think we have to include the complete mapping, be we should point to it ("please note - we use OpenSSL cipher names yada yada, link to mapping")

config/settings.yml.example
```diff
@@ -6,6 +6,9 @@
 #:ssl_certificate: ssl/certs/fqdn.pem
 #:ssl_ca_file: ssl/certs/ca.pem
 #:ssl_private_key: ssl/private_keys/fqdn.key
+# If you need to disable certain cipher suits use this option.
+# Only use this option if you know what you are doing!
```
dmitri-d: This isn't informative. If you think you need something along those lines, I would suggest a url to openssl ciphers...

dmitri-d:
Looks good, although rubocop isn't happy.

config/settings.yml.example
tbrisker: replaced the warning with a link to the ssl docs.

tbrisker:
@witlessbird Fixed your comment, rubocop was complaining about the line being too long so I broke it into several. Please re-review.

dmitri-d:
Merged in b73b71a. Thanks, @tbrisker!

tbrisker:
docs pr: https://github.com/theforeman/theforeman.org/pull/609

PR #382, opened 2016-02-17
Fixes #13761 - added leases support to virsh provider

lzap:
This patch also splits configuration into two individual module configuration
files. This allows to use two different libvirt network for DHCP and DNS.

Note this needs most recent libvirt that stores leases now in JSON format
rather than plaintext:

```
cat /var/lib/libvirt/dnsmasq/virbr0.status
[
    {
        "ip-address": "192.168.122.68",
        "mac-address": "52:54:00:13:05:13",
        "client-id": "01:52:54:00:13:05:13",
        "expiry-time": 1455723598
    }
]
```

This takes advantage of that and directly parses the JSON. The filename is
configurable. Example:

```
curl -sk http://localhost:8448/dhcp/192.168.122.0 | json_reformat
{
    "reservations": [
        {
            "hostname": "test01.local.lan",
            "ip": "192.168.122.14",
            "mac": "52:54:00:01:4a:01"
        }
    ],
    "leases": [
        {
            "ip": "192.168.122.68",
            "mac": "52:54:00:13:05:13",
            "starts": "2016-02-17 14:41:51 UTC",
            "ends": "2016-02-17 15:39:58 UTC",
            "state": "active"
        }
    ]
}
```

There is no lease start stored in the JSON data, therefore the module puts
current time there.

### DIFF
--- a/config/settings.d/dhcp_virsh.yml.example
+++ b/config/settings.d/dhcp_virsh.yml.example
@@ -2,7 +2,9 @@
 #
 # Configuration file for libvirtd-specific 'virsh' dhcp provider
 #
-# There's currently no configuration options for this provider.
-# Virsh network name is a global parameter that can be set
-# in the main settings.yml file in 'virsh_network' parameter.
-#
+
+# Libvirt network to use
+#:network: default
+
+# Leases file (managed via /usr/libexec/libvirt_leaseshelper)
+#:leases: /var/lib/libvirt/dnsmasq/virbr0.status
--- a/test/dhcp_virsh/virsh_provider_test.rb
+++ b/test/dhcp_virsh/virsh_provider_test.rb
@@ -7,7 +7,7 @@ class VirshProviderTest < Test::Unit::TestCase
   def setup
     @service = Proxy::DHCP::SubnetService.new
     @subnet_store = @service.subnets = Proxy::MemoryStore.new
-    @server = ::Proxy::DHCP::Virsh::Provider.new.initialize_for_testing(:virsh_network => 'default',
+    @server = ::Proxy::DHCP::Virsh::Provider.new.initialize_for_testing(:network => 'default',
                                                                         :name => "127.0.0.1", :service => @service)
     @dump_xml = <<EODUMPXML
 <network>
@@ -29,15 +29,24 @@ class VirshProviderTest < Test::Unit::TestCase
   </ip>
 </network>
 EODUMPXML
+
+    @json_leases = [{
+        "ip-address" => "192.168.122.22",
+        "mac-address" => "52:54:00:13:05:12",
+        "client-id" => "01:52:54:00:13:05:12",
+        "expiry-time" => 1_455_723_598
+    }]
   end
 
-  class DhcpVirshProviderForTesting <  Proxy::DHCP::Virsh::Provider
-    attr_reader :network
+  def test_default_settings
+    assert_equal 'default', Proxy::DHCP::Virsh::Provider.new.network
+    assert_equal '/var/lib/libvirt/dnsmasq/virbr0.status', Proxy::DHCP::Virsh::Provider.new.leases
   end
 
   def test_virsh_provider_initialization
-    Proxy::SETTINGS.stubs(:virsh_network).returns('another_one')
-    assert_equal 'another_one', DhcpVirshProviderForTesting.new.network
+    ::Proxy::DHCP::Virsh::Plugin.load_test_settings(:network => 'some_network', :leases => 'leases.txt')
+    assert_equal 'some_network', Proxy::DHCP::Virsh::Provider.new.network
+    assert_equal 'leases.txt', Proxy::DHCP::Virsh::Provider.new.leases
   end
 
   def test_should_load_subnets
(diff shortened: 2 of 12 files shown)


test/dns_virsh/dns_virsh_test.rb
```diff
@@ -4,9 +4,8 @@
 
 class DnsVirshTest < Test::Unit::TestCase
   def test_virsh_provider_initialization
-    ::Proxy::SETTINGS.stubs(:virsh_network).returns('some_network')
-    server = Proxy::Dns::Virsh::Record.new
-    assert_equal "some_network", server.network
+    ::Proxy::Dns::Virsh::Plugin.settings.stubs(:network).returns('some_network')
```
dmitri-d: Please use `::Proxy::Dns::Virsh::Plugin.load_test_settings(:network => 'some_network')` instead. Could you add a test to verify default settings too?

test/dhcp_virsh/virsh_provider_test.rb
```diff
@@ -29,15 +29,18 @@ def setup
...
+        "mac-address" => "52:54:00:13:05:12",
+        "client-id" => "01:52:54:00:13:05:12",
+        "expiry-time" => 1455723598
+    }]
   end
 
   def test_virsh_provider_initialization
-    Proxy::SETTINGS.stubs(:virsh_network).returns('another_one')
-    assert_equal 'another_one', DhcpVirshProviderForTesting.new.network
+    ::Proxy::DHCP::Virsh::Plugin.settings.stubs(:network).returns('another_one')
```
dmitri-d: see my comment re: use of 'load_test_settings' instead.

dmitri-d: Also, add a test for default settings please.

test/dns_virsh/dns_virsh_test.rb
lzap: Sure, although I think it's useless. Changing a default value is indeed intrusive and it's quite obvious (without a test).

lzap:
Fixed all.

config/settings.d/dhcp_virsh.yml.example
```diff
@@ -2,7 +2,9 @@
 #
 # Configuration file for libvirtd-specific 'virsh' dhcp provider
 #
-# There's currently no configuration options for this provider.
-# Virsh network name is a global parameter that can be set
-# in the main settings.yml file in 'virsh_network' parameter.
-#
+
+# Libvirt network to use
+#:network: default
```
dmitri-d, line 7: Just realised: as you moved the 'network' setting, you'll need to add setting migrations for both virsh_dhcp and virsh_dns modules. There are a few examples in https://github.com/theforeman/smart-proxy/blob/develop/extra/migrations.

dmitri-d:
Also, rubocop isn't happy.

lzap:
Ok fixed Rubocop and added migration that gets the network setting and creates the YAML files. Not sure if I am supposed to store the main YAML file without the option.

```
[lzap@lzapx smart-proxy]$ ruby extra/migrate_settings.rb 
using:
  config file: /home/lzap/work/smart-proxy/config/settings.yml,
  modules config dir: /home/lzap/work/smart-proxy/config/settings.d,
  working dir: /home/lzap/work/smart-proxy/tmp,
  migrations dir: /home/lzap/work/smart-proxy/extra/migrations,
  migrations state path: /home/lzap/work/smart-proxy/config/migration_state

Running migrations...
migrate_monolithic_config
migrate_dns_settings
migrate_dhcp_settings
migrate_virsh_settings_config
Writing result to /home/lzap/work/smart-proxy/tmp/migrate_virsh_settings_config/dst/settings.d/dhcp_virsh.yml
Writing result to /home/lzap/work/smart-proxy/tmp/migrate_virsh_settings_config/dst/settings.d/dns_virsh.yml
[lzap@lzapx smart-proxy]$ cat /home/lzap/work/smart-proxy/tmp/migrate_virsh_settings_config/dst/settings.d/*_virsh.yml
---
:network: default
:leases: "/var/lib/libvirt/dnsmasq/virbr0.status"
---
:network: default
```

dmitri-d:
I wonder if instead of parsing the leases file, we should use ruby-libvirt API to query for leases: https://www.libvirt.org/ruby/api/Libvirt/Network.html#method-i-dhcp_leases. I'd prefer not to parse 3rd-party files if at all possible. This would also have a side-effect of simplifying the configuration a little bit, as there would be no need to specify the location of the leases file. 

There seems to be an issue in the latest release but it has been fixed in the master: http://libvirt.org/git/?p=ruby-libvirt.git;a=commit;h=c2d4192ebf28b8030b753b715a72f0cdf725d313.

Thoughts?

@lzap --  also, could you add tests for migrations pls?

lzap:
> I wonder if instead of parsing the leases file, we should use ruby-libvirt API to query for leases: https://www.libvirt.org/ruby/api/Libvirt/Network.html#method-i-dhcp_leases. I'd prefer not to parse 3rd-party files if at all possible. This would also have a side-effect of simplifying the configuration a little bit, as there would be no need to specify the location of the leases file. 

That's totally valid point, but the format is dictated by libvirt and
not dnsmasq. It's a thing that is being maintained by
`/usr/libexec/libvirt_leaseshelper` script which is part of libvirt.

Our plugin is based on CLI interface and we'd need to parse output of:

```
sudo virsh net-dhcp-leases default
 Expiry Time          MAC address        Protocol  IP address
 Hostname        Client ID or DUID
 -------------------------------------------------------------------------------------------------------------------
```

I don't think this is worth the effort, I will be implementing proper
dnsmasq plugin soon: http://projects.theforeman.org/issues/13806 which
gives new flexibility of using dnsmasq directly.

Remember this plugin is named virsh and it was done as dirty one for
development. We can start a new one called libvirt with proper Ruby
bindings perhaps, but I'd like to avoid refactoring this one.

Using Ruby binding is indeed interesting, I will perhaps prototype this
next week. Let's hold on with this patch for now.

## 

Later,
 Lukas #lzap Zapletal

lzap:
Ok I will actually bring new PR which will refactor virsh providers into new providers called `libvirt` (finally a proper name). I will include this feature there (DHCP leases parsing) as well, therefore I will reuse the ticket.

PR #383, opened 2016-02-18
refs #13638 - fix .empty? check in pkg:generate_source task

domcleal:
Caught pretty quickly by Jenkins when building the smart proxy packages.


domcleal:
PR test run manually: http://ci.theforeman.org/job/test_proxy_develop_pull_request/1062/

I've also added this rake task to the PR tests so we catch it in future.

dmitri-d:
Merged in at 6ca411c. Thanks, @domcleal!

PR #386, opened 2016-02-29
Fixes #13942: Backport {a,ptr}_record_conflicts functions

ekohl:
This backports from 450e0b694ead84961db9c1ffd4256919423484f6 but leaves out all the dnscmd changes. It also correctly ignores IPv6 addresses in a_record_conflicts. That's taken from 05d87379a132968d045258249c5e765e6aa8824a which hasn't made it to develop yet.

### DIFF
--- a/modules/dns_common/dns_common.rb
+++ b/modules/dns_common/dns_common.rb
@@ -1,3 +1,6 @@
+require 'resolv'
+require 'ipaddr'
+
 module Proxy::Dns
   class Error < RuntimeError; end
   class NotFound < RuntimeError; end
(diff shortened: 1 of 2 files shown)


dmitri-d:
@ekohl: I assume you need this for power_dns provider?

ekohl:
@witlessbird Yes, I wrote https://github.com/theforeman/smart_proxy_dns_powerdns/commit/06ce7a2b838d0d7c1eef5e9bc1ec7c64793b002a assuming it was already in 1.11.

modules/dns_common/dns_common.rb
```diff
@@ -24,5 +27,43 @@ def dns_find key
...
+     else
+       raise Proxy::Dns::Error.new("Not a PTR record: '#{ptr}'")
+     end
+    end
+
+    # conflict methods return values:
+    # no conflict: -1; conflict: 1, conflict but record / ip matches: 0
+    def a_record_conflicts(fqdn, ip)
+      if ip_addr = to_ipaddress(ip)
+        addresses = resolver.getaddresses(fqdn).reject { |a| a =~ Resolv::IPv6::Regex }
```
helge000: Great, I could also backport #379

ekohl: I think I'm going to rewrite this to collect IPv4. First I'll submit a PR to get it into develop. Then I'll update this PR.

helge000: I suppose you're referring using [`#Resolv::DNS::Resource`](http://ruby-doc.org/stdlib-1.9.3/libdoc/resolv/rdoc/Resolv/DNS/Resource.html)? Performance wise, the regex might be faster..

ekohl: No, I'm meaning `resolver.getaddresses(fqdn).collect { |a| a =~ Resolv::IPv4::Regex }`.

helge000: Right, this is better

ekohl: With this you mean what I'm proposing rather than the current code in the commit, correct?

helge000: Sorry, should have been more clear. Using `#collent` with IPv4 regex makes more sense to me also; though I currently see no issues with your code since `#getaddesses` only gets us A or AAAA records.

ekohl: I agree it makes no difference for results so it's minor.

mmoll:
what's the status here? we're already at RC3...

helge000:
@mmoll: I could base a PR on this commit handling dnscmd

dmitri-d:
This PR has additional functionality and changes that are not present in the develop branch. Could you make a PR against develop?

domcleal:
> what's the status here? we're already at RC3...

Please avoid merging this between RC3 and .0, which will be in the next week or two if all goes well.  It'd be fine for .1.

ekohl:
@domcleal does that mean develop hasn't been branched to 1.11-stable?

domcleal:
It has, but I meant this PR specifically which is against 1.11-stable.  The ones against develop can go in at any point, but I'd like to avoid merging new bits to 1.11-stable between the last RC and .0.

ekohl:
Oh right, that makes sense.

ekohl:
@witlessbird and this functionality is part of https://github.com/theforeman/smart-proxy/pull/371.

mmoll:
what's the status here?

ekohl:
Rebased. Since #371 was merged, this functionality is now in develop and I think it can be merged.

dmitri-d:
@ekohl: you'll need to rebase again, tests are failing on 1.8.7 due to the new version of webmock not supporting that platform (has been fixed). 

Also, is this still relevant? I reckon we are somewhere mid-way through the release cycle (@domcleal, what do you think?)...

domcleal:
> Also, is this still relevant? I reckon we are somewhere mid-way through the release cycle (@domcleal, what do you think?)...

I'd normally say it should wait, but this looks to be very low risk as it's simply a set of new functions, and has test coverage - no behaviour changes that we need to worry about.  I'm OK with it.

ekohl:
> @ekohl: you'll need to rebase again, tests are failing on 1.8.7 due to the new version of webmock not supporting that platform (has been fixed).

It's been fixed in develop, not on 1.11-stable.

domcleal:
Good point, I've done the cherry pick to 1.11-stable now.

ekohl:
Rebased

dmitri-d:
Merged as 8b6096b. Thanks, @ekohl.

PR #387, opened 2016-02-29
Fixes #13947 - removed html views for dhcp and root modules

dmitri-d:


domcleal:
Merged as 2578671, thanks @witlessbird.

PR #388, opened 2016-03-02
refs #13421 - fix regression: explicitly return nil

helge000:
Fixes forman witch expects JSON response. See [discussion](https://github.com/theforeman/smart-proxy/pull/379#issuecomment-187442932) in #379


dmitri-d:
merged at 0d35a44. Thanks, @helge000!

PR #390, opened 2016-03-09
fixes #14618 - Introduced configurable initialization of modules and providers

dmitri-d:


dmitri-d:
Work in progress, although most functionality has been added. 

This adds functionality to customize/override configuration and initialization of plugins and providers. 
- ability to update/add computed settings (as opposed to purely config-file based settings).
- ability to add and execute custom validators
- ability to start background services during plugin initialization

Configuration/initialization logic has been moved to https://github.com/theforeman/smart-proxy/pull/390/files#diff-11cebac8a6625c7bb985371a365a48e3R163 and https://github.com/theforeman/smart-proxy/pull/390/files#diff-11cebac8a6625c7bb985371a365a48e3R196. 

Plugin dsl adds support for overriding the classes above, but also provides ability to individually override validator, dependency, dependency injection wirings loader classes. 

Some questions that need answers:
- I noticed that one of the common issues is referring to wrong plugin class when retrieving provider settings. I'd like to merge main plugin and provider settings (per provider).
- Right now all runtime dependencies (think a bunch of `require`s that used to go in the `after_activation` block) are loaded before validators are being executed (to load custom validators classes). Instead, validator loader can perform that, which would allow us to delay loading of classes until after we know the configuration is valid.

Deep dive, examples, and docs are coming. I'll be happy to do a code walk through to help with the review(s). 

@ekohl, @helge000, @GregSutcliffe, @domcleal: thoughts?

ekohl:
It would help me if I saw what would change for me as a plugin author because it's hard for me to derive the changes from this diff.

dmitri-d:
I will port https://github.com/theforeman/smart-proxy/pull/376 PR to highlight the changes.

dmitri-d:
[test]

dmitri-d:
@ekohl: pls see https://github.com/theforeman/smart-proxy/pull/406 that demonstrates most of the changes in this PR.

dmitri-d:
I'm aware of build failures under 1.8.7 and am looking into them.

lib/proxy/pluggable.rb
```diff
@@ -1,131 +1,137 @@
 module Proxy::Pluggable
-  def self.included(base)
-    base.send(:include, InstanceMethods)
-    base.send(:extend, ClassMethods)
+  attr_reader :plugin_name, :version, :after_activation_blk, :class_loader
+
+  # Methods below define DSL for defining plugins
+  def after_activation(&blk)
```
domcleal, line 5: Is this method meant to be deprecated?  There are references in the loading code that it's "legacy".

lib/proxy/pluggable.rb
```diff
@@ -1,131 +1,137 @@
...
-      self.class.plugin_name
-    end
+  def requires(plugin_name, version_spec)
+    self.dependencies << ::Proxy::Dependency.new(plugin_name, cleanup_version(version_spec))
+  end
 
-    def version
-      self.class.version
-    end
+  def bundler_group(name)
```
domcleal, line 13: missing test coverage?

lib/proxy/plugin_initializer.rb
```diff
@@ -1,36 +1,287 @@
...
+    logger.error(a_message, a_backtrace)
+    members.each do |m|
+      ::Proxy::LogBuffer::Buffer.instance.failed_module(m.plugin_name, a_message)
+    end
   end
 
-  def instantiate_plugins(loaded_plugins)
-    loaded_plugins.dup.map {|p| p.update(:instance => p[:class].new)}
+  def notify_group_about_failure
+    #providers.each {|p| p.stop }
```
domcleal: commented code - this should be implemented properly, or removed

lib/proxy/settings.rb
```diff
@@ -20,4 +20,9 @@ def self.load_plugin_settings(defaults, settings_file, settings_directory = nil)
     end
     ::Proxy::Settings::Plugin.new(defaults, settings)
   end
+
+  def self.read_settings_file(settings_file, settings_directory = nil)
+    YAML.load(File.read(File.join(settings_directory || ::Proxy::SETTINGS.settings_directory, settings_file))) || {}
+  end
 end
+
```
domcleal: Nitpick, stray line.

lib/smart_proxy_main.rb
```diff
@@ -9,20 +9,22 @@
 require 'webrick/https'
 require 'daemon'
 
-require 'checks'
 require 'proxy/log'
 require 'proxy/settings'
 require 'proxy/settings/plugin'
 require 'proxy/settings/global'
 require 'proxy/dependency_injection'
+require 'proxy/default_di_wirings'
```
domcleal: Perhaps require this only from lib/proxy/plugin_initializer.rb, which is where it's used, rather than at the app level.  Probably the same for default plugin validators?  (Saves also doing it in the tests.)

modules/dhcp_isc/dhcp_isc_main.rb
```diff
@@ -5,25 +5,15 @@ module Proxy::DHCP::ISC
   class Provider < ::Proxy::DHCP::Server
     include Proxy::Util
 
-    def initialize
-      super(Proxy::DhcpPlugin.settings.server)
-      @config_file = Proxy::DHCP::ISC::Plugin.settings.config
-      @leases_file = Proxy::DHCP::ISC::Plugin.settings.leases
+    def initialize(provider_settings = OpenStruct.new, main_settings = OpenStruct.new, service = nil)
```
domcleal: Why is only this class refactored in this way?  Perhaps leave it for now, and do them all together?

lib/proxy/plugin.rb
```diff
@@ -77,28 +19,38 @@ def find_provider(provider_name)
 #  requires :foreman_proxy, ">= 1.5.develop"
 #  requires :another_plugin, "~> 1.3.0"
 #  default_settings :first => 'first', :second => 'second'
+#
+# load_classes 'a/b', 'a/c', ...
```
domcleal: Nitpick, indentation is out - should match above+below.

domcleal:
1) The changes to the public plugin API are difficult to understand, particularly the DI terminology and to understand how the new APIs are used. This would be a good time to improve the documentation, both to help me review and for future use.

The wiki page would be a good start, provided it has a warning to say that this is >= 1.12 only.  I'm also quite concerned to understand which APIs (if any) are changing behaviour in this patch, are being deprecated, or are being removed.

dmitri-d:
The docs are in plans for the next week. I would suggest taking a look at https://github.com/theforeman/smart-proxy/pull/406 to see how plugins can use the new functionality (puppet_proxy_legacy module is probably one of the better examples).

lib/proxy/pluggable.rb
dmitri-d, line 5: Not at the moment, although it can be deprecated if I add support for blocks as means of loading classes, specifying di wiring, etc, all of which require class(es) atm.

modules/dhcp_isc/dhcp_isc_main.rb
dmitri-d: I cannot remember anymore, probably because I used that for testing. There's a PR for dhcp changes: https://github.com/theforeman/smart-proxy/pull/409.

dmitri-d:
Done with the first pass at the docs: http://projects.theforeman.org/projects/foreman/wiki/How_to_Create_a_Smart-Proxy_Plugin

dmitri-d:
Rebased, fixed issues pointed out above (other than changes to dhcp_isc module). Also:
- ::Proxy::Pluggable.load_classes now accepts a block as a parameter
- DI wirings no longer use block parameter (as it was redundant)
- renamed ::Proxy::Pluggable.load_runtime_configuration to ::Proxy::Pluggable.load_programmable_settings (the former was too ambiguous)

>  I'm also quite concerned to understand which APIs (if any) are changing behaviour in this patch, are being deprecated, or are being removed.

No existing APIs have been removed, deprecated, or changed behaviour (or not intentionally).

lib/proxy/pluggable.rb
```diff
@@ -1,131 +1,129 @@
...
   end
 
-  module ClassMethods
-    attr_reader :plugin_name, :version, :after_activation_blk, :bundler_group_name
+  def load_dependency_injection_wirings(class_name_to_use)
+    @di_wirings_loader_class = class_name_to_use
+  end
 
-    # Methods below define DSL for defining plugins
+  def load_programmable_settings(class_name_to_use)
```
domcleal: Can this accept a block as well, in the same way as load_classes?

Or even only accept a block?  It seems needlessly complex to create and instantiate classes just to call a method in both cases.

lib/proxy/pluggable.rb
```diff
@@ -1,131 +1,129 @@
...
-    def loading_failed(message)
-      raise ::Proxy::PluginLoadingAborted, message
-    end
+  def load_validators(class_name_to_use)
+    @validator_loader_class = class_name_to_use
   end
 
-  module ClassMethods
-    attr_reader :plugin_name, :version, :after_activation_blk, :bundler_group_name
+  def load_dependency_injection_wirings(class_name_to_use)
```
domcleal: ditto, a block here might be simpler?

lib/proxy/pluggable.rb
```diff
@@ -1,131 +1,129 @@
...
-      end
-    end
+    settings.each {|setting| validations << {:name => validator_name, :predicate => predicate, :args => validator_args, :setting => setting} }
+  end
 
-    def loading_failed(message)
-      raise ::Proxy::PluginLoadingAborted, message
-    end
+  def load_validators(class_name_to_use)
+    @validator_loader_class = class_name_to_use
```
domcleal: and ditto, a block or even just accept a hash of validator names to classes.  I don't see the need for the extra class, instantiation and method.

lib/proxy/pluggable.rb
domcleal, line 13: Seems to be fixed at https://github.com/theforeman/smart-proxy/pull/390/files#diff-192cbec985fe34a2e50b9569df0975f4R43

modules/dhcp_isc/dhcp_isc_main.rb
domcleal: Okay, can you revert it for now please?

lib/proxy/pluggable.rb
dmitri-d: While having a class is much heavier than passing in a block, I prefer the former as it's a) testable (or easier to test), b) doesn't pollute plugin class with configuration logic/executable code (this is the nastiest example I have, there are other cases that aren't as bad, but still: https://github.com/theforeman/smart-proxy/pull/406/commits/5e1b066e520f29e7d5b9dee1dfb1466f350ff261#diff-9b357977669579ae92d74c22420de796R7)

lib/proxy/pluggable.rb
dmitri-d: Similar reasoning to the above, and an example to go along: https://github.com/theforeman/smart-proxy/pull/406/commits/5e1b066e520f29e7d5b9dee1dfb1466f350ff261#diff-9b357977669579ae92d74c22420de796R53

lib/proxy/pluggable.rb
dmitri-d: This would work, considering that validator classes can be loaded in load_classes or even directly.

lib/proxy/pluggable.rb
dmitri-d, line 13: I added a few tests, could one of those...

lib/proxy/pluggable.rb
domcleal: Ah I see, that makes more sense with the complexity of that method.  I wonder if keeping the block as an option's best for consistency, allowing either to be used.

Is there any reason to require a class (and then instantiate it) over just passing something that responds to the correct method, so it'd be a class method or pass an instance of the class?

dmitri-d: I can add support for blocks in load_programmable_settings and load_dependency_injection_wirings methods, although I'm concerned a bit that those will be abused.

Not sure I understand you correctly -- how would you specify an instance of a class (or anything else other than a class) in these spots?

domcleal: > Not sure I understand you correctly -- how would you specify an instance of a class (or anything else other than a class) in these spots?

Yeah, it's not possible now because `.new` is called.  I was trying to understand why that was needed, when this could just accept something that responds to `#load_programmable_settings` and not call `.new` first.  It's just the instantiation of the class seems like a peculiar interface in Ruby, where just responding to methods is usually enough to implement an interface.

Rather than `@programmable_settings.new.load_programmable_settings`, I'd expect `@programmable_settings.load_programmable_settings`, then either change `load_programmable_settings` to a class method or pass an instance of the Proxy::Example::RuntimeConfiguration in here.

lzap:
When you are done with review, let me know. I would like to test this patch with my plugin that has a background service. Until now, there was no way to properly stop it.

dmitri-d:
- added support for blocks in `::Proxy::Pluggable#load_classes`, `::Proxy::Pluggable#load_programmable_settings`, and `::Proxy::Pluggable#load_dependency_injection_wirings`
- `::Proxy::Pluggable#load_validators` now expects a map of validator names to classses
- `::Proxy::Pluggable#class_loader`, `::Proxy::Pluggable#programmable_settings`, and `::Proxy::Pluggable#di_wirings` now return an instance of a class/block wrapper
- reverted changes in `dhcp_isc/dhcp_isc_main.rb` and related tests

lib/proxy/error.rb
```diff
@@ -1,5 +1,12 @@
 module Proxy::Error
-  class BadRequest < StandardError; end
-  class Unauthorized < StandardError; end
+  class HttpError < StandardError
```
domcleal: Is this related?  I can't see where it's used.

lib/proxy/pluggable.rb
```diff
@@ -1,131 +1,170 @@
...
+    else
+      @class_loader.new
     end
+  end
 
-    def validate_presence(*settings)
-      settings.each { |setting| validators << ::Proxy::PluginValidators::Presence.new(self, setting) }
+  class SettingsProcWrapper
+    def initialize(a_blk); @a_blk = a_blk; end
+    def load_programmable_settings(settings); @a_blk.call(settings); end
```
domcleal: Can you expand these out rather than using semicolons?  It's more readable - also in classes above and below.

lib/proxy/pluggable.rb
```diff
@@ -1,131 +1,170 @@
...
+      SettingsProcWrapper.new(@programmable_settings)
+    else
+      @programmable_settings.new
     end
+  end
 
-    # End of DSL
+  class DiWiringsProcWrapper
+    def initialize(a_blk); @a_blk = a_blk; end
+    def load_dependency_injection_wirings(container, settings); @a_blk.call(container, settings); end
```
domcleal: particularly here

test/dhcp_isc/server_isc_test.rb
```diff
@@ -56,8 +56,11 @@ def test_omcmd_server_connect
     omio = OMIO.new
     IO.expects(:popen).with("/bin/sh -c 'fakeshell 2>&1'", "r+").returns(omio)
     @dhcp.omcmd('connect')
-    assert_equal "port 999", omio.input_commands[1]
-    assert_equal "server 192.168.122.1", omio.input_commands[0]
+    assert_equal 'key key_name "key_secret"', omio.input_commands[0]
```
domcleal: Unrelated changes?

test/dns_dnscmd/dnscmd_test.rb
```diff
@@ -86,7 +86,7 @@ def test_dns_zone_matches_second_best_match_if_zone_name_equals_host_name
   end
 
   def test_dns_zone_matches_sole_available_zone
-    assert_equal('sole.domain', Proxy::Dns::Dnscmd::Record.new.match_zone('host.foo.bar.sole.domain', ["sole.domain"]))
+    assert_equal('sole.domain', Proxy::Dns::Dnscmd::Record.new('localhost', 86_400).match_zone('host.foo.bar.sole.domain', ["sole.domain"]))
```
domcleal: ditto, unrelated?

test/plugins/presence_validator_test.rb
```diff
@@ -1,26 +1,24 @@
 require 'test_helper'
+require 'ostruct'
```
domcleal: Unused?

test/plugins/plugin_test.rb
```diff
@@ -0,0 +1,165 @@
...
+    http_rackup_path File.expand_path("http_config.ru", File.expand_path("../", __FILE__))
+    https_rackup_path File.expand_path("http_config.ru", File.expand_path("../", __FILE__))
+
+    default_settings :enabled => 'https'
+    plugin :test12, '1.0'
+  end
+  def test_plugin_loads_http_rack_path
+    assert !TestPlugin12.http_enabled?
+    assert TestPlugin12.https_enabled?
+    assert TestPlugin12.plugin_name, 'test12'
```
domcleal: `assert_equal 'test12', TestPlugin12.plugin_name`

test/plugins/plugin_test.rb
```diff
@@ -0,0 +1,165 @@
...
+
+    # Ensure that the content is read from 'http_config.ru'
+    File.stubs(:read).returns("require 'test12/test12_api'")
+
+    assert_equal TestPlugin12.http_rackup, ''
+    assert_equal TestPlugin12.https_rackup, "require 'test12/test12_api'"
```
domcleal: argument order's wrong: expected, actual

lib/proxy/plugin_initializer.rb
```diff
@@ -1,36 +1,288 @@
...
+    return if failed?
+    return unless @plugin.uses_provider?
+
+    used_providers = [@plugin.settings.use_provider].flatten.map(&:to_sym)
+    providers = all_plugins_and_providers.select {|p| used_providers.include?(p[:name].to_sym)}
+
+    not_available = used_providers - providers.map {|p| p[:name].to_sym}
+    return @providers = providers.map {|p| p[:class]} if not_available.empty?
+
+    fail_group_with_message("Disabling all modules in the group '#{member_names}': following providers are not available #{not_available}")
```
domcleal: `member_names` here should be the group name?  Or should it be a list of members - if so, it should probably be printed more clearly with a join(", ") or something, as it appears as:

```
Disabling all modules in the group '[:hello]' due 
```

lib/proxy/plugin_initializer.rb
```diff
@@ -1,36 +1,288 @@
...
+    return if failed?
+    members.each {|p| p.module_loader_class.new(p, di_container).configure_plugin }
+    @state = :running
+  rescue Exception => e
+    stop_services
+    fail_group(e)
+  end
+
+  def fail_group(an_exception)
+    fail_group_with_message("Disabling all modules in the group '#{member_names}' due to a failure in one of them: #{an_exception}", an_exception.backtrace)
```
domcleal: ditto, re member_names/group name

dmitri-d:
Rebased, fixed the issues pointed out in the review.

test/dhcp_isc/server_isc_test.rb
```diff
@@ -40,7 +39,7 @@ def test_isc_provider_initialization
     ::Proxy::DhcpPlugin.load_test_settings(:server => 'a_server')
     ::Proxy::DHCP::ISC::Plugin.load_test_settings(:config => 'config_file', :leases => 'leases_file',
                                                   :omapi_port => '7777', :key_name => 'key_name',
-                                                  :key_secret => 'key_secret')
+                                                  :key_secret => 'key_secret', )
```
domcleal: This breaks the test on Ruby 1.8.7, but like the previous comment, I don't think this file needs any changes.  Could you revert it to what's in develop please?

test/dns_dnscmd/dnscmd_test.rb
```diff
@@ -128,6 +128,6 @@ def test_dnscmd_enum_zones_parses_primary_zones_only
       "f.e.e.d.8.b.d.0.1.0.0.2.ip6.arpa",
       "bar.domain.local",
       "domain.local",
-      "TrustAnchors"], Proxy::Dns::Dnscmd::Record.new.enum_zones
+      "TrustAnchors"], Proxy::Dns::Dnscmd::Record.new('localhost', 86_400).enum_zones
```
domcleal: ditto, not sure this is related?

dmitri-d:
fixed the two issues above.

domcleal:
Would you mind updating the wiki page again to reflect any recent changes (e.g. support for blocks)?

domcleal:
Merged as 342eeda, thanks @witlessbird.

dmitri-d:
> Would you mind updating the wiki page again to reflect any recent changes (e.g. support for blocks)?

Done.

PR #392, opened 2016-03-11
Fixes #14153 - Allow configuration of SSL_CLIENT_CERT header name

mateusz-gozdek-sociomantic:
In passenger 5, `passenger_set_header` nginx option is adding HTTP_ prefix to each header, so we need to be able to set proper header name in order to validate client certificate.

### DIFF
--- a/lib/proxy/helpers.rb
+++ b/lib/proxy/helpers.rb
@@ -25,11 +25,15 @@ module Proxy::Helpers
 
   # read the HTTPS client certificate from the environment and extract its CN
   def https_cert_cn
-    certificate_raw = request.env['SSL_CLIENT_CERT'].to_s
-    log_halt 403, 'could not read client cert from environment' if certificate_raw.empty?
+	  certificate_raw        = request.env[Proxy::SETTINGS.ssl_client_cert_header]
+	  certificate_header     = "-----BEGIN CERTIFICATE-----\n"
+	  certificate_body       =  certificate_raw.to_s.split(/[ |\t]/)[2..-3].join("\n")
+	  certificate_footer     = "\n-----END CERTIFICATE-----"
+	  certificate_normalized = certificate_header + certificate_body + certificate_footer
+    log_halt 403, 'could not read client cert from environment' if certificate_normalized.empty?
 
     begin
-      certificate = OpenSSL::X509::Certificate.new certificate_raw
+      certificate = OpenSSL::X509::Certificate.new certificate_normalized
       if certificate.subject && certificate.subject.to_s =~ /CN=([^\s\/,]+)/i
         $1
       else
--- a/lib/sinatra/authorization.rb
+++ b/lib/sinatra/authorization.rb
@@ -32,7 +32,7 @@ module Sinatra
 
       before do
         if ['yes', 'on', '1'].include? request.env['HTTPS'].to_s
-          if request.env['SSL_CLIENT_CERT'].to_s.empty?
+          if request.env[Proxy::SETTINGS.ssl_client_cert_header].to_s.empty?
             log_halt 403, "No client SSL certificate supplied"
           end
         else
--- a/config/settings.yml.example
+++ b/config/settings.yml.example
@@ -6,6 +6,7 @@
 #:ssl_certificate: ssl/certs/fqdn.pem
 #:ssl_ca_file: ssl/certs/ca.pem
 #:ssl_private_key: ssl/private_keys/fqdn.key
+#:ssl_client_cert_header: SSL_CLIENT_CERT
 
 # Use this option only if you need to disable certain cipher suites.
 # Note: we use the OpenSSL suite name, take a look at:
--- a/lib/proxy/settings/global.rb
+++ b/lib/proxy/settings/global.rb
@@ -11,7 +11,8 @@ module ::Proxy::Settings
       :bind_host => "*",
       :log_buffer => 2000,
       :log_buffer_errors => 1000,
-      :ssl_disabled_ciphers => []
+      :ssl_disabled_ciphers => [],
+      :ssl_client_cert_header => 'SSL_CLIENT_CERT',
     }
 
     HOW_TO_NORMALIZE = {


lib/proxy/helpers.rb
```diff
@@ -25,7 +25,8 @@ def log_halt code=nil, exception=nil
 
   # read the HTTPS client certificate from the environment and extract its CN
   def https_cert_cn
-    certificate_raw = request.env['SSL_CLIENT_CERT'].to_s
+    ssl_client_cert_header = Proxy::SETTINGS.ssl_client_cert_header || 'SSL_CLIENT_CERT'
```
dmitri-d: You can avoid nil checks by defining a default value for the setting, which can be done here: https://github.com/theforeman/smart-proxy/blob/develop/lib/proxy/settings/global.rb#L3.

lib/sinatra/authorization.rb
```diff
@@ -32,7 +32,8 @@ def authorize_with_ssl_client
 
       before do
         if ['yes', 'on', '1'].include? request.env['HTTPS'].to_s
-          if request.env['SSL_CLIENT_CERT'].to_s.empty?
+          ssl_client_cert_header = Proxy::SETTINGS.ssl_client_cert_header || 'SSL_CLIENT_CERT'
```
dmitri-d: ditto re: nil value.

lib/proxy/helpers.rb
```diff
@@ -25,7 +25,8 @@ def log_halt code=nil, exception=nil
...
+    certificate_raw = request.env[ssl_client_cert_header].to_s.gsub(/   /, "\n")
```
dmitri-d: Why is this needed?

dmitri-d:
[test]

lib/proxy/helpers.rb
mateusz-gozdek-sociomantic: In my environment, nginx/1.8.1, Phusion Passenger 5.0.26, I was getting  `nested asn1 error` while parsing SSL_CLIENT_CERT. it was because somehow, new lines was replaced with spaces. I reverted that, to valid PEM format. It will not affect standalone running service. I know it's not a best solution, it would be better to implement what I suggested in [issue](http://projects.theforeman.org/issues/14153).

lib/proxy/helpers.rb
mateusz-gozdek-sociomantic: Updated as you suggested, thank you!

lib/proxy/helpers.rb
dmitri-d: This suggests an issue in your environment. How is the value for `$ssl_client_cert` being generated?

mateusz-gozdek-sociomantic: With this line in nginx server section for foreman-proxy:

```
passenger_set_header  X-SSL-Client-Cert $ssl_client_cert;
```

dmitri-d: This is a known issue with nginx: https://forum.nginx.org/read.php?2,236546,236546#msg-236546.

As header continuation can use either a single space or a tab, something like this would be a better way for stripping of whitespaces:

`request.env[Proxy::SETTINGS.ssl_client_cert_header].to_s.gsub(/\s+/, "\n")`

Lastly, could you add a comment explaining a reason for this line please.

mateusz-gozdek-sociomantic: Thank you for digging out that issue, unfortunately, `/\s+/` instead of `/   /` will not work, because `-----BEGIN CERTIFICATE-----` contains space. How about `/[\t]+/` as nginx is [adding \t](https://github.com/nginx/nginx/blob/master/src/event/ngx_event_openssl.c#L3280) to cert ?

mateusz-gozdek-sociomantic:
`gsub(/[\t]+/, "\n")` is required, because nginx is adding
`\t` after each line of `$ssl_client_cert`, which makes certificate
invalid with PEM format. Nginx can't add `\n` character instead of `\t`, because it would be invalid with [HTTP Header format](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html).

lib/proxy/helpers.rb
dmitri-d: I'd prefer not to make this fix specific to nginx (or this version of it) and handle the issue in accordance with rfc:

```
def strip_header_continuation_from_ssl_cert(cert_header_value)
  begin_cert = "-----BEGIN CERTIFICATE-----"
  end_cert = "-----END CERTIFICATE-----"
  cert_body_with_header_continuation = cert_header_value.slice((begin_cert.length)..-(end_cert.length+1))
  begin_cert + "\n" + cert_body_with_header_continuation.gsub(/\s/, "\n") + "\n" + end_cert
end
```

I think the new line before "END CERTIFICATE" is needed, could you please check that? Also, please note that the regex in gsub now matches single white spaces (per rfc).

mateusz-gozdek-sociomantic: I tested `/[\t]+/` both on webrick and nginx and it's working properly. If you want this to be RFC compatible, how about `[ ]{2,}|\t+` to remove `two or more spaces` or `one or more tabs`? It's also working with both webrick and nginx and it's more bulletproof.

dmitri-d: Hmm, according to https://www.w3.org/Protocols/rfc2616/rfc2616-sec2.html (see section 2.2 Basic Rules, look for "folded onto multiple lines" line) header continuation uses either a single space or a tab. Are you trying to handle something that's nginx-specific?

mateusz-gozdek-sociomantic: From my point of view, creating another function, just to make sure the certificate is in valid format is unnecessary. We need to skip one space case, because that one space is included in PEM header. Maybe we can replace ` `with`\s`.

dmitri-d: I'm not going to insist on a particular implementation, as long as:
- the code is readable (one-liner, or a method)
- the code is testable
- solving the general issue rather then tailored to a particular corner case

mateusz-gozdek-sociomantic: Ok, I thought about it during weekend and I worked out something like this:

```
def strip_header_continuation_from_ssl_cert(ssl_client_cert)
  cert = "-----BEGIN CERTIFICATE-----\n"
  #skip header and footer of certificate from splitting with [2..-3]
  cert += ssl_client_cert.split(/[ |\t]+/)[2..-3].join("\n")
  cert += "\n-----END CERTIFICATE-----"
  cert
end
```

it should be valid for RFC, quite readable and testable.

dmitri-d: RFC specifies single white space or a tab for header continuation, a '+' in `split(/[ |\t]+/)` isn't needed. I need to think over if we want to handle cases where header continuation is invalid, but I'm leaning towards "no". 

No need for the `cert` line before the end,  the result of the expression on the previous line is the same. I wouldn't build a string by reassigning it to itself -- I haven't looked, but I'd bet this incurs additional copying. Something like `begin + ssl_client_cert.split(/[ |\t]+/)[2..-3].join("\n") + end` is probably preferable.

Lastly, could you add a test for this? `sinatra/ssl_client_verification_test.rb` looks like a good spot.

mateusz-gozdek-sociomantic: I am sorry for adding that `+`, I was looking for `HTTP message header`, not `HTTP header`, my bad.

So, you prefer:

```
def strip_header_continuation_from_ssl_cert(ssl_client_cert)
  cert_header = "-----BEGIN CERTIFICATE-----\n"
  cert_footer = "\n-----END CERTIFICATE-----"
  cert_header + ssl_client_cert.split(/[ |\t]/)[2..-3].join("\n") + cert_footer
end
```

or just:

```
  "-----BEGIN CERTIFICATE-----\n" + ssl_client_cert.split(/[ |\t]/)[2..-3].join("\n") + "\n-----END CERTIFICATE-----"
```

?
I prefer first version, because second one is a bit to long for me. 

About test. I never wrote ruby tests before, but I will try to write it.

dmitri-d: Either one is ok with me. Let me know if you need help with tests.

mateusz-gozdek-sociomantic: I don't really know how to write those tests. Should I test if `strip_header_continuation_from_ssl_cert` function returns valid certificate from invalid, with fixtures containing those certificates certificates? or should i edit `test_https_cert` test to check also syntax of certificates?

dmitri-d: I'd write two tests: one that verifies that strip_header_continuation_from_ssl_cert method removes extraneous spaces from a cert when they are present. And another one that tests that the cert is left intact when it has no header continuation artefacts.

mateusz-gozdek-sociomantic:
Updated as we discussed in comments. But I don't really know how to write tests for this.

mateusz-gozdek-sociomantic:
BTW, this raw certificate fix + tests should be applied to [foreman](https://github.com/theforeman/foreman) too.

roidelapluie:
[test]

mateusz-gozdek-sociomantic:
It may be hard for me to provide unit tests for that one, so if someone could add it, it would be really nice, as they will be required for Foreman as well.

lib/proxy/helpers.rb
```diff
@@ -25,11 +25,15 @@ def log_halt code=nil, exception=nil
 
   # read the HTTPS client certificate from the environment and extract its CN
   def https_cert_cn
-    certificate_raw = request.env['SSL_CLIENT_CERT'].to_s
-    log_halt 403, 'could not read client cert from environment' if certificate_raw.empty?
+	  certificate_raw        = request.env[Proxy::SETTINGS.ssl_client_cert_header]
```
ekohl, line 28: The indenting looks odd here

ekohl (changes requested):
I can't speak for @witlessbird but I'd start with having a look at ```test/sinatra/trusted_hosts_test.rb```. Personally I'd try to add the expectation that it would be called with the correct parameters. I generally also like integration tests where I just pass in a certificate and verify it returns the correct result, but not sure where those should be placed.

lzap:
Hello, it's been a while. This is a critical portion of our code (every request executes this bit) and also security-related. Good test coverage is a must. I am closing for now, please come back later when you figure it out. Thanks for the contribution so far!

PR #394, opened 2016-03-11
Fixes #12295 -- Added support for log rotation

dmitri-d:
SIGUSR1 now can be used to re-open log file when file-based logger is used.

### DIFF
--- a/lib/proxy/log.rb
+++ b/lib/proxy/log.rb
@@ -14,19 +14,13 @@ Logger.class_eval { alias_method :write, :info }
 
 module Proxy
   module Log
-    @@logger = nil
-
     def logger
-      @@logger ||= ::Proxy::Log.logger
+      ::Proxy::LogBuffer::Decorator.instance
     end
+  end
 
-    def self.default_logger(log_file)
-      # We keep the last 6 10MB log files
-      ::Logger.new(log_file, 6, 1024*1024*10)
-    end
-
+  class LoggerFactory
     def self.logger
-      log_file = ::Proxy::SETTINGS.log_file
       if log_file.casecmp('STDOUT').zero?
         if SETTINGS.daemon
           puts "Settings log_file=STDOUT and daemon=true are incompatible, exiting..."
@@ -44,7 +38,16 @@ module Proxy
         logger = default_logger(log_file)
       end
       logger.level = ::Logger.const_get(::Proxy::SETTINGS.log_level.upcase)
-      ::Proxy::LogBuffer::Decorator.new(logger)
+      logger
+    end
+
+    def self.default_logger(log_file)
+      # We keep the last 6 10MB log files
+      ::Logger.new(log_file, 6, 1024*1024*10)
+    end
+
+    def self.log_file
+      ::Proxy::SETTINGS.log_file
     end
   end
 
--- a/lib/proxy/log_buffer/decorator.rb
+++ b/lib/proxy/log_buffer/decorator.rb
@@ -1,32 +1,55 @@
 require 'proxy/log_buffer/decorator'
 require 'proxy/log_buffer/buffer'
+require 'thread'
 
 module Proxy::LogBuffer
   class Decorator
-    def initialize(logger, buffer = Proxy::LogBuffer::Buffer.instance)
+    def self.instance
+      @@instance ||= new(::Proxy::LoggerFactory.logger, ::Proxy::LoggerFactory.log_file)
+    end
+
+    def initialize(logger, log_file, buffer = Proxy::LogBuffer::Buffer.instance)
       @logger = logger
       @buffer = buffer
+      @log_file = log_file
+      @mutex = Mutex.new
+      @roll_log = false
+    end
+
+    # due to synchronization can't re-open the log from the signal trap
+    def roll_log
+      @roll_log = true
+    end
+
+    def handle_log_rolling
+      return if @log_file.casecmp('STDOUT') == 0 || @log_file.casecmp('SYSLOG') == 0
+      @roll_log = false
+      @logger.close rescue nil
+      @logger = ::Proxy::LoggerFactory.logger
     end
 
     def add(severity, message = nil, progname = nil, backtrace = nil)
-      severity ||= UNKNOWN
-      if message.nil?
-        if block_given?
-          message = yield
-        else
-          message = progname
+      @mutex.synchronize do
+        handle_log_rolling if @roll_log
+        severity ||= UNKNOWN
+        if message.nil?
+          if block_given?
+            message = yield
+          else
+            message = progname
+          end
         end
-      end
-      # add to the logger first
-      @logger.add(severity, message)
-      @logger.add(::Logger::Severity::DEBUG, backtrace) if backtrace
-      # add add to the buffer
-      if severity >= @logger.level
-        # we accept backtrace, exception and simple string
-        backtrace = backtrace.is_a?(Exception) ? backtrace.backtrace : backtrace
-        backtrace = backtrace.respond_to?(:join) ? backtrace.join("\n") : backtrace
-        rec = Proxy::LogBuffer::LogRecord.new(nil, severity, message, backtrace)
-        @buffer.push(rec)
+        # add to the logger first
+        @logger.add(severity, message)
+        @logger.add(::Logger::Severity::DEBUG, backtrace) if backtrace
+        # add add to the buffer
+        if severity >= @logger.level
+          # we accept backtrace, exception and simple string
+          backtrace = backtrace.is_a?(Exception) ? backtrace.backtrace : backtrace
+          backtrace = backtrace.respond_to?(:join) ? backtrace.join("\n") : backtrace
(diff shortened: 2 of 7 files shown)


lib/proxy/log.rb
```diff
@@ -35,6 +30,7 @@ def self.logger
         logger = ::Logger.new(STDOUT)
       elsif log_file.casecmp('SYSLOG').zero?
         begin
+          @@needs_log_rolling = false
```
domcleal: This appears to be unused, should it be used instead of `@logger.instance_of?` in `handle_log_rolling`?

lib/proxy/log.rb
```diff
@@ -35,6 +30,7 @@ def self.logger
         logger = ::Logger.new(STDOUT)
```
domcleal, line 29: STDOUT probably doesn't need rolling either.

lib/proxy/log_buffer/decorator.rb
```diff
@@ -1,32 +1,53 @@
...
+
     def initialize(logger, buffer = Proxy::LogBuffer::Buffer.instance)
       @logger = logger
       @buffer = buffer
+      @mutex = Mutex.new
+      @log_rolled = false
+    end
+
+    # due to synchronization can't re-open the log from the signal trap
+    def log_rolled
```
domcleal: Nitpick, this method name is a little unclear as it implies it's been done - perhaps `roll_log`?

lib/proxy/log_buffer/decorator.rb
```diff
@@ -1,32 +1,53 @@
...
     end
 
     def add(severity, message = nil, progname = nil, backtrace = nil)
-      severity ||= UNKNOWN
-      if message.nil?
-        if block_given?
-          message = yield
-        else
-          message = progname
+      @mutex.synchronize do
```
domcleal, line 32: Is this to prevent the old logger being written to during the rolling?

If so, we could perhaps remove it and allow the old logger to be written to for that brief period - httpd works this way and logrotate is then usually configured with `delaycompress`.  The idea's that the old log gets written to as the pre-rotation workers finish off their requests and the log only gets compressed on the next rotation.  The old file descriptor's still valid so loggers still work, it's just that the file gets moved to a `.1` name.

dmitri-d, line 32: That, but mainly because @logger is referred to more than once. The mutex guarantees us that all writes are done to the _same_ logger. For example, without this synchronization it would be possible for https://github.com/theforeman/smart-proxy/pull/394/files#diff-6ae09024c8b8cebdf00278e6e789bc93R41 and https://github.com/theforeman/smart-proxy/pull/394/files#diff-6ae09024c8b8cebdf00278e6e789bc93R42 to end up in different log files. There's also the awkwardness of multiple re-initialization of @logger in the event of log rolling.

It's not completely thread-safe atm (I don't think boolean assignment and checks are atomic), but the mutex makes the whole thing more deterministic (which in turns hurts my head less).

dmitri-d:
- renamed `Decorator#log_rolled` to `Decorator#roll_log`
- removed unused `@@needs_log_rolling` in the `Log` module
- STDOUT-based logs are not being rolled anymore

domcleal:
I'm experiencing hangs when processing HTTP requests after sending a USR1 signal to the smart proxy while trying to test this feature.  Here's some info, hopefully you can reproduce it:
1. `:log_file: /tmp/proxy.log`
2. `be bin/smart-proxy` running in the foreground in one window under Ruby 2.1.0
3. requests to /features are working fine, logging OK too
4. pkill -f -USR1 bin/smart-proxy
5. subsequent calls to /features just block after the GET request is issued
6. the smart proxy process still has the same file descriptor open to /tmp/proxy.log

Stack trace issued through a TTIN signal shows this:

<pre>
Starting thread dump for current Ruby process
=============================================
Thread TID-10048280
/home/dcleal/code/foreman/smart-proxy/lib/proxy/signal_handler.rb:22:in `backtrace'
/home/dcleal/code/foreman/smart-proxy/lib/proxy/signal_handler.rb:22:in `block (2 levels) in install_ttin_trap'
/home/dcleal/code/foreman/smart-proxy/lib/proxy/signal_handler.rb:20:in `each'
/home/dcleal/code/foreman/smart-proxy/lib/proxy/signal_handler.rb:20:in `block in install_ttin_trap'
/home/dcleal/code/foreman/smart-proxy/lib/launcher.rb:131:in `call'
/home/dcleal/code/foreman/smart-proxy/lib/launcher.rb:131:in `join'
/home/dcleal/code/foreman/smart-proxy/lib/launcher.rb:131:in `launch'
bin/smart-proxy:6:in `<main>'
Thread TID-28920800
/home/dcleal/.rvm/rubies/ruby-2.1.0/lib/ruby/2.1.0/webrick/server.rb:170:in `select'
/home/dcleal/.rvm/rubies/ruby-2.1.0/lib/ruby/2.1.0/webrick/server.rb:170:in `block in start'
/home/dcleal/.rvm/rubies/ruby-2.1.0/lib/ruby/2.1.0/webrick/server.rb:32:in `start'
/home/dcleal/.rvm/rubies/ruby-2.1.0/lib/ruby/2.1.0/webrick/server.rb:160:in `start'
/home/dcleal/.rvm/gems/ruby-2.1.0@proxy/gems/rack-1.5.5/lib/rack/handler/webrick.rb:14:in `run'
/home/dcleal/.rvm/gems/ruby-2.1.0@proxy/gems/rack-1.5.5/lib/rack/server.rb:264:in `start'
/home/dcleal/code/foreman/smart-proxy/lib/launcher.rb:126:in `block in launch'
Thread TID-28920640
/home/dcleal/.rvm/rubies/ruby-2.1.0/lib/ruby/2.1.0/webrick/server.rb:170:in `select'
/home/dcleal/.rvm/rubies/ruby-2.1.0/lib/ruby/2.1.0/webrick/server.rb:170:in `block in start'
/home/dcleal/.rvm/rubies/ruby-2.1.0/lib/ruby/2.1.0/webrick/server.rb:32:in `start'
/home/dcleal/.rvm/rubies/ruby-2.1.0/lib/ruby/2.1.0/webrick/server.rb:160:in `start'
/home/dcleal/.rvm/gems/ruby-2.1.0@proxy/gems/rack-1.5.5/lib/rack/handler/webrick.rb:14:in `run'
/home/dcleal/.rvm/gems/ruby-2.1.0@proxy/gems/rack-1.5.5/lib/rack/server.rb:264:in `start'
/home/dcleal/code/foreman/smart-proxy/lib/launcher.rb:127:in `block in launch'
Thread TID-28224700
/home/dcleal/.rvm/rubies/ruby-2.1.0/lib/ruby/2.1.0/webrick/utils.rb:174:in `sleep'
/home/dcleal/.rvm/rubies/ruby-2.1.0/lib/ruby/2.1.0/webrick/utils.rb:174:in `block in initialize'
</pre>

dmitri-d:
Fixed an issue here: https://github.com/witlessbird/smart-proxy/blob/12295-log-rotation/lib/proxy/log_buffer/decorator.rb#L25. A '@' was missing in `@log_file.casecmp('SYSLOG')` expression, which lead to a crash and webrick freezing.

Will add a test.

dmitri-d:
Now with tests.

lib/proxy/log.rb
```diff
@@ -14,19 +14,13 @@
 
 module Proxy
   module Log
-    @@logger = nil
-
     def logger
```
domcleal, line 17: Is this retained for compatibility only?  Perhaps issue a warning if called?

lib/proxy/log_buffer/decorator.rb
```diff
@@ -1,32 +1,54 @@
...
+      @roll_log = false
+    end
+
+    # due to synchronization can't re-open the log from the signal trap
+    def roll_log
+      @roll_log = true
+    end
+
+    def handle_log_rolling
+      return if (@log_file.casecmp('STDOUT') == 0 || @log_file.casecmp('SYSLOG') == 0)
```
domcleal: rubocop fails here, the brackets around the conditional should be removed

domcleal:
Thanks, working better now.

The proxy seems to leak fds when rotating a log file, every time I send a signal and make a request, an additional fd is opened for the log and the old ones aren't getting cleaned up.  After also rotating the file a few of them were indeed closed, but at peak I had about 50 fds open to the same log file after sending signals.

Is it possible/best to explicitly close the old logger down?  It appears to be relying on the GC only.

dmitri-d:
Fixed rubocop warning.

dmitri-d:
Logger no longer leaks fds -- Logger#stop is being called during log rolling.

lib/proxy/log.rb
dmitri-d, line 17: ATM this is the main interface to logger, a warning on every call would be quite annoying. I haven't given up on switching to events, at which point this would go away...

domcleal, line 17: Ah yes, the module's included in other places, thanks.

domcleal:
Merged as 7cc71a1, thanks @witlessbird!

PR #396, opened 2016-03-18
[DO NOT MERGE] Fixes #14142 - fix for a race condition when using webrick and listening on both http and https ports

dmitri-d:
This switches to thin. There are still a couple of issues here:
- SSL config is different, it's impossible to pass SSL options directly. As a result there's no way to force server's cipher ordering
- ssl client cert authorization isn't working atm: cert ca check has to be done by smart-proxy


dmitri-d:
Another issue is packaging: I don't think it would make sense to use passenger with smart-proxy if we are to use thin (pretty fast as is).

dmitri-d:
I looked into using puma, but at a brief glance it appears pretty different: atm I'm not even sure how I can assemble a single web application from a bunch of modules.

domcleal:
How about keeping using WEBrick but avoid Rack::Server?  It seems the issue is pretty much in Rack's server startup support.

dmitri-d:
Atm we are relying on rack for the composition of the app out of modules. To get rid completely of rack I'd need to figure out how to do this with sinatra only, probably a bigger change than switching to thin (pretty trivial).

dmitri-d:
Having said that, I like the idea -- I'll see how much work getting rid of rack would be (although I also like the idea of not using webrick).

domcleal:
Perhaps even keeping rack, but not using that method relying on class-level variables in the rack handlers, e.g. https://github.com/theforeman/smart-proxy/compare/develop...domcleal:14142-rack-server-race.  This still uses the rack WEBrick handler which helps set up a few environment bits, otherwise it's pretty lightweight I believe.

If you're interested, I'll submit a PR with that.

dmitri-d:
Yeah, sinatra and rack seem to be very closely integrated, not sure it is possible for us to get rid of rack. Which leaves the question of should we continue using webrick? Or switch to thin (or something else)?

domcleal:
I've no real opinion about one server versus the other, all other things being equal.  The SSL points raised against thin in the PR description are concerning, however.  Packaging is always going to be additional work, but all solvable.

If there's not a compelling reason to switch then I suppose I'd prefer to keep the status quo and just attempt to fix the bug.

dmitri-d:
No compelling reason, more of an opportunity to do a switch. What attracts me to thin is its speed and a chance to simplify installs for people who deploy smart-proxy under passenger (looks like thin is just as fast if not faster).

dmitri-d:
@domcleal: can you do a PR with your changes?

domcleal:
Sure, will do.  I noticed today that my current branch isn't actually working for requests, so I need to take a closer look at that first.

dmitri-d:
@domcleal: I'm curious about this bit: https://github.com/theforeman/smart-proxy/compare/develop...domcleal:14142-rack-server-race#diff-b8fbcbffabe3ae5c9935bb21e3e94852R117. The concurrency issue is in rack's webrick handler (around `@server` singleton class var: https://github.com/rack/rack/blob/master/lib/rack/handler/webrick.rb#L31). Lemme know if you want to me look into that...

domcleal:
> I'm curious about this bit: develop...domcleal:14142-rack-server-race#diff-b8fbcbffabe3ae5c9935bb21e3e94852R117. The concurrency issue is in rack's webrick handler (around @server singleton class var:

Yep, looks like it was that line that needed changing.  The `Rack::Server.new(app)` was incorrect and entirely unnecessary - because it's Rack::Server I was trying to remove.  Simply changing it to the Rack::Builder app instance generated in the http/https_app methods fixed the issue.

The intention as you say is to avoid that class variable and method in the server handler, which is the one that Rack::Server relies on.  The handler itself is still used as an interface between Rack and WEBrick, but we don't need its class methods, only the instance level ones.

domcleal:
#397 is opened with my proposed fix.

dmitri-d:
closing in favour of https://github.com/theforeman/smart-proxy/pull/397.

PR #397, opened 2016-03-21
fixes #14142 - launch WEBrick directly, avoiding Rack::Server

domcleal:
Rack::Server calls server-specific handlers such as
Rack::Handler::WEBrick which relies on a class-level variable that
causes race conditions when launching two WEBrick instances via
Rack::Server within the same process.

Launching WEBrick with the Rack::Builder-prepared app prevents the race
between the two instances, ensuring both start up correctly.  The Rack
WEBrick handler is still used to interface between WEBrick and the app
built with Rack.

### DIFF
--- a/lib/launcher.rb
+++ b/lib/launcher.rb
@@ -22,13 +22,14 @@ module Proxy
         end
       end
 
-      Rack::Server.new(
+      {
         :app => app,
         :server => :webrick,
-        :Host => SETTINGS.bind_host,
+        :BindAddress => SETTINGS.bind_host,
         :Port => SETTINGS.http_port,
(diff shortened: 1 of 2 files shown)


dmitri-d:
Looks good. A nitpick: looks like 5 second delay here https://github.com/theforeman/smart-proxy/blob/develop/lib/proxy/signal_handler.rb#L32 can go now...

domcleal:
Thanks, I simplified the INT trap installation to remove the whole Rack version check too, as none of it will be installed as Rack::Server's removed.

dmitri-d:
Merged in at e5b6697. Thanks, @domcleal!

PR #398, opened 2016-03-23
Fixes #9169 - Add option to ignore supersede next-server statement

akhirai:
This patch introduces a new option "dont_pass_nextserver" into tftp.yml.

To set this option true, all new reservations for the subnet does not include "supersede server.next-server" statement into dhcpd.leases.
So, "next-server" value in each subnet specified in dhcpd.conf becomes valid.

It is necessary to modify both foreman and smart-proxy, I send the same titled PR to each.

### DIFF
--- a/config/settings.d/tftp.yml.example
+++ b/config/settings.d/tftp.yml.example
@@ -5,3 +5,6 @@
 #:tftproot: /var/lib/tftpboot
 # Defines the TFTP Servername to use, overrides the name in the subnet declaration
 #:tftp_servername: tftp.domain.com
+# By enabling this option, all new reservations for this subnets will not include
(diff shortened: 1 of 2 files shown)


dmitri-d:
Wouldn't it be better to have this setting in foreman? Perhaps in subnet ui? It's awkward needing to set a flag and restart the proxy ony to send a flag to foreman to override a default behaviour (that probably should be controlled per-interface?).

akhirai:
Thank you for the comment.

I withdraw this patch, 
 and I will revised proposal in Foreman side.

domcleal:
Thanks @akhirai, I'll close this one and leave the Foreman one open for revision.

PR #399, opened 2016-03-23
Fixes #13551: Add IP address validation in DNS API

ekohl:

PR #400, opened 2016-03-23
Fixes #13552: Validate DNS names passsed to the DNS API

ekohl:
Includes https://github.com/theforeman/smart-proxy/pull/399 because they modify the same code and splitting them and then merging is a lot of useless work.

### DIFF
--- a/modules/dns/dns_api.rb
+++ b/modules/dns/dns_api.rb
@@ -18,11 +18,14 @@ module Proxy::Dns
       log_halt(400, "'create' requires fqdn, value, and type parameters") if fqdn.nil? || value.nil? || type.nil?
 
       begin
+        validate_dns_name!(fqdn)
+
         case type
         when 'A'
           ip = IPAddr.new(value, Socket::AF_INET).to_s
           server.create_a_record(fqdn, ip)
         when 'PTR'
+          validate_reverse_dns_name!(value)
           server.create_ptr_record(fqdn, value)
         else
           log_halt(400, "unrecognized 'type' parameter: #{type}")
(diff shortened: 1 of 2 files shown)


ekohl:
Rebased due to a merge conflict.

modules/dns/dns_api.rb
```diff
@@ -35,16 +38,35 @@ class Api < ::Sinatra::Base
...
+        end
       rescue Proxy::Dns::NotFound => e
         log_halt 404, e
       rescue => e
         log_halt 400, e
       end
     end
+
+    def validate_dns_name!(name)
+      raise Proxy::Dns::Error.new("Invalid DNS name #{name}") unless name =~ /^([a-zA-Z0-9]([-a-zA-Z0-9]+)?\.?)+$/
```
ekohl, line 64: This is the quite strict version of the regex, I didn't check how Foreman validates DNS names. It doesn't allow underscores, which technically isn't allowed according to the RFCs but usually does work. It also allows some invalid things like ..

dmitri-d, line 64: I think this is ok: 
- For A/AAAA record creation this is more of a precaution, as the server is going to do its own verification
- Not sure how valuable this is for PTR record creation -- fqdn is used for detection of duplicate records only
- Similarly for record deletion -- I doubt that servers perform fqdn validation, as all they need is a match/no match against known host names...

ekohl, line 64: I haven't verified this, but I strongly suspect you can insert anything into the nsupdate command right now so this is intended as a security measure. It would be hard to exploit, but I think it is possible. Same goes for dnscmd which executes it using popen.

dmitri-d:
[test]

dmitri-d:
Merged in at 5d68f49. Thanks, @ekohl!

PR #402, opened 2016-03-30
Fixes #14394 - set :ServerSoftware to nil

brandonweeks:
Initial attempt at reducing the information disclosed by WEBrick in the HTTP server header.

Options I can think of:
- Set :ServerSoftware to "WEBrick." This is a best practice across many major sites, e.g. only returning "Apache" as the header. However given WEBrick's tiny install base, this would effectively uniquely identify Smart Proxy servers exposed to the internet.
- Set the :ServerSoftware to nil, causing WEBrick to return an empty server header.
- Monkey patch WEBrick::HTTPResponse#setup_header to not set the server header. This is the most ideal result, however I don't know how to accomplish it without copying the entire method into the Smart Proxy codebase. https://github.com/ruby/ruby/blob/trunk/lib/webrick/httpresponse.rb#L222 Is there a clean way to do this?

This is just a rough draft, I'd appreciate input on the best way to accomplish this.

### DIFF
--- a/lib/launcher.rb
+++ b/lib/launcher.rb
@@ -28,6 +28,7 @@ module Proxy
         :BindAddress => SETTINGS.bind_host,
         :Port => SETTINGS.http_port,
         :Logger => ::Proxy::LogBuffer::Decorator.instance,
+        :ServerSoftware => 'WEBrick',
         :daemonize => false
       }
     end
@@ -52,6 +53,7 @@ module Proxy
           :BindAddress => SETTINGS.bind_host,
           :Port => SETTINGS.https_port,
           :Logger => ::Proxy::LogBuffer::Decorator.instance,
+          :ServerSoftware => 'WEBrick',
           :SSLEnable => true,
           :SSLVerifyClient => OpenSSL::SSL::VERIFY_PEER,
           :SSLPrivateKey => load_ssl_private_key(SETTINGS.ssl_private_key),


dmitri-d:
I think my preferred approach would be #1 -- secure enough (I don't think setting the header to Webrick would give away much) and simple. #2 should work ok, but I'm always a bit leery of using nils. #3 is doable, but not sure it's worth it, and it's going to be the most fragile: 

```
class Webrick::HTTPResponse
  alias_method :setup_header_original, :setup_header
  def setup_header
    setup_header_original()
    @header['server'] = "whatever"
  end
end
```

or something like that.

brandonweeks:
https://www.shodan.io/search?query=WEBrick+port%3A%228443%22

I'm leaning towards option one, it should be noted that smart proxies exposed to the internet are incredibly easy to find. PR updated with option one as well.

dmitri-d:
The commit message is a bit misleading now o_O.

dLobatog:
@brandonweeks Perhaps we can set it through the settings.yaml file - and default to nil? I'm not very comfortable with offering a default that can be so easily searchable.

@witlessbird What would be the problem with serving `nil` by default?

dmitri-d:
I don't like the idea of a setting: most people won't ever touch this. I dislike nil because there's a high chance the rest of webrick codebase expects something in that place, nil can potentially cause trouble (it works now, however). If there are objections to using "webrick" for :ServerSoftware, I would suggest using an empty string instead.

dLobatog:
My only objection is what @brandonweeks mentioned above - it makes smart-proxies too easy to spot online. An empty string or a nil are both fine in my opinion, the point is just not to show what we're running.

shlomizadok:
@brandonweeks - what's the status of this PR? Can you please rebase?

dmitri-d:
closing in favour of https://github.com/theforeman/smart-proxy/issues/482

PR #404, opened 2016-04-04
Fixes #14461 - removed :method_missing from Proxy::Settings::Plugin class

dmitri-d:
OpenStruct always returned 'nil' when calling methods for non-existing keys. Before ruby 2.3 methods for keys were defined during OpenStruct instantiation.
  Starting with ruby 2.3 methods for members of the OpenStruct are defined lazily, via method_missing. Our override of the method was breaking OpenStruct (and
  as a consequence Proxy::Settings::Plugin).

### DIFF
--- a/lib/proxy/settings/plugin.rb
+++ b/lib/proxy/settings/plugin.rb
@@ -9,7 +9,5 @@ module ::Proxy::Settings
       @used_defaults = @defaults.keys - settings.keys
       super(@defaults.merge(settings))
     end
-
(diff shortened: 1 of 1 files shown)


GregSutcliffe:
Tested locally having hit this on Ruby 2.3/Archlinux - fixes the issue for me. :shipit:

mmoll:
Fixes the problem also here (Debian/testing already switched to 2.3) :+1:

domcleal:
Merged as b215df8, thanks @witlessbird.

PR #406, opened 2016-04-13
fixes #13667 - use puppet 4.0 classes api

dmitri-d:
- Replaces https://github.com/theforeman/smart-proxy/pull/376
- Uses changes to the core introduced in https://github.com/theforeman/smart-proxy/pull/390
- This PR is functionality almost identical to https://github.com/theforeman/smart-proxy/pull/376, with the exception of puppet and augeas gems not being loaded when puppet_api provider is enabled.


mmoll:
Maybe there's a problem with my setup, but I'm getting

E, [2016-04-16T23:12:57.681813 #26029] ERROR -- : Disabling all modules in the group '[:puppet_proxy_puppet_api, :puppet]' due to a failure in one of them: undefined method `to_sym' for [:puppet_ssl_ca, :puppet_ssl_cert, :puppet_ssl_key]:Array

```
root@puppetserver:/usr/share/foreman-proxy/config/settings.d# cat puppet_proxy_puppet_api.yml 
---
# URL of the puppet master itself for API requests.
:puppet_url: https://puppetserver.kvedulv.de:8140
#
# SSL certificates used to access the puppet API
:puppet_ssl_ca: /etc/puppetlabs/puppet/ssl/certs/ca.pem
:puppet_ssl_cert: /etc/puppetlabs/puppet/ssl/certs/puppetserver.kvedulv.de.pem
:puppet_ssl_key: /etc/puppetlabs/puppet/ssl/private_keys/puppetserver.kvedulv.de.pem
```

dmitri-d:
Not sure what's going on here. Could you enable debug logging and post the results please? I assume you pulled this branch into your local install? Did you run "rake migrate_settings" (although puppet_proxy_puppet_api.yml looks good to me)?

mmoll:
I did take this PR as a patch to the most recent nightly DEB package and updated the config maually, after running the rake job the behaviour didn't change, the only relevant debug log snippet I'm seeing:
D, [2016-04-16T23:56:24.816966 #27424] DEBUG -- : 'puppet' settings: 'customrun_args': -ay -f -s, 'customrun_cmd': /bin/false, 'enable
d': https, 'puppet_conf': /etc/puppetlabs/puppet/puppet.conf, 'puppet_version': 4.4, 'puppetssh_command': /usr/bin/puppet agent --onet
ime --no-usecacheonfailure, 'puppetssh_sudo': false, 'puppetssh_wait': false, 'use_provider': [:puppet_proxy_puppet_api]
D, [2016-04-16T23:56:24.819225 #27424] DEBUG -- : 'puppet_proxy_puppet_api' settings: 'classes_retriever': apiv3, 'customrun_args': -a
y -f -s, 'customrun_cmd': /bin/false, 'enabled': false, 'environments_retriever': apiv3, 'puppet_conf': /etc/puppetlabs/puppet/puppet.
conf, 'puppet_ssl_ca': /etc/puppetlabs/puppet/ssl/certs/ca.pem, 'puppet_ssl_cert': /etc/puppetlabs/puppet/ssl/certs/puppetserver.kvedu
lv.de.pem, 'puppet_ssl_key': /etc/puppetlabs/puppet/ssl/private_keys/puppetserver.kvedulv.de.pem, 'puppet_url': https://puppetserver.k
vedulv.de:8140, 'puppet_version': 4.4, 'puppetssh_command': /usr/bin/puppet agent --onetime --no-usecacheonfailure, 'puppetssh_sudo': 
false, 'puppetssh_wait': false, 'use_provider': [:puppet_proxy_puppet_api]
E, [2016-04-16T23:56:24.819601 #27424] ERROR -- : Disabling all modules in the group '[:puppet_proxy_puppet_api, :puppet]' due to a fa
ilure in one of them: undefined method `to_sym' for [:puppet_ssl_ca, :puppet_ssl_cert, :puppet_ssl_key]:Array

dmitri-d:
It should show a stacktrace if debug logging is enabled. Can you try checking out and copying the whole source tree into the install directory?

mmoll:
> Can you try checking out and copying the whole source tree into the install directory?

Did that now and copied in the old config dir, however still getting the same message and no backtrace.

dmitri-d:
That log looks pretty weird: I see a bunch of settings (for example 'puppet_conf', 'puppetssh_command') that shouldn't be in puppet.yml anymore. Check that migrations executed correctly (you may need to specify path to the config dir). Check that log_level in the main config file set to DEBUG.

Before spending time troubleshooting your setup I'd like to make sure that the baseline source install works. If you could: check out this branch, modify the puppet-related config to suit your needs, do bundle install and then bundle exec bin/smart-proxy (you may want to use rvm).

mmoll:
The migration rake task does not work very well in that type of install, fixed that now. Output is more or less still the same:

```
W, [2016-04-17T00:40:24.220243 #4314]  WARN -- : Couldn't find settings file /etc/foreman-proxy/settings.d/foreman_proxy.yml. Using default settings.
D, [2016-04-17T00:40:24.221950 #4314] DEBUG -- : 'foreman_proxy' settings: 'enabled': true (default)
D, [2016-04-17T00:40:24.223415 #4314] DEBUG -- : 'tftp' settings: 'enabled': https, 'tftproot': /srv/tftp
D, [2016-04-17T00:40:24.224830 #4314] DEBUG -- : 'puppetca' settings: 'enabled': https, 'puppetdir': /etc/puppetlabs/puppet, 'ssldir': /etc/puppetlabs/puppet/ssl
D, [2016-04-17T00:40:24.227313 #4314] DEBUG -- : 'puppet' settings: 'enabled': https, 'puppet_version': 4.3.1, 'use_provider': [:puppet_proxy_puppet_api]
D, [2016-04-17T00:40:24.230156 #4314] DEBUG -- : 'puppet_proxy_puppet_api' settings: 'classes_retriever': apiv3, 'enabled': false, 'environments_retriever': apiv3, 'puppet_ssl_ca': /etc/puppetlabs/puppet/ssl/certs/ca.pem, 'puppet_ssl_cert': /etc/puppetlabs/puppet/ssl/certs/puppetserver.kvedulv.de.pem, 'puppet_ssl_key': /etc/puppetlabs/puppet/ssl/private_keys/puppetserver.kvedulv.de.pem, 'puppet_url': https://puppetserver.kvedulv.de:8140, 'puppet_version': 4.3.1, 'use_provider': [:puppet_proxy_puppet_api]
E, [2016-04-17T00:40:24.230596 #4314] ERROR -- : Disabling all modules in the group '[:puppet_proxy_puppet_api, :puppet]' due to a failure in one of them: undefined method `to_sym' for [:puppet_ssl_ca, :puppet_ssl_cert, :puppet_ssl_key]:Array
I, [2016-04-17T00:40:24.230670 #4314]  INFO -- : Successfully initialized 'foreman_proxy'
I, [2016-04-17T00:40:24.230730 #4314]  INFO -- : Successfully initialized 'tftp'
I, [2016-04-17T00:40:24.230807 #4314]  INFO -- : Successfully initialized 'puppetca'
```

mmoll:
I really won't rule out it's something in my setup...
@alexjfisher iirc you tested the previous PR by building a patched RPM and that works, could you re-test this PR in the same manner?

alexjfisher:
I took a backup of /etc/foreman-proxy and uninstalled the https://github.com/theforeman/smart-proxy/pull/376 version.
After installing the new RPM, I reinstated my configuration backup before starting the service.
This failed with ...

```
ERROR -- : Disabling all modules in the group '[:puppet_proxy_puppet_api, :puppet]' due to a failure in one of them: Provider 'puppet_proxy_puppet_api' settings conflict with the main plugin's settings: [:puppet_ssl_ca]
```

Which having taken a closer look, is fair enough ;)  This PR includes https://github.com/theforeman/smart-proxy/pull/390

So...

I reinstalled the old https://github.com/theforeman/smart-proxy/pull/376 version and made sure everything was still working.  Installing my new RPM I built for this PR, and letting the config migration happen this time, I get...

```
ERROR -- : Disabling all modules in the group '[:puppet_proxy_mcollective, :puppet_proxy_puppet_api, :puppet]' due to a failure in one of them: undefined method `to_sym' for [:puppet_ssl_ca, :puppet_ssl_cert, :puppet_ssl_key]:Array
```

... which is the same error @mmoll has encountered with his DEB.

dmitri-d:
@mmoll, @alexjfisher: thanks for testing! Found and fixed an issue -- I introduced a bug in plugin validator DSL when I was fixing 1.8.7-related issues. Could you test this again when you have a chance please?

alexjfisher:
@witlessbird Looks better this time.  At least for the puppet proxy (not sure if I'll get an opportunity soon to test the puppet-ca proxy part).

Importing environments still fails if any individual environment contains no classes. (eg if your base module path contains no modules).

dmitri-d:
@mmoll: is the issue you are seeing for legacy or puppet api-based provider? Could you post the smart-proxy log when that happens?

mmoll:
@witlessbird do you mean me or @alexjfisher? :)

dmitri-d:
Apologies, I meant @alexjfisher.

alexjfisher:
With a puppet 4 master.

```
ERROR -- : Failed to show puppet classes: Proxy::Error::HttpError
```

I am currently using quite an old foreman though (1.8.2).  Apologies if the bug is there (and maybe even fixed in recent versions) and not in the smart-proxy code.

dmitri-d:
@alexjfisher: fixed the issue with smart-proxy returning an error when listing classes for an empty environment (had to handle puppet resource api returning a 404 in such a case).

mmoll:
Did some light testing against Puppetserver 2.3 and it seems to work fine :tada:

dmitri-d:
[test]

alexjfisher:
@witlessbird I've now tested this with a puppetserver 2 based puppet CA too.  No problems encountered.

alexjfisher:
`ERROR -- : Couldn't enable plugin puppet: Gem loading error: cannot load such file -- augeas`

The puppet-agent package doesn't require ruby-augeas like the puppet 3 agent package did, so I think the foreman-proxy RPMs and debs will need to instead.

domcleal:
Could you rebase this please @witlessbird?

dmitri-d:
rebased.

domcleal:
Sorry, it needs another rebase, perhaps after b0ae4e0.

dmitri-d:
rebased.

modules/puppet_proxy/configuration_loader.rb
```diff
@@ -0,0 +1,20 @@
+module ::Proxy::Puppet
+  class ConfigurationLoader
+    def load_programmable_settings(settings)
+      raise ::Proxy::Error::ConfigurationError, "Parameter ':puppet_version' is expected to have a non-empty value" if settings[:puppet_version].to_s.empty?
```
domcleal, line 4: use `validates_presence` in the plugin definition?

config/settings.d/puppet.yml.example
```diff
@@ -3,54 +3,15 @@
...
-:puppetssh_command: /usr/bin/puppet agent --onetime --no-usecacheonfailure
-# wait for the command to finish (and capture exit code), or detach process and return 0
-# Note: enabling this option causes the Foreman web UI to be blocked when executing puppetrun,
-# with timeout from the Browser and/or Foreman's REST client after 60 seconds.
-:puppetssh_wait: false
-# With which user should the proxy connect
-#:puppetssh_user: root
-#:puppetssh_keyfile: /etc/foreman-proxy/id_rsa
+#   puppet_proxy_puppetrun   (for puppetrun/kick, deprecated in Puppet 3)
+#   puppet_proxy)mcollective (uses mco puppet)
```
domcleal: typo, underscore instead of `)`

config/settings.d/puppet_proxy_legacy.yml.example
```diff
@@ -0,0 +1,18 @@
+---
```
domcleal, line 7: Could you add some comments to the top of this and the api config file to indicate when they're used?  Perhaps also to puppet.yml near puppet_version so people know they exist.  It may be unclear to users which is in use.

modules/puppet_proxy/configuration_loader.rb
```diff
@@ -0,0 +1,20 @@
...
+
+      use_provider = settings[:use_provider]
+      use_provider = [use_provider].compact unless use_provider.is_a?(Array)
+      use_provider << (settings[:puppet_version].to_s >= "4.0" ? :puppet_proxy_puppet_api : :puppet_proxy_legacy)
```
domcleal, line 8: Some explicit debug logging to explain which provider was selected and why might be good here, to help point users to the right config file and know which behaviour's in use.

modules/puppet_proxy/configuration_loader.rb
dmitri-d, line 4: Can't rely on that here -- validators are run after settings are loaded and processed, and I need :puppet_version setting value here. It's either run validators several times at different stages of plugin initialization process (there are issues with this though) or do on the spot checks like this one.

modules/puppet_proxy/configuration_loader.rb
dmitri-d, line 8: I can log what providers are going to be used, would that work?

domcleal, line 8: Yes, that'd help.

modules/puppet_proxy/puppet_api.rb
```diff
@@ -43,30 +21,31 @@ def puppet_setup(opts = {})
...
   end
 
   get "/environments/:environment" do
     content_type :json
     begin
-      env = Proxy::Puppet::Environment.find(params[:environment])
-      log_halt 404, "Not found" unless env
+      env = environment_retriever.get(params[:environment])
       {:name => env.name, :paths => env.paths}.to_json
+    rescue  Proxy::Puppet::EnvironmentNotFound
```
domcleal: Nitpick, stray second space character before `Proxy::..`, ditto on line 46

modules/puppet_proxy_common/api_request.rb
```diff
@@ -0,0 +1,45 @@
...
+          http.verify_mode = OpenSSL::SSL::VERIFY_PEER
+        end
+
+        if ssl_cert && !ssl_cert.to_s.empty? && ssl_key && !ssl_key.to_s.empty?
+          http.cert = OpenSSL::X509::Certificate.new(File.read(ssl_cert))
+          http.key  = OpenSSL::PKey::RSA.new(File.read(ssl_key), nil)
+        end
+      end
+
+      path = [uri.path.end_with?("/") ? uri.path.slice(0..-2) : uri.path, path.start_with?("/") ? path.slice(1..-1) : path].join('/')
```
domcleal: `uri.path.chomp('/')` may be easier to read for the end_with? part

modules/puppet_proxy_common/custom_validators.rb
```diff
@@ -0,0 +1,12 @@
+module ::Proxy::Puppet
+  class Validators
+    class UrlValidator < ::Proxy::PluginValidators::Base
+      def validate!(settings)
+        raise ::Proxy::Error::ConfigurationError, "Setting 'puppet_url' is expected to contain a url for puppet server" if settings[:puppet_url].to_s.empty?
+        URI.parse(settings[:puppet_url])
+      rescue URI::InvalidURIError
+        raise ::Proxy::Error::ConfigurationError.new("Setting 'puppet_url' expected to contain a url for puppet server contains an invalid url")
```
domcleal: `Setting 'puppet_url' contains an invalid url for puppet server`

modules/puppet_proxy_customrun/puppet_proxy_customrun_plugin.rb
```diff
@@ -0,0 +1,10 @@
+module Proxy::PuppetCustomrun
+  class Plugin < Proxy::Provider
+    default_settings :customrun_cmd => '/bin/false', :customrun_args => "-ay -f -s"
```
domcleal: Perhaps just leave these out and require them to be set?  I don't think they're useful defaults, just a vague example from the config file.

modules/puppet_proxy_legacy/puppet_config_environments_retriever.rb
```diff
@@ -24,7 +24,7 @@ def all
     end
     if env.values.compact.empty?
       # fall back to defaults - we probably don't use environments
-      env[:production] = conf[:main][:modulepath] || conf[:master][:modulepath] || '/etc/puppet/modules'
+      env[:production] = (conf[:main][:modulepath] rescue nil) || (conf[:master][:modulepath] rescue nil) || '/etc/puppet/modules'
```
domcleal: What are the rescues for specifically?  If it's to catch nil values (e.g. `conf[:master]` missing) then we should handle that more explicitly, so as to let genuine exceptions be raised and logged - falling back to defaults here could have nasty results with Foreman's class importer.

config/settings.d/puppet.yml.example
```diff
@@ -3,54 +3,15 @@
...
+#   puppet_proxy_puppetssh   (run puppet over ssh)
+#   puppet_proxy_salt        (uses salt puppet.run)
+#   puppet_proxy_customrun   (calls a custom command with args)
+#:use_provider: puppet_proxy_puppetrun
 
 # Which user to invoke sudo as to run puppet commands
 #:puppet_user: root
```
domcleal: This doesn't seem to be used anywhere now, it looks like it used to be a fallback which is now migrated into both puppetrun/mcollective config files.

test/puppet/mcollective_test.rb
```diff
@@ -1,58 +1,49 @@
...
     @mcollective.stubs(:which).with("sudo", anything).returns("/usr/bin/sudo")
     @mcollective.stubs(:which).with("mco", anything).returns("/usr/bin/mco")
-    Proxy::Puppet::Plugin.settings.stubs(:mcollective_user).returns("peadmin")
+    ::Proxy::PuppetMCollective::Plugin.load_test_settings(:mcollective_user => "peadmin")
     @mcollective.expects(:shell_command).with(["/usr/bin/sudo", "-u", "peadmin", "/usr/bin/mco", "puppet", "runonce", "-I", "host1", "host2"]).returns(true)
-    assert @mcollective.run
+    assert @mcollective.run(["host1", "host2"])
   end
 
   def test_run_command_with_mcollective_user_should_take_precedence
```
domcleal: This test is obsolete, there's no :puppet_user setting any more.

modules/puppet_proxy_common/custom_validators.rb
dmitri-d: Hahaha, that's a good one.

test/puppet/puppet_legacy_runtime_configuration_test.rb
```diff
@@ -0,0 +1,87 @@
...
+    configuration = LegacyProviderConfigurationForTesting.new({})
+    assert_equal :cached_legacy_parser, configuration.load_programmable_settings(:use_cache => true)[:classes_retriever]
+  end
+
+  def test_load_programmable_settings_sets_legacy_parser
+    configuration = LegacyProviderConfigurationForTesting.new({})
+    assert_equal :legacy_parser, configuration.load_programmable_settings(:use_cache => false)[:classes_retriever]
+  end
+
+  def test_load_programmable_settings_sets_config_file_for_environments_retriever_for_puppet_before_3_2
```
domcleal: I think it's missing a test case for using `:api_v2` with `:puppet_version => "3.2"` and no override.

domcleal:
@alexjfisher re https://github.com/theforeman/smart-proxy/pull/406#issuecomment-216524978:

> ERROR -- : Couldn't enable plugin puppet: Gem loading error: cannot load such file -- augeas

If you're using Puppet Server 2 then the Augeas dependency shouldn't be loaded at all, I couldn't reproduce this in a quick test from source. I wonder if you can pastebin corresponding debug logs so we can check it's configured correctly?

domcleal:
@witlessbird thanks, first pass review done and I'll do more thorough testing on the next round.  A few general comments:

1) The new env/class retriever clearly has no cache, since we're not parsing and tracking individual files. Have you checked whether Puppet Server itself performs any caching by default, and if this might lead to pre-caching performance again on class imports?

I recall caching was enabled [briefly in 3.7 and disabled later](https://docs.puppet.com/puppet/3.7/reference/release_notes.html#new-default-value-environmenttimeout--0), so I'd guess the default behaviour with Puppet Server 2 would also be without any caching (in PS or in the smart proxy).

2) This would be a good time to tidy up some settings names during the migration - since they're split into new separate config files, module name prefixes should be removed where appropriate.  Some ideas:

```
:puppetssh_command => :command
:customrun_cmd => :command
:puppet_use_environment_api => :use_environment_api
```

3) puppet_api's /run call should be tested (test/puppet/puppet_api_test.rb)

alexjfisher:
@domcleal Sorry, my mistake.  The error was on a server I'd been testing the older https://github.com/theforeman/smart-proxy/pull/376 PR out on.

As mentioned in this PRs very first
https://github.com/theforeman/smart-proxy/pull/406#issue-148107992 ,this should no longer be an issue.

domcleal:
Good, thanks for confirming.

dmitri-d:
responded to all feedback but things in "@witlessbird thanks, first pass review done and I'll do more thorough testing on the next round. A few general comments: ... " note.

also added tests for https://github.com/theforeman/smart-proxy/pull/406/files#diff-9b357977669579ae92d74c22420de796R25. Looks like there was a long-standing bug in `use_environment_api?`: it would return true if environmentpath in puppet configuration wasn't empty.

dmitri-d:
added tests for puppet_api's /run call.

dmitri-d:
> 1) The new env/class retriever clearly has no cache, since we're not parsing and tracking individual files. Have you checked whether Puppet Server itself performs any caching by default, and if this might lead to pre-caching performance again on class imports?

Based on experience I'd say so -- first calls take a bit of time, all subsequent are instantaneous. Also, puppet server returns an ETAG with their new environment-classes api, which would suggest use of cache. I'll look into implementation details to confirm.  

> I recall caching was enabled briefly in 3.7 and disabled later, so I'd guess the default behaviour with Puppet Server 2 would also be without any caching (in PS or in the smart proxy). 

Using cache on our side would depend on PS supplying ETAG with their responses, which is already done for environment-classes end point but AFAIK not for environments. This would only be needed if  PS is slow enough to warrant use of cache on our end.

domcleal:
Setting WoC as I think the setting name comment remains unaddressed.

dmitri-d:
Nods, renaming settings will take a bit of work...

domcleal:
A rebase is also needed due to the MCollective provider merge.

dmitri-d:
Rebased, renamed configuration parameter names (I think I got all of them).

modules/puppet_proxy_legacy/plugin_configuration.rb
```diff
@@ -0,0 +1,109 @@
...
+      end
+
+      container_instance.dependency :puppet_initializer, lambda {Proxy::PuppetLegacy::Initializer.new(settings[:puppet_conf]) }
+
+      case settings[:classes_retriever]
+        when :cached_future_parser
+          require 'puppet_proxy_legacy/class_scanner_eparser'
+          require 'puppet_proxy_legacy/puppet_cache'
+          container_instance.singleton_dependency :class_retriever_impl, (lambda do
+            ::Proxy::PuppetLegacy::PuppetCache.new(c.get_dependency(:environment_retriever_impl), ::Proxy::PuppetLegacy::ClassScannerEParser.new(nil, c.get_dependency(:puppet_initializer)))
```
domcleal: `c` is undefined in this context, and I suspect in all of the ones below too:

<pre>
E, [2016-05-17T11:16:11.588725 #19944] ERROR -- : Couldn't enable 'puppet_proxy_legacy': undefined local variable or method `c' for #<Proxy::PuppetLegacy::PluginConfiguration:0x000000036e7230>
D, [2016-05-17T11:16:11.588814 #19944] DEBUG -- : ["/home/dcleal/code/foreman/smart-proxy/modules/puppet_proxy_legacy/plugin_configuration.rb:73:in `block in load_dependency_injection_wirings'", "/home/dcleal/code/foreman/smart-proxy/lib/proxy/dependency_injection.rb:16:
in `call'", "/home/dcleal/code/foreman/smart-proxy/lib/proxy/dependency_injection.rb:16:in `initialize'", "/home/dcleal/code/foreman/smart-proxy/lib/proxy/dependency_injection.rb:50:in `new'", "/home/dcleal/code/foreman/smart-proxy/lib/proxy/dependency_injection.rb:50:in
 `singleton_dependency'", "/home/dcleal/code/foreman/smart-proxy/modules/puppet_proxy_legacy/plugin_configuration.rb:72:in `load_dependency_injection_wirings'", "/home/dcleal/code/foreman/smart-proxy/lib/proxy/plugin_initializer.rb:190:in `block in wire_up_dependencies'"
, "/home/dcleal/code/foreman/smart-proxy/lib/proxy/plugin_initializer.rb:189:in `each'", "/home/dcleal/code/foreman/smart-proxy/lib/proxy/plugin_initializer.rb:189:in `wire_up_dependencies'", "/home/dcleal/code/foreman/smart-proxy/lib/proxy/plugin_initializer.rb:179:in `
configure_plugin'", "/home/dcleal/code/foreman/smart-proxy/lib/proxy/plugin_initializer.rb:67:in `block in configure'", "/home/dcleal/code/foreman/smart-proxy/lib/proxy/plugin_initializer.rb:67:in `each'", "/home/dcleal/code/foreman/smart-proxy/lib/proxy/plugin_initializ
er.rb:67:in `configure'", "/home/dcleal/code/foreman/smart-proxy/lib/proxy/plugin_initializer.rb:135:in `block in initialize_plugins'", "/home/dcleal/code/foreman/smart-proxy/lib/proxy/plugin_initializer.rb:135:in `each'", "/home/dcleal/code/foreman/smart-proxy/lib/proxy
/plugin_initializer.rb:135:in `initialize_plugins'", "/home/dcleal/code/foreman/smart-proxy/lib/launcher.rb:120:in `launch'", "bin/smart-proxy:6:in `<main>'"]
</pre>

lib/proxy/plugin_initializer.rb
```diff
@@ -24,9 +24,14 @@ def resolve_providers(all_plugins_and_providers)
     providers = all_plugins_and_providers.select {|p| used_providers.include?(p[:name].to_sym)}
 
     not_available = used_providers - providers.map {|p| p[:name].to_sym}
-    return @providers = providers.map {|p| p[:class]} if not_available.empty?
 
-    fail_group_with_message("Disabling all modules in the group '#{member_names.join(', ')}': following providers are not available #{not_available}")
+    if not_available.empty?
+
```
domcleal: Nitpick, blank line

modules/puppet_proxy_legacy/plugin_configuration.rb
dmitri-d: ugh, yeah, this is remnant from when these lambdas had a parameter. I wonder if I should write tests that verify that each of the dependencies can be instantiated; atm I have for the dependencies that have settings as parameters (to validate that parameters are being passed correctly).

domcleal: Yep, that'd be valuable I think.

dmitri-d:
Fixed legacy provider wirings. Added di wiring tests for legacy and puppet api providers.

domcleal:
There are some test failures appearing on Puppet < 3.2 where the EParser class defined in test/puppet/future_parser_classes_retriever_test.rb doesn't have the same initialiser arity as the real class. The comment on the dummy class suggests it isn't needed to execute the tests, but that'll have to change - and perhaps only load it when < 3.2?

dmitri-d:
I moved Proxy::PuppetLegacy::ClassScannerEParser definition at https://github.com/theforeman/smart-proxy/pull/406/files#diff-f9930bafb2ea975c80645d3f4f0b23b5R5 outside of the conditional, only with `require 'puppet/pops'` remaining there. I also removed the dummy class definition from future_parser_classes_retriever_test.rb (the test suite only gets executed with puppet > 3.2).

modules/puppet_proxy_common/custom_validators.rb
```diff
@@ -0,0 +1,12 @@
+module ::Proxy::Puppet
+  class Validators
+    class UrlValidator < ::Proxy::PluginValidators::Base
+      def validate!(settings)
```
domcleal, line 4: This is missing a check for the `:if` predicate in the plugin description:

```
return true unless evaluate_predicate(settings)
```

(causing it to fail when :puppet_url is unset, i.e. for config file based environment detection)

dmitri-d:
Fixed the issue with predicates for custom validators -- they no longer need to explicitly evaluate the predicate. Updated and expanded the tests.

domcleal:
[test] - one of the slaves ran out of disk space.

domcleal:
Merged as df0fc72, thanks @witlessbird.

PR #408, opened 2016-04-19
Fixes #14719 - Add setting to allow TLSv1.

smithj4:
Php in both RHEL 6 & 7 cannot connect to the Foreman smart-proxy if TLSv1 is not allowed. Setting php to use TLSv1 or above will only work if you have curl 7.34 or newer. RHEL6 comes with curl 7.19 and RHEL7 comes with curl 7.29. See:

http://php.net/manual/en/function.curl-setopt.php#115993

### DIFF
--- a/lib/launcher.rb
+++ b/lib/launcher.rb
@@ -53,7 +53,7 @@ module Proxy
       # This is required to disable SSLv3 on Ruby 1.8.7
       ssl_options |= OpenSSL::SSL::OP_NO_SSLv2 if defined?(OpenSSL::SSL::OP_NO_SSLv2)
       ssl_options |= OpenSSL::SSL::OP_NO_SSLv3 if defined?(OpenSSL::SSL::OP_NO_SSLv3)
-      ssl_options |= OpenSSL::SSL::OP_NO_TLSv1 if defined?(OpenSSL::SSL::OP_NO_TLSv1)
+      ssl_options |= OpenSSL::SSL::OP_NO_TLSv1 if !SETTINGS.enable_tls_v1 && defined?(OpenSSL::SSL::OP_NO_TLSv1)
 
       {
         :app => app,
--- a/lib/proxy/settings/global.rb
+++ b/lib/proxy/settings/global.rb
@@ -11,7 +11,8 @@ module ::Proxy::Settings
       :bind_host => "*",
       :log_buffer => 2000,
       :log_buffer_errors => 1000,
-      :ssl_disabled_ciphers => []
+      :ssl_disabled_ciphers => [],
+      :enable_tls_v1 => false
     }
 
     HOW_TO_NORMALIZE = {
--- a/config/settings.yml.example
+++ b/config/settings.yml.example
@@ -13,6 +13,11 @@
 # for more information.
 #:ssl_disabled_ciphers: [CIPHER-SUITE-1, CIPHER-SUITE-2]
 
+# Use this option only if you need to enable TLSv1, the default value is false.
+# Note: setting this option will lower security as there are known vulnerabilities
+# with TLSv1, only set this if absolutely necessary.
+#:enable_tls_v1: false
+
 # Hosts which the proxy accepts connections from
 # commenting the following lines would mean every verified SSL connection allowed
 # HTTPS: test the certificate CN


dmitri-d:
I don't think allowing TLSv1 is the way to go. A configuration option to enable TLSv1 (should be disabled by default) would be better. A comment for the option should include a warning that enabling TLSv1 lowers security.

smithj4:
I agree that a config setting would be better, but I am not a ruby developer and I wouldn't know where to begin patching the foreman smart-proxy code to add such an option. I will also be unavailable for the next 2 weeks. Something needs to be done to fix this though, since the whole point of a REST API is to allow 3rd parties access and the current code will not even allow a PHP client on a relatively recent RHEL7 host to connect. Better security is great, but also counterproductive if it completely blocks all access because of supported protocol compatibility.

Our foreman server & proxy are on a firewalled private network so there is little danger in allowing TLSv1 for us. TLSv1.1 is also vulnerable, yet #401 removed the line that also disabled it. That line should probably be put back if a config setting is added.

brandonweeks:
@smithj4 I think I did a pretty good job of justifying #401 on the PR. That PR was a modification to the changes I made in #351 which originally was intended to disable both TLSv1 and TLSv1.1, alas it is simply not possible to disable TLSv1.1 when running under Ruby 1.8.7.

@witlessbird how would you like a setting for this to look?

dmitri-d:
@smithj4: I second @brandonweeks -- TLSv1.1 is going to be disabled as soon as we stop supporting ruby 1.8.7. 
@brandonweeks: not sure: I don't think we need to bother with enabling SSLv*, the setting can be limited to TLS only. Considering the low number of available versions of TLS, we could have dedicated settings, e.g. "enable_tls_v1" (true/false), enable tls_v1_1 (true/false). Alternatively, we can have "disabled_tls" (probably not a great name?) and list disabled versions of TLS there.

lib/launcher.rb
```diff
@@ -46,7 +46,9 @@ def https_app
         # This is required to disable SSLv3 on Ruby 1.8.7
         ssl_options |= OpenSSL::SSL::OP_NO_SSLv2 if defined?(OpenSSL::SSL::OP_NO_SSLv2)
         ssl_options |= OpenSSL::SSL::OP_NO_SSLv3 if defined?(OpenSSL::SSL::OP_NO_SSLv3)
-        ssl_options |= OpenSSL::SSL::OP_NO_TLSv1 if defined?(OpenSSL::SSL::OP_NO_TLSv1)
+        if SETTINGS.enable_tls_v1 == false
+          ssl_options |= OpenSSL::SSL::OP_NO_TLSv1 if defined?(OpenSSL::SSL::OP_NO_TLSv1)
+        end
```
dmitri-d, line 48: This is a bit too verbose, could you change to something like:
`ssl_options |= OpenSSL::SSL::OP_NO_TLSv1 if (!SETTINGS.enable_tls_v1 && defined?(OpenSSL::SSL::OP_NO_TLSv1))`

lib/proxy/settings/global.rb
```diff
@@ -12,7 +12,8 @@ class Global < ::OpenStruct
       :virsh_network => 'default',
       :log_buffer => 2000,
       :log_buffer_errors => 1000,
-      :ssl_disabled_ciphers => []
+      :ssl_disabled_ciphers => [],
+      :enable_tls_v1 => false,
```
dmitri-d: extra comma at the end.

mmoll:
[test]

lib/launcher.rb
```diff
@@ -46,7 +46,7 @@ def https_app
         # This is required to disable SSLv3 on Ruby 1.8.7
         ssl_options |= OpenSSL::SSL::OP_NO_SSLv2 if defined?(OpenSSL::SSL::OP_NO_SSLv2)
         ssl_options |= OpenSSL::SSL::OP_NO_SSLv3 if defined?(OpenSSL::SSL::OP_NO_SSLv3)
-        ssl_options |= OpenSSL::SSL::OP_NO_TLSv1 if defined?(OpenSSL::SSL::OP_NO_TLSv1)
+        ssl_options |= OpenSSL::SSL::OP_NO_TLSv1 if (!SETTINGS.enable_tls_v1 && defined?(OpenSSL::SSL::OP_NO_TLSv1))
```
dmitri-d: Could you remove the parenthesis around the condition? Rubocop doesn't like them.

brandonweeks:
Wait, I can't reproduce this on either CentOS 6 or 7. Using the following script I can connect to a smart-proxy with only TLSv1.1 and 1.2 configured. Can you provide a test case? The PHP documents aren't really valid given how much RedHat backports patches, especially for SSL/TLS support. 

```
<?php
$curl = curl_init();
curl_setopt_array($curl, array(
    CURLOPT_RETURNTRANSFER => 1,
    CURLOPT_URL => 'https://smart-proxy:8443',
));
$resp = curl_exec($curl);
echo $resp;
curl_close($curl);
?>
```

fubarwrangler:
The following PHP on Scientific Linux 6 (RHEL-compatable) fails.

```
    $url = "https://puppet:8443/"
    $caFile = "ca.pem";
    $keyFile = "{$hostname}-private.pem";
    $certFile = "{$hostname}.pem";

    $ch = curl_init($url);

    curl_setopt($ch, CURLOPT_URL, $url);
    curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
    curl_setopt($ch, CURLOPT_SSLKEY, $keyFile);
    curl_setopt($ch, CURLOPT_CAINFO, $caFile);
    curl_setopt($ch, CURLOPT_SSLCERT, $certFile);

    $d = curl_exec($ch);
    if(!$d) {
        $err = curl_error($ch);
        echo "Error: $err";
        return array();
    }

```

as does running curl from the command-line without explicitly setting the TLS version via command-line flag

dLobatog:
[test]

dmitri-d:
@fubarwrangler: You should be able to specify the TLS version to use via `CURLOPT_SSLVERSION` set to 5 (tls 1.1) or 6 (tls 1.2).

fubarwrangler:
@witlessbird: I see now, there is a very recent (1 month ago) redhat update that includes this option.

[Bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=1255920)
[Update](https://rhn.redhat.com/errata/RHBA-2016-0842.html)

It seems that Scientific Linux hasn't ported to 6.8 yet or I'm missing it, so this is still an issue (but I can easily upgrade when it is available). However, given the conservative nature of RHEL this issue could persist in organizations that don't update their clients for some time and is still worth fixing IMO.

nitinz:
Allowing TLSv1 would also fix pxe kickstart file pull from clients. It currently fails, if the foreman-proxy is serving templates. I dont find any specific option to force anaconda to use any version. Can you expedidate fix on this bug?

fyi, This bug is reproducible on rhel6/Ruby-1.8.7 and rhel7/Ruby-2.0.0.

dmitri-d:
Closing this PR: there were no other reports of this issue, and it's been fixed in rhel 6 php package too.

PR #411, opened 2016-04-27
Fixes #14862 - pin webmock version for Ruby <1.9.3

mmoll:
this is also needed in 1.11 and 1.10


domcleal:
Merged as cfb732d, thanks @mmoll.

PR #414, opened 2016-05-05
Fixes #12633 - Pxegrub2 variant and multiple configs

lzap:
Foreman Proxy must be enhanced with new PXE variant "Pxegrub2" to allow
deploying configuration files in "grub2/" folder. In addition to that, Pxegrub
(version 1) must deploy two configuration files instead of one now to support
both BIOS and UEFI naming conventions (grub/menu.lst vs grub/efidefault and
grub/menu.lst.01aabbccddeeff vs grub/01-AA-BB-CC-DD-EE-FF). Also, Pxegrub
variant will use new base directory "grub/" instead of "boot/grub/" for
consistency with "grub2" and "pxelinux.cfg/".

More details at:
http://projects.theforeman.org/projects/foreman/wiki/PXE_Booting_UEFI

### DIFF
--- a/modules/tftp/server.rb
+++ b/modules/tftp/server.rb
@@ -7,52 +7,59 @@ module Proxy::TFTP
     # Creates TFTP pxeconfig file
     def set mac, config
       raise "Invalid parameters received" if mac.nil? || config.nil?
-
-      FileUtils.mkdir_p pxeconfig_dir
-
-      File.open(pxeconfig_file(mac), 'w') {|f| f.write(config) }
-      logger.debug "TFTP: entry for #{mac} created successfully"
+      pxeconfig_file(mac).each do |file|
+        write_file file, config
+      end
+      true
     end
 
     # Removes pxeconfig files
     def del mac
-      file = pxeconfig_file(mac)
-      if File.exist?(file)
-        FileUtils.rm_f file
-        logger.debug "TFTP: entry for #{mac} removed successfully"
-      else
-        logger.debug "TFTP: Skipping a request to delete a file which doesn't exists"
+      pxeconfig_file(mac).each do |file|
+        delete_file file
       end
+      true
     end
 
-    # Gets the contents of a pxeconfig file
+    # Gets the contents of one of pxeconfig files
     def get mac
-      file = pxeconfig_file(mac)
-      if File.exist?(file)
-        config = File.open(pxeconfig_file(mac), 'r') {|f| f.readlines }
-        logger.debug "TFTP: entry for #{mac} read successfully"
-      else
-        logger.debug "TFTP: Skipping a request to read a file which doesn't exists"
-        raise "File #{file} not found"
-      end
-      config
+      file = pxeconfig_file(mac).first
+      read_file(file)
     end
 
     # Creates a default menu file
     def create_default config
       raise "Default config not supplied" if config.nil?
-
-      FileUtils.mkdir_p File.dirname pxe_default
-      File.open(pxe_default, 'w') {|f| f.write(config) }
-      logger.debug "TFTP: #{pxe_default} entry created successfully"
+      pxe_default.each do |file|
+        write_file file, config
+      end
+      true
     end
 
-    protected
     # returns the absolute path
-    # TODO:
     def path(p = nil)
       p ||= Proxy::TFTP::Plugin.settings.tftproot
       return (p =~ /^\//) ? p : Pathname.new(File.expand_path(File.dirname(__FILE__))).join(p).to_s
+    end
+
+    def read_file(file)
+      raise("File #{file} not found") unless File.exist?(file)
+      File.open(file, 'r') {|f| f.readlines }
+    end
+
+    def write_file(file, contents)
+      FileUtils.mkdir_p(File.dirname(file))
+      File.open(file, 'w') {|f| f.write(contents)}
+      logger.debug "TFTP: #{file} created successfully"
+    end
+
+    def delete_file(file)
+      if File.exist?(file)
+        FileUtils.rm_f file
+        logger.debug "TFTP: #{file} removed successfully"
+      else
+        logger.debug "TFTP: Skipping a request to delete a file which doesn't exists"
+      end
     end
   end
 
@@ -61,22 +68,34 @@ module Proxy::TFTP
       "#{path}/pxelinux.cfg"
     end
     def pxe_default
-      "#{pxeconfig_dir}/default"
+      ["#{pxeconfig_dir}/default"]
     end
     def pxeconfig_file mac
-      "#{pxeconfig_dir}/01-"+mac.gsub(/:/,"-").downcase
+      ["#{pxeconfig_dir}/01-"+mac.gsub(/:/,"-").downcase]
     end
   end
 
   class Pxegrub < Server
     def pxeconfig_dir
-      path.to_s
+      "#{path}/grub"
     end
     def pxe_default
-      "#{pxeconfig_dir}/boot/grub/menu.lst"
+      ["#{pxeconfig_dir}/menu.lst", "#{pxeconfig_dir}/efidefault"]
     end
     def pxeconfig_file mac
-      "#{pxeconfig_dir}/menu.lst.01"+mac.gsub(/:/,"").upcase
+      ["#{pxeconfig_dir}/menu.lst.01"+mac.gsub(/:/,"").upcase, "#{pxeconfig_dir}/01-"+mac.gsub(/:/,'-').upcase]
+    end
+  end
+
+  class Pxegrub2 < Server
+    def pxeconfig_dir
+      "#{path}/grub2"
+    end
+    def pxe_default
+      ["#{pxeconfig_dir}/grub.cfg"]
+    end
+    def pxeconfig_file mac
+      ["#{pxeconfig_dir}/grub.cfg-"+mac.gsub(/:/,'-').downcase]
     end
   end
 
@@ -85,10 +104,10 @@ module Proxy::TFTP
       "#{path}/ztp.cfg"
     end
     def pxe_default
-      pxeconfig_dir
+      [pxeconfig_dir]
     end
     def pxeconfig_file mac
-      "#{pxeconfig_dir}/"+mac.gsub(/:/,"").upcase
+      ["#{pxeconfig_dir}/"+mac.gsub(/:/,"").upcase]
     end
   end
 
@@ -97,15 +116,14 @@ module Proxy::TFTP
       "#{path}/poap.cfg"
     end
     def pxe_default
-      pxeconfig_dir
+      [pxeconfig_dir]
     end
     def pxeconfig_file mac
-      "#{pxeconfig_dir}/"+mac.gsub(/:/,"").upcase
+      ["#{pxeconfig_dir}/"+mac.gsub(/:/,"").upcase]
     end
   end
 
   def self.fetch_boot_file dst, src
-
     filename    = boot_filename(dst, src)
     destination = Pathname.new(File.expand_path(filename, Proxy::TFTP::Plugin.settings.tftproot)).cleanpath
     tftproot    = Pathname.new(Proxy::TFTP::Plugin.settings.tftproot).cleanpath
--- a/test/tftp/tftp_api_test.rb
+++ b/test/tftp/tftp_api_test.rb
@@ -14,12 +14,35 @@ class TftpApiTest < Test::Unit::TestCase
 
   def setup
     Proxy::TFTP::Plugin.load_test_settings(:tftproot => "/some/root")
-    @args = { :pxeconfig => "foo" }
+    @args = {
+      :pxeconfig => "foo",
+      :menu => "bar"
+    }
   end
 
   def test_instantiate_syslinux
     obj = app.helpers.instantiate "syslinux", "AA:BB:CC:DD:EE:FF"
     assert_equal "Proxy::TFTP::Syslinux", obj.class.name
+  end
+
+  def test_instantiate_pxegrub
+    obj = app.helpers.instantiate "pxegrub", "AA:BB:CC:DD:EE:FF"
+    assert_equal "Proxy::TFTP::Pxegrub", obj.class.name
+  end
+
+  def test_instantiate_pxegrub2
+    obj = app.helpers.instantiate "pxegrub2", "AA:BB:CC:DD:EE:FF"
+    assert_equal "Proxy::TFTP::Pxegrub2", obj.class.name
+  end
+
+  def test_instantiate_ztp
+    obj = app.helpers.instantiate "ztp", "AA:BB:CC:DD:EE:FF"
+    assert_equal "Proxy::TFTP::Ztp", obj.class.name
+  end
+
+  def test_instantiate_poap
+    obj = app.helpers.instantiate "poap", "AA:BB:CC:DD:EE:FF"
+    assert_equal "Proxy::TFTP::Poap", obj.class.name
   end
 
   def test_instantiate_nonexisting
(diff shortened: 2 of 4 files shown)


modules/tftp/server.rb
```diff
@@ -7,33 +7,35 @@ class Server
...
     def set mac, config
       raise "Invalid parameters received" if mac.nil? || config.nil?
-
       FileUtils.mkdir_p pxeconfig_dir
-
-      File.open(pxeconfig_file(mac), 'w') {|f| f.write(config) }
-      logger.info "TFTP: entry for #{mac} created successfully"
+      pxeconfig_file(mac).each do |file|
+        File.open(file, 'w') {|f| f.write(config) }
+        logger.info "TFTP: #{file} created successfully"
```
dmitri-d: Can you change "info"-level logging with "debug" one (here and below)?

modules/tftp/server.rb
```diff
@@ -7,33 +7,35 @@ class Server
...
-        FileUtils.rm_f file
-        logger.debug "TFTP: entry for #{mac} removed successfully"
-      else
-        logger.info "TFTP: Skipping a request to delete a file which doesn't exists"
+      pxeconfig_file(mac).each do |file|
+        if File.exist?(file)
+          FileUtils.rm_f file
+          logger.info "TFTP: #{file} removed successfully"
+        else
+          logger.warn "TFTP: Skipping a request to delete a file which doesn't exists"
```
dmitri-d: This one too?

test/tftp/tftp_api_test.rb
```diff
@@ -26,22 +26,64 @@ def test_api_can_fetch_boot_file
   end
 
   def test_api_can_create_syslinux_tftp_reservation
```
dmitri-d: I think these tests should be broken into two or three suites:
- a suite to test Proxy::TFTP::Server, in particular to verify that files are created (with content we expect), deleted, etc. I'd avoid mocking altogether for this, and rely on Tempfile instead (which are automatically deleted during garbage collection/on vm exit)
- a suite of tests to verify paths for subclasses of ::Proxy::TFTP::Server
- a suite of tests to verify tftp api (ok to mock Proxy::TFTP::Server interface)

modules/tftp/server.rb
lzap: > Can you change "info"-level logging with "debug" one (here and below)?

Changed all three then.

## 

Later,
 Lukas #lzap Zapletal

modules/tftp/server.rb
lzap: > > -          logger.warn "TFTP: Skipping a request to delete a file which doesn't exists"
> 
> This one too?

This one is actually correct I think - this is a reverse of previous
orchestration and it should always work correctly. We should definitely
leave a warning message when there was an inconsistency (user e.g. added
TFTP association for previously provisioned host).

## 

Later,
 Lukas #lzap Zapletal

dmitri-d: It doesn't look like this has any consequences for/may require an action from the user. I don't think we should log stuff at the level above "debug" ("info" for settings on startup is an exception) unless some action may be required (the level denotes the urgency of such an action).

dmitri-d: Having said that, I'm not _that_ against it. I _would like_ to reduce the amount of logging in smart-proxy, at least at levels above "debug".

lzap: Ignore the above if your comment was for line no 22.

test/tftp/tftp_api_test.rb
lzap: I have broken into two suites - one is testing the API, the other one is testing the Server and its implementations. Both paths and code logic are executed. I am keeping stubbing because when I started with tempfiles solution, it actually was a longer code. TFTP module does not change many often.

lzap:
Rebased and added those instantiation tests which I believe are useless.

modules/tftp/server.rb
lzap: > Having said that, I'm not _that_ against it. I _would like_ to reduce the amount of logging in smart-proxy, at least at levels above "debug".  

I generally agree, sure this code path seems to be sufficient to keep.

## 

Later,
 Lukas #lzap Zapletal

lzap: > It doesn't look like this has any consequences for/may require an action from the user. I don't think we should log stuff at the level above "debug" ("info" for settings on startup is an exception) unless some action may be required (the level denotes the urgency of such an action).

Don't agree, having investigated some Satellite 6 deployments with
complex setups, any kind of information is very helpful to find out what
happened. This one line can for example confirm that a volume was
unmounted at given point, or directory was wiped out etc.

## 

Later,
 Lukas #lzap Zapletal

lzap:
Failed because of the bundler issue.

lzap:
It looks like JSON responses of the API changed, I need to add asserts for that and compare with develop.

lzap:
Added, it looks like previously the API was handing over return values of `logger` calls, I changed this to be always empty.

test/tftp/tftp_server_test.rb
```diff
@@ -0,0 +1,104 @@
...
+    @content = "file content"
+    @file_handle = mock
+    @file_handle.stubs(:write).with(@content)
+    @file_handle.stubs(:readlines).returns(@content)
+    Proxy::TFTP::Plugin.settings.stubs(:tftproot).returns(@rootdir)
+    setup_paths
+  end
+
+  def test_set
+    FileUtils.stubs(:mkdir_p).returns(true)
```
dmitri-d: If you want to use mocks for these tests, can you extract file-related operations into methods in `Proxy::TFTP::Server` class? For example, around https://github.com/lzap/smart-proxy/blob/pxegrub2/modules/tftp/server.rb#L10 would become something like:

```
def write_config_for_mac(dir, file, config)
  FileUtils.mkdir_p dir
  File.open(File.join(dir, file), 'w') {|f| f.write(config) }
end
```

etc.

Such a refactor would both improve readability (both code and tests) and reduce the fragility of mocks...

lzap: > Such a refactor would both improve readability (both code and tests) and reduce the fragility of mocks...

Excellent idea.

## 

Later,
 Lukas #lzap Zapletal

lzap:
Fixed tests, it's now much more simple with the two little helper methods!

test/tftp/tftp_api_test.rb
```diff
@@ -14,60 +14,99 @@ def app
...
+
+  def test_api_can_create_defatult
+    Proxy::TFTP::Syslinux.any_instance.expects(:create_default).with(@args[:menu]).returns(true)
+    post "/create_default", @args
+    assert last_response.ok?
+  end
+
+  def test_api_can_fetch_boot_file
+    Proxy::Util::CommandTask.stubs(:new).returns(true)
+    FileUtils.stubs(:mkdir_p).returns(true)
```
dmitri-d: these two stubs aren't needed anymore.

test/tftp/tftp_server_test.rb
```diff
@@ -0,0 +1,110 @@
...
+
+  def pxe_default_files
+    @pxe_default_files.collect{|f| File.join(@rootdir, f)}
+  end
+
+  def test_set
+    pxe_config_files.each do |file|
+      @subject.stubs(:write_file).with(file, @content)
+    end
+    @subject.expects(:write_file).with(pxe_config_files.first, @content).once
```
dmitri-d: This doesn't look right to me -- first you stub two calls, and then you set an expectation on one of them. I _think_ what you meant here was to set expectations on two "write_file" calls that must be called...

dmitri-d: Actually that depends on the number of the config files -- the expectations will need to be set similarly to how calls are stubbed atm.

test/tftp/tftp_server_test.rb
```diff
@@ -0,0 +1,110 @@
...
+    @subject.set @mac, @content
+  end
+
+  def test_del
+    pxe_config_files.each do |file|
+      File.stubs(:exist?).with(file).returns(true)
+      FileUtils.stubs(:rm_f).with(file)
+    end
+    @subject.del @mac
```
dmitri-d, line 34: This test doesn't verify anything, as there are no expectations, only stubbed calls.

test/tftp/tftp_server_test.rb
```diff
@@ -0,0 +1,110 @@
...
+  end
+
+  def test_get
+    file = pxe_config_files.first
+    File.stubs(:exist?).with(file).returns(true)
```
dmitri-d: Perhaps move file existence check inside "read_file" method?

test/tftp/tftp_server_test.rb
dmitri-d, line 34: Also, perhaps move file existence check and removal into a dedicated method, similar to write_file and read_file?

test/tftp/tftp_server_test.rb
```diff
@@ -0,0 +1,110 @@
...
+    @subject.expects(:read_file).with(file).returns(@content)
+    assert_equal @content, @subject.get(@mac)
+  end
+
+  def test_create_default
+    FileUtils.stubs(:mkdir_p).returns(true)
```
dmitri-d: No need for this stub anymore.

test/tftp/tftp_server_test.rb
```diff
@@ -0,0 +1,110 @@
...
+    pxe_default_files.each do |file|
+      @subject.stubs(:write_file).with(file, @content)
+    end
+    @subject.expects(:write_file).with(pxe_default_files.first, @content)
```
dmitri-d: Same issue as I pointed out in the comment for https://github.com/theforeman/smart-proxy/pull/414/files#diff-a25bd017a8ff9ea98b4d47183ffa2d50R23.

test/tftp/tftp_server_test.rb
lzap: It's a limitation of Mocha, I can't create multiple expectations with different parameters. I can only do what I propose or expect N calls of the method without checking parameters.

test/tftp/tftp_server_test.rb
lzap: Yup

test/tftp/tftp_server_test.rb
lzap: Yeah

lzap:
Amended all your remarks except the multiple expects/stubs problem. Let me know how you would like to handle it.

test/tftp/tftp_server_test.rb
dmitri-d: Mocha supports mocking of repeated calls with different parameter values, including verification of the order of calls:
 :

```
  class A
    def a(arg); end
    def b
      a(:first)
      a(:second)
    end
  end

  def test_multiple_mocks_disregarding_call_order
    under_test = A.new

    under_test.expects(:a).with(:first)
    under_test.expects(:a).with(:second).once

    under_test.b
  end

  def test_multiple_mocks_in_order
    under_test = A.new
    call_sequence = sequence('test_sequence')

    under_test.expects(:a).with(:first).in_sequence(call_sequence)
    under_test.expects(:a).with(:second).in_sequence(call_sequence)

    under_test.b
  end
```

lzap: Oh interesting, not sure what made me believe it does not! That does make sense of course. Amended.

modules/tftp/server.rb
```diff
@@ -7,76 +7,98 @@ class Server
     # Creates TFTP pxeconfig file
     def set mac, config
       raise "Invalid parameters received" if mac.nil? || config.nil?
-
       FileUtils.mkdir_p pxeconfig_dir
```
dmitri-d: I don't think this is needed anymore -- the dir is being created in write_file method.

test/tftp/tftp_server_test.rb
```diff
@@ -0,0 +1,105 @@
+require 'test_helper'
+require 'tftp/server'
+require 'tftp/tftp_plugin'
+
+module TftpGenericServerSuite
+  def setup
+    @rootdir = "/some/root"
+    @mac = "aa:bb:cc:dd:ee:ff"
+    @content = "file content"
+    FileUtils.stubs(:mkdir_p).returns(true)
```
dmitri-d: I don't think this stub is needed: mkdir_p call is made in set method; write_file takes care of creating of the dir now.

dmitri-d:
Looks good. It would be useful to add tests that verify that  Proxy::TFTP::Server#read_file, Proxy::TFTP::Server#write_file, and Proxy::TFTP::Server#delete_file methods do what we expect them to.

dmitri-d:
[test]

lzap:
Amended both comments.

> Looks good. It would be useful to add tests that verify that  Proxy::TFTP::Server#read_file, Proxy::TFTP::Server#write_file, and Proxy::TFTP::Server#delete_file methods do what we expect them to.

Well, I added those to actually avoid testing of the File.open and read
stuff.

## 

Later,
 Lukas #lzap Zapletal

dmitri-d:
> Well, I added those to actually avoid testing of the File.open and read stuff.

Kinda -- mocks and stubs should ideally be as close as possible to the class that is being tested (to improve readability and reduce brittleness of tests). Having said that, the logic that you extracted into dedicated method should be tested too -- you verified interfaces to read_file, etc, but not if they actually work.

To test those extracted methods you can verify changes they do to the file system, or mock lower-level calls. Personally I would prefer the former approach. Mocking of lower-level calls should be ok in this case though, as while it is more brittle (chances are these tests will have to updated every time read_file, write_file, or delete_file change), there is going to be only one test (I think) per method, which contains the amount of pain these mocks can cause.

lzap:
> To test those extracted methods you can verify changes they do to the file system, or mock lower-level calls. Personally I would prefer the former approach. Mocking of lower-level calls should be ok in this case though, as while it is more brittle (chances are these tests will have to updated every time read_file, write_file, or delete_file change), there is going to be only one test (I think) per method, which contains the amount of pain these mocks can cause.

Done the filesystem path.

## 

Later,
 Lukas #lzap Zapletal

test/tftp/tftp_server_test.rb
```diff
@@ -0,0 +1,146 @@
...
+
+  def test_create_default
+    pxe_default_files.each do |file|
+      @subject.expects(:write_file).with(file, @content).once
+    end
+    @subject.create_default @content
+  end
+end
+
+class HelersServerTest < Test::Unit::TestCase
```
dmitri-d: Spelling?

test/tftp/tftp_server_test.rb
```diff
@@ -0,0 +1,146 @@
...
+    tmp_filename = File.join(Dir.tmpdir(), 'foreman-proxy-tftp-server-write-file.txt')
+    @subject.send(:write_file, tmp_filename, "test")
+    assert_equal "test", File.open(tmp_filename, "rb").read
+  ensure
+    File.unlink(tmp_filename) if tmp_filename
+  end
+
+  def test_delete_file
+    tmp_filename = File.join(Dir.tmpdir(), 'foreman-proxy-tftp-server-write-file.txt')
+    File.open(tmp_filename, 'w') {|f| f.write("test")}
```
dmitri-d: Probably not needed for this test?

test/tftp/tftp_server_test.rb
```diff
@@ -0,0 +1,146 @@
...
+    @subject = Proxy::TFTP::Server.new
+  end
+
+  def test_path_with_settings
+    Proxy::TFTP::Plugin.settings.expects(:tftproot).returns("/some/root")
+    assert_equal "/some/root", @subject.send(:path)
+  end
+
+  def test_path
+    assert_match /file.txt/, @subject.send(:path, "file.txt")
```
dmitri-d: Here, and other protected methods -- dropping 'protected' would allow you to call methods directly instead of relying on 'send'.

dmitri-d:
@lzap: please take a look, tests are failing under 1.8.7.

lzap:
All fixed.

dmitri-d:
@lzap: there's also a couple of rubocop failures.

lzap:
> @lzap: there's also a couple of rubocop failures. 

Holy, no my day. Refactored.

## 

Later,
 Lukas #lzap Zapletal

dmitri-d:
Merged as b1cec0e. Thanks, @lzap.

PR #416, opened 2016-05-09
Fixes #14952 - isc_dhcp respects :subnets setting now.

dmitri-d:


modules/dhcp_common/server.rb
```diff
@@ -19,8 +19,9 @@ class Server
     include Proxy::Log
     include Proxy::Validations
 
-    def initialize(name)
-      @name    = name
+    def initialize(name, managed_subnets)
```
domcleal: This might break the API for plugins who inherit from this class, as they'll need to change their constructors. Perhaps avoid refactoring this now, or set a default value for `managed_subnets` to `Proxy::DhcpPlugin.settings.subnets`?

dmitri-d:
Made managed_subnets parameter of the Proxy::DHCP::Server constructor optional.

modules/dhcp_native_ms/dhcp_native_ms_main.rb
```diff
@@ -10,11 +10,12 @@ class Provider < ::Proxy::DHCP::Server
     def initialize_for_testing(params)
       @name = params[:name] || @name
       @service = params[:service] || service
+      @subnets = params[:subnets] || @subnets
```
domcleal: Should this be `@managed_subnets` to match the instance var in the parent class?

dmitri-d: Indeed it should be. Fixed.

domcleal:
Merged as f9514f6, thanks @witlessbird.

PR #419, opened 2016-05-13
Fixes #15028 - log buffer capacity reported correctly

lzap:
Added a test for this.

### DIFF
--- a/lib/proxy/log_buffer/buffer.rb
+++ b/lib/proxy/log_buffer/buffer.rb
@@ -30,8 +30,6 @@ module Proxy::LogBuffer
   end
 
   class Buffer
-    attr_reader :size, :size_tail
-
     def self.instance
       @@buffer ||= Buffer.new
     end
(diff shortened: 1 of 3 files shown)


lib/proxy/log_buffer/buffer.rb
```diff
@@ -86,10 +86,22 @@ def failed_module(a_module, message)
       @failed_modules[a_module] = message
     end
 
+    def size
```
domcleal, line 87: Also remove the attr_readers from the class, else the methods will be defined twice.

lzap, line 87: > Also remove the attr_readers from the class, else the methods will be defined twice.

Right, missed it. Thanks. Done!

## 

Later,
 Lukas #lzap Zapletal

domcleal:
Merged as f9c9e7a, thanks @lzap.

PR #421, opened 2016-05-20
Fixes #15110 - used default setting values are marked now.

dmitri-d:


domcleal:
Merged as dd2fc60, thanks @witlessbird.

PR #423, opened 2016-05-31
Fixes #15240 - empty arrays no longer trigger validation failures

dmitri-d:


domcleal:
Thanks, I've cherry picked your other commit back to 1.11-stable to include the test too: 698dc39.

PR #424, opened 2016-05-31
Fixes #15240 - empty arrays no longer trigger validation failures

dmitri-d:
```
On 1.8.7 (only there) [].to_s == '', which lead to presence validator failures.
```

### DIFF
--- a/lib/proxy/plugin_validators.rb
+++ b/lib/proxy/plugin_validators.rb
@@ -34,7 +34,11 @@ module ::Proxy::PluginValidators
   class Presence < Base
     def validate!(settings)
       setting_value = settings[@setting_name]
-      raise ::Proxy::Error::ConfigurationError, "Parameter '#{@setting_name}' is expected to have a non-empty value" if setting_value.to_s.empty?
+
+      empty_value = setting_value.nil?
+      empty_value ||= setting_value.empty? if setting_value.is_a?(String)
+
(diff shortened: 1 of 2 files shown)


dmitri-d:
[test]

domcleal:
I think you've answered this in the commit message, but to be sure - we're saying that an empty array is present in this change?  I was partly expecting `[]` to fail the presence validator still.

dmitri-d:
Correct, a non-nil value or a non-empty string is considered "present". 

I think I see what you are saying though: it feels a bit strange that empty strings and empty arrays are being treated differently. I think this behaviour is due to empty array being a valid default value in most cases (at least the ones we have in smart-proxy), while empty string isn't. 

I can change the logic for arrays (add hashmaps too) to match that for strings, but it would require a migration, as it is potentially a breaking change.

domcleal:
> I think this behaviour is due to empty array being a valid default value in most cases (at least the ones we have in smart-proxy), while empty string isn't.

Ah, that's a very good point - :dhcp_subnets does default to `[]`, so that ought to be valid to specify in the config file too.

I guess we can revisit it with another new validator if we come across a case when we want to require at least one element, but it's probably correct to leave it as-is. Thanks.

domcleal:
Merged as 222cf17, thanks @witlessbird.

PR #426, opened 2016-06-06
fixes #7766 - MS DHCP code scales poorly

pikelly:
Looking at the history of this ticket it is un clear whether another solution is required but, anyway, here is some code that we use in our organization to address this problem. This has been used in a production since 2013 but obviously I have forward ported it so it needs further testing in its new plugin based container running on 1.9 ruby.

### DIFF
--- a/modules/dhcp_native_ms/dhcp_native_ms_main.rb
+++ b/modules/dhcp_native_ms/dhcp_native_ms_main.rb
@@ -11,10 +11,12 @@ module Proxy::DHCP::NativeMS
       @name = params[:name] || @name
       @service = params[:service] || service
       @managed_subnets = params[:subnets] || @managed_subnets
+      @options_cache = {}
       self
     end
 
     def initialize
+      @options_cache = {}
       super(Proxy::DhcpPlugin.settings.server, Proxy::DhcpPlugin.settings.subnets)
     end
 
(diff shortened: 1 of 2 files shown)


dmitri-d:
I'd prefer to wait with these changes -- I'm getting close to being done with a rewrite of ms dhcp module to use native ms dhcp interface (via ffi). If that proves too slow or problematic for w/e reason, we can come back to this?

pikelly:
An ffi based fast solution would be absolutely great! A caching solution may be required on top of this because even using the M$ snapin things can be a little slow :-( I am look forward to your commit.

modules/dhcp_native_ms/dhcp_native_ms_main.rb
```diff
@@ -195,11 +198,12 @@ def load_subnets
       service.add_subnets(*find_all_subnets)
     end
 
-    def execute cmd, msg=nil, error_only=false
-      tsecs = 5
+    def execute cmd, msg=nil, error_only=false, dumping=false
+      tsecs = 10
```
alexjfisher, line 202: I've also found that 5 seconds isn't enough...

dmitri-d:
If you are curious: got to the point when ruby bindings are functional and cover most of the functionality required: https://github.com/witlessbird/ruby-dhcpsapi. Pls. see integration tests for examples for now.

dmitri-d:
closing in favour of https://github.com/theforeman/smart-proxy/pull/445

PR #427, opened 2016-06-07
fixes #15323 - correct Puppet SSH provider name

domcleal:


dLobatog:
Merged as fd75364e432ebaec0ccecc591d3566c3dfbb0947, thanks @domcleal!

PR #432, opened 2016-06-08
fixes #15348 - close check is now correct for std_out and std_err

pikelly:


dmitri-d:
[test]

dmitri-d:
Merged as b0858f4. Thanks, @pikelly.

PR #433, opened 2016-06-10
fixes #15373 - windows ping called with very small timeout

pikelly:
The value of 1 millisecond usually works but occasionally it is just too small.


dmitri-d:
Merged as 5d1b7cb. Thanks, @pikelly.

PR #434, opened 2016-06-13
fixes #15379 - add subnet usage api

timogoebel:


dmitri-d:
A few notes/thoughts:
- i think `get :/network` can be used instead of adding a new end-point: the existing end-point already serves some of the information you are returning in the new call and in my view it would be logical to extend it to provide more details about a given subnet. As long as these additions do not change data that is currently returned by this end-point old clients should continue to function without any issues.
- I'm not sure how much value is in showing general available/used/total counts. On isc dhcpd dynamic addresses can only be assigned from a pool (or pools). Reserved addresses shouldn't come from pools (or there will be problems). To be useful we'd need to show the number of available/used dynamic addresses, the number of reserved addresses, and the total number of hosts in a subnet. I don't think we should attempt to calculate the number of available addresses outside of pools -- will get expensive quickly, esp on networks larger than /24. 
  MS dhcp is a bit different: dynamic addresses can only come from a scope but have to be outside of any of the excluded ranges. To make things more complicated: reserved ip can be any address within the scope, including excluded ip ranges, and address outside of the scope, if it is within the same subnet as the scope. Ms provides a way to query for available ips in a scope via cmdlet: https://technet.microsoft.com/en-us/library/jj590701(v=wps.630).aspx. As with isc dhcpd I think providing information about unused addresses outside of scopes would be quite expensive. 
  Both isc and ms servers can have multiple pools/scopes per subnet, all of them will need to be queried to show per-subnet availability information.
- What's the use-case for the to/from filter in the call?

lzap:
Timo, if you plan elaborating this please reopen. Thanks for the effort!

PR #436, opened 2016-06-17
fixes #15244 - provide a CNAME api to the DNSCMD provider

pikelly:
This is similar to a part of a solution that is used in my organization. There is a rake task that scans all hosts for a cnames host parameter and pushes out these aliases into the Microsoft DHCP servers scattered across the world.

This implementation is not optimal as the collision checks for A, AAAA and CNAME entries should all check for the other types as an A and CNAME entry can collide. I would argue that cname management is an ancillary function so does not need so much attention. In addition the names given to aliases tend to be specialized so will generally not be given to an ordinary machine. The rake task also does additional collision detection to ensure that A and CNAME entries are not already present.

### DIFF
--- a/modules/dns/dns_api.rb
+++ b/modules/dns/dns_api.rb
@@ -27,6 +27,8 @@ module Proxy::Dns
         when 'AAAA'
           ip = IPAddr.new(value, Socket::AF_INET6).to_s
           server.create_aaaa_record(fqdn, ip)
+        when 'CNAME'
+          server.create_cname_record(fqdn, value)
         when 'PTR'
           validate_reverse_dns_name!(value)
           server.create_ptr_record(fqdn, value)
@@ -56,6 +58,8 @@ module Proxy::Dns
           server.remove_a_record(name)
         when 'AAAA'
           server.remove_aaaa_record(name)
+        when 'CNAME'
+          server.remove_cname_record(name)
         when 'PTR'
           validate_reverse_dns_name!(name)
           server.remove_ptr_record(name)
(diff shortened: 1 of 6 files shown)


modules/dns/dns_api.rb
```diff
@@ -27,6 +27,8 @@ class Api < ::Sinatra::Base
         when 'AAAA'
           ip = IPAddr.new(value, Socket::AF_INET6).to_s
           server.create_aaaa_record(fqdn, ip)
+        when 'CNAME'
+          server.create_cname_record(fqdn, value)
```
dmitri-d, line 31: We'll need a default implementation for create_cname_record (probably returning an error "This provider doesn't support creation of CNAME records" or something similar), as it's going to take some time for other providers to catch up...

modules/dns/dns_api.rb
```diff
@@ -56,6 +58,8 @@ class Api < ::Sinatra::Base
           server.remove_a_record(name)
         when 'AAAA'
           server.remove_aaaa_record(name)
+        when 'CNAME'
+          server.remove_cname_record(name)
```
dmitri-d, line 62: ditto.

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -41,6 +41,21 @@ def create_aaaa_record(fqdn, ip)
       end
     end
 
+    def create_cname_record(fqdn, target)
```
dmitri-d, line 44: Could you add tests for create_cname_record and remove_cname_record methods?

pikelly:
I have added the tests that you requested and the base class methods that call an exception.

dmitri-d:
Merged as 2b498a4. Thanks, @pikelly.

PR #437, opened 2016-06-17
fixes #1172 - proxy should return leases and reservations

pikelly:
This is my second attempt to get this code included. The last attempt was some time ago :-) You may want to read the earlier pull request or just judge it on its own merits.

This modification has been working in my organisation for some time and seems to separate out the leases and reservations correctly.

### DIFF
--- a/modules/dhcp_native_ms/dhcp_native_ms_main.rb
+++ b/modules/dhcp_native_ms/dhcp_native_ms_main.rb
@@ -91,28 +91,28 @@ module Proxy::DHCP::NativeMS
     private :find_or_create_vendor_name
 
     def find_subnet_dhcp_records(subnet)
-      cmd = "scope #{subnet.network} show reservedip"
+      cmd = "scope #{subnet.network} show client 1"
       msg = "Enumerated hosts on #{subnet.network}"
 
       to_return = []
       # Extract the data
       execute(cmd, msg).each do |line|
-        #     172.29.216.6      -    00-a0-e7-21-41-00-
-        if line =~ /^\s+([\w\.]+)\s+-\s+([-a-f\d]+)/
-          ip  = $1
+        #192.168.205.4    - 255.255.255.128- 00-1e-68-65-55-f8   -4/23/2013 6:41:21 PM   -D-
+        #192.168.205.5    - 255.255.255.128-00-1b-24-93-35-09   - NEVER EXPIRES        -U-  host.brs.company.com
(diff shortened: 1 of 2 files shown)


modules/dhcp_native_ms/dhcp_native_ms_main.rb
```diff
@@ -91,28 +91,28 @@ def find_or_create_vendor_name vendor, exception
...
+          # Some mac addresses appear to be more than 6 bytes!
+          mac = mac[0,17].gsub!(/-/,":")
           begin
-            opts = {:subnet => subnet, :ip => ip, :mac => mac}
-            opts.merge!(loadRecordOptions(opts))
+            opts = {:subnet => subnet, :ip => ip, :mac => mac, :name => name}
             logger.debug opts.inspect
-            if opts.include? :hostname
-              to_return << Proxy::DHCP::Reservation.new(opts.merge(:deleteable => true))
+            if kind == 'D' && expire !~ /^INACTIVE|^NEVER/
```
dmitri-d: Are you sure about `kind == 'D'` bit? I'm seeing 'N' for clients that support both bootp and dhcp? I think this flag is an "or" of various options and netsh isn't very good at parsing those...

My understanding is that we can treat any client without expiration date on its address as a "reserved ip".

domcleal:
Admin bits: I sent you an invite to the GitHub org @pikelly, which will enable Jenkins to automatically test your PRs when you file or update them. Just accept it and set your membership to public on [this page](https://github.com/orgs/theforeman/teams/members) and it should take effect on future git pushes. Cheers!

modules/dhcp_native_ms/dhcp_native_ms_main.rb
pikelly: You may well be right on this point. This logic works for my organisation but this may not be generally true.
I will update the push to remove the "D" clause as just using the INACTIVE|NEVER test would return the same host list.

dmitri-d:
[test]

pikelly:
If I wish to resubmit this pull request without the "D" then what is the right procedure? Do I close this PR and open another? Do I push another commit onto this branch? I expect that creating a new commit with a commit amend and then  doing a forced push will break the PR as the commit's  hash will change.

dmitri-d:
@pikelly: you can make the change, squash the commits and force-push the changeset (what I'd do, as it's fastest). Or you can close this PR and open a new one, up to you really.

ohadlevy:
IMHO better to push, as you dont need to spam redmine with the new pr number

pikelly:
I have force pushed the commit minus the "D" check.

dmitri-d:
merged as a832ac2. Thanks, @pikelly.

PR #439, opened 2016-06-20
fixes #15452 - disable DDNS updates for Native MS DHCP provider

pikelly:
I also renamed the file **dhcp_ms_native.yml.example** to
**dhcp_native_ms.yml.example** as the name must match the class name.

### DIFF
--- a/config/settings.d/dhcp_native_ms.yml.example
+++ b/config/settings.d/dhcp_native_ms.yml.example
@@ -1,0 +1,13 @@
+---
+#
+# Configuration file for Windows-specific 'netsh' dhcp provider
+#
+# This parameter informs the proxy that you wish to disable dynamic DNS updates for all hosts
+# that the smart proxy creates. This will slightly slow the host creation process but will
+# ensure that the DHCP server will not create or delete DNS entries on behalf of these clients.
+# If you can turn this feature off at the scope level then that would be a better solution.
+# Can be true or false. Defaults to true.
+#
+# :disable_ddns: true
+#
+# use :server setting in dhcp.yml if you are managing a dhcp server which is not localhost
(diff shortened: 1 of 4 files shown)


config/settings.d/dhcp_native_ms.yml.example
```diff
@@ -0,0 +1,12 @@
+---
+#
+# Configuration file for Windows-specific 'netsh' dhcp provider
+#
+# Uncomment the following line if you wish to disable dynamic DNS updates for all hosts that
+# the smart proxy creates. This will slightly slow the host creation process but will ensure
+# that the DHCP server will not create or delete DNS entries on behalf of these clients. If
+# you can turn this feature off at the scope level then that would be a better solution.
+#
+# :disable_ddns: true
```
dmitri-d, line 11: Can you define a default value for `disable_ddns` in https://github.com/theforeman/smart-proxy/blob/develop/modules/dhcp_native_ms/dhcp_native_ms_plugin.rb too?

pikelly, line 11: This I have now done. Plugins are very cool but are a little confusing at first.

dmitri-d:
Don't know if saw this already, there's a how-to on smart-proxy plugin writing which should help: http://projects.theforeman.org/projects/foreman/wiki/How_to_Create_a_Smart-Proxy_Plugin

modules/dhcp_native_ms/dhcp_native_ms_plugin.rb
```diff
@@ -2,6 +2,8 @@ module ::Proxy::DHCP::NativeMS
   class Plugin < ::Proxy::Provider
     plugin :dhcp_native_ms, ::Proxy::VERSION
 
+    default_settings :disable_ddns => false
```
dmitri-d: Current convention is that config files examples have parameters set to their default values (with comments reflecting this). I also think that ddns updates can be disabled by default (as this is preferable configuration when used with foreman). Could you set `disable_ddns` to true here?

pikelly:
The default is now **true** for disable_ddns, and the example config file shows this.

dmitri-d:
Merged as 36956f8. Thanks, @pikelly.

PR #440, opened 2016-07-04
fixes #15572 - check for missing vendor class in native ms was wrong

pikelly:
It was just wrong :-)

### DIFF
--- a/modules/dhcp_native_ms/dhcp_native_ms_main.rb
+++ b/modules/dhcp_native_ms/dhcp_native_ms_main.rb
@@ -223,20 +223,20 @@ module Proxy::DHCP::NativeMS
 
     def report msg, response, error_only
       if response.grep(/completed successfully/).empty?
-        if response.grep /class name being used is unknown/
+        if !response.grep(/class name being used is unknown/).empty?
           logger.warn "Vendor class not found"
         else
-          logger.error "Netsh failed:\n" + response.join("\n")
+          logger.error "Netsh failed:\n" + response.join()
         end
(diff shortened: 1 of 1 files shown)


dLobatog:
Just a comment for the next reviewer...
If `grep(xxx)` returns something, then both are positive
If it doesn't find the string, it returns `[]`, which is still positive! 

Let's run [test] s again, failure was unexpected `OpenSSL::SSL::SSLError: SSL_read:: decryption failed or bad record mac` on 1.9.3

modules/dhcp_native_ms/dhcp_native_ms_main.rb
```diff
@@ -223,17 +223,17 @@ def execute cmd, msg=nil, error_only=false
...
         msg.sub! /Removed/,    "remove"
         msg.sub! /Added/,      "add"
         msg.sub! /Enumerated/, "enumerate"
         msg.sub! /Queried/,    "query"
         msg  = "Failed to #{msg}"
-        msg += "Vendor class not found" if response.grep /class name being used is unknown/
+        msg += "Vendor class not found" if !response.grep(/class name being used is unknown/).empty?
         msg += ": No entry found" if !response.grep(/not a reserved client/).empty?
         match = response.grep(/used by another client/)
         msg += ": #{match}" if !match.size.empty?
```
dmitri-d: Could you fix this line while we are here? Should be `!match.empty?`

dLobatog:
Merged as 3cd0744c2d0392ae41deebfe9a1589b79732bfbd, thanks @pikelly!

PR #409, opened 2016-04-22
Fixes #2687 - Leases file is no longer being parsed on every request

dmitri-d:
This is work in progress and currently only isc dhcpd running on linux is supported (support for BSD is coming). Please ignore the failing tests, these are due to changes in internal API.

PR #441, opened 2016-07-20
WIP #15613 - Add DHCPv6 support to proxy

xprazak2:
Initial attempt for DHCPv6 on proxy.

TODO:
- [x] Only one Libvirt network supported - check if it is v4 or v6 before suggesting ip
- [ ] Ad file observers for libvirt
- [ ] Check for reserved v6 ips (multicast)
- [ ] ISC
- [ ] native MS
- [ ] tests

### DIFF
--- a/lib/proxy/validations.rb
+++ b/lib/proxy/validations.rb
@@ -1,4 +1,7 @@
+require 'ipaddr'
+
 module Proxy::Validations
+  include Proxy::Log
 
   MAC_REGEXP_48BIT = /\A([a-f0-9]{1,2}:){5}[a-f0-9]{1,2}\z/i
   MAC_REGEXP_64BIT = /\A([a-f0-9]{1,2}:){19}[a-f0-9]{1,2}\z/i
@@ -14,14 +17,42 @@ module Proxy::Validations
 
   # validates the ip address
   def validate_ip ip
-    raise Error, "Invalid IP Address #{ip}" unless ip =~ /(\d{1,3}\.){3}\d{1,3}/
+    IPAddr.new(ip)
     ip
+  rescue invalid_address_error
+    raise Error, "Invalid IP Address #{ip}"
+  end
+
+  # we may not want to raise error in some cases bu skip the ip instead
+  def soft_validate_ip ip
+    IPAddr.new(ip)
+    ip
+  rescue invalid_address_error
+    logger.debug("Invalid IPv6 address #{ip} detected, skipping")
+    nil
+  end
+
+  # validate prefix for IPv6 subnet
+  def validate_v6_prefix prefix
+    prefix = prefix.to_i
+    raise Error, "Invalid IPv6 prefix #{prefix}" if (prefix < 1 || prefix > 127)
+    prefix
   end
 
   # validates the mac
   def validate_mac mac
     raise Error, "Invalid MAC #{mac}" unless valid_mac?(mac)
     mac.downcase
+  end
+
+  def validate_mac_or_uid mac, uid
+    raise Error, "MAC or UID must be present" if [mac, uid].none?
+    raise Error, "Only one of MAC, UID must be present" if [mac, uid].all?
+    if mac
+      { :mac => validate_mac(mac) }
+    else
+      { :uid => uid }
+    end
   end
 
   def validate_subnet subnet
@@ -39,4 +70,11 @@ module Proxy::Validations
     record
   end
 
+  private
+
+  def invalid_address_error
+    # IPAddr::InvalidAddressError is undefined for ruby 1.9
+    return IPAddr::InvalidAddressError if IPAddr.const_defined?('InvalidAddressError')
+    ArgumentError
+  end
 end
--- a/modules/dhcp/dhcp_api.rb
+++ b/modules/dhcp/dhcp_api.rb
@@ -31,7 +31,7 @@ class Proxy::DhcpApi < ::Sinatra::Base
   get "/?" do
     begin
       content_type :json
-      server.subnets.map{|s| {:network => s.network, :netmask => s.netmask, :options => s.options}}.to_json
+      server.subnets.map{|s| {:network => s.network, :netmask => s.netmask, :options => s.options, :prefix => s.prefix}}.to_json
     rescue => e
       log_halt 400, e
     end
--- a/modules/dhcp_common/record.rb
+++ b/modules/dhcp_common/record.rb
@@ -5,7 +5,7 @@ module Proxy::DHCP
   # represent a DHCP Record
   class Record
 
-    attr_reader :ip, :mac, :subnet, :options
+    attr_reader :ip, :mac, :uid, :subnet, :options
     include Proxy::DHCP
     include Proxy::Log
     include Proxy::Validations
@@ -13,7 +13,7 @@ module Proxy::DHCP
     def initialize options = {}
       @subnet  = validate_subnet options[:subnet]
       @ip      = validate_ip options[:ip]
-      @mac     = validate_mac options[:mac]
+      set_mac_or_uid(validate_mac_or_uid options[:mac], options[:uid])
       @deleteable = options.delete(:deleteable) if options[:deleteable]
       @options = options
     end
@@ -22,8 +22,12 @@ module Proxy::DHCP
       @subnet.network
     end
 
+    def v6?
+      subnet.v6?
+    end
+
     def to_s
-      "#{ip} / #{mac}"
+      "#{ip} / #{prefix}"
     end
 
     def inspect
@@ -54,5 +58,11 @@ module Proxy::DHCP
       true
     end
 
+    private
+
+    def set_mac_or_uid(hash)
+      var = hash.keys.first
+      instance_variable_set "@#{var}", hash[var]
+    end
   end
 end
--- a/modules/dhcp_common/subnet.rb
+++ b/modules/dhcp_common/subnet.rb
@@ -10,16 +10,15 @@ require 'tmpdir'
 module Proxy::DHCP
   # Represents a DHCP Subnet
   class Subnet
-    attr_reader :network, :netmask, :server, :timestamp
+    attr_reader :network, :server, :timestamp
     attr_accessor :options
 
     include Proxy::DHCP
     include Proxy::Log
     include Proxy::Validations
 
-    def initialize network, netmask, options = {}
+    def initialize network, options = {}
       @network = validate_ip network
-      @netmask = validate_ip netmask
       @options = {}
 
       @options[:routers] = options[:routers].each{|ip| validate_ip ip } if options[:routers]
@@ -48,15 +47,27 @@ module Proxy::DHCP
     end
 
     def to_s
-      "#{network}/#{netmask}"
+      raise NotImplementedError, "Method 'to_s' must be implemented"
     end
 
-    def cidr
-      IPAddr.new(netmask).to_i.to_s(2).count("1")
+    def prefix
+      raise NotImplementedError, "Method 'prefix' must be implemented"
+    end
+
+    def netmask
+      raise NotImplementedError, "Method 'netmask' must be implemented"
+    end
+
+    def valid_range
+      raise NotImplementedError, "Method 'valid_range' must be implemented"
+    end
+
+    def v6?
+      is_a? Proxy::DHCP::Ipv6
     end
 
     def range
-      r=valid_range
+      r = valid_range
       "#{r.first}-#{r.last}"
     end
 
@@ -74,7 +85,7 @@ module Proxy::DHCP
       @file = File.new(@filename,'r+') rescue File.new(@filename,'w+')
 
       # this returns the index in the file
-      return @file.readlines.first.to_i rescue 0
+      return index_from_file(@file)
     end
 
     def write_index_and_unlock index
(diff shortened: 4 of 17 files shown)


dmitri-d:
@xprazak2: We should chat about this when you have a moment.

dLobatog:
@witlessbird Let us know what comes out of that chat here :smile: - I'm interested in the implementation and would be happy to help with coding & testing as I'm already familiar with how this works in Foreman.

cc-ing @timogoebel too

dmitri-d:
@xprazak2, @dLobatog, @timogoebel: There are a few things I would like to discuss:
- the interface with backends will need to be updated to deal with ipv6, which is a good opportunity to bring it up to date with the latest changes (https://github.com/theforeman/smart-proxy/pull/409 and https://github.com/witlessbird/smart-proxy/tree/use-dhcps-api)
- The first PR above needs to be merged before this work, as it significantly changes our interaction with isc dhcpd
- I expect significant changes (please see https://github.com/witlessbird/smart-proxy/tree/use-dhcps-api if you are curious) the to ms dhcp provider are going to be introduced early next week, this is going to affect this work as well

lib/proxy/validations.rb
```diff
@@ -14,8 +16,17 @@ def valid_mac? mac
 
   # validates the ip address
   def validate_ip ip
-    raise Error, "Invalid IP Address #{ip}" unless ip =~ /(\d{1,3}\.){3}\d{1,3}/
+    IPAddr.new(ip)
     ip
+  rescue IPAddr::InvalidAddressError
```
timogoebel: The exception thrown differs between ruby versions, please add something like this: https://github.com/theforeman/foreman/blob/79c0664a008afe0fea54a0b5256426f3867a02bc/app/models/subnet.rb#L288-L292

timogoebel:
@xprazak2 : I just had a brief look at the code. Thanks for taking over this issue. Please keep me posted with your progress. I'd be happy to help.

I currently paused my work on the IPv6 stuff to wait for the EFI PRs from @lzap . I think they are a requirement to test orchestration as a whole.

I agree, that we should merge the PR from @witlessbird before this. Let's try to speed this up a little. :-)

timogoebel:
@xprazak2 : #409 hat just been merged. I think, you can go ahead with this. Curious to see an update.

xprazak2:
This is what I have so far: I refactored libvirt provider and cleaned up libvirt_main, there are no file observers yet. I made changes to isc only to find out omshell lacks support for IPv6. I asked on isc-users list if anyone knows about plans to add it but I haven't received any replies yet. In the meantime, I will move on to ms provider.

timogoebel:
@xprazak2 : Just fyi, I started with the Foreman parts: https://github.com/timogoebel/foreman/tree/15612-ipv6-dhcpv6
This is still very wip.

lib/proxy/validations.rb
```diff
@@ -14,8 +17,26 @@ def valid_mac? mac
 
   # validates the ip address
   def validate_ip ip
-    raise Error, "Invalid IP Address #{ip}" unless ip =~ /(\d{1,3}\.){3}\d{1,3}/
+    IPAddr.new(ip)
```
dmitri-d, line 20: Please don't rely on implicit validation during object instantiation. Instead:

```
raise "Error" if (ip =~ =~ Resolv::AddressRegex).nil?
```

lib/proxy/validations.rb
```diff
@@ -14,8 +17,26 @@ def valid_mac? mac
...
+    ip
+  rescue invalid_address_error
+    raise Error, "Invalid IP Address #{ip}"
+  end
+
+  # we may not want to raise error in some cases bu skip the ip instead
+  def soft_validate_ip ip
```
dmitri-d, line 27: That is a very confusing name. Can rely on "!" do signal presence of side-effects in calls. Then `validate_ip` becomes `validate_ip!` and `soft_validate_ip` becomes `validate_ip`.

dmitri-d, line 27: Also, not need for logging here.

lib/proxy/validations.rb
```diff
@@ -24,6 +45,16 @@ def validate_mac mac
     mac.downcase
   end
 
+  def validate_mac_or_uid mac, uid
```
dmitri-d, line 48: Please break this up into two methods

lib/proxy/validations.rb
```diff
@@ -39,4 +70,11 @@ def validate_record record
     record
   end
 
+  private
```
dmitri-d, line 73: No need to use 'private' anywhere other than in libraries. All it does is here is making it harder to test things.

modules/dhcp/dhcp_api.rb
```diff
@@ -31,7 +31,7 @@ def load_subnet_data
   get "/?" do
     begin
       content_type :json
-      server.subnets.map{|s| {:network => s.network, :netmask => s.netmask, :options => s.options}}.to_json
+      server.subnets.map{|s| {:network => s.network, :netmask => s.netmask, :options => s.options, :prefix => s.prefix}}.to_json
```
dmitri-d, line 34: I think the api controller needs:
- change the interface to providers -- drop `load_subnet` and `load_subnet_data` calls. If you look at the recent and pending changes you'll notice that those aren't being used. In other words, one call to provider per operation.
- keep ipv4 and ipv6 calls separate (by possibly prefixing the latter with 'ipv6'?)
  - You may have broken current (and older) versions of foreman with this implementation of `get /`
  - create/delete/free ip calls are going to be different for ipv4 and ipv6, better to keep them separate to avoid guessing which needs to be called
  - Similarly for record retrieval: At least dhcp option calls are going to be ipv4/ipv6-specific, probably easier to keep those separate

modules/dhcp_common/record.rb
```diff
@@ -5,15 +5,15 @@ module Proxy::DHCP
   # represent a DHCP Record
   class Record
```
dmitri-d, line 6: I think the record class can be removed, it's mostly being used as a container for options hash. Additionally: 
- subnet should be stored as a string: it's super annoying having to construct a Subnet object every time a lease or resevation is constructed, only so the subnet address string can be used later).
- Ipv4 and ipv6 can be kept separate, would also allow for meaningful initialization (no set_mac_or_uid-like methods)
  - validation should be moved out of constructor (if required at all -- we only create these objects from data received from isc/ms/etc backends)

modules/dhcp_common/subnet.rb
```diff
@@ -48,15 +47,27 @@ def include? ip
     end
 
     def to_s
-      "#{network}/#{netmask}"
+      raise NotImplementedError, "Method 'to_s' must be implemented"
     end
 
-    def cidr
-      IPAddr.new(netmask).to_i.to_s(2).count("1")
+    def prefix
```
dmitri-d, line 53: Please remove these:
- internal code not meant to be extended
- you will get an "undefined method" error anyway

modules/dhcp_common/subnet.rb
```diff
@@ -10,16 +10,15 @@
...
-    attr_reader :network, :netmask, :server, :timestamp
+    attr_reader :network, :server, :timestamp
     attr_accessor :options
 
     include Proxy::DHCP
     include Proxy::Log
     include Proxy::Validations
 
-    def initialize network, netmask, options = {}
+    def initialize network, options = {}
```
dmitri-d, line 20: I think this class can be removed too in favour of ipv4 and ipv6-specific ones:
- unused_ip code should be moved out from Subnet -- at least two providers have their own implementation. It's going to be ipv4/ipv6 specific too.
- Everywhere else subnet class is either used as a container for options hash (which is broken) or for its ip address/mask

modules/dhcp_isc/configuration_loader.rb
```diff
@@ -48,11 +32,55 @@ def load_dependency_injection_wirings(container, settings)
...
+      require 'dhcp_isc/isc_file_parser6'
       require 'dhcp_isc/configuration_file'
       require 'dhcp_isc/leases_file'
       require 'dhcp_isc/isc_state_changes_observer'
       require 'dhcp_isc/dhcp_isc_main'
+      require 'dhcp_common/subnet/ipv4'
+      require 'dhcp_common/subnet/ipv6'
+    end
+
+    private
```
dmitri-d, line 45: no need for private here

modules/dhcp_isc/configuration_loader.rb
```diff
@@ -48,11 +32,55 @@ def load_dependency_injection_wirings(container, settings)
...
+    def versions
+      %w(4 6)
+    end
+
+    def dep_name(name, ver)
+      "#{name}#{ver}".to_sym
+    end
+
+    def load_wirings_for_files(container, settings)
+      versions.each do |ver|
```
dmitri-d, line 56: `[4, 6].each do |ver|`

modules/dhcp_isc/dhcp_isc_plugin.rb
```diff
@@ -2,16 +2,19 @@ module ::Proxy::DHCP::ISC
...
-                     :omapi_port => '7911'
+    default_settings :config4 => '/etc/dhcp/dhcpd.conf',
+                     :leases4 => '/var/lib/dhcpd/dhcpd.leases',
+                     :omapi_port => '7911',
+                     :config6 => '/etc/dhcp/dhcpd6.conf',
+                     :leases6 => '/var/lib/dhcpd/dhcpd6.leases'
 
     requires :dhcp, ::Proxy::VERSION
-    validate_readable :config, :leases
+    validate_readable :config4, :leases4, :config6, :leases6
```
dmitri-d, line 12: What if only ipv4 or ipv6 is enabled/used?

modules/dhcp_isc/isc_file_parser.rb
```diff
@@ -7,10 +7,11 @@ module Proxy::DHCP::ISC
   class FileParser
```
dmitri-d, line 7: As mentioned above, I think it's better to keep ipv6 and ipv4 data structures separate. Because of this big chunks of code will move from this class into ipv4/ipv6 specific ones, which might reduce it to a module?

modules/dhcp_libvirt/configuration_loader.rb
```diff
@@ -2,16 +2,22 @@ module Proxy::DHCP::Libvirt
   class PluginConfiguration
     def load_dependency_injection_wirings(container, settings)
       container.dependency :memory_store, ::Proxy::MemoryStore
-      container.dependency :subnet_service, (lambda do
+      container.singleton_dependency :subnet_service, (lambda do
```
dmitri-d, line 5: why?

dmitri-d:
To summarise: 
- Now is a good opportunity to change the interface between providers and the api controller, we should do it. Otherwise there will be way more work after this.
- ipv4 and ipv6 operations and data structures can be splt from each other: this will make implementation and testing easier
- I would suggest to add/update tests early, esp. around various parsers

dmitri-d:
> I made changes to isc only to find out omshell lacks support for IPv6. I asked on isc-users list if anyone knows about plans to add it but I haven't received any replies yet. 

I can't find much (only looked very briefly) on omapi support for ipv6. Looks like isc is developing a new dhcp server http://kea.isc.org/wiki, perhaps it has better ipv6 support? 

> In the meantime, I will move on to ms provider.

Please use changes in https://github.com/theforeman/smart-proxy/pull/445, not the 'develop' branch.

dmitri-d:
I finally had a chance to play with omshell talking to a dhcpd6 instance. It appears to mostly work ok  (haven't tried setting options on host object, but that bit should work). 

The main disconnect I found is between ipv4 host ids and ipv6 duids (host-identifier option vs. option dhcp-client-identifier). While it is possible to set/modify the former, it is impossible to do the same for ipv6 duids, which restricts how forman-managed hosts can be deployed (see https://tools.ietf.org/html/rfc3315#section-11 for more information).

That aside, here are some examples. Locating a record via hardware address:

```
> new host
obj: host
> set hardware-address = 01:00:80:a2:55:67:34
obj: host
hardware-address = 01:00:80:a2:55:67:34
> open
obj: host
hardware-address = 01:00:80:a2:55:67:34
ip-address = 20:01:0d:b8:00:01:00:01:00:00:00:00:00:00:07:01
dhcp-client-identifier = 00:01:00:01:14:ed:66:c1:08:00:27:94:08:40
name = "test"
hardware-type = 00:00:00:01
> close
obj: <null>
```

Creating a new host record:

```
> new host
obj: host
> set name = "test-another"
obj: host
name = "test-another"
> set ip-address = 20:01:0d:b8:00:01:00:01:00:00:00:00:00:00:07:02
obj: host
name = "test-another"
ip-address = 20:01:0d:b8:00:01:00:01:00:00:00:00:00:00:07:02
> set hardware-address = 01:00:80:a2:55:67:35
obj: host
name = "test-another"
ip-address = 20:01:0d:b8:00:01:00:01:00:00:00:00:00:00:07:02
hardware-address = 01:00:80:a2:55:67:35
> create
obj: host
name = "test-another"
ip-address = 20:01:0d:b8:00:01:00:01:00:00:00:00:00:00:07:02
hardware-address = 01:00:80:a2:55:67:35
hardware-type = 00:00:00:00
> close
obj: <null>
> new host
obj: host
> set name = "test-another"
obj: host
name = "test-another"
> open
obj: host
name = "test-another"
ip-address = 20:01:0d:b8:00:01:00:01:00:00:00:00:00:00:07:02
hardware-address = 01:00:80:a2:55:67:35
hardware-type = 00:00:00:00
```

Host record removal:

```
obj: host
name = "test-another"
ip-address = 20:01:0d:b8:00:01:00:01:00:00:00:00:00:00:07:02
hardware-address = 01:00:80:a2:55:67:35
hardware-type = 00:00:00:00
> remove
obj: <null>
> new host
obj: host
> set hardware-address = 01:00:80:a2:55:67:35
obj: host
hardware-address = 01:00:80:a2:55:67:35
> open
can't open object: not found
obj: host
hardware-address = 01:00:80:a2:55:67:35
```

dmitri-d:
> I can't find much (only looked very briefly) on omapi support for ipv6. Looks like isc is developing a new dhcp server http://kea.isc.org/wiki, perhaps it has better ipv6 support?

kea doesn't have external api other than for server status and restarts. Unless we are prepared to invest time into adding this functionality it's not  an option, at least for the time being.

lzap:
@witlessbird did you get any updates on the ISC DHCP mailing list in this regard?

dmitri-d:
@lzap see my comment above.

dmitri-d:
I don't think we reached an impasse here -- most of stuff is doable, it just needs doing. Removing the label.

timogoebel:
What's the status here, guys? @xprazak2, are you still working on this?

dmitri-d:
@timogoebel: I started working on this.

xprazak2:
@witlessbird, I am sorry I could not finish this in timely manner, but priorities shifted 😞. If you want to take over, I have no objections.

dmitri-d:
I'm closing this PR -- I think ipv6 support will require a different approach to the one we currently use.

PR #442, opened 2016-07-22
Fixes #15791 - no more intermittent test failures in ms dhcp test

dmitri-d:


domcleal:
Merged as bce15a5, thanks @witlessbird.

PR #444, opened 2016-07-27
Fixes #15865 - pin json_pure to < 2.0.0 for RUBY_VERSION < 2.0.0

dmitri-d:


bundler.d/puppet.rb
```diff
@@ -1,4 +1,5 @@
 group :puppet_proxy_legacy do
+  gem 'json_pure', '< 2.0.0' if RUBY_VERSION < '2.0.0'
```
domcleal: I think this needs to be set `:require => false` as packaged installations probably won't use json_pure - they're more likely to use the version of json that's in MRI and ignore this dependency of puppet/hiera.

Testing on a 3.x installation causes this:

```
E, [2016-07-28T10:04:10.290317 #11936] ERROR -- : Disabling all modules in the group ['puppet_proxy_legacy', 'puppet'] due to a failure in one of them: Gem loading error: cannot load such file -- json_pure
```

dmitri-d:
Added `:require => false` for json_pure gem.

domcleal:
Merged as 5a38c11, thanks @witlessbird.

PR #447, opened 2016-08-03
Fixes #15942 - dns module and providers use updated di config.

dmitri-d:


domcleal:
Is it possible to keep the old DI structure for a release or two and issue a deprecation warning?  Plugins currently use modules/dns_common/dependency_injection/dependencies.rb, so removing it and changing this will cause another core to plugin API break.

dmitri-d:
I'd like to move to the new initialization, as it makes testing easier (was done in order to add tests to infoblox). I was going to do PRs for power_dns and route53 plugins if that helps...

dmitri-d:
Actually, I think I can keep dns_common/dependency_injection/dependencies.rb around, just point it to the di container used by the module... 

Let me know which way is preferable...

domcleal:
> Actually, I think I can keep dns_common/dependency_injection/dependencies.rb around, just point it to the di container used by the module...

Yes, this is what I was thinking of.

dmitri-d:
Restored dns_common/dependency_injection/dependencies.rb and modified it to use module's di container.

modules/dns_common/dependency_injection/dependencies.rb
```diff
@@ -1,9 +1,10 @@
-require 'dns_common/dependency_injection/container'
-
 module Proxy::Dns
   module DependencyInjection
     class Dependencies
-      extend Proxy::Dns::DependencyInjection::Wiring
+      extend Proxy::DependencyInjection::Wiring
+      def self.container_instance
```
domcleal, line 6: Do you want to add a logger debug/warning message into here, so it can be removed more safely in a later release?

dmitri-d, line 6: Nods, let's do it.

dmitri-d:
Added a deprecation warning and rebased.

domcleal:
Merged as 69afd34, thanks @witlessbird. Could you also update https://github.com/theforeman/smart_proxy_dns_plugin_template please?

dmitri-d:
> Could you also update https://github.com/theforeman/smart_proxy_dns_plugin_template please?

will do.

PR #449, opened 2016-08-04
fixes #15985 - Pxegrub2 tftp filename prefix

timogoebel:
@lzap : Mind to take a look?

### DIFF
--- a/modules/tftp/server.rb
+++ b/modules/tftp/server.rb
@@ -96,7 +96,7 @@ module Proxy::TFTP
       ["#{pxeconfig_dir}/grub.cfg"]
     end
     def pxeconfig_file mac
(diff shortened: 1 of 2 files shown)


lzap:
I will test this with Grub2 from RHEL7, maybe it was patched and starting with MAC without 01, I was testing all my entries in the design table.

timogoebel:
> I will test this with Grub2 from RHEL7

Thanks. That's what I also tested this with. I also found this in some SUSE, Oracle and some other distro's documentation.

lzap:
I can confirm this now. Thanks.

lzap:
Merged as d4b60b36842ce5203cc009863ef89e21d71176b5, thank you!

PR #453, opened 2016-08-12
Fixes #16099 - in daemon mode module threads continue to execute

dmitri-d:
```
This also fixes the issue with missed isc lease file updates.
```

### DIFF
--- a/lib/launcher.rb
+++ b/lib/launcher.rb
@@ -10,12 +10,16 @@ module Proxy
       SETTINGS.daemon_pid
     end
 
+    def http_enabled?
+      !SETTINGS.http_port.nil?
+    end
+
     def https_enabled?
       SETTINGS.ssl_private_key && SETTINGS.ssl_certificate && SETTINGS.ssl_ca_file
     end
 
     def http_app
-      return nil if SETTINGS.http_port.nil?
+      return nil unless http_enabled?
       app = Rack::Builder.new do
         ::Proxy::Plugins.instance.select {|p| p[:state] == :running && p[:http_enabled]}.each do |p|
           instance_eval(p[:class].http_rackup)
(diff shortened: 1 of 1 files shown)


iNecas:
Looks reasonable. If I read it correctly, the issue was the `Process.daemon` closing the open file descriptors that got open during the plugin initialization. Right?

dmitri-d:
[test]

dmitri-d:
No, fork call preserves file descriptors in children processes.

If fork is called and there are executing threads there, and the parent process calls exit , all running threads are going to be terminated (as part of the exit).

dmitri-d:
[test]

iNecas:
Ah, I've been dealing with the problems with ordering in the past, but I always though it was about the file descriptors, but looking back, the threads were probably the reason for behavior I saw. Anyway, the solution always was forking sooner. So :+1: from me

lib/launcher.rb
```diff
@@ -119,18 +119,18 @@ def webrick_server(app)
...
         check_pid
         Process.daemon
         write_pid
       end
 
+      ::Proxy::PluginInitializer.new(::Proxy::Plugins.instance).initialize_plugins
+
+      http_app = http_app()
+      https_app = https_app()
+      raise Exception.new("Both http and https are disabled, unable to start.") if http_app.nil? && https_app.nil?
```
domcleal: Could you keep this check before forking, if possible? I think the pre-reqs for HTTP/HTTPS apps could be checked trivially before forking occurs - it's just checking for presence of various settings. The more errors that are raised before forking, the easier it is for users to know about and debug failed startups.

dmitri-d: Hmm. I think this is going to be tricky -- here we effectively check that there is at least one module successfully configured using the appropriate port. Ports used (:https_enabled and :http_enabled properties in elements of ::Proxy::Plugins hash) themselves are set during plugin initialization.

domcleal: I don't think it checks if modules are present - only that `SETTINGS.http_port` is given for HTTP, or that `SETINGS.ssl_private_key` etc are given for HTTPS (curiously, not https_port).  The apps will still be returned even if no modules are configured, I think.

dmitri-d: Also, not sure I get the comment re: ease of debugging of failed startups in daemon mode. The failures are going to be logged in the log file independent of where this check is. Console logger can't be used in daemon mode either...

dmitri-d: > I don't think it checks if modules are present - only that SETTINGS.http_port is given for HTTP, or that SETINGS.ssl_private_key etc are given for HTTPS (curiously, not https_port). The apps will still be returned even if no modules are configured, I think.

I see what you mean. Yes, I can move these checks to before forking.

domcleal: If the check's done before forking, executing bin/smart-proxy shows this exception on stdout and exits with a non-zero exit code which service managers like systemd and even init scripts will pick up immediately. If it moves to after forking, then it forks away, but the process stops and the error's only shown in the log.

dmitri-d:
Moved http/https presence check to before forking, rebased.

dmitri-d:
[test]

lib/launcher.rb
```diff
@@ -119,18 +123,19 @@ def webrick_server(app)
     end
 
     def launch
-      ::Proxy::PluginInitializer.new(::Proxy::Plugins.instance).initialize_plugins
-
-      http_app = http_app()
-      https_app = https_app()
-      raise Exception.new("Both http and https are disabled, unable to start.") if http_app.nil? && https_app.nil?
+      raise Exception.new("Both http and https are disabled, unable to start.") unless (http_enabled? || https_enabled?)
```
domcleal: rubocop suggests the parentheses should be removed from the `unless` clause

dmitri-d:
Fixed rubocop warning and rebased.

domcleal:
Merged as 350b4cf, thanks @witlessbird.

PR #454, opened 2016-08-16
Fixes #16122 -- added equality mehtod for subnets and records.

dmitri-d:


dmitri-d:
[test]

iNecas:
:+1:

test/dhcp/record_test.rb
```diff
@@ -1,6 +1,5 @@
 require 'test_helper'
 require 'dhcp_common/dhcp_common'
-require 'dhcp_common/server'
```
domcleal, line 3: This is necessary to run the test file standalone, else it errors with: `NameError: uninitialized constant Proxy::DHCP::Subnet`

test/dhcp/record_test.rb
```diff
@@ -15,20 +14,12 @@ def test_record_should_have_a_subnet
     assert_kind_of Proxy::DHCP::Subnet, @record.subnet
   end
 
-  def test_should_return_a_record_kind
```
domcleal: Why delete this?

test/dhcp/record_test.rb
```diff
@@ -56,7 +47,16 @@ def test_should_not_save_invalid_subnets
     end
   end
 
-  def test_options_should_be_a_hash
-    assert_kind_of Hash, @record.options
```
domcleal: I guess this also tested the initialiser?

test/dhcp/subnet_test.rb
```diff
@@ -63,10 +59,6 @@ def test_should_not_save_invalid_netmask
     end
   end
 
-  def test_options_should_be_a_hash
-    assert_kind_of Hash, @subnet.options
```
domcleal: ditto, this tested the initialiser?

domcleal:
I don't think any of the the removals are warranted in the current patch, just add the equality method?

test/dhcp/record_test.rb
dmitri-d: By default the constructor argument is a hash, otherwise options will be initialized with w/e arguments were passed to the constructor. All this test does is verifies that setup code passes in hash as a parameter.

test/dhcp/record_test.rb
dmitri-d: This is the only place where #kind is invoked. I'll remove the method too.

dmitri-d:
removed changes other than related to added equality methods.

test/dhcp/record_test.rb
```diff
@@ -59,4 +56,17 @@ def test_should_not_save_invalid_subnets
   def test_options_should_be_a_hash
     assert_kind_of Hash, @record.options
   end
+
+  def test_equality
+    assert_equal Proxy::DHCP::Record.new(:subnet => @subnet, :ip => @ip, :mac => @mac, :option1 => 'one'),
+                 Proxy::DHCP::Record.new(:subnet => @subnet, :ip => @ip, :mac => @mac, :option1 => 'one')
+    assert Proxy::DHCP::Record.new(:subnet => @subnet, :ip => @ip, :mac => @mac, :option1 => 'one') !=
```
domcleal: prefer `assert_not_equal` here and also in the subnet test - like `assert_equal` above, it'll show better error messages when the test fails than true/false.

test/dhcp/subnet_test.rb
```diff
@@ -108,4 +100,18 @@ def test_unused_ip_should_return_ip_from_within_the_range
...
     File.delete('test/tmp/foreman-proxy_192.168.0.0_24.tmp')
   end
+
+  def test_options_should_be_a_hash
+    assert_kind_of Hash, @subnet.options
+  end
+
+  def test_equality
+    assert_equal ::Proxy::DHCP::Subnet.new(@network, @netmask, :domain_name => 'a.b.c'), ::Proxy::DHCP::Subnet.new(@network, @netmask, :domain_name => 'a.b.c')
+    assert ::Proxy::DHCP::Subnet.new(@network, @netmask, :domain_name => 'a.b.c') !=
```
domcleal: ditto, assert_not_equal

domcleal:
The branch also needs rebasing now.

dmitri-d:
Switched to assert_not_equal instead of assert calls and rebased.

domcleal:
Merged as 0f39b11, thanks @witlessbird.

PR #456, opened 2016-08-17
refs #16021 - monitor modifications on entire directory

domcleal:
inotify file watches are based on the inode, so after the leases file
is recreated the watch no longer fires. Later versions of rb-inotify
will properly allow for the watch to be re-registered after the file is
recreated (issue #16140).


dmitri-d:
Merged as c4314a8. Thanks, @domcleal. .

PR #457, opened 2016-08-17
Fixes #16145 - Removed unnecessary tests for common dhcp entities

dmitri-d:


domcleal:
Merged as dd93f60, thanks @witlessbird.

PR #459, opened 2016-09-01
 fixes #16410 - fix bug if multiple lines contain word "server" fixes #16410

pronix:
error happend whith this patch https://github.com/Katello/forklift/pull/263
because in cat /etc/ipa/default.conf

```
#File modified by ipa-client-install

[global]
basedn = dc=example,dc=com
realm = EXAMPLE.COM
domain = example.com
server = centos7-freeipa-server.example.com
host = centos7-katello-nightly.example.com
xmlrpc_uri = https://centos7-freeipa-server.example.com/ipa/xml
enable_ra = True
```

and returned 2 string by 

```
grep server /etc/ipa/default.conf | cut -f2 -d"="
 centos7-freeipa-server.example.com
 https://centos7-freeipa-server.example.com/ipa/xml
```

### DIFF
--- a/sbin/foreman-prepare-realm
+++ b/sbin/foreman-prepare-realm
@@ -22,7 +22,7 @@ die()   { echo "$@" 1>&2; exit 1; }
 [ ! -z $1 ] || usage
 [ ! -z $2 ] || usage
 
-SERVER=$(grep server /etc/ipa/default.conf | cut -f2 -d"=")
+SERVER=$(grep 'server =' /etc/ipa/default.conf | cut -f2 -d"=")
 
 if ipa --version 2>&1 | grep -q 'VERSION'
 then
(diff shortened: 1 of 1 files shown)


jaekel:
If I understand the problem correctly I would suggest a slightly different solution.

Instead of searching for “server = “ you could search for lines that begin with the word “server”. (like with a regex /^server(\w=)/ or something)

This would still work if someone left out the spaces in the file and wrote “server=blabla.bla”

dmitri-d:
I think @jaekel has a very good suggestion.

pronix:
@jaekel @witlessbird  updated pattern, but file is generated by ipa - so format should not change

dmitri-d:
merged in e02ca04. Thanks, @pronix!

PR #463, opened 2016-10-03
Fixes #16766 - allow non hash-rocket syntax in rubocop

dmitri-d:


.rubocop.yml
```diff
@@ -46,3 +46,8 @@ Metrics/ParameterLists:
 
 Style/Lambda:
   Enabled: false
+
+# enable both hash-rocket and ruby19 hashmap syntax
+Style/HashSyntax:
+  Enabled: false
```
domcleal, line 52: Could you enable the cop, but set EnforcedStyle to `no_mixed_keys`? This supports both hash rockets and the 1.9+ style, but at least prevents both occurring within the same hash (https://github.com/bbatsov/rubocop/blob/master/lib/rubocop/cop/style/hash_syntax.rb#L44).

domcleal (changes requested):

.rubocop.yml
dmitri-d: This is messier than I expected: smart-proxy is on an older version of rubocop which only has 'ruby19_no_mixed_keys". When enabled it insists on changing of hash syntax in most places, a change I'd like to avoid until 1.8.7 support has been dropped completely/unambiguously. Ditto for upgrading of rubocop.

Perhaps I can leave this setting disabled for now? Alternatively you could switch to 1.8 syntax in bundler file.

domcleal: Sure, leave it disabled for now, thanks for checking.

domcleal:
Merged, thanks @witlessbird.

.rubocop.yml
daviddavis: Yea, [I just added it a couple months ago](https://github.com/bbatsov/rubocop/pull/3441). It's 0.43+ only.

PR #464, opened 2016-10-11
Fixes #16860 - use blank 404 page

tstrachota:


dmitri-d (changes requested):
This only fixes the issue for the root module, leaving the rest of the modules unaffected. To resolve this for all modules you'll need to change the global error handler (I think), something like: ```

```
class ::Sinatra::Base
  error Sinatra::NotFound do
   content_type 'application/json'
    [404, '']
  end
end
```

Not sure if it makes sense to return different types of content_type.

tstrachota:
@witlessbird updated

lib/sinatra/default_not_found_page.rb
```diff
@@ -0,0 +1,6 @@
+class Sinatra::Base
+  error Sinatra::NotFound do
+    content_type 'application/json'
+    [404, '']
```
dmitri-d: We should return a brief description here, otherwise it can be hard to differentiate between a "url not found" and api "not found" errors.

tstrachota:
@witlessbird thanks, I added message `'Requested url was not found'`

tstrachota:
[test]

dmitri-d:
Merged as a2b8c3b. Thanks, @tstrachota!

PR #465, opened 2016-10-17
Fixes #16964 - Proxy::Dns::Record#dns_find works with ipv6 params

dmitri-d:


modules/dns_nsupdate/dns_nsupdate_main.rb
```diff
@@ -55,23 +55,23 @@ def do_create(id, value, type)
...
 
     def do_remove(id, type)
       nsupdate_connect
-
-      raise Proxy::Dns::NotFound.new("Cannot find DNS entry for #{id}") unless dns_find(id)
       nsupdate "update delete #{id} #{type}"
-
       nsupdate_disconnect
       nil
     end
```
domcleal, line 79: Part two of the ticket was about a missing `ensure` block here, so the omshell processes were leaking. Either add it here or separate into a new ticket if you prefer.

domcleal (changes requested):

domcleal:
I'll try and look at the test failures on Ruby 2.0/2.1, the change to CFLAGS in the latest rkerberos release isn't taking effect.

modules/dns_nsupdate/dns_nsupdate_main.rb
dmitri-d: An ensure block isn't needed for this implementation, as exception is raised before shell is spawned.

dmitri-d:
Thanks, there's an update coming -- get_name can differentiate between ipv4 and ipv6 addresses. If you haven't started looking into the build failure by then, I can take a look at it.

modules/dns_nsupdate/dns_nsupdate_main.rb
domcleal: I think it's worth adding anyway to match the create method - then if another bug occurs later, it's taken care of.

dmitri-d:
- nsupdate provider now always closes om shell.
- when resolving a host name it is now possible to request ip address family.

domcleal:
[test] with rkerberos 0.1.5.

domcleal:
[test]

dmitri-d:
- fixed the issue discovered by Robert Frank (pls. see http://projects.theforeman.org/issues/16964)
- fixed issues related to the one above in record collision detection methods (Proxy::DNS::Record#a_record_conflicts and Proxy::DNS::Record#aaaa_record_conflicts)

dmitri-d:
The failure on ruby 2.3.0 is due to https://bugs.ruby-lang.org/issues/11910, which has been fixed in 2.3.1. We should switch to 2.3.1, as it's the stable release of 2.3.x -- I can take a look at it in the morning.

mmoll:
The reason to stay on 2.3.0 was that Ubuntu/xenial was coming with it, however, 2.3.1 is available from xenial-updates, so I'm :+1: to this, if the version requirement is bumped after merging this for xenial packages.

domcleal (approved):

domcleal:
[test] with 2.3.1.

arogge:
I'd like to suggest another change:
If `get_name(a_ptr)` used `Resolv::DNS#get_resource(ptr, Resolv::DNS::Resource::IN::PTR)` instead of `Resolv::DNS#get_name(ptr_to_ip(ptr))` you could also remove the ugly `ptr_to_ip()` as there is no need to know the IP anymore.

domcleal:
@arogge thanks, that's neater. I'd prefer to do that in a second PR though to prevent this snowballing into a refactor beyond the bug it was intended to fix.

domcleal:
Merged, thanks @witlessbird.

PR #467, opened 2016-10-19
Fixes #17172 - SSH BMC network details implementation

lzap:
This adds few dummy implementations to the SSH provider in order to prevent
Foreman WebUI ugly errors. Without network credentials returned, BMC SSH won't
work at all.

Core counterpart: https://github.com/theforeman/foreman/pull/3953

### DIFF
--- a/modules/bmc/ssh.rb
+++ b/modules/bmc/ssh.rb
@@ -60,6 +60,36 @@ module Proxy
       def powercycle
         ssh(@powercycle)
       end
+
+      def ip
+        host
+      end
+
+      # the following are dummy implementations
+
+      def mac
+        ''
+      end
+
+      def gateway
+        ''
+      end
+
+      def netmask
+        ''
+      end
+
+      def bootpxe(reboot=false, persistent=false)
+      end
+
+      def bootdisk(reboot=false, persistent=false)
+      end
+
+      def bootbios(reboot=false, persistent=false)
+      end
+
+      def bootcdrom(reboot=false, persistent=false)
+      end
     end
   end
 end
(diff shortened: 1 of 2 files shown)


dmitri-d:
Could you create a dedicated ticket for this an reference it in the commit message please?

domcleal:
I'm not sure I understand what the "credentials" are either? Do you mean implementation?

lzap:
Yeah, reworded whole commit and filed a ticket.

modules/bmc/ssh.rb
```diff
@@ -60,6 +60,36 @@ def powerstatus
...
       end
+
+      def ip
+        host
+      end
+
+      # the following are dummy implementations
+
+      def mac
+        ''
```
dmitri-d, line 71: Perhaps instead of empty strings appropriate dummy values would work better? Something like `00:00:00:00:00:00` for mac address, `127.0.0.1` for gateway, etc?

dmitri-d (commented):
Probably worth adding tests for stubbed methods to ensure they have implementations.

modules/bmc/ssh.rb
lzap: These will show up in the WebUI.

Testing these stubs makes no sense, this is a provider aimed for testing purposes, it has no real use.

dmitri-d: > Testing these stubs makes no sense, this is a provider aimed for testing purposes, it has no real use.

Is this PR required then? I'm not suggesting testing stubs, but rather that SSH provider implements methods required for foreman's correct behaviour.

lzap: Ok, added those.

dmitri-d:
Merged in 012ccb5. Thanks, @lzap! .

PR #468, opened 2016-10-20
Fixes #17045 - hostgroup templates URLs proxied correctly

lzap:
It was not working as the route did not match it.

### DIFF
--- a/modules/templates/template_proxy_request.rb
+++ b/modules/templates/template_proxy_request.rb
@@ -8,15 +8,17 @@ module Proxy::Templates
     include Proxy::Log
 
     def get_template(kind, env, params)
-      url = Proxy::Templates::Plugin.settings.template_url
-      proxy_ip = URI.parse(url).host
-      opts = params.clone.merge(:url => url)
+      template_url = Proxy::Templates::Plugin.settings.template_url
+      proxy_ip = URI.parse(template_url).host
+      opts = params.clone.merge(:url => template_url)
       opts.delete("kind")
+      opts.delete("template")
+      opts.delete("hostgroup")
       opts.delete("splat")
       opts.delete("captures")
       proxy_headers = extract_request_headers(env).merge("X-Forwarded-For" => "#{env['REMOTE_ADDR']}, #{proxy_ip}")
       proxy_req = request_factory.create_get("/unattended/#{kind}", opts, proxy_headers)
-      logger.debug "Retrieving a template from %s/%s" % [url, proxy_req.path]
+      logger.debug "Retrieving a template from %s%s" % [uri, proxy_req.path]
       logger.debug "HTTP headers: #{proxy_headers.inspect}"
       res = send_request(proxy_req)
 
--- a/modules/templates/templates_api.rb
+++ b/modules/templates/templates_api.rb
@@ -14,9 +14,15 @@ class Proxy::TemplatesApi < Sinatra::Base
     end
   end
 
+  get "/:kind/:template/:hostgroup" do |kind, template, hostgroup|
+    log_halt(500, "Failed to retrieve #{kind} hostgroup template for #{params.inspect}: ") do
+      Proxy::Templates::TemplateProxyRequest.new.get_template([kind, template, hostgroup].join('/'), request.env, params)
+    end
+  end
+
   get "/:kind" do |kind|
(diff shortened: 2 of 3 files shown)


modules/templates/templates_api.rb
```diff
@@ -14,9 +14,9 @@ class Proxy::TemplatesApi < Sinatra::Base
     end
   end
 
-  get "/:kind" do |kind|
+  get "/:kind/?:template?/?:hostgroup?" do |kind, template, hostgroup|
```
dmitri-d: I think creating a dedicated route for hostgroup templates would be a better approach: the `template` and `hostgroup` parameters will no longer be optional, which would allow for better handling of malformed urls (atm it's  valid to pass `kind` and `template` parameters only).

test/templates/template_api_test.rb
```diff
@@ -30,4 +30,11 @@ def test_api_can_ask_for_a_template
     assert_match("An API template", last_response.body)
   end
 
+  def test_api_can_ask_for_a_hostgroup_template
+    Proxy::Templates::TemplateProxyRequest.any_instance.expects(:get_template).returns("An API template")
```
dmitri-d: This test could be improved by adding a call parameter validation.

test/templates/template_api_test.rb
```diff
@@ -30,4 +30,11 @@ def test_api_can_ask_for_a_template
...
+    get "/template/hostgroup1/template1", @args
+    assert last_response.ok?, "Last response was not ok"
+    assert_match("An API template", last_response.body)
```
dmitri-d, line 40: This assertion duplicates the expectation set on line 34 and can be removed.

dmitri-d (changes requested):

lzap:
Good points, amended both changes.

test/templates/template_api_test.rb
```diff
@@ -18,16 +18,22 @@ def setup
...
     Proxy::Templates::TemplateProxyRequest.any_instance.expects(:get_template).returns("An API template")
     get "/provision", @args
-    assert last_response.ok?, "Last response was not ok"
+    assert last_response.ok?, "Last response was ok"
     assert_match("An API template", last_response.body)
   end
 
+  def test_api_can_ask_for_a_hostgroup_template
+    Proxy::Templates::TemplateProxyRequest.any_instance.expects(:get_template).with("kind/temp/hg", anything, anything).returns("An API template")
+    get "/kind/temp/hg", @args
```
dmitri-d: I don't see any verification (here and in the test above) that the contents of @args is being passed into TemplateProxyRequest. Ideally, there should be a dedicated test that verifies that request parameters are being passed further along. At the very least, could you skip args or pass an empty hash in these two tests instead?

lzap: Hey, I don't understand, isn't this tested on line 34 here?

lzap: Wait, I see I don't actually proxy the request correctly. Hmm.

lzap: Ok, I introduced webmock to test it end-to-end. Added a test for token and template_url.

modules/templates/template_proxy_request.rb
```diff
@@ -8,15 +8,17 @@ class TemplateProxyRequest < ::Proxy::HttpRequest::ForemanRequest
...
+      opts = params.clone.merge(:url => template_url)
       opts.delete("kind")
+      opts.delete("template")
+      opts.delete("hostgroup")
       opts.delete("splat")
       opts.delete("captures")
       proxy_headers = extract_request_headers(env).merge("X-Forwarded-For" => "#{env['REMOTE_ADDR']}, #{proxy_ip}")
       proxy_req = request_factory.create_get("/unattended/#{kind}", opts, proxy_headers)
-      logger.debug "Retrieving a template from %s/%s" % [url, proxy_req.path]
+      logger.debug "Retrieving a template from %s%s" % [uri, proxy_req.path]
```
lzap, line 21: FYI this was incorrect, the debug message was showing template_url instead foreman_url. I also renamed the `url` variable to reflect this is template_url not to confuse with `uri` which is foreman_url.

dmitri-d:
Merged as 3b81b1d. Thanks, @lzap!

PR #469, opened 2016-10-20
Fixes #17037 - Resolver::DNS#getresources is used now.

dmitri-d:
It replaces calls to Resolver::DNS#getaddresses and Resolver::DNS#getnames in dns module and its providers.


arogge:
Can we please also deprecate `ptr_to_ip()` and remove it together with `find_dns()` in a future release? AFAICT `find_dns()` is the only function that still calls it.

dmitri-d:
> Can we please also deprecate ptr_to_ip() and remove it together with find_dns() in a future release? AFAICT find_dns() is the only function that still calls it.

Proxy::DNS::Record#dns_find has been deprecated. Proxy::DNS::Record#ptr_to_ip is required in some non-core dns providers and cannot be removed.

domcleal:
Merged, thanks @witlessbird.

PR #472, opened 2016-11-04
Fixes #3034 - Better handling of errors during tftp downloads

dmitri-d:


lzap:
My biggest concern is change of behaviour. Currently we check timestamp reported by HTTP server and if the file was not modified, it does not re-download. If I read the code correctly, this will re-download every time which is a regression as users have slow networks (VPNs etc) and initial kernel/image download is usually a problem but thank to the timestamp comparison it is not big deal (you wait once for each OS associated in Foreman). The change introduced will be painful for them.

dmitri-d:
> Currently we check timestamp reported by HTTP server and if the file was not modified, it does not re-download.

Where do we do that?

modules/tftp/http_downloads.rb
```diff
@@ -0,0 +1,297 @@
...
+            @status.file_length = response['content-length'].to_i rescue 0
+            @status.update_downloaded(start_position) if response.code == '206'
+            open(@dst_path, (response.code == '206' ? 'ab' : 'wb')) do |io|
+              if @stop
+                @status.stopped = true
+                break
+              end
+              response.read_body {|b| @status.update_downloaded(b.length); io.write(b)}
+            end
+          end
```
lzap, line 96: _-I don't see any IO seek in the target file in this method, how this code can correctly resume downloads? This cannot work.-_ Scratch that, I mis-read the code, in case of partial it appends.

I think we are reinventing wheel with this, it would be easier to spawn wget on the background and only watching the files growing for each individual status API call rather than implementing this. This code looks fragile.

Isn't synchronized queue better approach than mutex with synchronized blocks everywhere? Such a job queue could easily handle scheduling of re-downloads (new download could be always scheduled three times, worker thread would pick them from the queue on time skipping already finished jobs). Plus queue could help to decouple implementation from the rest of the code.

modules/tftp/http_downloads.rb
```diff
@@ -0,0 +1,297 @@
...
+        end
+      end
+
+      def restart
+        start(File.size?(@dst_path))
+      end
+
+      def range_header(range_start)
+        {'Range' => "bytes=#{range_start}-"}
+      end
```
lzap, line 106: Not all HTTP servers support ranges, we need to handle that correctly and add a test for this.

ohadlevy:
I wonder if we can keep it simply to becoming a blocking action? new installs (e.g. when the kernel/initrd are not deployed) will have a longer progress bar, future provisioning events will be faster.

if this is a real problem (e.g takes more than a minute or some other threshold) IMHO we should discuss having a provisioning setup phase in foreman, where we pre-check that all of requirements for provisioning are met, and only then allow provisioning to start on a host.

my concern is that we only solve it on the smart proxy side without foreman side will not be enough to alleviate the pain^H^H usability users are facing when setting up new operating system for provisioning.

modules/tftp/http_downloads.rb
dmitri-d: I do check the response, pls. take another look at the code.

lzap:
> Where do we do that?

`wget -c` compares -timestamps- (I was wrong) _sizes_ and if remote is bigger, it does not re-download. Now, your implementation should do the same (I'd prefer explicit check rather than opening the file for write), can you add a test for this case?

lzap:
Whoops sorry.

modules/tftp/http_downloads.rb
dmitri-d: > I don't see any IO seek in the target file in this method, how this code can correctly resume downloads? This cannot work.

'Append' mode is used on successful resumes, which appends to the end of an existing file, no need for a seek call.

The code dealing with actually downloading stuff is pretty small (and well tested). But it gives me ability to respond to different modes of failures (transient failures vs. permanent), which is very useful. 

> Isn't synchronized queue better approach than mutex with synchronized blocks everywhere?

I would still need synchronization, as there's a couple of pieces of shared state (id to download and url to download mappings; synchronization around destination dir creation, etc).

dmitri-d:
> I wonder if we can keep it simply to becoming a blocking action? new installs (e.g. when the kernel/initrd are not deployed) will have a longer progress bar, future provisioning events will be faster.

I would rather not, as that would tie down a thread in a webserver.

> my concern is that we only solve it on the smart proxy side without foreman side will not be enough to alleviate the pain^H^H usability users are facing when setting up new operating system for provisioning.

Yes, you absolutely need changes on foreman side, I don't think they are huge though.

lzap:
Blocking is not good idea, I am with Dmitri. I think better approach then this is to download kernel/initram disk during OS creation or update or during content promotion and then having some status flag next to OS or CV with "Ready for provisioning" informing the user that all files were downloaded correctly.

ohadlevy:
@witlessbird a couple of options for non blocking:
1. introduce callbacks once the download is complete (you can still use wget in that case)
2. use dynflow (just like in rex plugins)

@lzap sounds like we are agreeing this should be part of a larger design on foreman side as well?

dmitri-d:
@ohadlevy: I'd rather not.
1. This is rest api. If we want reactive api, we should move away from rest (still not using callbacks though).
2. I'm not about to introduce a workflow framework (and all that it entails) where basic promises work better.

ohadlevy:
@witlessbird for option 1. couldn't we have an API end point to tell if us if there is a download in progress? (e.g. think poor mans task management?) this could simply check if wget (or another process) is still running?
to me the overhead of rewriting wget and concurrent ruby makes it less appealing TBH.

dmitri-d:
> wget -c compares -timestamps- (I was wrong) sizes and if remote is bigger, it does not re-download.

I guess it makes sense for current implementation; Although I think these should be two different calls -- one to check if files differ (only by filesize though), and second to download a file.

lzap:
I don't know if it's the time for bigger refactoring of this, TFTP downloading is a pain but a change in core to move TFTP downloading from Host Creation to OS creation might be as big patch as this one (+584 −61) maybe even smaller. All we need to do is move the orchestration step and at the beginning of host creation check if the file were already downloaded raising error if not.

Anyway, this is perhaps wrong place to discuss this - this should be in the ticket (and I made such a note in it already).

dmitri-d:
> for option 1. couldn't we have an API end point to tell if us if there is a download in progress? (e.g. think poor mans task management?) this could simply check if wget (or another process) is still running?

https://github.com/theforeman/smart-proxy/pull/472/files#diff-0f25ac5f8eadf527321f922b2f55b140R42

> rewriting wget and concurrent ruby makes it less appealing TBH.

As I pointed out, handling downloads isn't that hard, but I get much better error handling out it. Concurrent ruby is already in proxy; using promises makes handling of asynchronous responses much easier to understand and test (which is the main driver behind it). Also, you just suggested using a workflow framework, this is orders of magnitude lower overhead compared to dynflow?

lzap:
> As I pointed out, handling downloads isn't that hard, but I get much better error handling out it. Concurrent ruby is already in proxy; using promises makes handling of asynchronous responses much easier to understand and test (which is the main driver behind it).

Debugging TFTP downloading at users and customers was always a big pain and I am not sure I will find a confidence to switch from single `wget` call that just works to 500 lines of code with multiple threads (hidden dragons). We have unit tests, sure, but our real-world end-to-end testing was always a pain and I am little bit concerned about this.

To be more constructive, do you think it is hard to decouple the implementation and provide a `wget` based one along with the new Ruby-based one so if there is an issue (and data corruption which leads to nasty random errors when Anaconda is booting up), we can instruct users to switch to simple wget implementation? I think the status API can be as simple as "is it finished yet? yes or no" for the "fallback" implementation, we don't need fancy progress bars.

dmitri-d:
> do you think it is hard to decouple the implementation and provide a wget based one along with the new Ruby-based one so if there is an issue

Yes, I think it's going to be pretty hard, as they won't have much in common at all. Please look at the code, and if you see spots where tests can be improved, please let me know. FWIW I don't think this code is risky at all -- getting a file over http isn't that hard. In fact it's probably more reliable, as it tries pretty hard to download the requested file (including resuming downloads after a crash/restart).

dmitri-d:
> I think the status API can be as simple as "is it finished yet? yes or no" for the "fallback" implementation, we don't need fancy progress bars.

It's not much harder to keep track of download vs. "done"/"not done" reporting (one line of code).

dmitri-d:
[test]

dmitri-d:
Added a staleness check when attempting to download a locally existing file.

test/tftp/tftp_api_test.rb
```diff
@@ -105,10 +122,24 @@ def test_api_can_create_defatult
     assert last_response.ok?
   end
 
-  def test_api_can_fetch_boot_file
-    Proxy::TFTP.expects(:fetch_boot_file).with('/some/root/boot/file','http://localhost/file').returns(true)
-    post "/fetch_boot_file", :prefix => '/some/root/boot/file', :path => 'http://localhost/file'
+  def test_api_can_dowload_boot_file
```
timogoebel: nitpick, typo: download

dmitri-d:
rebased

dmitri-d:
and fixed the typo in test/tftp/tftp_api_test.rb.

timogoebel:
@lzap , @ohadlevy , @witlessbird : How are we going to proceed here?
What bugs me (in the real world) is mostly, that when tftp files don't get downloaded correctly because of some errors (mirror broken, IPS, Firewall, ... you name it) a user won't find out that there was a problem. The files are even created in the tftp directory - but empty. That makes this even harder to spot ("Timo, why is provisioning not working? Look, the files are even there. But they are not picked up."). This is a pain to debug and I want to improve this. The code seems reasonable to me and I do prefer it over a wget approach because we have more control about what's going on and can log appropriate error messages to help users fix problems.
I think, what we most importantly need is user feedback in Foreman. Smart Proxy should report if it can't start downloading a file to Foreman and we should show the error in the UI. Preferably with a meaningful error message. It should not be possible to provision a host when the tftp files aren't deployed properly. We can surely improve this, but other than missing error messages the current approach has served us well imho. Let's not overcomplicate this.

ohadlevy:
@timogoebel my preferred option is to "do it right" in foreman. I would like to see us splitting the provisioning related attributes in foreman into a new object, which we will allow a user to setup provision profile (or some better name) - this will configure os, templates, enabled proxies etc.
when setting it up, we will ensure provisioning is ready (including downloading tftp files) this imho will simplify the pain with initial provisioning setup, remove the problem of os changes via facter and then you can't rebuild and will in overall reduce the clutter in the host object.

timogoebel:
> I would like to see us splitting the provisioning related attributes in foreman into a new object

Honestly I don' think this is a good idea, but maybe I just can't imaging the outcome. Maybe we should chat about that over a beer (or some other beverage of your choice) in Belgium. ;-)

> my preferred option is to "do it right" in foreman

That's always my preferred solution, too. However I think this PR is a step in that direction. My current feeling is, that we currently discuss too much in the project and don't get things done. An iterative approach like with the notifications would be better. In my opinion @witlessbird did a great job with the code and it improves the current situation. So let's not block this. It won't stop us from improving other areas later.

dmitri-d:
rebased.

dmitri-d:
FWIW: Unless the whole thing is going to be moved into foreman, these (or similar) changes would be required on the proxy side. Not sure what are other options are available to us.

dmitri-d:
Would it be possible to get a clarification on "do it right"? What needs be changed in this PR to get there?

lzap:
Apologize I am late in reply, but I wrote my concerns already. The API is a bit complex, I can imagine we only need to know if a download job is finished or not. We already have a lot of ugly "status" calls in Katello when Foreman queries backend systems every second. If we had a messaging, that's another story, but we don't have that.

If the API is simpler, we could keep the old wget implementation as an alternative. In the enterprise, there are firewalls and HTTP proxies that sometimes misbehave (breaking rules, adding unnecessary delays etc), I believe wget have lots of workarounds already implemented. But I am fine with new pure ruby implementation, if we had a chance to keep the old one as an alternative. I am mostly worried about debugging this at the customer, with all respect to @witlessbird coding skills. I am just extra defensive as I spend most of the time working on Satellite 6 customer cases last couple of months.

But more importantly, downloading kernel when build is triggered is too late and we need to change this in the future ("do it right"). Then this patch becomes less relevant, there will be plenty of time between when OS is created and new host provisioned and also progress will not be that important. Sure, it's nice to have, but with that amount of code, threads and storing metadata, I simply see this patch as a pricey workaround.

Please do not take it as I am blocking this, I am just not doing formal review and testing of this patch. If you plan to do, make sure you test this against Katello's "ondemand" policy which makes Pulp kickstart repository to return HTTP redirects for all files (including kernel/initram disk) to verify the Ruby library handles that correctly.

dmitri-d:
> I can imagine we only need to know if a download job is finished or not. We already have a lot of ugly "status" calls in Katello when Foreman queries backend systems every second.

http://projects.theforeman.org/issues/3034 explicitly mentions tracking of progress. Implementation-wise there is little difference between done/not done response and progress tracking. Until we have bi-directional communication between smart-proxy and Foreman, I don't see how we can get away from using some form of polling.

> we could keep the old wget implementation as an alternative

Wget implementation comes with its own issues. I don't want to maintain both, esp. considering how wildly different both are. I appreciate the desire to simplify support, this implementation was done with this goal in mind: it more testable, it has better error reporting, and it attempts to resolve interrupted downloads on its own.

> But more importantly, downloading kernel when build is triggered is too late and we need to change this in the future ("do it right"). 

This sounds good to me, and I think changes in this PR would support such a change in Forman:  tracking of download status, including errors would most certainly be useful.

>  make sure you test this against Katello's "ondemand" policy which makes Pulp kickstart repository to return HTTP redirects for all files (including kernel/initram disk) to verify the Ruby library handles that correctly.

This implementation doesn't handle redirects, I'll look into adding support for them.

lzap:
We have a report (not yet confirmed) on misbehaviour of wget that could make this PR more relevant. We still need confirmation by our consultant at the customer site (for the record the thread is named "Satellite serving RHEL 7.2 kernel during RHEL 7.3 kickstart"), but my assumption what happened is:

* customer created new OS named RHEL 7.3 against 7.2 repos
* customer initiated provisioning of this OS
* wget downloaded kernel and image onto smart-proxy
* customer fixed the OS and changed the repos to real 7.3
* customer initiated provisioning of this OS
* wget started downloading kernel and image:
  * init RAM disk size differs, the remote is bigger = redownloaded
  * kernel image differs, but the remote is *smaller* = wget will not download
* Anaconda blows up with "Unknown XFS file system" due to mismatch of kernel and init ram disk with modules (https://bugzilla.redhat.com/show_bug.cgi?id=1169034)

It looks like this is known bug with solution article: https://access.redhat.com/solutions/2115301

Therefore I suggest to add extra test - try to redownload completely different file, smaller or bigger than the one that is already downloaded. Ideally we should be also take creation or modification time into account if that's provided by the HTTP server.

dmitri-d:
The size check is already there: https://github.com/theforeman/smart-proxy/pull/472/files#diff-a7b380254558ab8c9965ee50c789bf46R125. I'll add more tests though. 

Instead of date/time, I'd use etags instead (but that would mean keeping metadata around).

One way to completely resolve issues in this area is to extract media and tftp service into a dedicated service that manages the whole lifecycle of bootstrap files.

lzap:
I can confirm the mis-behaviour of `wget`, I think the PR is back in business, I will take a look on that once I do my huge post-FOSDEM and illness backlog. Please add as many tests as possible around this (remote file is bigger or smaller). Also keep in mind that we need to work reliably with very limited HTTP server that basically only provides size of file, or maybe not even that (not all HTTP servers provides that but it's rather uncommon - at least we should raise error or warning if we encounter that).

FYI the relevant man page part is:

> Beginning with Wget 1.7, if you use ‘-c’ on a file which is of equal size as the one on the server, Wget will refuse to download the file and print an explanatory message. The same happens when the file is smaller on the server than locally (presumably because it was changed on the server since your last download attempt)—because “continuing” is not meaningful, no download occurs.
> 
> On the other side of the coin, while using ‘-c’, any file that’s bigger on the server than locally will be considered an incomplete download and only (length(remote) - length(local)) bytes will be downloaded and tacked onto the end of the local file. This behavior can be desirable in certain cases—for instance, you can use ‘wget -c’ to download just the new portion that’s been appended to a data collection or log file.
> 
> However, if the file is bigger on the server because it’s been changed, as opposed to just appended to, you’ll end up with a garbled file. Wget has no way of verifying that the local file is really a valid prefix of the remote file. You need to be especially careful of this when using ‘-c’ in conjunction with ‘-r’, since every file will be considered as an "incomplete download" candidate.
> 
> Another instance where you’ll get a garbled file if you try to use ‘-c’ is if you have a lame HTTP proxy that inserts a “transfer interrupted” string into the local file. In the future a “rollback” option may be added to deal with this case.

When remote file was replaced with different content and it was bigger, wget screwed things by only downloading the remaining bytes appending them. That's the issue we are currently facing.

Since wget does not seem work correctly, I have no problems with Ruby rewrite anymore. Again, I will get back to this after some time, so if @timogoebel is interested reviewing this, go ahead!

timogoebel:
Glad, this PR is back in business. I'll happily review once @witlessbird thinks, this is ready.

I've been thinking about the Foreman side of this. In Foreman we do want to know, when a TFTP download is finished correctly and only then continue with other orchestration steps. If an error occurs, we want to fail the orchestration step (and rollback the queue). On the other hand, we don't want to use continuous polling in Foreman to request the download status from smart-proxy. I'm curious why we haven't considered using websockets or long-polling or something like this to do the job. Any opinions in this regard?

dmitri-d:
rebased, added tests for local copy staleness check.

Gemfile
```diff
@@ -1,6 +1,9 @@
 source 'https://rubygems.org'
 
 gemspec
+# this cannot go into gemspec, bundler_ext fails to load this gem.
+# see http://projects.theforeman.org/issues/16760
+gem 'concurrent-ruby', '~> 1.0', require: 'concurrent'
```
lzap, line 6: I am not sure I understand, the ticket does not clears this to me. Gemspec does not support "require" statement I assume.

modules/tftp/http_downloads.rb
```diff
@@ -0,0 +1,318 @@
...
+        begin
+          return url_to_download[src_url] if url_to_download.key?(src_url)
+          unless restart
+            FileUtils.mkdir_p(Pathname.new(filepath).parent)
+            create_metadata(metadata_filepath, src_url, id)
+          end
+          start_or_restart_download(filepath, tmp_filepath, metadata_filepath, src_url, id, restart)
+          id
+        rescue Exception => e
+          logger.debug("Error creating metadata for download from '#{src_url}'", e)
```
lzap, line 242: Catching too much, this will log error about metadata for non-metadata exceptions.

modules/tftp/http_downloads.rb
```diff
@@ -0,0 +1,318 @@
...
+            remove_file(metadata_filepath)
+          end
+        end
+        .on_error do |reason|
+          logger.debug("Error when downloading '#{src_url}'", reason)
+          download_in_progress.status.last_error = reason
+          remove_file(tmp_filepath)
+          remove_file(metadata_filepath)
+        end
+        .then { m.synchronize { url_to_download.delete(src_url) }}
```
lzap, line 271: Will the `then` block be executed if error occurs? It does not looks like.

Also, `url_to_download` is synchronized here, but not all manipulation is synchronized (e.g. line 254 above).

modules/tftp/plugin_configuration.rb
```diff
@@ -0,0 +1,14 @@
...
+  class PluginConfiguration
+    def load_classes
+      require 'tftp/server'
+      require 'tftp/http_downloads'
+      require 'tftp/dependency_injection'
+      require 'tftp/tftp_api'
+    end
+
+    def load_dependency_injection_wirings(container_instance, settings)
+      container_instance.singleton_dependency :http_downloads, lambda {::Proxy::TFTP::HttpDownloads.new(settings[:tftproot])}
```
lzap, line 11: The code is using the default thread pool, I'd like to have this configurable for HTTP downloads. I assume that the default Ruby Concurrent thread pool is something like by default it creates CPU cores + N which is not appropriate for network downloading. I think the decent default number of concurrent downloads is 2-5. I'd definitely would let user to decide depending on the network (VLAN, VPN, slow or low latency network).

modules/tftp/http_downloads.rb
```diff
@@ -0,0 +1,318 @@
...
+      def file_length=(length)
+        @m.synchronize { @file_length = length }
+      end
+
+      def update_downloaded(chunk_length)
+        @m.synchronize { @downloaded += chunk_length }
+      end
+
+      def progress
+        (fl = file_length) == 0 ? 0 : ((downloaded.to_f/fl) * 100).round(2).to_i
```
lzap, line 46: Isn't this code supposed to be in synchronized block? You are accessing both numbers at once. The same goes for `to_hash` method down below. If your goal was only to keep atomicity of reads and writes, consider using `AtomicFixnum` class instead which will be cleaner.

modules/tftp/http_downloads.rb
```diff
@@ -0,0 +1,318 @@
...
+    end
+
+    def is_local_copy_stale?(filepath)
+      return true unless File.exist?(filepath)
+      to_return = Concurrent::Promise.new { new_downloader.size }
+                                     .on_success { |result| result == 0 || result != File.size(filepath) }
+                                     .on_error { true }
+                                     .execute
+                                     .value(10) # don't block for too long on the check, fail it if we don't receive a response in 10 seconds
+      to_return.nil? ? true : to_return # Concurrent::Promise#value returns nil if the promise is in 'pending' state
```
lzap, line 130: Can be simplifed to `... || true`

modules/tftp/http_downloads.rb
```diff
@@ -0,0 +1,318 @@
...
+    end
+
+    attr_reader :status, :m
+
+    def initialize(dst_path, url, status = Status.new)
+      @status = status
+      @dst = dst_path
+      @url = URI.parse(url)
+      @downloader = nil
+      @m = Mutex.new
```
lzap, line 120: Add descriptive name to all mutexes in this patch with comments what is the purpose. While this one is easier to understand, the one below is not that clear. I would like to prevent future changes mis-use of mutexes which is a typical hard to debug error.

modules/tftp/http_downloads.rb
```diff
@@ -0,0 +1,318 @@
...
+
+    # download_in_progress is used in tests only
+    attr_reader :m, :url_to_download, :id_to_download, :downloader_class, :download_in_progress
+
+    def initialize(dst_dir, downloader_class = HttpDownload)
+      @url_to_download = {}
+      @id_to_download = {}
+      @dst_dir = Pathname.new(dst_dir).cleanpath.to_s
+      @downloader_class = downloader_class
+      @m = Mutex.new
```
lzap, line 191: Ditto. Add descriptive name to all mutexes in this patch with comments what is the purpose. This one looked like thread-safe hash at first sight (I'd use Concurrent::Hash instead in this case). But then I see huge lock in `do_download` method, please explain in the comment.

modules/tftp/http_downloads.rb
```diff
@@ -0,0 +1,318 @@
...
+          http.request(get) do |response|
+            raise ::Proxy::TFTP::HttpError.new(response.code, response.message) unless response.is_a?(Net::HTTPSuccess)
+            @status.file_length = response['content-length'].to_i rescue 0
+            @status.update_downloaded(start_position) if response.code == '206'
+            open(@dst_path, (response.code == '206' ? 'ab' : 'wb')) do |io|
+              if @stop
+                @status.stopped = true
+                break
+              end
+              response.read_body {|b| @status.update_downloaded(b.length); io.write(b)}
```
lzap, line 94: Write first and then update status with amount of bytes written. If there is a failure (disk full), this will end up with more consistent state.

lzap (dismissed):
I find this needlessly complex, this could be as simple as synchronized queue and thread pool of workers picking up tasks downloading whole files and another sync queue accepting status information for status requests. This would work without any shared state, therefore almost without synchronization.

lzap:
Don't get me wrong, there's a lot of good code (actual downloading, handling errors, storing metadata), but I don't like the highlevel design with promises and many synchronization points. Consider to refactor this to something more robust.

lzap:
Also please rename the commit and PR subject to something more appropriate. This is async download with new status HTTP endpoint now. It is worth giving it some description as it's quite unclear from the issue how the implementation should look like. Thanks.

modules/tftp/http_downloads.rb
dmitri-d: No. This uses a mutex, you'll get a "ThreadError: deadlock" if you attempt to acquire a mutex multiple times from the same thread.  This is fine as ```file_length``` and ```downloaded``` are initialized/updated at different times and individually synchronized.

modules/tftp/http_downloads.rb
dmitri-d: > Will the then block be executed if error occurs?

yes.

> Also, url_to_download is synchronized here, but not all manipulation is synchronized (e.g. line 254 above).

```start_or_restart_download``` is only called from under a mutex, none of the calls here need synchronization (or can use it, as it is a mutex). The promise created in ```#then``` is executed in a different thread, and needs a dedicated synchronization.

This does warrant a comment, however.

File: Gemfile (hunk 4 above)
Comment (dmitri-d) [reply Gemfile:6]: This is related to difference between bundler and bundler_ext in handling loading errors. The ticket has the details.

**Comment by dmitri-d** [comment]
> I find this needlessly complex, this could be as simple as synchronized queue and thread pool of workers picking up tasks downloading whole files and another sync queue accepting status information for status requests. This would work without any shared state, therefore almost without synchronization.

This *is* using a threadpool, with HttpDownload:: Downloader class a worker in that pool. Using a queue for keeping track of status is a good idea, but I picked a different implementation and I don't want to change now. Additionally, such a queue needs to be bounded, but that raises questions re: lost updates, esp. completion ones.

**Comment by lzap** [comment]
> This is using a threadpool, with HttpDownload:: Downloader class a worker in that pool. Using a queue for keeping track of status is a good idea, but I picked a different implementation and I don't want to change now. Additionally, such a queue needs to be bounded, but that raises questions re: lost updates, esp. completion ones.

Thread pool was not *the* point of my comment at all. What's important is amount of shared state thus synchronization, complexity of the code and use of futures pattern which is obviously not the best tool for this kind of job. I spent an hour figuring out code flows and synchronization points. I can imagine you spent much more time on this code which makes you uncomfortable changing it from the ground up, but this is not an argument. Let's find why this approach is better than my proposal:

```
Download Request -> [URL channel*] -> Worker N -> [Status buffers] -> Status request
```

Jobs can be serialized to metadata files on changes (started or finished). Status buffers as well ideally using timer to prevent major data loss, if this was your point ("lost updates"). We can keep track of statuses of all failed jobs for the last of 24 hours, all missing ones can be considered as finished. To be honest, I'd rather exclude metadata serialization from this patch completely, added value is questionable.

Now, what I propose can be implemented without a single synchronization block (using thread-safe types from Concurrent Ruby). The workflow is simplified to single class (download worker) that does a single job while pushing out status info. It is quite a change and I don't doubt it at all, but I believe it is worth the effort. The most relevant code which handles downloading stays as-is.

* By channel I really mean thread-safe implementation from Ruby Concurrent library, or any kind of thread-safe FIFO data structure.

dmitri-d:
> What's important is amount of shared state thus synchronization, complexity of the code and use of futures pattern which is obviously not the best tool for this kind of job.

Shared state is limited to download stats and download-id to download job mappings. You'll need to maintain the latter in order to prevent multiple downloads of the same content. I disagree re: promises, this is exactly how they should be used.

As for the queue for status updates: it needs to be bounded; probably the easiest way to manage overflow in a queue is to drop least important items, which means a bounded priority queue is a good fit. There isn't a priority queue implementation in the standard library, and I won't introduce a dependency just for that. Insertions in a priority queue are definitely harder to understand than synchronization code in this PR. 

There still will be synchronization to support old state cleanups.

Most importantly, however, is that I would entertain the idea of rewriting this, had it come in November when this PR was first introduced. After five months I will absolutely not change the implementation, other than to fix bugs -- I'm sick and tired of how long the process' been dragging on.

lzap:
> As for the queue for status updates: it needs to be bounded; probably the easiest way to manage overflow in a queue is to drop least important items, which means a bounded priority queue is a good fit. There isn't a priority queue implementation in the standard library, and I won't introduce a dependency just for that. Insertions in a priority queue are definitely harder to understand than synchronization code in this PR.

There is no reason to do status updates via (bounded) queue, this could be simply synchronized hash of structs holding status data. I think I've called it queues once, changed it to "status buffers" in my last comment.

> There still will be synchronization to support old state cleanups.

Synchronized hash provides you safe iterators, so no. Anyway, before you respond, let's end this. This is not going to anywhere.

> Most importantly, however, is that I would entertain the idea of rewriting this, had it come in November when this PR was first introduced. After five months I will absolutely not change the implementation, other than to fix bugs -- I'm sick and tired of how long the process' been dragging on.

There was exactly one code comment in this PR by Timo, it was a typo/nitpick. Other than that generic and overall concerns and discussion about the approach. It was just you rebasing it several times, you did this yourself, nobody asked. More than that, I raised my concerns literally *three days* after you filed the PR, including my "Isn't synchronized queue better approach...".

Please do my comments from my today's review when appropriate, there are several minor issues I found which are worth changes. Then let me know when you're done and I will dismiss my review and put my hands off this. I am not merging this excercise in concurrent ruby futures. Age of the PR will not change my mind, it's completely irrelevant.

dmitri-d:
> There is no reason to do status updates via (bounded) queue, this could be simply synchronized hash of structs holding status data. I think I've called it queues once, changed it to "status buffers" in my last comment.

You described your approach using the word "queue" from the very start, including today. You used "status buffers" twice, including the comment above (and I don't think you meant to use the word "buffer" if you had some sort of map in mind). I don't think you can blame me for thinking about an entirely different approach. 

Not passing around a global state hash was a conscious decision -- I opted for keeping download state close to HttpDownload object that's responsible for said download. 

> More than that, I raised my concerns literally three days after you filed the PR, including my "Isn't synchronized queue better approach...".

And I disagreed with all of them, including your objection to not using wget. And synchronized hash isn't a queue.

lzap:
I won't play this word game, I am done here.

PR #473, opened 2016-11-08
fixes #17262 - verify the JSON request body contains a hash

domcleal:
The json 2.x gem now parses strings and other non-hash root nodes, so
the data type need to be checked when parsing JSON bodies.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,5 +1,9 @@
 inherit_from: .rubocop_todo.yml
 
+Metrics/ClassLength:
+  Exclude:
+    - 'test/**/*'
+
 # Offense count: 11
 Lint/AmbiguousRegexpLiteral:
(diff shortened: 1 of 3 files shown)


domcleal:
http://ci.theforeman.org/job/test_proxy_develop_pull_request/1397/ is a manual test run, but an issue on Ruby 1.9.3 remains. Will send a second pull request for that soon.

dmitri-d:
OT: should we continue testing on 1.9.3? I haven't heard any complaints (yet) to my suggestion to officially discontinue 1.9.3?

domcleal:
Sure, if you like - change it in foreman-infra.

dmitri-d:
will do.

dmitri-d:
Merged as c896ba3. Thanks, @domcleal.

PR #474, opened 2016-11-11
Fixes #16079 - subnet recursive parsing

lzap:
This patch adds recursive parsing, but there is one test failing and I am not
sure how to handle that. Do we care about pools inside subnets? I think we
don't, in that case I will just fix the failing test. Opinion?

### DIFF
--- a/modules/dhcp_isc/isc_file_parser.rb
+++ b/modules/dhcp_isc/isc_file_parser.rb
@@ -58,7 +58,19 @@ module Proxy::DHCP::ISC
       ret_val
     end
 
-    SUBNET_BLOCK_REGEX = /subnet\s+([\d\.]+)\s+netmask\s+([\d\.]+)\s*\{\s*([\w-]+\s*\{[^{}]*\}\s*|[\w-][^{}]*;\s*)*\}/
+    SUBNET_BLOCK_REGEX = %r{
+      subnet\s+(?<subnet>[\d\.]+)\s+
+      netmask\s+(?<netmask>[\d\.]+)\s*
+        (?<body>
+          \{
+            (?:
+              (?> [^{}]+ )
+              |
+              \g<body>
+            )*
+          \}
+        )
+      }x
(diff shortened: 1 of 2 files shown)


dmitri-d:
The test that fails is verifies parsing of subnet's range and which is returned with other subnet data (when available). I don't think we can drop that.

lzap:
Range can be declared either with or without pool block. We currently do not support multiple ranges at all, so if I understand this correctly, this test makes sure that range declared outside pool block (inside subnet) wins to all other ranges defined in pool blocks.

Now the problem is I had to start looking for "pool" block to workaround that, that is essentially another rather complex regexp which I would like to avoid. How about changing this behavior to "first range wins"?

lzap:
What I could is I could simply ignore everything within the pool block, would that work?

lzap:
Ok pushed the change, this is explicit bit of code that ignores all content in `pool { ... }` which does not break current behavior. Still it fixes the bug that this patch aims for. We definitely need LALR parser here, but I am actually doing a plugin for `dnsmasq` during evenings that will be much simpler than for ISC. A good alternative that we can deliver fast than breaking whole parsing...

dmitri-d:
[test]

dmitri-d:
I think ignoring the contents of pool blocks is ok as quick-fix for parsing problems.

lzap:
Now the biggest question is if the regexps will work on 1.9. They should... let's see, I only tested this on 2.0.

lzap:
Oh it looks like green, rubocop green locally as well.

dmitri-d:
Develop branch no longer supports 1.8.7 and 1.9.3.

lzap:
Wow green, to be honest I did not expect this to pass.

dmitri-d:
Merged in 7f016ec. Thanks, @lzap!

lzap:
Also added warning block into docs to warn users: https://github.com/theforeman/theforeman.org/pull/744

PR #475, opened 2016-11-14
Fixes #15874 - added logging of 'X_REQUEST_ID' http headers

dmitri-d:


dmitri-d:
Some notes:
- for standard logger I replaced process id with request_id. I can restore the process id if it's important. Should I surround request id with <>'s (to help with parsing)?
- I can return request id with smart-proxy responses if this would be useful

lib/proxy/log.rb
```diff
@@ -20,6 +20,46 @@ def logger
   end
 
   class LoggerFactory
+    class LogFormatter
+      Format = "%s, [%s%s] %5s -- %s: %s\n".freeze
+
+      def call(severity, time, progname, msg)
+        Format % [severity, format_datetime(time), request_id[0..5], severity, progname, msg2str(msg)]
```
domcleal: The first `severity` field used to be a single letter (`severity[0..0]` in stdlib) so it's now duplicated and affects the indentation. Previously:

<pre>
D, [2016-11-10T13:34:04.946305 #20953] DEBUG -- : require_ssl_client_verification: skipping, non-HTTPS request
I, [2016-11-10T13:34:04.946767 #20953]  INFO -- : 192.168.122.1 - - [10/Nov/2016 13:34:04] "GET /tftp/serverName HTTP/1.1" 200 30 0.0007
</pre>


Now:

<pre>
DEBUG, [2016-11-15T13:12:52.052872 ] DEBUG -- : Rack::Handler::WEBrick is mounted on /.
INFO, [2016-11-15T13:12:52.052942 ]  INFO -- : WEBrick::HTTPServer#start: pid=4326 port=8443
</pre>

</pre>

lib/proxy/log.rb
```diff
@@ -20,6 +20,46 @@ def logger
...
+      end
+
+      def format_datetime(time)
+        time.strftime("%Y-%m-%dT%H:%M:%S.%6N ".freeze)
+      end
+
+      def msg2str(msg)
```
domcleal: Should this inherit Logger::Formatter to avoid copying this implementation?

modules/root/root_api.rb
```diff
@@ -14,6 +14,7 @@ class Proxy::RootApi < Sinatra::Base
 
   get "/version" do
     begin
+
```
domcleal: Not needed?

lib/proxy/request_id_middleware.rb
```diff
@@ -0,0 +1,14 @@
+module Proxy
+  class RequestIdMiddleware
+    def initialize(app)
+      @app = app
+    end
+
+    def call(env)
+      Thread.current.thread_variable_set(:request_id, env['HTTP_X_REQUEST_ID']) if env.has_key?('HTTP_X_REQUEST_ID')
+      to_ret = @app.call(env)
+      Thread.current.thread_variable_set(:request_id, nil)
```
domcleal: This seems to break logging of the request ID in a default WEBrick configuration, I can't see the request ID at all in the logs.

It appears from adding debug that the thread var's being cleared by this line _before_ the log formatter's called, so the thread variable is always nil. Commenting it out fixes the logging, but will obviously lead to data leaking between requests.

domcleal (changes requested):

domcleal:
> for standard logger I replaced process id with request_id. I can restore the process id if it's important. Should I surround request id with <>'s (to help with parsing)?

I don't think so, it seems fine to me as-is.

lib/proxy/log.rb
dmitri-d: No point, as the methods that I need access to are private (grrr).

lib/proxy/request_id_middleware.rb
dmitri-d: This isn't going to work for logging outside of rack app: the middleware is injected after webrick passes control to it, but before the controller code is invoked. There isn't much I can about this short of changing webrick, which I'd like to avoid.

What I could do however is to log the request (probably via additional middleware) and disable debug logging of requests in webrick/rack?

lib/proxy/log.rb
domcleal: A subclass should be able to call the private method (without needing `send`).

dmitri-d: You are indeed correct.

lib/proxy/request_id_middleware.rb
domcleal: Yes, I think I see the issue now, that the Rack::CommonLogger defers logging until the very last minute by using a BodyProxy class. The CommonLogger middleware passes back a BodyProxy class wrapping the original `body` with a block that logs when the body is closed. By this point, this RequestIdMiddleware has long since completed.

While I'm not completely adverse to reimplementing request logging, it might be simpler to either:
1. Use Rack::BodyProxy here too - pass a BodyProxy with block back out of `call` that will wipe the thread variable, which should then be after CommonLogger.
2. Subclass Rack::CommonLogger, override `log` to wipe the thread variable.

dmitri-d: Thanks for looking into this -- I didn't think to look into BodyProxy, but it explains why it's always one of the last thing to get logged. I also think using BodyProxy to clear the thread variable is probably the best approach.

dmitri-d:
- LoggerFactory::LogFormatter now inherits from ::Logger::Formatter, which eliminated most of the code duplication
- request_id thread variable is now cleared when request is closed (via Rack::BodyProxy)

Note: I should point out that only the first 6 digits of the request id are logged, the whole thing (64 digits) would take too much space.

lib/proxy/log.rb
```diff
@@ -20,6 +20,30 @@ def logger
   end
 
   class LoggerFactory
+    class LogFormatter < ::Logger::Formatter
+      Format = "%s, [%s%s] %5s -- %s: %s\n".freeze
+
+      def call(severity, time, progname, msg)
+        Format % [severity[0..0], format_datetime(time), request_id[0..5], severity, progname, msg2str(msg)]
```
domcleal: Could you use eight characters of the request ID? Mostly to match what Foreman currently logs - if the same ID is used then it'll be identical in both log files.

domcleal (changes requested):

dmitri-d:
Eight characters of the request id are now being logged. Rebased.

domcleal (approved):

domcleal (changes requested):
Still using the first six rather than eight characters?

dmitri-d:
Actually pushed the change now.

domcleal (approved):

domcleal:
Thanks, that's better

There's a comment on the ticket stating:

> Make sure this gets recorded in structured format in logger buffer as well, so we can easily pull out logs (e.g. errors) via the API and show them.

Would you be able to add this?

dmitri-d:
> Make sure this gets recorded in structured format in logger buffer as well, so we can easily pull out logs (e.g. errors) via the API and show them.

Request_ids are now being stored when available in the log buffer.

domcleal:
Thank you @witlessbird, merged.

PR #476, opened 2016-11-17
Fixes #17387 - SubnetService#find_subnet is O(log n) now.

dmitri-d:

PR #477, opened 2016-11-22
fixes #17387 - SubnetService#find_subnet has constant time lookup

domcleal:
find_subnet is now approximately constant with the number of subnets
configured, using hash lookups of possible network prefixes for the
given IP address until the most specific prefix is found. Benchmark
results:

    find_subnet (1 subnets, 200 hosts)
                              1.665k (± 9.5%) i/s -     16.437k in   9.991955s
    find_subnet (5 subnets, 1000 hosts)
                            331.898  (± 7.5%) i/s -      3.298k in   9.999109s
    find_subnet (50 subnets, 10000 hosts)
                             31.363  (± 6.4%) i/s -    313.000  in  10.005986s
    find_subnet (500 subnets, 100000 hosts)
                              3.078  (± 0.0%) i/s -     31.000  in  10.072384s
    find_subnet (5000 subnets, 1000000 hosts)
                              0.301  (± 0.0%) i/s -      4.000  in  13.298996s

add_subnet only checks for an identical network prefix instead of
overlapping prefixes with #find_subnet, speeding it up considerably.
Benchmark results:

        add_subnet (1)     41.389k (±17.4%) i/s -    378.638k in   9.792687s
        add_subnet (5)     10.278k (±11.2%) i/s -     99.588k in   9.931308s
       add_subnet (50)      1.062k (± 9.3%) i/s -     10.470k in   9.991114s
      add_subnet (500)    105.161  (± 9.5%) i/s -      1.042k in  10.007969s
     add_subnet (1000)     53.826  (± 7.4%) i/s -    536.000  in  10.012055s
    add_subnet (15000)      3.424  (± 0.0%) i/s -     35.000  in  10.241879s

Replaces #476.

### DIFF
--- a/bundler.d/development.rb
+++ b/bundler.d/development.rb
@@ -6,4 +6,6 @@ group :development do
   gem 'single_test'
   gem 'pry'
   gem 'rubocop', '0.38.0' if  RUBY_VERSION > "1.9.2"
+  gem 'benchmark-ips'
+  gem 'ruby-prof'
 end
--- a/modules/dhcp_common/subnet_service.rb
+++ b/modules/dhcp_common/subnet_service.rb
@@ -2,9 +2,11 @@ module Proxy::DHCP
   class SubnetService
     include Proxy::Log
 
+    SEARCH_MASKS = (0..31).map { |bit| ~(1 << bit) }
+
     attr_reader :m, :subnets, :leases_by_ip, :leases_by_mac, :reservations_by_ip, :reservations_by_mac, :reservations_by_name
 
-    def initialize(subnets, leases_by_ip, leases_by_mac, reservations_by_ip, reservations_by_mac, reservations_by_name)
+    def initialize(leases_by_ip, leases_by_mac, reservations_by_ip, reservations_by_mac, reservations_by_name, subnets = {})
       @subnets = subnets
       @leases_by_ip = leases_by_ip
       @leases_by_mac = leases_by_mac
@@ -15,15 +17,17 @@ module Proxy::DHCP
     end
 
     def self.initialized_instance
-      new(::Proxy::MemoryStore.new, ::Proxy::MemoryStore.new, ::Proxy::MemoryStore.new,
+      new(::Proxy::MemoryStore.new, ::Proxy::MemoryStore.new,
           ::Proxy::MemoryStore.new, ::Proxy::MemoryStore.new, ::Proxy::MemoryStore.new)
     end
 
     def add_subnet(subnet)
       m.synchronize do
-        raise Proxy::DHCP::Error, "Unable to add subnet #{subnet}" if find_subnet(subnet.network)
+        key = subnet.ipaddr.to_i
+        raise Proxy::DHCP::Error, "Unable to add subnet #{subnet}" if subnets.key?(key)
         logger.debug("Added a subnet: #{subnet.network}")
-        subnets[subnet.network] = subnet
+        subnets[key] = subnet
+        subnet
       end
     end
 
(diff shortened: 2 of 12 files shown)


bundler.d/development.rb
```diff
@@ -6,4 +6,5 @@
   gem 'single_test'
   gem 'pry'
   gem 'rubocop', '0.38.0' if  RUBY_VERSION > "1.9.2"
+  gem 'benchmark-ips'
```
dmitri-d, line 9: Could you keep the profiler gem too?

modules/dhcp_common/subnet_service.rb
```diff
@@ -2,9 +2,11 @@ module Proxy::DHCP
   class SubnetService
     include Proxy::Log
 
+    SEARCH_MASKS = (0..31).map { |bit| ~(1 << bit) }
```
dmitri-d, line 5: Good call :)

test/dhcp/subnet_service_test.rb
```diff
@@ -49,6 +49,59 @@ def test_find_subnet
...
+  def test_find_subnet_with_cidr_mask
+    subnets = [Proxy::DHCP::Subnet.new("192.168.0.0", "255.255.255.192"),
+               Proxy::DHCP::Subnet.new("192.168.0.64", "255.255.255.192"),
+               Proxy::DHCP::Subnet.new("192.168.0.128", "255.255.255.192"),
+               Proxy::DHCP::Subnet.new("192.168.0.192", "255.255.255.192")]
+    @service.add_subnets(*subnets)
+    assert_equal subnets[2], @service.find_subnet("192.168.0.131")
+  end
+
+  def test_find_subnet_with_cidr_mask_and_match_in_a_group_of_one_network
```
dmitri-d: This test is probably not needed -- I originally added it to test binary-search for when a partition contained a single network only.

test/dhcp/subnet_service_test.rb
```diff
@@ -49,6 +49,59 @@ def test_find_subnet
...
+  def test_find_subnet_with_cidr_mask_if_subnet_is_undefined
+    subnets = [Proxy::DHCP::Subnet.new("192.168.0.0", "255.255.255.192"),
+               Proxy::DHCP::Subnet.new("192.168.0.64", "255.255.255.192"),
+               Proxy::DHCP::Subnet.new("192.168.0.192", "255.255.255.192")]
+    @service.add_subnets(*subnets)
+    assert_nil @service.find_subnet("192.168.0.131")
+  end
+
+  # the address we are looking for is in a /27 subnet
+  def test_find_subnet_with_mixed_cidr_masks_deep
```
dmitri-d: These two names aren't great (esp. seeing that you added comments here). Perhaps something like ```test_find_subnet_with_host_in_255_255_255_224_subnet``` or something like that?

domcleal: Renamed to include the target subnet in the test name.

test/dhcp/subnet_service_test.rb
domcleal: Test removed.

bundler.d/development.rb
domcleal: Added ruby-prof, but I've left the profiling script itself out (of limited use I think.)

domcleal:
Updated the PR to address the test and ruby-prof comments.

modules/dhcp_common/subnet_service.rb
domcleal: It did make a noticeable difference, even though it's a trivial operation. Every little calculation does seem to matter (I also removed a slightly superfluous `& 0xff` from ipv4_to_i too, it saved a small amount of work.)

dmitri-d: Nods. Even accessing class var vs. local var (heap vs. stack, probably) makes a difference.

dmitri-d:
Merged as c16b176. Thanks a bunch, @domcleal!

PR #478, opened 2016-11-24
Fixes #2482: Global search in DHCP reservations/leases using MAC and/…

korekhov:
…or hostname

### DIFF
--- a/modules/dhcp/dhcp_api.rb
+++ b/modules/dhcp/dhcp_api.rb
@@ -62,6 +62,18 @@ class Proxy::DhcpApi < ::Sinatra::Base
     end
   end
 
+  # global search fo MAC/IP/hostname across all networks
+  get "/search/:record" do
+    begin
+      content_type :json
+      records = @server.global_search(params[:record])
+      log_halt 404, "Record #{params[:record]} not found" unless records
+      records.to_json
+    rescue => e
+      log_halt 400, e
+    end
+  end
+
   get "/:network/:record" do
     begin
       content_type :json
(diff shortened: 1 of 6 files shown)


dmitri-d:
I did some isc dhcpd source code spelunking today and found out that:
 - one hw address maps to multiple reservations
 - one name (*not* dhcp option 12) maps to one reservation
 - one hw address maps to one lease (effectively; client will get a lease with the highest preference if multiple leases are active)
 - one ip maps to one lease

Taking into account current dhcp api, we can simplify subnet_service by dropping subnet key from leases_by_mac and reservations_by_mac maps (but keeping in mind that these will be one-to-many mappings now). This change will mean that you only need to add all_reservations_by_ip and all_leases_by_ip mappings (good, as maintaining a hash isn't free, especially so for a large number of keys).

I should be able to make this change within next couple of weeks.

Thoughts on the new api endpoint. I'm not a fan a "search" resource at all. I think "records" resource with filters would work much better (also would adhere to rest, for w/e its worth):
  - explicit filters would mean no need to guess what criteria user is searching on (also faster), something like:
```
GET /dhcp/records?mac=00:01:02:03:04:05
```
  - probably needs additional filters for record types (leases vs. reservations)
```
GET /dhcp/records?type=reservation
```
those two can be used together:
```
GET /dhcp/records?mac=00:01:02:03:04:05&type=reservation
```

Lastly, this is seach-like operation, it should return an empty set when no records were found.

korekhov:
Yes, I do like "records" proposal. However, I'm not developer/programmers enough to get that implemented :(

Feel free to take this issue over.

korekhov:
Added the following response in case of record not found by search:
{
  "record": "10.10.10.254",
  "result": "not found"
}

PR #479, opened 2016-12-01
Fixes #17554 - inotify events can't overflow the observer anymore.

dmitri-d:


domcleal:
One test appears to be failing, possibly intermittently as it was only shown on Ruby 2.0.0:

<pre>
unsatisfied expectations:
- expected exactly once, not yet invoked: #&lt;StateChangesObserverTest::ObserverForTesting:0x64f7968>.do_leases_modified(any_parameters)
 (Test::Unit::AssertionFailedError)
/var/lib/workspace/workspace/test_proxy_develop_pr_core/puppet/3.4.0/ruby/2.0.0/test/dhcp_isc/state_changes_observer_test.rb:191:in `test_event_loop_with_modified_event'    
</pre>

Is the idea of this change to coalesce the inotify events in the observer, decoupling it from the code processing the events and parsing?

There are comments today on the ticket that the user's _replacing_ the entire file instead of modifying (appending) to it like ISC DHCP does, which would explain why it's performing so badly. Do you want to address that separately (if possible) and only improve the inotify queue here?

dmitri-d:
Hmm. For some reason I stopped receiving notifications from redmine.

> Is the idea of this change to coalesce the inotify events in the observer, decoupling it from the code processing the events and parsing?

Nods, pretty much. The main goal was to limit the rate of incoming messages: sequential updates can be treated as one; re-create event takes precedence of previous modify events.

> Do you want to address that separately (if possible) and only improve the inotify queue here?

Let me think about this. I don't think it's possible to do what Konstantin is asking as it will break our interface with dhcpd, but perhaps there's something else that can be done.

dmitri-d:
- Updated event loop tests to use test's thread (instead of executing event loop in a worker thread).
- events are constants now.

dmitri-d:
I have a helper script to generate a large volume of lease file updates, not sure if it's worth keeping it (I couldn't trigger inotify queue overflow with it though.)

test/bmc/bmc_api_test.rb
```diff
@@ -384,7 +384,7 @@ def test_api_can_pass_wrong_data_type_and_get_415_error
     Proxy::BMC::IPMI.any_instance.stubs(:bootbios).returns(true)
     put "/#{@host}/chassis/config/bootdevice/bios", "".to_json, "CONTENT_TYPE" => "application/json"
     assert_equal 415, last_response.status
-    assert_equal 'Invalid JSON content in body of request: data must be a hash, not String', last_response.body
+    assert last_response.body.start_with?('Invalid JSON content in body of request')
```
domcleal, line 387: Why change this?

modules/dhcp_isc/isc_state_changes_observer.rb
```diff
@@ -3,27 +3,128 @@ module ISC
...
+              @last_event = RECREATED
+            else
+              return false
+            end
+          end
+          return true
+        end
+
+        def pop
+          @m.synchronize { to_return = @last_event; @last_event = NONE; to_return}
```
domcleal: Split this into separate lines please.

test/dhcp_isc/state_changes_observer_test.rb
```diff
@@ -94,14 +184,42 @@ def test_update_subnet_service_with_dhcp_records_should_delete_leases_with_dupli
     assert @service.find_lease_by_mac("192.168.0.0", "00:11:22:33:44:66")
   end
 
+  def test_event_loop_with_started_event
+    @observer.kill_event_loop
```
domcleal: Should this be in the teardown instead of observer.worker.exit?

domcleal (changes requested):

test/bmc/bmc_api_test.rb
dmitri-d: JSON gem appears to return slightly different error messages depending on the version which was causing test failure on my machine.

test/dhcp_isc/state_changes_observer_test.rb
dmitri-d: We could, although there might be a slight slowdown associated with waiting for the worker thread to exit. I'll make the change to keep things easier to understand.

dmitri-d:
- broke up synchronized block in Proxy::DHCP::ISC::IscStateChangesObserver::Events#pop into multiple lines
- state_changes_observer_test now uses ObserverForTesting#kill_event_loop in teardown

test/dhcp_isc/state_changes_observer_test.rb
domcleal: Should these lines be removed now?

dmitri-d: I need the event loop stopped for these tests, otherwise it consumes test events before tests can get to them.

modules/dhcp_isc/isc_state_changes_observer.rb
```diff
@@ -3,27 +3,132 @@ module ISC
...
         @service = subnet_service
+        @events = events
+        @event_loop_active = true
+        @worker = Thread.new { event_loop }
       end
 
-      def monitor_started
+      def start
+        events.started
+        worker.run
```
domcleal: Should this call `do_start` rather than queuing a STARTED event and using the worker? Currently the module is starting up immediately and then parsing the file in the background, while it used to be synchronous during module startup.

modules/dhcp_isc/inotify_leases_file_observer.rb
```diff
@@ -32,19 +32,20 @@ def monitor_leases
...
       logger.warn "Queue overflow occured when monitoring #{leases_filename}, restarting monitoring", e
       observer.leases_recreated
+      sleep 60
       retry
     rescue Exception => e
       logger.error "Error occured when monitoring #{leases_filename}", e
     end
 
     def start
-      observer.monitor_started
```
domcleal, line 41: Also update the kqueue impl.

domcleal (changes requested):
Seems to work well otherwise.

dmitri-d:
- isc_state_changes_observer executes #start method synchronously now
- updated kqueue implementation.

domcleal (approved):

domcleal:
Merged, thanks @witlessbird.

PR #480, opened 2016-12-08
fixes #4917 - active directory realmprovider with radcli

martencassel:
- Connect using a keytab.
- Create host: 
Domain join a new server and set a new random password.
- Rebuild 
Reset a new random password.
- Delete
Delete the computer account.

The ad realm provider is using the radcli library.

* adcli - https://www.freedesktop.org/software/realmd/adcli/adcli.html
* radcli - https://github.com/martencassel/radcli


dmitri-d:
I opened a PR that introduces realm module providers: https://github.com/theforeman/smart-proxy/pull/481

dmitri-d:
Thanks for the PR! My apologies about the timing of the PR above -- you'll need to make additional changes to get your code working with the changes I'm proposing. 

I briefly looked at the code in your PR, but didn't see any tests. Could you add some to cover the functionality of your provider?

martencassel:
Yes sure. I will add tests ASAP.

helge000:
Thanks @martencasselm, the AD provider is of interest to me as well. In think adcli is a valid and more simple approach then using WinRM; mainly because of the IMHO quite complicated setup of secured remote PS calls.

> Would it be a good idea to implement a native ruby extension that uses the above library for this secnario ? The extension will be used by the ad provider. I'm not aware of any existing libraries.

Me neither (apart from WinRM; but that does not count). And I would answer with a big **no**. Since we rely on many external commands for windows providers (`dncmd`, `netsh`) we should definaly not create (and maintain) a new library.

My main reason for choosing WinRM in #378 where:
- The goal to use Powershell directly and replace those awkward to parse commands
- Legacy commands might go away in the future
- Elminination to run SmartProxy on Windows (aslo not strait forward)

To test this: I never grasped what needs to changed in Foreman to make this work @witlessbird? I only did direct api calls to smart proxy for WinRM.

dmitri-d:
> Since we rely on many external commands for windows providers (dncmd, netsh) we should definaly not create (and maintain) a new library.

Just dnscmd now, ms dhcp provider uses ruby bindings for native dhcps api staring with 1.14. 

Personally, my preference would be to not rely on shell commands for intergration: it's easy to do, but error handling is subpar, testing is harder, and causes support issues. I mentioned this to @martencassel already, but I think interfacing with AD via ldap might be a good alternative?

helge000:
> ms dhcp provider uses ruby bindings for native dhcps api staring with 1.14.

Great!

>Personally, my preference would be to not rely on shell commands for intergration

If you got another impression from my comment, it was definitely not intentional. For the same reasons you named and the reasons I listed above in favor for WinRM.

However, using a pure ldap method for host creation might not be so simple: 
- I am not sure wherever or not other parts happen apart from LDAP on account creation
- AFAIK not a supported method from MS for host creation and therefore a no go for our prod env (at when AD has Exchange schema extension, as in our case)

While I think it is defiantly doable, I fear the hours spend digging through MSDN/MSTN could be spend much more productive for smart proxy (and we would need to maintain it).

dmitri-d:
> However, using a pure ldap method for host creation might not be so simple...

I'll take a closer look at adcli sources. At a quick glance it looked like a wrapper around ldap calls, but that may very well not be the case.

martencassel:
After playing around directly with krb5 and ldap in c i realized why not use the source code from adcli...

I started hacking on a ruby extension that wraps the adcli "library". The static library exists in the source code repo of adcli but is not distributed standalone. I think it supports most of what we want, setting a random password, delete accounts.

And the library inside adcli is rather neat and easy to use.

See more at https://github.com/martencassel/radcli 

What do you think ?

dmitri-d:
> What do you think ?

I think that's a pretty good way to go about this!

helge000:
Good idea to reuse adcli! I've had my issues with it as I am unable to build the gem. Opened issues over at radcli repo

martencassel:
The ruby adcli extension "radcli" that will support the realm ad provider is here https://github.com/martencassel/radcli.

dmitri-d:
FYI: https://github.com/theforeman/smart-proxy/pull/481 has been merged into develop branch.

helge000:
> FYI: #481 has been merged into develop branch.

Thanks for the heads up, @witlessbird. @martencassel did a great job packaging adcli in a gem; I can confirm it works with tests I've done.

I think I'll be able to warp this up, thanks to the work and tests added in #481 

However, I/we am/are struggling with adding unit tests for it; I think it would involve a lot of mocking LDAP calls. I am not that deep into LDAP to do any good there...

How you think we should proceed?

dmitri-d:
> However, I/we am/are struggling with adding unit tests for it; I think it would involve a lot of mocking LDAP calls. I am not that deep into LDAP to do any good there...
> 
> How you think we should proceed?

It's a bit hard to say without seeing the latest code. I could suggest creating a local abstraction for all ldap-related calls and mock those calls instead. Tests would probably look quite similar to ms_native dhcp provider (https://github.com/theforeman/smart-proxy/blob/develop/test/dhcp_ms_native/server_ms_test.rb), in particular that suite's handling of dhcpsapi implementation.

martencassel:
Hi,
Currently working on completing radcli library tests and error handling.

modules/realm/activedirectory.rb
```diff
@@ -0,0 +1,207 @@
...
+    digits = '0123456789'
+    #punctuation = '!"#$%&\'()*+,-./:;<=>?@[\\]^_`{|}~'
+    #characters = (digits + ascii_letters + punctuation + ' ')
+    characters = (digits + ascii_letters + ' ')
+    if pwd_len.nil?
+      pwd_len=22
+    end
+    upper_bound = characters.length - 1
+    rndpwd = ''
+    r = Random.new
```
dekimsey, line 186: If I understand this section correctly, this whole section be simplified to:

```ruby
characters = (digits + ascii_letters).split('')
1.step(pwd_len) do
  rndpwd += characters.sample()
end
return rndpwd
```

martencassel:
Hi
The realm ad provider will be implemented as a smart-proxy plugin project at:

https://github.com/martencassel/smart_proxy_realm_ad_plugin

ekohl:
Does that mean we can close this PR?

martencassel:
Hi closing this.
https://github.com/martencassel/smart_proxy_realm_ad_plugin
// Marten

sandersr:
Any idea when the plugin will be ready?

martencassel:
No. working on it... :-)

martencassel:
Hi, @sandersr  @helge000  @witlessbird 
Made lots of progress this week!!
Have to do lots of testing against a active directory domain and integrate it with foreman.
https://github.com/martencassel/smart_proxy_realm_ad_plugin

/Regards
Marten

sandersr:
That's great news. Thank you very much for investing your time to make it work!

PR #482, opened 2016-12-20
Fixes #14394 - Undisclose WEBrick server version

shlomizadok:
Replaces #402, which seems abandoned.


dmitri-d:
I think in the discussion in the original PR it was agreed to use an empty string instead of 'WEBrick' as 'ServerSoftware' value. Could you do the same here please?

shlomizadok:
@witlessbird - yup... done :+1:

dmitri-d:
Merged as 97dd525. Thanks, @shlomizadok!

PR #484, opened 2016-12-26
fixes #17823 -  MS DHCP record deletion is returning the stdout

alex131290:
The stdout which is being returned by the server, is causing to parsing
issues by Foreman and thus the host is not being deleted at the first
attempt while the DHCP record is already gone

### DIFF
--- a/modules/dhcp_native_ms/dhcp_native_ms_main.rb
+++ b/modules/dhcp_native_ms/dhcp_native_ms_main.rb
@@ -20,8 +20,8 @@ module Proxy::DHCP::NativeMS
       mac = record.mac.gsub(/:/,"")
       msg = "Removed DHCP reservation for #{record.name} => #{record.ip} - #{record.mac}"
       cmd = "scope #{subnet.network} delete reservedip #{record.ip} #{mac}"
-
       execute(cmd, msg)
+      nil
     end
 
     def add_record options={}


timogoebel:
@alex131290 : Thanks for your PR. It seems like our smart-proxy team is on christmas holidays. Hang tight for a final review ;-)

[test] - This triggers CI testing.

timogoebel (approved):

timogoebel (dismissed):
Oh, i just saw: Please file your PR agains the "develop" branch and not the "stable" branch. We are cherry-picking fixes from develop to the stable branches.

dmitri-d:
@alex131290, cc: @timogoebel: please keep this PR against 1.13 branch -- the fix doesn't apply to 1.14 and newer releases.

alex131290:
@timogoebel The problem is that the develop branch has a lot of changes in the DHCP module, which are related to 1.14 release, so i can't open the PR against this branch..
I'm not sure that it's a good idea since the fix should go to 1.13 release only, it's a patch (I.E we'll create a new fix on top of this release.

What do you think?

alex131290:
@witlessbird Yep.
I didn't find any documentation regarding this scenario in the ["Contribute docs"](https://theforeman.org/handbook.html#Codingstandards), anyway, this is what i thought in the first place.

Who's going to merge it?

dmitri-d:
> Who's going to merge it?

One of the maintainers will take a look at it when they are back from the break.

alex131290:
@witlessbird Thanks!
So I'm going to leave the PR as is, right?

timogoebel:
> So I'm going to leave the PR as is, right?

Yes, please do. Sorry for the noise and confusion.

dmitri-d:
Merged as 8ccd349. Thanks, @alex131290!

PR #485, opened 2016-12-28
fixes #17856 - support multiple dhcp records for same ip

timogoebel:
This is the smart-proxy implementation for https://github.com/theforeman/foreman/pull/4129.

The code works fine so far in my tests.

Comments very welcome.

### DIFF
--- a/modules/dhcp/dhcp_api.rb
+++ b/modules/dhcp/dhcp_api.rb
@@ -62,15 +62,51 @@ class Proxy::DhcpApi < ::Sinatra::Base
     end
   end
 
+  # Deprecated, returns a single record
   get "/:network/:record" do
+    begin
+      content_type :json
+
+      logger.warn('GET dhcp/:network/:record endpoint has been deprecated and will be removed in future versions. '\
+                  'Please use GET dhcp/:network/mac/:mac_address or GET dhcp/:network/ip/:ip_address instead.')
+
+      load_subnet
+      load_subnet_data
+
+      record = server.find_record(@subnet.network, params[:record])
+      log_halt 404, "No DHCP record for #{params[:network]}/#{params[:record]} found" unless record
+      record.to_json
+    rescue => e
+      log_halt 400, e
+    end
+  end
+
+  # returns an array of records for an ip address
+  get "/:network/ip/:ip_address" do
     begin
       content_type :json
 
       load_subnet
       load_subnet_data
 
-      record = server.find_record(@subnet.network, params[:record])
-      log_halt 404, "DHCP record #{params[:network]}/#{params[:record]} not found" unless record
+      records = server.find_records_by_ip(@subnet.network, params[:ip_address])
+      log_halt 404, "No DHCP records for IP #{params[:network]}/#{params[:ip_address]} found" unless records
+      records.map(&:options).to_json
+    rescue => e
+      log_halt 400, e
+    end
+  end
+
+  # returns a record for a mac address
+  get "/:network/mac/:mac_address" do
+    begin
+      content_type :json
+
+      load_subnet
+      load_subnet_data
+
+      record = server.find_record_by_mac(@subnet.network, params[:mac_address])
+      log_halt 404, "No DHCP record for MAC #{params[:network]}/#{params[:mac_address]} found" unless record
       record.options.to_json
     rescue => e
       log_halt 400, e
@@ -95,17 +131,42 @@ class Proxy::DhcpApi < ::Sinatra::Base
     end
   end
 
-  # delete a record from a network
+  # deprecated, delete a record from a network
   delete "/:network/:record" do
     begin
       load_subnet
       load_subnet_data
 
+      logger.warn('DELETE dhcp/:network/:record endpoint has been deprecated and will be removed in future versions. '\
+                  'Please use DELETE dhcp/:network/mac/:mac_address or DELETE dhcp/:network/ip/:ip_address instead.')
+
       record = server.find_record(@subnet.network, params[:record])
-      log_halt 404, "DHCP record #{params[:network]}/#{params[:record]} not found" unless record
-      server.del_record @subnet, record
-    rescue Proxy::DHCP::InvalidRecord
-      log_halt 404, "DHCP record #{params[:network]}/#{params[:record]} not found"
+      log_halt 404, "No DHCP record for #{params[:network]}/#{params[:record]} found" unless record
+      server.del_record(@subnet, record).to_json
+    rescue Exception => e
+      log_halt 400, e
+    end
+  end
+
+  # deletes all records for an ip address from a network
+  delete "/:network/ip/:ip_address" do
+    begin
+      load_subnet
+      load_subnet_data
+
+      server.del_records_by_ip(@subnet, params[:ip_address]).to_json
+    rescue Exception => e
+      log_halt 400, e
+    end
+  end
+
+  # delete a record for a mac address from a network
+  delete "/:network/mac/:mac_address" do
+    begin
+      load_subnet
+      load_subnet_data
+
+      server.del_record_by_mac(@subnet, params[:mac_address]).to_json
     rescue Exception => e
       log_halt 400, e
     end
--- a/modules/dhcp_common/record/deleted_reservation.rb
+++ b/modules/dhcp_common/record/deleted_reservation.rb
@@ -3,8 +3,6 @@ require 'dhcp_common/record/reservation'
 module Proxy::DHCP
   # represent a deleted DHCP Record
   class DeletedReservation < Reservation
-    attr_reader :name
-
     def initialize options = {}
       @name = options[:name] || options[:hostname] || raise("Must define a name: #{options.inspect}")
       @options = options
--- a/modules/dhcp_common/server.rb
+++ b/modules/dhcp_common/server.rb
@@ -56,10 +56,32 @@ module Proxy::DHCP
     end
 
     def find_record(subnet_address, an_address)
-      service.find_host_by_ip(subnet_address, an_address) ||
-        service.find_host_by_mac(subnet_address, an_address) ||
-        service.find_lease_by_ip(subnet_address, an_address) ||
-        service.find_lease_by_mac(subnet_address, an_address)
+      records_by_ip = find_records_by_ip(subnet_address, an_address)
+      return records_by_ip.first unless records_by_ip.empty?
+      find_record_by_mac(subnet_address, an_address)
+    end
+
+    def find_record_by_mac(subnet_address, mac_address)
+      service.find_host_by_mac(subnet_address, mac_address) ||
+        service.find_lease_by_mac(subnet_address, mac_address)
+    end
+
+    def find_records_by_ip(subnet_address, ip)
+      hosts = service.find_hosts_by_ip(subnet_address, ip)
+      return hosts if hosts
+      lease = service.find_lease_by_ip(subnet_address, ip)
+      return [lease] if lease
+      []
+    end
+
+    def del_records_by_ip(subnet, ip)
+      records = find_records_by_ip(subnet.network, ip)
+      records.map { |record| del_record(subnet, record) }
+    end
+
+    def del_record_by_mac(subnet, mac_address)
+      record = find_record_by_mac(subnet.network, mac_address)
+      del_record(subnet, record)
     end
 
     def unused_ip(subnet, mac_address, from_address, to_address)
@@ -93,6 +115,8 @@ module Proxy::DHCP
     # Delete keys with string names before adding them back with symbol names,
     # otherwise there will be duplicate information.
     def add_record options = {}
+      related_macs = options.delete("related_macs") || []
+      logger.debug "Ignoring duplicates for macs: #{related_macs.inspect}" unless related_macs.empty?
       options = clean_up_add_record_parameters(options)
 
       validate_ip(options[:ip])
@@ -105,12 +129,16 @@ module Proxy::DHCP
       raise(Proxy::DHCP::Error, "DHCP implementation does not support Vendor Options") if vendor_options_included?(options) && !vendor_options_supported?
 
       # try to figure out if we already have this record
-      record = service.find_host_by_ip(subnet.network, options[:ip]) || service.find_host_by_mac(subnet.network, options[:mac])
-      unless record.nil?
+      similar_records = find_similar_records(subnet, options)
+      similar_records.each do |record|
         if Record.compare_options(record.options, options)
           # we already got this record, no need to do anything
           logger.debug "We already got the same DHCP record - skipping"
           raise Proxy::DHCP::AlreadyExists
+        elsif related_macs.include?(record.mac)
+          logger.debug "Record is an allowed duplicate - skipping"
+          logger.debug "request: #{options.inspect}"
+          logger.debug "local:   #{record.options.inspect}"
         else
           logger.warn "Request to create a conflicting DHCP record"
           logger.debug "request: #{options.inspect}"
@@ -120,6 +148,13 @@ module Proxy::DHCP
       end
 
       Proxy::DHCP::Reservation.new(options)
+    end
+
+    def find_similar_records(subnet, options)
+      records = []
+      records << service.find_hosts_by_ip(subnet.network, options[:ip])
+      records << service.find_host_by_mac(subnet.network, options[:mac])
+      records.flatten.compact.uniq { |record| record.mac }
     end
 
     def clean_up_add_record_parameters(in_options)
@@ -137,7 +172,9 @@ module Proxy::DHCP
       to_return.delete("subnet") # Not a valid key; remove it to prevent conflict with :subnet
       subnet = to_return.delete("network")
 
-      to_return.merge!(:hostname => hostname || name, :subnet => subnet, :ip => ip, :mac => mac)
+      to_return.merge!(:hostname => hostname || name,
+                       :name => name || hostname,
+                       :subnet => subnet, :ip => ip, :mac => mac)
     end
 
     def vendor_options_included? options
--- a/modules/dhcp_common/subnet_service.rb
+++ b/modules/dhcp_common/subnet_service.rb
@@ -79,7 +79,8 @@ module Proxy::DHCP
 
     def add_host(subnet_address, record)
       m.synchronize do
-        reservations_by_ip[subnet_address, record.ip] = record
+        records_for_ip = (reservations_by_ip[subnet_address, record.ip] ||= [])
+        records_for_ip << record
         reservations_by_mac[subnet_address, record.mac] = record
         reservations_by_name[record.name] = record
       end
@@ -96,7 +97,8 @@ module Proxy::DHCP
 
     def delete_host(record)
       m.synchronize do
-        reservations_by_ip.delete(record.subnet.network, record.ip)
+        reservations_by_ip[record.subnet.network, record.ip].delete(record)
+        reservations_by_ip.delete(record.subnet.network, record.ip) if reservations_by_ip[record.subnet.network, record.ip].empty?
         reservations_by_mac.delete(record.subnet.network, record.mac)
         reservations_by_name.delete(record.name)
       end
@@ -115,7 +117,7 @@ module Proxy::DHCP
       m.synchronize { leases_by_ip[subnet_address, ip_address] }
     end
 
-    def find_host_by_ip(subnet_address, ip_address)
+    def find_hosts_by_ip(subnet_address, ip_address)
       m.synchronize { reservations_by_ip[subnet_address, ip_address] }
     end
 
--- a/modules/dhcp_isc/dhcp_isc_main.rb
+++ b/modules/dhcp_isc/dhcp_isc_main.rb
@@ -43,9 +43,9 @@ module Proxy::DHCP::ISC
       options = record.options
       # TODO: Extract this block into a generic dhcp options helper
       statements = []
-      statements << "filename = \\\"#{options[:filename]}\\\";"  if options[:filename]
-      statements << bootServer(options[:nextServer])             if options[:nextServer]
-      statements << "option host-name = \\\"#{record.name}\\\";" if record.name
+      statements << "filename = \\\"#{options[:filename]}\\\";" if options[:filename]
+      statements << bootServer(options[:nextServer])            if options[:nextServer]
+      statements << "option host-name = \\\"#{options[:hostname] || record.name}\\\";"
 
       statements += solaris_options_statements(options)
       statements += ztp_options_statements(options)
--- a/modules/dhcp_isc/isc_file_parser.rb
+++ b/modules/dhcp_isc/isc_file_parser.rb
@@ -23,9 +23,9 @@ module Proxy::DHCP::ISC
       # scan for host statements
       conf.scan(/host\s+(\S+\s*\{[^}]+\})/) do |host|
         if match = host[0].match(/^(\S+)\s*\{([^\}]+)/)
-          hostname = match[1]
+          name = match[1]
           body  = match[2]
-          opts = {:hostname => hostname}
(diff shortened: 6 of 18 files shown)


modules/dhcp_common/record/deleted_reservation.rb
```diff
@@ -3,10 +3,9 @@
 module Proxy::DHCP
   # represent a deleted DHCP Record
   class DeletedReservation < Reservation
-    attr_reader :name
-
     def initialize options = {}
       @name = options[:name] || options[:hostname] || raise("Must define a name: #{options.inspect}")
+      @hostname = options[:hostname] || options[:name] || raise("Must define a name: #{options.inspect}")
```
dmitri-d: This is confusing. @hostname isn't being used anywhere in the class. It gives an impression of being subtly different from @name, but if one looks at how these are originally assigned, both @name and @hostname are the same (most of the time). To add to this, ```record.hostname``` call doesn't use @hostname, but will call @options[:hostname] instead.

modules/dhcp_common/record/reservation.rb
```diff
@@ -6,6 +6,7 @@ class Reservation < Record
 
     def initialize options = {}
       @name = options[:name] || options[:hostname] || raise("Must define a name: #{options.inspect}")
+      @hostname = options[:hostname] || options[:name] || raise("Must define a name: #{options.inspect}")
```
dmitri-d: ditto.

modules/dhcp_common/server.rb
```diff
@@ -122,6 +124,13 @@ def add_record options = {}
       Proxy::DHCP::Reservation.new(options)
     end
 
+    def find_similar_record(subnet, options, related_macs)
+      # we ignore ip results if the mac is part of the related macs
+      record = service.find_host_by_ip(subnet.network, options[:ip])
+      record = nil if record && related_macs.include?(record.mac)
```
dmitri-d: This implies a certain order of records in isc dhcpd configuration and/or leases file: Atm subnet service maps one record to one ip address, for isc dhcpd this record will have the mac address of the last record with a given ip address that was parsed. 

This is going to cause support problems. I think a better approach would be to extend subnet service to support multiple records per ip address. Note that ms dhcpd and infoblox providers do not rely on subnet service and will have to be modified as well.

modules/dhcp_isc/dhcp_isc_main.rb
```diff
@@ -43,9 +43,9 @@ def om_add_record(record)
       options = record.options
       # TODO: Extract this block into a generic dhcp options helper
       statements = []
-      statements << "filename = \\\"#{options[:filename]}\\\";"  if options[:filename]
-      statements << bootServer(options[:nextServer])             if options[:nextServer]
-      statements << "option host-name = \\\"#{record.name}\\\";" if record.name
+      statements << "filename = \\\"#{options[:filename]}\\\";"      if options[:filename]
+      statements << bootServer(options[:nextServer])                 if options[:nextServer]
+      statements << "option host-name = \\\"#{record.hostname}\\\";" if record.hostname
```
dmitri-d: I think it would be more clear if the logic of "hostname" option initialization was explicit here: ```options[:hostname] || record.name``` or something like that. It wouldn't require a check either, as name attribute is required during record instantiation.

modules/dhcp_isc/isc_file_parser.rb
```diff
@@ -23,12 +23,13 @@ def parse_config_and_leases_for_records(conf)
...
         if match = host[0].match(/^(\S+)\s*\{([^\}]+)/)
-          hostname = match[1]
+          name = match[1]
           body  = match[2]
-          opts = {:hostname => hostname}
+          opts = {:name => name}
           body.split(";").each do |data|
             opts.merge!(parse_record_options(data))
           end
+          opts[:hostname] = name if opts[:hostname].nil? || opts[:hostname].empty?
```
dmitri-d: You could: ```opts[:hostname] = name``` after line 28; Line 30 will overwrite it if "hostname" option is defined.  This way no conditionals are required.

modules/dhcp_isc/isc_file_parser.rb
```diff
@@ -152,6 +153,8 @@ def parse_record_options text
           options[:nextServer] = ns
         when /^supersede server.filename\s+=\s+"(\S+)"/
           options[:filename] = $1
+        when /^supersede host-name\s+=\s+"(\S+)"/
```
dmitri-d, line 157: We should probably support a "normal" "host-name" option too: atm it is assumed that both record name and "host-name" option are the same.

modules/dhcp_isc/isc_state_changes_observer.rb
```diff
@@ -166,9 +166,10 @@ def update_subnet_service_with_dhcp_records(records)
               if dupe = service.find_host_by_mac(record.subnet_address, record.mac)
                 service.delete_host(dupe)
               end
-              if dupe = service.find_host_by_ip(record.subnet_address, record.ip)
-                service.delete_host(dupe)
-              end
+              #TODO: What is the aim of this?
```
dmitri-d: This bit checks if subnet service already has host records with the same mac or ip address, and removes them. I haven't checked, but I assume this is how isc dhcpd handles such conflicts.

dmitri-d (changes requested):

modules/dhcp_isc/isc_state_changes_observer.rb
timogoebel: @witlessbird : I think, this is a very bad approach. When you have two records with the same ip address, conflict detection breaks because the first record is deleted and the second is never saved so smart-proxy doesn't know anything about either of them.
Is it ok if I remove this completely or am I missing something more?

modules/dhcp_isc/isc_file_parser.rb
timogoebel: Not sure, I follow. Do you mean an option without the "supersede"?

Something like this?
```
host wiki {
  hardware ethernet 00:00:0e:d2:da:be;
  fixed-address 192.168.2.5;
  option host-name "wiki";
}
```

modules/dhcp_isc/isc_state_changes_observer.rb
timogoebel: On second thought, let me check this once again...

modules/dhcp_isc/isc_file_parser.rb
dmitri-d: nods.

modules/dhcp_isc/isc_state_changes_observer.rb
dmitri-d: > the first record is deleted and the second is never saved

The second record takes precedence and is added on line 174.

timogoebel:
@witlessbird : Thanks for the review. You left some very helpful comments. We should chat about two things, though.

1. I don't think, we should delete duplicate records by ip anymore. Imagine this:

```
host bond.example.com-01 {
  dynamic;
  hardware ethernet 44:1e:a1:73:36:10;
  fixed-address 192.168.122.43;
        supersede server.next-server = c0:a8:00:01;
        supersede host-name = "bond.example.com";
}

host bond.example.com-02 {
  dynamic;
  hardware ethernet 44:1e:a1:73:36:11;
  fixed-address 192.168.122.43;
        supersede server.next-server = c0:a8:00:01;
        supersede host-name = "bond.example.com";
}
```

These two have the same ip address, but are totally different records. When the second record gets parsed, the first one would get deleted. We now will not find conflicts for the first record's mac address. Do you agree?

2. When adding a reservation, the code now loops over `find_similar_records` to check if the record already exists or is a duplicate. I changed subnet_service to you suggestion. The code is much better now, I think.
However, I can't follow what you mean by this? Care to elaborate a bit? Did you have something else in mind?
Anyways, I pushed the current state of the code. If you could take another look, that would be wonderful! Thanks.

> Note that ms dhcpd and infoblox providers do not rely on subnet service and will have to be modified as well.

dmitri-d:
> I don't think, we should delete duplicate records by ip anymore

I agree that the assumption of one record per ip isn't valid if we are to support bonded interfaces. This leads to a question of how to handle ```GET /<subnet-ip>/<record-ip>``` requests, however. At the moment this request returns a single a record, returning multiple records breaks the interface. Similarly, how do we handle ```DELETE /<subnet-ip>/<record-ip>``` records?

> Care to elaborate a bit?

I assume you meant my comment re: ms dhcp and infoblox. Those implementations do not rely on subnet service, but make direct calls to their respective backends. These implementations will need to be fixed in order to support one ip per many reservations mapping.

timogoebel:
> This leads to a question of how to handle GET /<subnet-ip>/<record-ip> requests, however. At the moment this request returns a single a record, returning multiple records breaks the interface. Similarly, how do we handle DELETE /<subnet-ip>/<record-ip> records?

The latter case is not used in core to my knowledge. The former is just used for conflict detection (here: https://github.com/theforeman/foreman/blob/develop/lib/net/dhcp/record.rb#L43). To do that properly, it would be best to return a list of records and not just a single record.
Is there a way to update the interface safely?
So basically, we need to decide if `find_record` can only return a single record or not.

> I assume you meant my comment re: ms dhcp and infoblox.

Yes. Sorry. I guess, I should have quoted you above my question.

> re: ms dhcp and infoblox. Those implementations do not rely on subnet service, but make direct calls to their respective backends. These implementations will need to be fixed in order to support one ip per many reservations mapping.

That's what I don't get. I have changed the subnet service (please see the code) and I can't see why this would break the ms dhcp provider. But maybe you have a different implementation in mind. I'd love to know. I am currently assuming, we won't support multiple ip addresses per record only for isc dhcpd.

dmitri-d:
> The latter case is not used in core to my knowledge.

I'm not sure how religiously everyone keeps foreman and smart-proxy versions synchronized. Breaking changes on the proxy side would break older foreman versions. There are also people there who use smart-proxy api in their scripts and such (I think of these cases as standalone proxies), such a change would break their code too.

 > Is there a way to update the interface safely?

Not at the moment. I'm thinking that this might be the time to introduce API versioning. It will have to be per module anyway, so such a change won't affect other modules. 

>  But maybe you have a different implementation in mind. I'd love to know.

I looked at your new changes, and yes, you are correct -- provider api doesn't need to change. Although I think it would be nice (from a developer perspective) to update the internal api to reflect what is expected: so find_record_by_ip should return a list, while find_record_by_mac should return a single record. 

>  I am currently assuming, we won't support multiple ip addresses per record only for isc dhcpd.

I think we should support this when we can, the limiting factor is whether a given backend supports multiple ips per record mapping. This might mean returning meaningful errors from providers when they do not allow for multiple reservations to share an ip, or update them to return multiple records from find_record_by_ip call.

modules/dhcp_common/subnet_service.rb
```diff
@@ -79,7 +79,8 @@ def add_lease(subnet_address, record)
 
     def add_host(subnet_address, record)
       m.synchronize do
-        reservations_by_ip[subnet_address, record.ip] = record
+        reservations_by_ip[subnet_address, record.ip] ||= []
+        reservations_by_ip[subnet_address, record.ip] << record
```
dmitri-d: This bit is performance sensitive. Could you change these to something like: 
```
records_for_ip = (reservations_by_ip[subnet_address, record.ip] ||= [])
records_for_ip << record
```

dmitri-d: Alternatively, you could change MemoryStore constructor to accept a default value for new keys.

modules/dhcp_common/subnet_service.rb
```diff
@@ -115,10 +117,16 @@ def find_lease_by_ip(subnet_address, ip_address)
       m.synchronize { leases_by_ip[subnet_address, ip_address] }
     end
 
-    def find_host_by_ip(subnet_address, ip_address)
+    def find_hosts_by_ip(subnet_address, ip_address)
       m.synchronize { reservations_by_ip[subnet_address, ip_address] }
     end
 
+    def find_host_by_ip(subnet_address, ip_address)
```
dmitri-d: You kept this for compatibility reasons, correct? Only isc dhcpd and libvirt providers use subnet service, I think updating both to use "find_hosts_by_ip" method is a better/safer approach: this changes internal model and I'm paranoid that some code will rely on one record per ip assumption. I'd much rather have an "undefined method" failure than something (potentially) subtle.

timogoebel:
@witlessbird : I just pushed my recent changes. I haven't tested them, yet. But unit tests indicate, that this could work. Tomorrow is a public holiday in Germany so I won't get to it until next week, but could you please take another look? Keen to hear your comments. Thanks.

dmitri-d:
I realized there as an alternative to api versioning: 
 - deprecate current ```get "/:network/:record"``` and ```delete "/:network/:record"``` api
 - introduce ```get /:network/mac/:mac_address```, ```get /:network/ip/:ip_address```, etc. 

Such a change would introduce better api (atm we have a "get" operation with search semantics, even though the caller knows if it uses mac or ip address to retrieve a record) and allow for the breaking change you are implementing. Thoughts?

timogoebel:
@witlessbird : I just tested this PR with the new endpoints and the Foreman PR. It worked very well. An unpatched Foreman instance continued to work as well.
If you could take another look, that'd be great.

modules/dhcp/dhcp_api.rb
```diff
@@ -95,17 +131,52 @@ def load_subnet_data
...
+
+  # deletes all records for an ip address from a network
+  delete "/:network/ip/:ip_address" do
+    begin
+      load_subnet
+      load_subnet_data
+
+      records = server.find_records_by_ip(@subnet.network, params[:ip_address])
+      log_halt 404, "No DHCP records for #{params[:network]}/#{params[:ip_address]} found" unless records
+      records.map { |record| server.del_record @subnet, record }.to_json
```
dmitri-d: I'd push all logic related to record deletion to the server: replace calling find_records_by_ip and then iteratively removing them via del_record with a single call along the lines of ```del_records_by_ip```. Concrete server implementation can deal with the details.

As it stands right now some implementations (infoblox, for example: https://github.com/theforeman/smart_proxy_dhcp_infoblox/blob/master/lib/smart_proxy_dhcp_infoblox/common_crud.rb#L49) end up making an unnecessary call simply to satisfy this controller-model interface.

modules/dhcp/dhcp_api.rb
```diff
@@ -95,17 +131,52 @@ def load_subnet_data
...
+
+  # delete a record for a mac address from a network
+  delete "/:network/mac/:mac_address" do
+    begin
+      load_subnet
+      load_subnet_data
+
+      record = server.find_record_by_mac(@subnet.network, params[:mac_address])
+      log_halt 404, "No DHCP record for #{params[:network]}/#{params[:mac_address]} found" unless record
+      server.del_record(@subnet, record).to_json
```
dmitri-d: ditto.

modules/dhcp/dhcp_api.rb
```diff
@@ -62,15 +62,51 @@ def load_subnet_data
...
+
+  # returns an array of records for an ip address
+  get "/:network/ip/:ip_address" do
+    begin
+      content_type :json
+
+      load_subnet
+      load_subnet_data
+
+      records = server.find_records_by_ip(@subnet.network, params[:ip_address])
```
dmitri-d, line 92: It *might be* useful to add ip address (and mac address below) validation here, as it *could* be helpful in understanding why a given record couldn't be found. Having said that, it's a retrieval operation which has no affect on persisted data -- returning a 404 or an empty list (and relying on the caller figuring out the cause) is ok too.

modules/dhcp_common/server.rb
```diff
@@ -56,10 +56,23 @@ def all_hosts(subnet)
...
+      service.find_host_by_mac(subnet_address, mac_address) ||
+        service.find_lease_by_mac(subnet_address, mac_address)
+    end
+
+    def find_records_by_ip(subnet_address, ip)
+      hosts = service.find_hosts_by_ip(subnet_address, ip)
+      return hosts if hosts
+      lease = service.find_lease_by_ip(subnet_address, ip)
+      return [lease] if lease
+      nil
```
dmitri-d: I think this should return an empty list (my mental map is: function returns a single object: an object or nil; function returns a list: a list of of objects or an empty list).

dmitri-d:
Looks pretty good, other than those minor things.

modules/dhcp/dhcp_api.rb
timogoebel: Yeah, we could issue an 422 if you queried the endpoint with something that doesn't look like an ip. I'd suggest doing this in a different PR, though.

modules/dhcp/dhcp_api.rb
timogoebel: Makes sense. Shall the server raise an exception if no records to delete are found? The API would then catch the exception and issue the 404?

dmitri-d: We are returning a 404 right now if no records were found. Having said that, I'm not sure if there's a point in raising an error if the end state is the same after the call -- record with a given mac/ip won't exist in either case.

timogoebel: Yes, that's why I was asking. I'd suggest removing the 404 completely for the time being.

dmitri-d: Sounds good to me.

timogoebel:
@witlessbird : I have implemented your comments. Thanks again for your invaluable comments! Mind to take yet another look?

modules/dhcp/dhcp_api.rb
```diff
@@ -95,17 +131,48 @@ def load_subnet_data
...
+
+  # deletes all records for an ip address from a network
+  delete "/:network/ip/:ip_address" do
+    begin
+      load_subnet
+      load_subnet_data
+
+      server.del_records_by_ip(@subnet, params[:ip_address]).to_json
+    rescue Proxy::DHCP::InvalidRecord
+      log_halt 404, "No DHCP records for #{params[:network]}/#{params[:ip_address]} found"
```
dmitri-d: I don't think the message and 404 status are appropriate in this case -- the only time this exception is raised when the record is not deletable (AKA lease), and that is probably going away at some point.

I looked at a few providers -- there doesn't seem to be an agreement on how to handle deletion of missing hosts (it wasn't required before these PR, however): ms and infoblox return without an error, isc will probably return with an error, and I'm not sure re: libvirt.

I'm ok with either returning or not returning 404 to the caller, but both cases will require changes to providers which now need to correctly report (or not report) the "missing record" error. If you don't have an opportunity to test required changes for isc and libvirt, I would suggest fixing this in a follow-up PR (which I can pick up).

modules/dhcp/dhcp_api.rb
```diff
@@ -95,17 +131,48 @@ def load_subnet_data
...
+  # delete a record for a mac address from a network
+  delete "/:network/mac/:mac_address" do
+    begin
+      load_subnet
+      load_subnet_data
+
+      server.del_record_by_mac(@subnet, params[:mac_address]).to_json
     rescue Proxy::DHCP::InvalidRecord
-      log_halt 404, "DHCP record #{params[:network]}/#{params[:record]} not found"
+      log_halt 404, "No DHCP record for #{params[:network]}/#{params[:mac_address]} found"
```
dmitri-d: ditto.

modules/dhcp_common/server.rb
```diff
@@ -105,12 +129,16 @@ def add_record options = {}
       raise(Proxy::DHCP::Error, "DHCP implementation does not support Vendor Options") if vendor_options_included?(options) && !vendor_options_supported?
 
       # try to figure out if we already have this record
-      record = service.find_host_by_ip(subnet.network, options[:ip]) || service.find_host_by_mac(subnet.network, options[:mac])
-      unless record.nil?
+      similar_records = find_similar_records(subnet, options)
```
dmitri-d, line 132: I just realized there is a bug here (always have been) -- find_similar_records only checks reservations, while there might be a conflict with a lease. Providers will catch this, but handling the conflict here would allow for a clearer error message. Again, I think a dedicated PR would be ok for this.

modules/dhcp_native_ms/dhcp_native_ms_main.rb
```diff
@@ -106,7 +106,31 @@ def find_record(subnet_address, ip_or_mac_address)
...
+      build_reservation_or_lease(client, subnet_address)
+    rescue DhcpsApi::Error => e
+      return nil if e.error_code == 20_013 # not found
+      raise e
+    end
+
+    def find_records_by_ip(subnet_address, ip_address)
+      client = dhcpsapi.get_client_by_ip_address(ip_address)
+      result = build_reservation_or_lease(client, subnet_address)
+      return unless result
```
dmitri-d: ```return [] unless result```

modules/dhcp_native_ms/dhcp_native_ms_main.rb
```diff
@@ -106,7 +106,31 @@ def find_record(subnet_address, ip_or_mac_address)
...
+      [result]
+    rescue DhcpsApi::Error => e
+      return nil if e.error_code == 20_013 # not found
```
dmitri-d: ```return [] if e.error_code == 20_013```

dmitri-d:
@timogoebel: let me know what you think...

modules/dhcp/dhcp_api.rb
timogoebel: I fully agree. Let's do it in another PR. I'd love to help, but my plate is kind of full right now.
Another idea may be to actually move the `load_subnet` and `load_subnet_data` to the server as well. I noticed, that infoblox doesn't need the subnet to delete a record for example.

modules/dhcp_common/server.rb
timogoebel: I agree. Let's do this in a separate PR.

modules/dhcp_native_ms/dhcp_native_ms_main.rb
timogoebel: Fixed.

modules/dhcp_native_ms/dhcp_native_ms_main.rb
timogoebel: Fixed.

timogoebel:
@witlessbird : Thanks for checking this again. I fixed the two return values I missed. Let's do the rest of this in dedicated PRs. I can definitely help with them. Either with reviewing or by writing a patch. I think, it's a good opportunity to fix those things as all the providers need updates anyways.

modules/dhcp/dhcp_api.rb
dmitri-d: Agree re: load_subnet* methods, they are vestiges of the old design. I do think, however, that we shouldn't raise 404s on Proxy::DHCP::InvalidRecord because of reasons above -- the common exception handler is sufficient.

I think we are in agreement to handle deletion of missing records in a separate PR, returning 404s should be a part of that change.

dmitri-d:
I think that issue with 404s on deletes is the only remaining bit!

timogoebel:
> I think that issue with 404s on deletes is the only remaining bit!

Oh, sorry. I misread your comments. Deleted the lines that were doing the extra excecption handling.

dmitri-d:
Merged in c7aa35f. Thanks, @timogoebel!

dmitri-d:
I created tickets for the issues we discussed in this PR:

- http://projects.theforeman.org/issues/18072 - removal of load_subnet and load_subnet_data methods
- http://projects.theforeman.org/issues/18071 - standardize handling of missing records on delete calls
- http://projects.theforeman.org/issues/18070 - new reservations should be checked for conflicts with leases
- http://projects.theforeman.org/issues/18069 - ip/mac address validation in api controller

timogoebel:
@witlessbird : Thanks a lot. I'll assign the tickets to me if I find the time to start work on them. I'll definitely help with reviewing them. :-)

PR #488, opened 2016-12-29
fixes  #17847 -  MS DHCP: delete the lease upon a host deletion

alex131290:
1) Added a functionality of a lease deletion
2) Handling better errors in the report generation - Instead of looking
for a hard coded string, replaced it with an exit code of the execution
    (A.K.A ERRORLEVEL in Windows)
3) Some cosmetics in the report function
4) Modified the execute method, so it'll get the exit value of the
process

### DIFF
--- a/modules/dhcp_native_ms/dhcp_native_ms_main.rb
+++ b/modules/dhcp_native_ms/dhcp_native_ms_main.rb
@@ -10,19 +10,31 @@ module Proxy::DHCP::NativeMS
     def initialize(name, subnets, subnet_service)
       super(name, subnets, subnet_service)
     end
+    
+    def del_reservation(record_name, record_ip, record_mac, subnet_network)
+      mac = record_mac.gsub(/:/,"")
+      msg = "Remove DHCP reservation for #{record_name} => #{record_ip} - #{record_mac}"
+      cmd = "scope #{subnet_network} delete reservedip #{record_ip} #{mac}"
+      execute(cmd, msg)
+      nil
+    end
+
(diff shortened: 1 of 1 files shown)


dmitri-d:
Thanks for the PR. Unfortunately, bugfixes only can be accepted into 1.13 branch at this point, and this PR implements new functionality which wasn't available in original release.

alex131290:
@witlessbird  Please consider it one more time, we're talking about a bug and not a feature, it's causing issues on MS DHCP.

dmitri-d:
It's classified as a feature in our bug-tracker. I'm sure that in some setups this behaviour is causing issues, it doesn't seem to be a problem for most users. 
Considering that a) In one month time even bug-fixes won't be accepted into 1.13 and b) a newer version of smart-proxy implements this feature there's little point in accepting such a big change.

alex131290:
@witlessbird Actually it's classified as a bug and not as a feature, look @ http://projects.theforeman.org/issues/17847
We need this fix.
In the release (14) you've implemented the DHCP/DNS in a completely different way, so this is a new feature, but we're talking about a patch into the 13 release.
I think it's reasonable, in every software development process, you'll have bug fixes which will go into earlier releases.
Other users, probably didn't notice, since it's not a trivial thing, this bug will happen only when you're provisioning and deleting hosts frequently and not letting them stand for long period of time.

dmitri-d:
You are referring to the issue you filed, the original ticket is http://projects.theforeman.org/issues/15554. 

I understand your frustration, but this is too late in the lifecycle and too big a feature for us to include it into 1.13 branch. This change isn't going to be merged.

PR #489, opened 2016-12-29
Fixes #17879 Add DNS CNAME support for nsupdate

guimaluf:
Smart-proxy API currently only supports DNS CNAME entries for DNSCMD

This commit creates and remove CNAME entries for nsupdate, making it
possible to assign alias to a host using nsupdate providers

### DIFF
--- a/modules/dns_nsupdate/dns_nsupdate_main.rb
+++ b/modules/dns_nsupdate/dns_nsupdate_main.rb
@@ -45,6 +45,17 @@ module Proxy::Dns::Nsupdate
       end
     end
 
+    def create_cname_record(fqdn, target)
+      case cname_record_conflicts(fqdn, target) #return -1, 0, 1
+        when 1
+          raise(Proxy::Dns::Collision, "'#{fqdn}' is already in use by #{target}")
+        when 0
+          return nil
+        else
+          do_create(fqdn, target, "CNAME")
+      end
+    end
+
     def do_create(id, value, type)
       nsupdate_connect
       nsupdate "update add #{id}. #{@ttl} #{type} #{value}"
@@ -67,6 +78,10 @@ module Proxy::Dns::Nsupdate
     def remove_ptr_record(ip)
       get_name!(ip)
       do_remove(ip, "PTR")
+    end
+
+    def remove_cname_record(fqdn_alias)
+      do_remove(fqdn_alias, "CNAME")
     end
 
     def do_remove(id, type)
--- a/test/dns_nsupdate/dns_nsupdate_test.rb
+++ b/test/dns_nsupdate/dns_nsupdate_test.rb
@@ -1,6 +1,5 @@
 require 'test_helper'
-require 'dns/dns'
-require 'dns_nsupdate/dns_nsupdate_plugin'
+require 'dns_common/dns_common'
 require 'dns_nsupdate/dns_nsupdate_main'
 require 'dns_nsupdate/dns_nsupdate_gss_main'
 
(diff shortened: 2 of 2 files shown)


mmoll:
[test]

modules/dns_nsupdate/dns_nsupdate_main.rb
```diff
@@ -69,6 +80,10 @@ def remove_ptr_record(ip)
       do_remove(ip, "PTR")
     end
 
+    def remove_cname_record(fqdn_alias)
```
ekohl, line 83: Bikeshedding: I don't like ```fqdn_alias```. ```fqdn``` is in line with other code but ```name``` might be less ambiguous.

ekohl (approved):
Looks like the correct implementation, but I haven't tested this against a real bind install. I am working on bats tests for the proxy in https://github.com/Katello/forklift/pull/350 and I'll see if it works when I test that against named.

test/dns_nsupdate/dns_nsupdate_test.rb
```diff
@@ -1,8 +1,6 @@
 require 'test_helper'
-require 'dns/dns'
-require 'dns_nsupdate/dns_nsupdate_plugin'
+require 'dns_common/dns_common'
 require 'dns_nsupdate/dns_nsupdate_main'
-require 'dns_nsupdate/dns_nsupdate_gss_main'
```
ekohl, line 5: This line is needed for the GSS tests to pass.

ekohl:
Verified to work

test/dns_nsupdate/dns_nsupdate_test.rb
mmoll: @guimaluf please re-add this line then ☝️

dmitri-d: These are needed for executing just this suite (as opposed to running all available tests). Please roll back this change

modules/dns_nsupdate/dns_nsupdate_main.rb
guimaluf: I agree `fqdn_alias` isn't the best variable name, but `name` is better then anything containing `alias`? I would like something more meaningful then `name`, since `alias` is a reserved keyword

test/dns_nsupdate/dns_nsupdate_test.rb
guimaluf: I'll add these lines(all 3?), but I'd to change this header, cause my local tests did not run at all!

ekohl: I've had the same problem since I don't have kerberos installed. In case you bundle install --without krb5.

I've thought about a rescue block. Then the tests fail, but at least the others will run. The alternative would be to split the gss tests into a separate file so you can test nsupdate without nsupdate_gss.

dmitri-d: I don't see any gss-specific tests in this suite, so this require can be safely dropped.

modules/dns_nsupdate/dns_nsupdate_main.rb
ekohl: https://en.wikipedia.org/wiki/CNAME_record also refers to it as name (and value). https://tools.ietf.org/html/rfc2181 refers to alias name. Since the rest of the code base uses fqdn let's leave it at fqdn_alias for now. At some point I'm going to write some patches to remove some duplication in this code anyway, let's fix it there.

test/dns_nsupdate/dns_nsupdate_test.rb
ekohl: I believe https://github.com/theforeman/smart-proxy/pull/492/commits/cb8112e021421f699183c4d23162ac89dd184367 fixes it properly.

guimaluf: Should I change it or leave the way it is?

ekohl: For now I'd revert your changes so the tests pass again.

dmitri-d:
Thanks for the PR! The PR at https://github.com/theforeman/smart-proxy/pull/492 is a better approach however, as it provides default implementations for create_* and delete_* methods for all providers. Closing this PR in favour of the refactoring in https://github.com/theforeman/smart-proxy/pull/492.

PR #490, opened 2016-12-29
Fixes #17880 - Follow same pattern for DNS CNAME methods signature

guimaluf:
All DNS methods uses fqdn parameters as the target of the action and the    second argument as the attribute that will be set to that target

Since CNAME methods does the opposite, this commit aim to fix that

### DIFF
--- a/modules/dns_common/dns_common.rb
+++ b/modules/dns_common/dns_common.rb
@@ -42,14 +42,14 @@ module Proxy::Dns
       end
     end
 
-    def create_cname_record(fqdn, target)
-      case cname_record_conflicts(fqdn, target) #returns -1, 0, 1
+    def create_cname_record(fqdn, host_alias)
+      case cname_record_conflicts(fqdn, host_alias) #returns -1, 0, 1
         when 1 then
-          raise(Proxy::Dns::Collision, "'#{fqdn} 'is already in use")
+          raise(Proxy::Dns::Collision, "'#{host_alias} 'is already in use")
         when 0 then
           return nil
         else
-          do_create(fqdn, target, "CNAME")
+          do_create(host_alias, fqdn, "CNAME")
       end
     end
 
@@ -76,8 +76,8 @@ module Proxy::Dns
       do_remove(fqdn, "AAAA")
     end
 
-    def remove_cname_record(fqdn)
-      do_remove(fqdn, "CNAME")
+    def remove_cname_record(host_alias)
+      do_remove(host_alias, "CNAME")
     end
 
     def remove_ptr_record(name)
@@ -170,10 +170,10 @@ module Proxy::Dns
       end
     end
 
-    def cname_record_conflicts(fqdn, target)
-      current = resolver.getresources(fqdn, Resolv::DNS::Resource::IN::CNAME)
+    def cname_record_conflicts(fqdn, host_alias)
+      current = resolver.getresources(host_alias, Resolv::DNS::Resource::IN::CNAME)
       return -1 if current.empty?
-      return 0 if current[0].name.to_s == target #There can only be one CNAME
+      return 0 if current[0].name.to_s == fqdn #There can only be one CNAME
       1
     end
 
--- a/test/dns/dns_api_test.rb
+++ b/test/dns/dns_api_test.rb
@@ -13,8 +13,8 @@ class DnsApiTestProvider
   def create_ptr_record(fqdn, ip)
     @fqdn = fqdn; @ip = ip; @type = 'PTR'
   end
-  def create_cname_record(fqdn, target)
-    @fqdn = fqdn; @target = target; @type = 'CNAME'
+  def create_cname_record(fqdn, host_alias)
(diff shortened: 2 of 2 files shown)


mmoll:
[test]

ekohl:
The annoying thing is that plugins can have implemented this feature using this API. Now the number of DNS plugins is limited, but we've already shipped this API in 1.13. I'm still working on implementing it for PowerDNS so I hadn't come across this yet.

This includes https://github.com/theforeman/smart-proxy/pull/489 and I see some other fixes in the imports. Let's first get those merged and then look at this again.

dmitri-d:
@guimaluf: could you rebase this PR and fix the conflicts please?

test/dns/dns_api_test.rb
```diff
@@ -149,10 +149,10 @@ def test_create_aaaa_record_is_shortened
   end
 
   def test_create_cname_record
-    post '/', :fqdn => 'test.com', :value => 'test1.com', :type => 'CNAME'
+    post '/', :fqdn => 'test.com', :value => 'alias.com', :type => 'CNAME'
```
guimaluf, line 152: This is the main important thing about this PR.
It'll change the signature of cname creation API.
I don't know if we should point out deprecation warning or keep a backwords compatibility

modules/dns_common/dns_common.rb
```diff
@@ -170,10 +170,10 @@ def aaaa_record_conflicts(fqdn, ip)
       end
     end
 
-    def cname_record_conflicts(fqdn, target)
-      current = resolver.getresources(fqdn, Resolv::DNS::Resource::IN::CNAME)
+    def cname_record_conflicts(fqdn, host_alias)
+      current = resolver.getresources(host_alias, Resolv::DNS::Resource::IN::CNAME)
```
ekohl, line 174: I don't think this is correct. If I interpret it correctly then ```fqdn``` is the name and ```target``` the value. https://github.com/theforeman/smart-proxy/blob/47732b9ef1375e7bc88b7056f497080477ac2e31/modules/dns/dns_api.rb#L30 passes ```fqdn```, ```value```. Then https://github.com/theforeman/smart-proxy/blob/47732b9ef1375e7bc88b7056f497080477ac2e31/modules/dns_common/dns_common.rb#L45-L46 does the same.

ekohl (changes requested):
I'm not sure this is even a bug but suspect you're interpreting the API incorrectly.

When you do a POST with fqdn=alias.example.com value=host.example.com then it should already work correctly. I believe I've verified this with bind and powerdns on a system level, using dig to verify.

modules/dns_common/dns_common.rb
guimaluf: It works already, but from my point of view this signature is incorrect.

`server.create_cname_record(fqdn, value)`
Actual CNAME implementations `fqdn` refers to host alias, while `value` refers to the true `fqdn`(`target`).

It makes more sense to use `fqdn=host.example.com, value=alias.example.com`, like IP attribution, `fqdn=host.example.com, value=10.0.0.1`.

Why CNAME is the only that would does the opposite?

ekohl: To me the DNS API is not DNS API but a Host to DNS translation API. That makes it inconsistent and hard to say which is 'correct' because it is a custom format.

The way I'd want it is that fqdn refers to what you query the server with and value what the server answers with. Since you query alias.example.com, that should be the value. In short: ```dig +short $type $fqdn``` should result in ```$value```. However, for PTRs this is already reversed. The way Foreman defines it this is that fqdn is the Host object and value is how it should point to the host object.

The current implementation already aligns with A/AAAA records. I'd say that PTR should be aligned with A/AAAA/CNAME rather than align CNAME with PTR. However, this is all breaking the API which we just can't do since this is all in released versions.

I'd propose to internally standardize the naming:
* ```fqdn``` is what is defined in the API and should only be in dns_api.rb
* ```value``` is what is defined in the API and should only be in dns_api.rb
* ```name``` is the DNS name you query for
* ```content``` is the result your DNS server should return

We should leave the current argument orders since this is the plugin API and breaking them is something we should avoid.

guimaluf: I get your point. Thanks for clarifying. 
So, should we standardize this or leave as it is?

dmitri-d: I agree with @ekohl, mostly. As this is internal api, deprecating a call and introducing a new one *is* an option: this way provider developers have about three months to catch up. Having said this, I'd rate good method parameter names higher than consistent parameter order, esp. when we only have a couple of parameters in these methods.

dmitri-d:
As the discussion seems to have stopped, I'm going to assume that we have settled on not making this change. @guimaluf: Please re-open the PR if you'd like to see it merged.

ekohl:
I do intend to submit a PR that makes the parameter names consistent like I suggested above.

ekohl:
Created https://github.com/theforeman/smart-proxy/pull/507 to fix this.

PR #491, opened 2016-12-30
Refs #17037 - Fix deprecation warning in dnscmd

ekohl:
522231bf3e4c918ee821a73acb3f70459ca57791 deprecated the use of ptr_record_conflicts with IPs, but didn't update this implementation.

Note that since I don't have a Windows install I didn't test this in production.

Should this also be backported to 1.14 once merged?


dmitri-d:
Merged as a0e36b3. Thanks, @ekohl!

dmitri-d:
> Should this also be backported to 1.14 once merged?

I think 1.14 can stay as is -- it will continue to work if with a nagging warning.

PR #492, opened 2017-01-03
Fixes #17906 - Move {create,remove}_*_record to dns_common

ekohl:
This includes https://github.com/theforeman/smart-proxy/pull/491 because it makes the diff of the actual patch smaller.

### DIFF
--- a/modules/dns_common/dns_common.rb
+++ b/modules/dns_common/dns_common.rb
@@ -20,6 +20,74 @@ module Proxy::Dns
       Resolv::DNS.new(:nameserver => @server)
     end
 
+    def create_a_record(fqdn, ip)
+      case a_record_conflicts(fqdn, ip) #returns -1, 0, 1
+        when 1 then
+          raise(Proxy::Dns::Collision, "'#{fqdn} 'is already in use")
+        when 0 then
+          return nil
+        else
+          do_create(fqdn, ip, "A")
+      end
+    end
+
+    def create_aaaa_record(fqdn, ip)
+      case aaaa_record_conflicts(fqdn, ip) #returns -1, 0, 1
+        when 1 then
+          raise(Proxy::Dns::Collision, "'#{fqdn} 'is already in use")
+        when 0 then
+          return nil
+        else
+          do_create(fqdn, ip, "AAAA")
+      end
+    end
+
+    def create_cname_record(fqdn, target)
+      case cname_record_conflicts(fqdn, target) #returns -1, 0, 1
+        when 1 then
+          raise(Proxy::Dns::Collision, "'#{fqdn} 'is already in use")
+        when 0 then
+          return nil
+        else
+          do_create(fqdn, target, "CNAME")
+      end
+    end
+
+    def create_ptr_record(fqdn, ptr)
+      case ptr_record_conflicts(fqdn, ptr) #returns -1, 0, 1
+        when 1 then
+          raise(Proxy::Dns::Collision, "'#{ptr}' is already in use")
+        when 0
+          return nil
+        else
+          do_create(ptr, fqdn, "PTR")
+      end
+    end
+
+    def do_create(name, value, type)
+      raise(Proxy::Dns::Error, "Creation of #{type} not implemented")
+    end
+
+    def remove_a_record(fqdn)
+      do_remove(fqdn, "A")
+    end
+
+    def remove_aaaa_record(fqdn)
+      do_remove(fqdn, "AAAA")
+    end
+
+    def remove_cname_record(fqdn)
+      do_remove(fqdn, "CNAME")
+    end
+
+    def remove_ptr_record(name)
+      do_remove(name, "PTR")
+    end
+
+    def do_remove(name, type)
+      raise(Proxy::Dns::Error, "Deletion of #{type} not implemented")
+    end
+
     def dns_find(key)
       logger.warn(%q{Proxy::Dns::Record#dns_find has been deprecated and will be removed in future versions of Smart-Proxy.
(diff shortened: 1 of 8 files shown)


ekohl:
This also fixes #17879, but doesn't add tests like https://github.com/theforeman/smart-proxy/pull/489 so I'd prefer to merge that first and then rebase on top of it.

guimaluf (approved):

test/dns_dnscmd/dnscmd_test.rb
```diff
@@ -81,33 +82,38 @@ def test_create_duplicate_cname_record_fails
   end
 
   def test_remove_address_records_with_longest_zone_match
+    Proxy::Dns::Dnscmd::Record.any_instance.expects(:get_ipv4_address!).with('host.foo.bar.domain.local')
     Proxy::Dns::Dnscmd::Record.any_instance.expects(:enum_records).with('bar.domain.local', 'host.foo.bar.domain.local',  'A').returns(['1.1.1.1','2.2.2.2'])
     Proxy::Dns::Dnscmd::Record.any_instance.expects(:remove_specific_record_from_zone).with('bar.domain.local', 'host.foo.bar.domain.local', '1.1.1.1', 'A').returns(true)
     Proxy::Dns::Dnscmd::Record.any_instance.expects(:remove_specific_record_from_zone).with('bar.domain.local', 'host.foo.bar.domain.local', '2.2.2.2', 'A').returns(true)
     assert_nil @server.remove_a_record('host.foo.bar.domain.local')
```
dmitri-d: I think provider tests can be broken up (which will make them simpler too): the methods that's been pulled into dns_common (create_* and delete_*) can be tested in their own suite. Provider's implementations of do_create and do_delete methods can be tested in their own suites, which will reduce the amount of mocking and make these tests less fragile.

test/dns_dnscmd/dnscmd_test.rb
```diff
@@ -81,33 +82,38 @@ def test_create_duplicate_cname_record_fails
...
   end
 
   def test_remove_a_record
-    Proxy::Dns::Dnscmd::Record.any_instance.expects(:remove_record).with('host.example.com','A').returns(nil)
+    Proxy::Dns::Dnscmd::Record.any_instance.expects(:get_ipv4_address!).with('host.example.com')
+    Proxy::Dns::Dnscmd::Record.any_instance.expects(:do_remove).with('host.example.com','A').returns(nil)
```
dmitri-d: As an illustration for the above, this and the next three tests should be the same for all providers and can be pulled into a dedicated suite.

modules/dns_common/dns_common.rb
```diff
@@ -20,6 +20,77 @@ def resolver
...
+          raise(Proxy::Dns::Collision, "'#{ptr}' is already in use")
+        when 0
+          return nil
+        else
+          do_create(ptr, fqdn, "PTR")
+      end
+    end
+
+    def do_create(name, value, type)
+      raise(Proxy::Dns::Error, "Creation of #{type} not implemented")
```
dmitri-d, line 68: I know this is what we have for do_remove, but is it actually that useful? By default ruby will throw "undefined method" exception, while this will result in a somewhat misleading error message (if you are getting this message do_create isn't implemented at all, not just for a particular record type).

I think having a test that validates that provider implements a required interface might be more helpful (for an example pls. see https://github.com/theforeman/smart-proxy/blob/develop/test/dhcp_isc/dhcp_isc_provider_interface_test.rb)?

ekohl:
Included https://github.com/theforeman/smart-proxy/pull/489 here because it fixes the tests.

modules/dns_common/dns_common.rb
ekohl: My thinking was that existing providers will have some create functions, but maybe not all. This would maintain compatibility with existing plugins. Maybe we could deprecate overriding the functions and start requiring this in the 1.16 window.

test/dns_dnscmd/dnscmd_test.rb
ekohl: Good point.

modules/dns_common/dns_common.rb
dmitri-d: Sorry, I don't think I explained what I had in mind well: I'm arguing that do_create and do_delete "abstract" methods aren't needed: ruby will throw an exception that's pretty clear when they aren't present, and tests would probably be a better safety net.

ekohl: I can see your point, but this way provides a better error message for a plugin which doesn't support CNAMEs. How about do_create and do_remove log a deprecation warning in 1.15 and be removed in 1.16?

ekohl:
Updated. I moved tests around as suggested.

This removes the get_* checks in remove_*_record that nsupdate had. I did that since dnscmd didn't had them either (and most likely plugins won't either) and it can mess with caching. That's why I think it's best to remove it for now. This does mean the API changes a bit since no 404 is returned but a 200, but Foreman handles those in the same way.

ekohl:
I should note that I currently don't have access to a working setup to test this, but after review I will make sure this is tested on bind before it's merged.

guimaluf (approved):

dmitri-d:
This looks good. I also verified that current implementation of dns_dnscmd dns provider doesn't raise an exception when an attempt to delete a non-existent record is made -- your changes to remove_* methods standardize this behaviour.

dmitri-d:
@ekohl: I tested this against bind, seems to be work ok. If you squash and rebase this, I can merge this in today.

ekohl:
@witlessbird done

dmitri-d:
Merged. Thanks, @ekohl!

PR #493, opened 2017-01-13
Fixes #13318 - Revoked certs always have state 'revoked'

dmitri-d:


domcleal:
Merged, thanks @witlessbird.

PR #494, opened 2017-01-16
Fixes #18071 - Deleting missing records doesn't generate errors

dmitri-d:


dmitri-d:
@timogoebel: Turned out that neither isc nor libvirt providers needed changes.

timogoebel (approved):
@witlessbird : Works great with isc-dhcp. Thanks.

PR #495, opened 2017-01-18
Fixes #18115 - Logger is initialized when Syslog isn't available.

dmitri-d:


dmitri-d:
Another change is that smart-proxy exists now if syslog is configured but not available. In such a configuration ::Proxy::SETTINGS.log_file doesn't contain a usable path and default log file path may not be accessible to smart-proxy.

lib/proxy/log.rb
```diff
@@ -32,43 +24,72 @@ def request_id
...
-        logger = ::Logger.new(STDOUT)
+        logger = stdout_logger
       elsif log_file.casecmp('SYSLOG').zero?
-        begin
-          logger = ::Syslog::Logger.new 'foreman-proxy'
-        rescue
-          logger = default_logger(log_file)
-          puts "'SYSLOG' logger is not supported on this platform, using file-based logger instead"
+        unless syslog_available?
+          puts "SYSLOG' logger is not supported on this platform, please use STDOUT or a file-based logger. Exiting..."
```
domcleal: Nitpick, typo introduced with a missing single quote before "SYSLOG".

domcleal (changes requested):

dmitri-d:
Fixed the typo and rebased.

domcleal (approved):

domcleal:
Merged, thanks @witlessbird.

PR #497, opened 2017-01-20
fixes #18177 - file descriptor leak in puppet autosign handling

pikelly:


modules/puppetca/puppetca_main.rb
```diff
@@ -22,7 +22,7 @@ def disable certname
       raise "No such file #{autosign_file}" unless File.exist?(autosign_file)
 
       found = false
-      entries = File.open(autosign_file, File::RDONLY).readlines.collect do |l|
+      entries = File.open(autosign_file, File::RDONLY){|f| f.readlines}.collect do |l|
```
dmitri-d: Why not use ```File.readlines(autosign_file)``` instead?

dmitri-d:
Could you also move https://github.com/theforeman/smart-proxy/blob/47732b9ef1375e7bc88b7056f497080477ac2e31/modules/puppetca/puppetca_main.rb#L37 and https://github.com/theforeman/smart-proxy/blob/47732b9ef1375e7bc88b7056f497080477ac2e31/modules/puppetca/puppetca_main.rb#L54 into "ensure" blocks while you are at it?

modules/puppetca/puppetca_main.rb
pikelly: Because I forgot that it inherits from IO. Dohhhh

mmoll:
what's the status here?

pikelly:
Is this acceptable? I have ensured that the descriptors are always closed by using a block

modules/puppetca/puppetca_main.rb
```diff
@@ -47,11 +46,12 @@ def disable certname
...
-      autosign = File.open(autosign_file, File::RDWR)
-      # Check that we don't have that host already
-      found = autosign.readlines.find { |line| line.chomp == certname }
-      autosign.puts certname unless found
-      autosign.close
+      open(autosign_file, File::RDWR) do |autosign|
+        # Check that we don't have that host already
+        found = false
+        autosign.each_line { |line| found = true if line.chomp == certname }
+        autosign.puts certname if found == false
```
ekohl: isn't ```unless found``` the same?

dmitri-d: ``` autosign.puts certname if autosign.any? {|l| l.chomp == certname}```

dmitri-d: nm, this is wrong. ```unless``` looks better there though.

ekohl: @witlessbird I think you're reversing the logic that should be unless.

pikelly: you are right.
Give me one minute. . . . .

dmitri-d: @ekohl: nods, and what I wrote just plain won't work.

pikelly: you are right.
Give me one minute. . . . .
Done

pikelly:
I have switched from "if found == false" to "unless found"

dmitri-d:
lgtm.

modules/puppetca/puppetca_main.rb
```diff
@@ -31,10 +31,9 @@ def disable certname
         end
       end.uniq.compact
       if found
-        autosign = File.open(autosign_file, File::TRUNC|File::RDWR)
-        autosign.write entries.join("\n")
-        autosign.write "\n"
-        autosign.close
+        open(autosign_file, File::TRUNC|File::RDWR) do |autosign|
+          autosign.write entries.join
```
ekohl, line 35: For my understanding: does this mean there is no more newline at the end of the file?

modules/puppetca/puppetca_main.rb
```diff
@@ -47,11 +46,12 @@ def disable certname
...
 
-      autosign = File.open(autosign_file, File::RDWR)
-      # Check that we don't have that host already
-      found = autosign.readlines.find { |line| line.chomp == certname }
-      autosign.puts certname unless found
-      autosign.close
+      open(autosign_file, File::RDWR) do |autosign|
+        # Check that we don't have that host already
+        found = false
+        autosign.each_line { |line| found = true if line.chomp == certname }
```
ekohl, line 52: Any reason to change the logic btw? The old code had ```found = autosign.readlines.find { |line| line.chomp == certname }```

pikelly: I cannot remember why I chose that new approach. Maybe it was to assure myself that autosign pointer was at EOF. 

It works for me and has been in use since January.

dmitri-d: @ekohl: are you ok with these changes?

ekohl: Personally I think the old code is much nicer and easier to read.

dmitri-d: @ekohl, @pikelly: I'll let you two sort this out.

pikelly: @ekohl: On further thought I believe that you are right and that the original 

found = autosign.readlines.find { |line| line.chomp == certname }

is better than the 

autosign.each_line { |line| found = true if line.chomp == certname }

You are welcome to fix this if you wish. Unfortunately I do not have the time to test this now. When I get some space then I will switch it back.

ekohl: That would be https://github.com/ekohl/smart-proxy/commit/af0a3a4fbbc81f438726406dfedb3cacbf5abce5. I also took the liberty to add a testcase for it.

pikelly: @witlessbird: We have sorted this out. We both agree that the ekohl's push request is fine. Please merge then we can close both requests. Thanks.

dmitri-d: @ekohl: could you open a new PR with the changes you made? I'll close this one then.

ekohl: That should be https://github.com/theforeman/smart-proxy/pull/522

dmitri-d:
Closing in favour of https://github.com/theforeman/smart-proxy/pull/522

PR #498, opened 2017-01-23
Fixes #18070 - add_record() checks for conflicts with leases now.

dmitri-d:


dmitri-d:
@timogoebel: updated add_record to check for conflicts with leases.

dmitri-d:
[test]

dmitri-d:
[test]

modules/dhcp_common/server.rb
```diff
@@ -118,44 +118,42 @@ def inspect
...
+      if similar_records.any? {|record| record == to_return}
+        # we already got this record, no need to do anything
+        logger.debug "We already got the same DHCP record - skipping"
+        raise Proxy::DHCP::AlreadyExists
+      end
+
+      unless similar_records.empty?
+        logger.warn "Request to create a conflicting DHCP record"
+        logger.debug "request: #{to_return.inspect}"
+        logger.debug "existing:   #{similar_records.inspect}"
```
timogoebel: Too many whitespaces between `:` and `#`

modules/dhcp_common/record.rb
```diff
@@ -34,25 +33,19 @@ def [] opt
...
 
-    # compare between incoming request and our existing record
-    # if our record has all requested attributes then we say they are comparable
-    def self.compare_options(record, request)
-      request.each do |k,v|
-        return false if record[k.to_sym] != v
-      end
-      true
+    def ==(other)
+      ip == other.ip && mac == other.mac && subnet == other.subnet && deleteable? == other.deleteable? && options == other.options
```
timogoebel: Should we check if other is really a record before comparison? Same for `Proxy::DHCP::Lease`.

modules/dhcp_common/server.rb
```diff
@@ -118,44 +118,42 @@ def inspect
     def add_record options = {}
       related_macs = options.delete("related_macs") || []
       logger.debug "Ignoring duplicates for macs: #{related_macs.inspect}" unless related_macs.empty?
-      options = clean_up_add_record_parameters(options)
+      name, ip_address, mac_address, subnet_address, options = clean_up_add_record_parameters(options)
 
-      validate_ip(options[:ip])
-      validate_mac(options[:mac])
-      raise(Proxy::DHCP::Error, "Must provide hostname") unless options[:hostname]
```
timogoebel, line 125: Is this safe to remove because we use "name" as a fallback? Should we add a check if name is set?

timogoebel (commented):
@witlessbird : Thanks. I've left some inline comments!

dmitri-d:
- restored name presence check in add_record method
- fixed formatting in https://github.com/theforeman/smart-proxy/pull/498#discussion-diff-97348267R143
- rewrote equality methods to remove duplication

timogoebel (approved):
Looks good and works great, thanks @witlessbird.

PR #499, opened 2017-01-24
Fixes #18209 - Destroy DHCP lease only when its found

dLobatog:
Currently it's possible to find a DHCP record in libvirt, but at
deletion time it might not be on the leases for that IP. Such operation
fails with a 400 error now with

=> #&lt;Libvirt::Error: Call to virNetworkUpdate failed: Requested
operation is not valid: couldn't locate a matching dhcp host entry in
network 'vagrant-libvirt'&gt;

To fix it, I've modified the search method to take into account the
leases list.

### DIFF
--- a/modules/dhcp_libvirt/dhcp_libvirt_main.rb
+++ b/modules/dhcp_libvirt/dhcp_libvirt_main.rb
@@ -92,5 +92,10 @@ module Proxy::DHCP::Libvirt
       logger.error msg = "Error removing DHCP record: #{e}"
       raise Proxy::DHCP::Error, msg
     end
+
+    def find_record(_, record)
+      return unless libvirt_network.leased_ip?(record.ip)
+      super(_, record)
+    end
   end
 end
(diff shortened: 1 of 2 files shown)


modules/dhcp_libvirt/dhcp_libvirt_main.rb
```diff
@@ -92,5 +92,10 @@ def del_record(_, record)
       logger.error msg = "Error removing DHCP record: #{e}"
       raise Proxy::DHCP::Error, msg
     end
+
+    def find_record(_, record)
```
dmitri-d, line 96: Proxy::DHCP::Libvirt::Provider#find_record returns either a Proxy::DHCP::Reservation or Proxy::DHCP::Lease. You can check if it's the latter in Proxy::DHCP::Libvirt::Provider#del_record instead  -- it skips a call to libvirt and is easier to test (could you add a test please too?).

DHCP provider interface has been recently changed (see https://github.com/theforeman/smart-proxy/blob/develop/modules/dhcp/dhcp_api.rb#L152 and https://github.com/theforeman/smart-proxy/blob/develop/modules/dhcp/dhcp_api.rb#L165). Could you update libvirt provider to support these -- it will need to overload del_records_by_ip and del_records_by_mac base-class implementations?

dmitri-d (changes requested):

mmoll:
what's the status here?

dmitri-d:
@dLobatog: ping

lzap:
Thanks for the patch, however libvirt provider was vastly refactored. Please rebase on top of develop and reopen if you still want this to get it through.

PR #500, opened 2017-01-24
Fixes #18210 - Nametag shouldnt be set if record has no name

dLobatog:
To set the nametag we call record.name, which might not be available.
In such a case, record.name fails with

] pry(#<Proxy::DHCP::Libvirt::LibvirtDHCPNetwork>)> record.name
0x005654f2c66798>

and the entire deletion fails. It should not call .name if it's not
there.


dmitri-d:
https://github.com/theforeman/smart-proxy/pull/498 introduces 'name' field for leases. I'd skip this in favour of the other PR.

dLobatog:
Missed that - thanks!

lzap:
Thanks, I was experiencing this lately a lot :-)

PR #501, opened 2017-01-25
Fixes #18026 - proxy can listen on multiple arbitrary interfaces

dmitri-d:


domcleal (changes requested):
Could you also change the example settings.yml to use an array syntax? It'll work for the single value and be clearer to users wanting to specify more interfaces.

dmitri-d:
Updated example config file to use array for "bind_host" setting.

dmitri-d:
rebased.

config/settings.yml.example
```diff
@@ -44,8 +44,8 @@
 #:daemon_pid: /var/run/foreman-proxy/foreman-proxy.pid
 
 # host and ports configuration
-# host to bind ports to (possible values: *, localhost, 0.0.0.0)
-#:bind_host: '*'
+# comma-separated list of interfaces to bind ports to (possible values: *, localhost, 0.0.0.0)
```
domcleal: Nitpick, it might be safest to say "array" rather than "comma-separated" so it isn't misinterpreted as `bind_host: eth1,eth` or something (i.e. comma-separated string).

domcleal (changes requested):

config/settings.yml.example
dmitri-d: Good point.

dmitri-d:
Replaced "a comma-sperated list of interface" with "an array of interfaces" in the comment to "bind_host" parameter.

domcleal (approved):

domcleal:
Merged, thanks @witlessbird.

PR #503, opened 2017-01-29
Fixes #9063 - Revert pinning of Rack to < 1.6

mmoll:


mmoll:
Ruby 1.8 is not supported by -proxy now anyway.

domcleal:
Merged, thanks @mmoll.

dLobatog:
Seems like now it's asking for Ruby 2.2 at least because it tries to get the latest version of Rack.

mmoll:
argh, this needs a `< 2` added, but I don't have time the next days...

mmoll:
or maybe if it works with 2.x, that <1 only for Ruby <2.2

PR #504, opened 2017-02-01
Fixes #18290 - increased puppet classes api timeout

dmitri-d:
But only when puppet server class cache is disabled. Additionally, a warning is logged to notify users about potenitally degraded performance.

### DIFF
--- a/modules/puppet_proxy_puppet_api/v3_environment_classes_api_classes_retriever.rb
+++ b/modules/puppet_proxy_puppet_api/v3_environment_classes_api_classes_retriever.rb
@@ -3,12 +3,12 @@ require 'concurrent'
 class Proxy::PuppetApi::V3EnvironmentClassesApiClassesRetriever
   include ::Proxy::Log
 
-  MAX_CLIENT_TIMEOUT = 15
+  DEFAULT_CLIENT_TIMEOUT = 15
   MAX_PUPPETAPI_TIMEOUT = 300
 
-  attr_reader :puppet_url, :ssl_ca, :ssl_cert, :ssl_key
+  attr_reader :puppet_url, :ssl_ca, :ssl_cert, :ssl_key, :api_timeout
 
-  def initialize(puppet_url, puppet_ssl_ca, puppet_ssl_cert, puppet_ssl_key, api = nil)
+  def initialize(puppet_url, puppet_ssl_ca, puppet_ssl_cert, puppet_ssl_key, api_timeout, api = nil)
     @etag_cache = {}
     @classes_cache = {}
     @futures_cache = {}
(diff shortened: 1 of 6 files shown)


modules/puppet_proxy_puppet_api/v3_environment_classes_api_classes_retriever.rb
```diff
@@ -61,7 +61,11 @@ def legacy_classes_format(response)
 
   def get_classes(environment)
     future = async_get_classes(environment)
-    future.value!(MAX_CLIENT_TIMEOUT)
+    cache_used = @m.synchronize { !!@etag_cache[environment] } #etags are only available when classes cache is enabled
+    client_timeout = cache_used ? DEFAULT_CLIENT_TIMEOUT : 2 * DEFAULT_CLIENT_TIMEOUT
```
domcleal: I think the client timeout if no cache is used should be equal to the API timeout (300 seconds), rather than 30 seconds. This would be closer to the behaviour of the resource_types provider, and ensures the client gets the response. (If no caching's in use, the response will always be lost.)

domcleal (changes requested):
Could you add the timeout to the module configuration, so it can be overridden by users if necessary?

modules/puppet_proxy_puppet_api/v3_environment_classes_api_classes_retriever.rb
dmitri-d: Do you know if there are there any reason(s) for keeping the class cache disabled on puppet server? If none, then I'd keep the timeout at a value which keeps foreman somewhat responsive (300 seconds will be pretty unusable, I think) and perhaps expand the warning to include a suggestion to enable the class cache...

domcleal: With the class cache enabled you need to either restart/reload the server on every code change or call an API, so it changes how environment deployments need to be made. It's also not enabled by default, so I don't think the default behaviour of the smart proxy can be too aggressive in assuming it is.

domcleal: > so I don't think the default behaviour of the smart proxy can be too aggressive in assuming it is.

Addendum... the default behaviour perhaps isn't so important, but it should be possible for the user to configure the smart proxy to work well when caching is disabled. It's very much valid to use Puppet Server without caching. For class imports, both Foreman's rake task and API mean the Foreman UI responsiveness isn't very important, it's normal for them to block.

If the smart proxy's timeout can be increased through a config to work well in very large environments then a lower default seems reasonable. Users with large environments can then choose between reconfiguring the timeout or enabling caching.

dmitri-d:
rebased, also:
 - added configurable client api timeout
 - updated example configuration file

domcleal (approved):

domcleal:
Merged, thanks @witlessbird.

PR #506, opened 2017-02-04
Fixes #18397 - Valid CNAME targets are valid DNS names

ekohl:


ekohl:
I did notice that an IPv4 address is currently considered a valid dns name so that should be considered a bug, but it's at least some validation.

dmitri-d:
Merged. Thanks, @ekohl!

PR #507, opened 2017-02-04
Fixes #18398: Use consistent DNS parameter naming

ekohl:
* fqdn is what is defined in the API and should only be in dns_api.rb
* value is what is defined in the API and should only be in dns_api.rb
* name is the DNS name you query for
* content is the result your DNS server should return

This should avoid confusion with types where the content is also a valid
hostname so fqdn is ambigious.

### DIFF
--- a/modules/dns_common/dns_common.rb
+++ b/modules/dns_common/dns_common.rb
@@ -20,64 +20,64 @@ module Proxy::Dns
       Resolv::DNS.new(:nameserver => @server)
     end
 
-    def create_a_record(fqdn, ip)
-      case a_record_conflicts(fqdn, ip) #returns -1, 0, 1
+    def create_a_record(name, content)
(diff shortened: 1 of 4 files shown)


dmitri-d:
I think that in most places these changes made method signatures more confusing: ```def create_a_record(fqdn, ip)``` is much clearer about its parameter expectations compared to ```def create_a_record(name, content)```. 

Perhaps instead of trying to come up with a common parameter naming convention we should rename parameters in cname-related method to more descriptive ones? For example: ```def create_cname_record(existing_fqdn, new_alias)```?

ekohl:
I agree with it on ```create_a_record``` that it's less readable, but my reasoning is that I want the ```Record``` class to be much closer to actual DNS. I'll first try to get other refactorings (like https://github.com/theforeman/smart-proxy/pull/511) in and then circle back here.

PR #508, opened 2017-02-06
Fixes #18395 - better error message with ms dhcp on win2008.

dmitri-d:


modules/dhcp_native_ms/dhcp_native_ms_main.rb
```diff
@@ -74,6 +74,7 @@ def unused_ip(subnet_address, mac_address, from_address, to_address)
       client = dhcpsapi.get_client_by_mac_address(subnet_address, mac_address) rescue nil
       return client[:client_ip_address] unless client.nil?
 
+      raise NoMethodError.new("DhcpsApi::Server#get_free_ip_address is not available on Windows Server 2008 and earlier versions.") unless dhcpsapi.api_level == DhcpsApi::Server::DHCPS_WIN2012_API
```
domcleal: 1. NoMethodError's not appropriate here, could you use Proxy::DHCP::Error?
1. Perhaps not necessary to change now, but if there are any future API versions then it'd be useful to either _blacklist_ 2008 and "none" API levels (rather than whitelist 2012), or have the API level comparable so it checks for versions < 2012.

bundler.d/windows.rb
```diff
@@ -1,5 +1,5 @@
 group :windows do
   gem 'highline', :platforms => [:mingw, :x64_mingw]
   gem 'win32-service', :platforms => [:mingw, :x64_mingw]
-  gem 'dhcpsapi', '~> 0.0'
+  gem 'dhcpsapi', '>= 0.0.11'
```
domcleal: and add `'< 1'` from the previous version specification

domcleal (changes requested):

modules/dhcp_native_ms/dhcp_native_ms_main.rb
dmitri-d: > NoMethodError's not appropriate here, could you use Proxy::DHCP::Error?

I'd like to be more specific with this error, perhaps even return a 501 instead of 400 for it? I can add something like  Proxy::DHCP::NotImplemented and use it here...

domcleal: Sure, that'd be even better.

dmitri-d:
rebased, and also:
 - added Proxy::DHCP::NotImplemented exception
 - that is raised on when get_free_ip address isn't available
 - and caught in api controller, which then returns 501
 - blacklisted DHCPS_WIN2008_API and DHCPS_NONE dhcps api configurations
 - updated dhcpsapi gem version to include '< 1.0.0'

domcleal (approved):

domcleal:
Merged, thanks @witlessbird.

PR #509, opened 2017-02-09
Fixes #18443: Introduce Webhook realm provider

joshmue:
Generic realm provider notifying other services/realms about the creation or deletion of hosts via HTTP Webhooks. (See redmine ticket, too)

This provider is able to send custumizable HTTP POST-requests ("Webhook") to a HTTP-Server. E.g.

```json
{"operation":"create","hostname":"host.lan","params":{"userclass":"User class",...}}
```

The HTTP-Server should respond with JSON, too. E.g. on creation:
```json
{"randompassword": "foo"}
```

As Webhooks are not a strict standard, the HTTP headers can be extended and the JSON keys in the body are also customizable.

There is opt-in support for a request signing procedure described in https://developer.github.com/webhooks/securing/

### DIFF
--- a/config/settings.d/realm_webhook.yml.example
+++ b/config/settings.d/realm_webhook.yml.example
@@ -1,0 +1,37 @@
+---
+# Host parameters, mandatory
+:host: localhost
+:port: 9090
+:path: /hooks
+
+# Will both default to true
+:use_ssl: false
+:verify_ssl: false
+
+# Additional headers
+:headers:
+  HTTP_X_ACME_SENDER: foreman
+  Content-Type: application/vnd.acme.hostevent+json
+
+# If signing is enabled an header within every request
+# will contain a HMAC hashed version of the request body to enable
+# the Webhook server to verify the origin using the shared secret.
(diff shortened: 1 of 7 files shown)


dmitri-d:
Thanks for your PR! 

Recent changes in smart-proxy's realm module were implemented (in part) to allow to keep realm providers separate from the core code. I would suggest creating a dedicated repository for webhooks realm provider and packaging the code as a gem. This wouldn't require significant changes, mostly things to support gem packaging. You can find a bit more information at http://projects.theforeman.org/projects/foreman/wiki/How_to_Create_a_Smart-Proxy_Plugin#Bundler-Configuration and http://projects.theforeman.org/projects/foreman/wiki/How_to_Install_a_Smart-Proxy_Plugin. 

Additionally, it might be useful to include in README docs links to projects, repositories, or example setups that use webhooks api to provide interface for realm services.

joshmue:
Thank you for the reply,

I will start working on this :+1: 

Closing this PR in favor of the gem approach...

PR #512, opened 2017-02-20
Fixes #18399 - use warning instead when cache init fails

dmitri-d:
With a clearer message. Currently, when such an issue occurs, users aren't clear about what they need to do to correct the error.


ekohl (approved):

domcleal:
Merged, thanks @witlessbird.

PR #514, opened 2017-03-06
Fixes #18799 - Handle empty string nsupdate dns_key

ekohl:


dmitri-d:
This check shouldn't be needed, there's a validator on 'dns_key' setting that is supposed to catch an empty string. I couldn't reproduce the bug, but if there's one it then it is probably the validator that is at fault.

PR #515, opened 2017-03-21
Fixes #18962 - moved isc file parser into dhcp_common

dmitri-d:
These changes are required in order to support https://github.com/theforeman/smart_proxy_dhcp_remote_isc


dmitri-d:
Any chance for a review? I'd like to get these changes in before 1.15 is branched.

dmitri-d:
@domcleal: any chance you could take a look at this before this Monday?

modules/dhcp_common/isc/omapi_provider.rb
```diff
@@ -1,10 +1,9 @@
 require 'time'
-require 'dhcp_isc/isc_common'
+require 'dhcp_common/server'
 
-module Proxy::DHCP::ISC
```
iNecas, line 4: Keeping the `ISC` namespace would read a bit better to me

iNecas:
The changes look reasonable. My only comment is regarding namespacing of the common ICS code, where I would expect the classes to be under 'ICS' submodule

modules/dhcp_common/isc/omapi_provider.rb
dmitri-d: ```::Proxy::DHCP::ISC``` namespace is reserved for the dhcp isc provider and can't be used for shared bits of code.

iNecas: What about `ISCCommon` or something similar?

dmitri-d:
- moved common dhcp isc code into CommonISC module.

iNecas:
I have not run the code but :+1: code-wise + the tests look to cover the behaviour quite well to trust the change. I'm not core proxy contributor though, perhaps @lzap or @timogoebel would be willing into having a look?

iNecas (approved):

timogoebel (approved):
Code looks good to me as well. I gave this a quick spin. Works well.

PR #516, opened 2017-03-28
Fixes #992 - Make NFS a recognized protocol

xprazak2:
Currently, if we use a medium with NFS path during provisioning, wget will fail because it cannot handle nfs.
```
16934 D, [2017-03-21T13:30:35.699595 ] DEBUG -- : Starting task: /usr/bin/wget --timeout=10 --tries=3 --no-check-certificate -nv -c "nfs://nfs-server/nfs-path/nfs-subdir/images/pxeboot/initrd.img" -O "/var/oprazak_tftproot/boot/NFS-7-x86_64-initrd.img"
16935 I, [2017-03-21T13:30:35.699901 ]  INFO -- : 192.168.100.136 - - [21/Mar/2017 13:30:35] "POST /tftp/fetch_boot_file HTTP/1.1" 200 - 0.0013
16937 D, [2017-03-21T13:30:35.708337 ] DEBUG -- : close: 192.168.100.136:58232
16938 E, [2017-03-21T13:30:35.709638 ] ERROR -- : [15621] nfs://nfs-server/nfs-path/nfs-subdir/images/pxeboot/initrd.img: Unsupported scheme ‘nfs’.
```

This allows using NFS in semi-automatic way:

1. create a Medium: path - nfs://nfs-server/nfs-medium
2. create Operating System: name - Custom, version - 7, Architecture - x86_64, add installation medium created in previous step
3. copy images/pxeboot/initrd.img and images/pxeboot/vmlinuz from your NFS-shared medium to /var/lib/tftpboot/boot on your proxy. Make sure the newly copied files have names corresponding to your OS (Custom-7-x86_64-initrd.img and Custom-7-x86_64-vmlinuz for OS created in step 2). Files should be owned by foreman-proxy user and have 544 permissions.
4. try to provision a new host with your NFS medium. When everything is configured correctly, proxy will serve the files you just copied and installer will mount the NFS medium for installation.

### DIFF
--- a/modules/tftp/server.rb
+++ b/modules/tftp/server.rb
@@ -2,6 +2,8 @@ require 'fileutils'
 require 'pathname'
 
 module Proxy::TFTP
+  extend Proxy::Log
+
   class Server
     include Proxy::Log
     # Creates TFTP pxeconfig file
@@ -135,7 +137,7 @@ module Proxy::TFTP
       ["#{pxeconfig_dir}/01-"+mac.gsub(/:/,"-").downcase+".ipxe"]
     end
   end
- 
+
   def self.fetch_boot_file dst, src
     filename    = boot_filename(dst, src)
     destination = Pathname.new(File.expand_path(filename, Proxy::TFTP::Plugin.settings.tftproot)).cleanpath
@@ -145,8 +147,18 @@ module Proxy::TFTP
     # Ensure that our image directory exists
     # as the dst might contain another sub directory
     FileUtils.mkdir_p destination.parent
+    choose_protocol_and_fetch src, destination
+  end
 
-    ::Proxy::HttpDownload.new(src.to_s, destination.to_s).start
+  def self.choose_protocol_and_fetch(src, destination)
+    case URI(src).scheme
+    when 'http', 'https', 'ftp'
+      ::Proxy::HttpDownload.new(src.to_s, destination.to_s).start
+    when 'nfs'
+      logger.debug "NFS as a protocol for installation medium detected."
+    else
+      raise "Cannot fetch boot file, unknown protocol for medium source path: #{src}"
+    end
   end
 
   def self.boot_filename(dst, src)
--- a/test/tftp/tftp_test.rb
+++ b/test/tftp/tftp_test.rb
@@ -49,4 +49,23 @@ class TftpTest < Test::Unit::TestCase
   def test_boot_filename_uses_dash_when_prefix_does_not_end_with_slash
     assert_equal "a/b/c-somefile", Proxy::TFTP.boot_filename('a/b/c', '/d/somefile')
   end
+
+  def test_choose_protocol_and_fetch_wget
+    ::Proxy::HttpDownload.any_instance.expects(:start).returns(true).times(3)
+    %w(http://proxy.test https://proxy.test ftp://proxy.test).each do |src|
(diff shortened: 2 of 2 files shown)


modules/tftp/server.rb
```diff
@@ -145,8 +145,18 @@ def self.fetch_boot_file dst, src
...
+    choose_protocol_and_fetch src
+  end
 
-    ::Proxy::HttpDownload.new(src.to_s, destination.to_s).start
+  def self.choose_protocol_and_fetch(src)
+    case src.downcase.split(':').first
+    when 'http', 'https', 'ftp'
+      ::Proxy::HttpDownload.new(src.to_s, destination.to_s).start
+    when 'nfs'
+      ::Proxy::LogBuffer::Decorator.instance.debug "NFS as a protocol for for installation medium detected."
```
ekohl: for for should be for

modules/tftp/server.rb
```diff
@@ -145,8 +145,18 @@ def self.fetch_boot_file dst, src
     # Ensure that our image directory exists
     # as the dst might contain another sub directory
     FileUtils.mkdir_p destination.parent
+    choose_protocol_and_fetch src
+  end
 
-    ::Proxy::HttpDownload.new(src.to_s, destination.to_s).start
+  def self.choose_protocol_and_fetch(src)
+    case src.downcase.split(':').first
```
dmitri-d: I think ```URI(src).scheme``` is a bit better here: if anything, no custom uri parsing and therefore no need for tests.

modules/tftp/server.rb
```diff
@@ -145,8 +145,18 @@ def self.fetch_boot_file dst, src
...
+    when 'http', 'https', 'ftp'
+      ::Proxy::HttpDownload.new(src.to_s, destination.to_s).start
+    when 'nfs'
+      ::Proxy::LogBuffer::Decorator.instance.debug "NFS as a protocol for installation medium detected."
```
dmitri-d: Please use ```logger.debug``` instead.

dmitri-d (changes requested):

dmitri-d:
Could you fix the failing test, and add tests to verify handling of http, https, and nfs uris please?

xprazak2:
I added tests, fixed the failing one, using `URI` and `logger.debug` now.

test/tftp/tftp_test.rb
```diff
@@ -49,4 +49,16 @@ def test_boot_filename_has_no_dash_when_prefix_ends_with_slash
...
+
+  def test_choose_protocol_and_fetch_wget
+    ::Proxy::HttpDownload.any_instance.expects(:start).returns(true).times(3)
+    %w(http://proxy.test https://proxy.test ftp://proxy.test).each do |src|
+      Proxy::TFTP.choose_protocol_and_fetch src, '/destination'
+    end
+  end
+
+  def test_choose_protocol_and_fetch_nfs
+    Proxy::TFTP.logger.expects(:debug).once
```
dmitri-d: I think the fact that we are logging at a specific level isn't important/is fragile here. I think a check that no exception is raised is in this case clearer in its intention and also less fragile.  Additionally, could add a test that verifies that an exception is raised when the url scheme isn't recognized?

xprazak2:
I check that nothing is raised instead of expecting a debug call and I added the missing test case.

dmitri-d:
Merged, thanks @xprazak2!

PR #518, opened 2017-04-05
Update register-service.rb

tek0011:
The question asking what user to run as was confusing for us as it defaults on the screen to "default user/logged in user", and was presented in a manner making a user think it's a yes or no question.  There are numerous other reasons why the windows service might fail to get created.  We thought for the longest time it was because the service creation file (this file), didnt allow a domain\username to be entered.  It wasn't until we looked at the file that we realized its not a yes or no question, but asking what user you want use (and only showing what the current default would be if you hit enter).

Also, account was spelled wrong.

### DIFF
--- a/extra/register-service.rb
+++ b/extra/register-service.rb
@@ -23,10 +23,10 @@ default_user         = ENV["USERDOMAIN"] + '\\' + default_user if ENV["USERDOMAI
 default_service_name = 'smart proxy'
 
 puts "This service must be run under an account that is a member of 'DHCP Administrators' group"
-puts 'The acount can be local or a domain account. If it is a domain account then use the domain\account syntax'
-service_name = ask("Enter the name of the service ")  {|s| s.default = default_service_name}
-user         = ask("Run this service as this user? ") {|u| u.default = default_user}
-pass1        = ask("Enter the user's password ") {|p| p.echo = "x"}
+puts 'The account can be local or a domain account. If it is a domain account then use the domain\account syntax'
+service_name = ask("Enter the name of the service. ")  {|s| s.default = default_service_name}
+user         = ask("Enter the user to run the service as, or press enter to use the following user: ") {|u| u.default = default_user}
+pass1        = ask("Enter the user's password. ") {|p| p.echo = "x"}
 
 description = 'Foreman Smart Proxy'
 description += " (#{service_name})" unless service_name == default_service_name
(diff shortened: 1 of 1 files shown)


dmitri-d:
Could you create a ticket for this at http://projects.theforeman.org/projects/smart-proxy/issues and update the commit message with the ticket number please?

tek0011:
@witlessbird  - I originally created this.  It was just a basic request on where to change documentation.  Should we just reopen?

http://projects.theforeman.org/issues/19112

dmitri-d:
@tek0011: that ticket looks sufficient. I reopened it and moved it into smart-proxy.

dmitri-d:
@tek0011: please update the commit message to include the ticket number (see theforeman-bot comment for the format), and push the changes again.

tek0011:
@witlessbird I am on mobile for the next 24-48 hours.  Might not be able to do this.  I made the changes remotely on github and don't believe there is a way to git commit --amend (within the website).

dmitri-d:
[test]

extra/register-service.rb
```diff
@@ -23,10 +23,10 @@
 default_service_name = 'smart proxy'
 
 puts "This service must be run under an account that is a member of 'DHCP Administrators' group"
-puts 'The acount can be local or a domain account. If it is a domain account then use the domain\account syntax'
-service_name = ask("Enter the name of the service ")  {|s| s.default = default_service_name}
-user         = ask("Run this service as this user? ") {|u| u.default = default_user}
-pass1        = ask("Enter the user's password ") {|p| p.echo = "x"}
+puts 'The account can be local or a domain account. If it is a domain account then use the domain\account syntax'
+service_name = ask("Enter the name of the service. ")  {|s| s.default = default_service_name}
+user         = ask("Enter the user to run the service as, or press enter to use the following user: ") {|u| u.default = default_user}
```
ekohl: Is the 'press enter' part needed? It isn't there for the service_name

tek0011: I suppose not.  I rewrote it like 10 times before I just stayed with that.  Basically, it just needs to be intuitive for us that dont really use windows often.  I suppose

> Enter the user to run the service as:  

would be acceptable.  The original wording made it seem like its a yes or no question that doent accept input, when it does.  Your call.

ekohl: I think that would be an very good text. @witlessbird do you agree?

dmitri-d: sounds good to me.

ekohl: @tek0011 mind amending the commit to use the new phrasing? Otherwise 👍

ekohl (approved):

tek0011:
Sorry.  I dont really know what the bot is looking for, nor can I git --amend commits.

ekohl:
You already committed it so you need to squash it. https://makandracards.com/makandra/527-squash-several-git-commits-into-a-single-commit describes how you can rewrite the branch. I'm sure it's no problem if you can't figure it out since it can be fixed on merge.

dmitri-d:
I rebased and squashed the commits, and added a redmine issue too. Merged in a645953. Thanks, @tek0011!

PR #524, opened 2017-05-03
Fixes #19441 - recursive descent parser for dhcpd config files.

dmitri-d:


dmitri-d:
This rewrite fixes all known parser issues related to handling of nested and conditional structures in dhcpd config files (please see http://projects.theforeman.org/issues/18674 and http://projects.theforeman.org/issues/17800). The parser now keeps track of dhcp options and attributes defined in groups, subnets, and shared networks, and makes uses them when parsing subnet and host records. This implementation makes it possible to resolve http://projects.theforeman.org/issues/14396.

dmitri-d:
@timogoebel: perhaps you'd be interested in reviewing this?

lzap:
Nice, that's a lot of work writing the parser. Also good timing after the 1.15 release, we should hurry up merging this and start testing this in develop. Overall it's clean and the parser is suprisingly readable (what I mean by that is the rsec syntax). My approach to the review would be just to test the code and merge it so we can all start testing this in real world scenarios ASAP, I gave the code quick read and I did not spot anything.

But I am missing one important piece - could you add small binary that will check syntax of configuration and return non-zero when error is found (ideally with some error message what's wrong)? Similarly what `dhcpd -t` does, so users are able to quickly test after do configuration change. Lots of users do manually modify reservations which are outside of Foreman control and when they make typo, it's usually too late when they find this.

Also are you adding the new dependency into packaging repo? That's blocker for us to merge this.

modules/dhcp_isc/dhcpd_conf_parser.rb
```diff
@@ -0,0 +1,7 @@
+require 'rsec'
+
+module ::Proxy::DHCP::ISC
+  class DhcpdConfParser < Parslet::Parser
+
```
lzap: Nit pick, I'd perhaps do `class DhcpdConfParser < Parslet::Parser; end` but not sure if Rubocop will like it.

dmitri-d:
Thanks for looking into this, @lzap.
 
- Removed modules/dhcp_isc/dhcpd_conf_parser.rb, as it wasn't being used anymore
- created extra/dhcpd_config_check.rb (not very useful atm).

However: the dhcpd config check script isn't very useful atm. The parser has been written in such a way that stuff it doesn't recognise/fails to parse ends up in IgnoredDeclaration/IgnoredBlock ast nodes, which are then simply skipped. This was done so we don't fail on things we don't care about (atm).

The config check script could print subnets and hosts that it found. Or it could output ignored nodes (although this can get messy). Another alternative is to parse all dhcpd syntax, in which case we could report on syntax errors. This would mean parsing things like option definitions, conditional expressions, a bunch more options, etc.

dmitri-d:
And, yes, I'm planning on packaging rsec gem.

lzap:
I haven't relized that the ignored decl/block has this overall effect, this is nice as it solves all unrecognized syntax in general. Then indeed the checker output is less useful, but still valuable tool in case of possible bug or something, I like it. I think this could be bin/ folder to have it on $PATH. I think printing ignored nodes can be helpful as well as printing subnets and other recognized structures for easier debugging, but only if its not much work. We can always grab configs from users. If you decide to extend it please add a test for it.

I am ready to give this a test and merge once we have rsec gem merged in packaging repo. No reason to wait, we need to start using this on real deployments.

lzap:
@witlessbird pls let me know if you are planning another rebase or not and I will start testing the patch, thanks.

dmitri-d:
@lzap: I'll package rsec gem tomorrow and will rebase the PR then too. I'll see what I can do about printing of parser output in dhcpd_config_check.rb script, but that should be largely independent of the rest of the code...

dmitri-d:
@lzap: rpm package for rsec: https://github.com/theforeman/foreman-packaging/pull/1652

lzap:
Understood, just let me know, I simply want to be testing the very same commit that gets merged :-)

lzap:
I gave it quick run and works just fine.

dmitri-d:
* updated dhcpd config file checker to return list of parsed subnets, hosts and leases, and ignored statements and blocks. Ignored statements and blocks also contain chain of parent ids to help identify their location.

lzap:
Works well here, we are good to go. Merging once tests are green. Great job, this was painful part of our DHCP module.

lzap:
@witlessbird there's a rubocop failure

dmitri-d:
* resolved rubocop complaints
* rebased

lzap:
Thanks @witlessbird great progress there!

PR #525, opened 2017-05-09
fixes #19480 - add systemd notify support

domcleal:
When started under systemd, the launcher now logs via the notify socket
API when all configured WEBrick servers have started so the service
state is reported accurately (rather than the daemonization point).

An example unit file is now shipped that uses a --no-daemonize
argument to override the config file, so misconfiguration will not
prevent the service from starting correctly.

### DIFF
--- a/extra/systemd/foreman-proxy.service
+++ b/extra/systemd/foreman-proxy.service
@@ -1,0 +1,12 @@
+[Unit]
+Description=Foreman Proxy
+Wants=basic.target
+After=basic.target network.target
+
+[Service]
+Type=notify
+User=foreman-proxy
+ExecStart=/usr/share/foreman-proxy/bin/smart-proxy --no-daemonize
+
+[Install]
+WantedBy=multi-user.target
--- a/lib/proxy/sd_notify.rb
+++ b/lib/proxy/sd_notify.rb
@@ -1,0 +1,30 @@
+require 'socket'
+
+# Implementation of libsystemd's sd_notify API, sends current state via socket
+module Proxy
+  class SdNotify
+    def active?
+      !ENV['NOTIFY_SOCKET'].nil?
+    end
+
+    def notify(message)
+      create_socket.tap do |socket|
+        socket.sendmsg(message.chomp + "\n") # ensure trailing \n
+        socket.close
+      end
+    end
+
+    def ready(state = 1)
+      notify("READY=#{state}")
+    end
+
+    private
+
+    def create_socket
+      raise 'Missing NOTIFY_SOCKET environment variable, is this process running under systemd?' unless active?
+      Socket.new(Socket::AF_UNIX, Socket::SOCK_DGRAM, 0).tap do |socket|
+        socket.connect(Socket.pack_sockaddr_un(ENV['NOTIFY_SOCKET']))
+      end
+    end
+  end
+end
--- a/lib/smart_proxy_main.rb
+++ b/lib/smart_proxy_main.rb
@@ -42,7 +42,7 @@ require 'sinatra/default_not_found_page'
 require 'webrick-patch'
 
 module Proxy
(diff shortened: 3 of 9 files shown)


extra/systemd/foreman-proxy.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Foreman Proxy
+Wants=basic.target
+After=basic.target network.target
+
+[Service]
+Type=notify
+User=foreman-proxy
+ExecStart=/usr/share/foreman-proxy/bin/smart-proxy --no-daemonize
+EnvironmentFile=-/etc/sysconfig/foreman-proxy
```
ekohl: This is RH specific. Debian has ```/etc/default``` but I'm wondering if the environment is even used. If users need it they can override it in ```/etc/systemd/system/foreman-proxy.service.d/override.conf```.

ekohl (commented):
Looks very good.

extra/systemd/foreman-proxy.service
domcleal: Good point, I had been thinking that packages for other OSes could patch the location, but preferring the systemd overrides seems a much better default. This can be added back in for RPM packages for their backwards-compatibility. Line removed.

lib/proxy/sd_notify.rb
```diff
@@ -0,0 +1,27 @@
...
+
+# Implementation of libsystemd's sd_notify API, sends current state via socket
+module Proxy
+  class SdNotify
+    def active?
+      !ENV['NOTIFY_SOCKET'].nil?
+    end
+
+    def notify(message)
+      socket.sendmsg(message.chomp + "\n") # ensure trailing \n
```
dmitri-d: You aren't closing the socket when you are done with it. Or is this intentional so other changes in state (restarts, for example) can be passed to systemd?

lib/smart_proxy_main.rb
```diff
@@ -43,6 +43,7 @@
 
 module Proxy
   SETTINGS = Settings.load_global_settings
+  SETTINGS.apply_argv(ARGV)
```
dmitri-d: More of a nitpick: I think Settings::Global#apply_argv call could be made from Settings.load_global_settings, mostly to keep settings initialization logic in one spot. If so, Settings.load_global_settings should probably be renamed to something like Settings.initialize_global_settings...

test/sd_notify_test.rb
```diff
@@ -0,0 +1,58 @@
+require 'test_helper'
+require 'proxy/sd_notify'
+
+class SdNotifyTest < Test::Unit::TestCase
+  def test_active_with_notify_socket
+    with_notify_socket('/run/systemd/notify') do
```
dmitri-d: Any reason you are using an actual systemd socket instead of a temporary one here?

lib/proxy/sd_notify.rb
domcleal: It was relying on the GC to finalise the object, which would autoclose it. I've changed it to explicitly close the socket as soon as it's done to free up the fd.

lib/smart_proxy_main.rb
domcleal: Thanks, updated as suggested.

test/sd_notify_test.rb
domcleal: It's actually a mocked environment variable, just set to a real life path. I've changed it to `/mock` to make this clearer.

domcleal:
PR updated, thanks for the reviews.

test/sd_notify_test.rb
dmitri-d: Ahh, yes, Proxy::SdNotify.new.active? only checks that the environment variable isn't nil. The change made it clear(er) though, thanks.

dmitri-d:
Merged, thanks @domcleal!

PR #526, opened 2017-05-09
fixes #19485 - pin sinatra to 1.x on Ruby 2.0-2.1

domcleal:


Gemfile
```diff
@@ -2,6 +2,8 @@ source 'https://rubygems.org'
 
 gemspec
 
+gem 'sinatra', '< 2' if RUBY_VERSION < '2.2'
```
dmitri-d, line 5: This duplicates sinatra dependency declared in gemspec file (https://github.com/theforeman/smart-proxy/blob/develop/smart_proxy.gemspec#L18). Could you remove it from there, but keep it here?

domcleal: Well it's a runtime dependency so needs to stay in the gemspec, but built gemspecs can't contain conditionals (such as `if RUBY_VERSION` etc.), so I think it'll have to be here too.

dmitri-d:
Merged, thanks @domcleal.

PR #527, opened 2017-05-19
Fixes #19597 - fallback DHCP hostname pattern

lzap:
I stumbled upon situation during host renaming with libvirt provider when hostname was not passed and it defaulted to '*lease*' which failed for incorrect characters. Apparently * is not allowed there, therefore I renaming this to something that will always work. Also adding MAC address so it's unique hostname.

To reproduce use libvirt provider, create a host in foreman using libvirt provider, then rename the hostname - it conflicts with entry which is created by libvirt itself. That's anoter story, in this patch I just want to fix the "fallback" name.


dmitri-d:
Merged, thanks, @lzap.

PR #528, opened 2017-05-20
Fixes #6022 - disable StrictHostKeyChecking

mmoll:


ekohl:
Should this be configurable?

mmoll:
Unsure, I guess at least the default of disabling the checking makes things much easier for the people using puppetrun with ssh.

lzap:
I tend to agree with @ekohl having this configurable is not a huge change and benefit is big. On the other hand, this is doable in `/etc/ssh/ssh_config` but I still find it useful to have it here as a sane default rather than puppetizing another file.

dmitri-d:
I'm on a fence re: configuration option. It's probably not that useful as a security measure for puppet-managed hosts...

mmoll:
so...?

dmitri-d:
My vote would be for not having a configurable 'StrictHostKeyChecking' option, at least at first. Unless everyone else is feeling strongly about this...

lzap:
I am fine with that.

lzap:
Thanks.

PR #529, opened 2017-05-23
Fixes #19599: del dhcp record returns empty body on success.

dmitri-d:


ekohl:
Should it also verify the HTTP status code? I'd expect a 204 No Content or 404 Not Found

dmitri-d:
> Should it also verify the HTTP status code? I'd expect a 204 No Content or 404 Not Found

I think a 204 would be appropriate for the newly added calls. I also think it should be handled in a dedicated PR. The old (and deprecated) call should remain as it was before the regression this PR is addressing was introduced.

ekohl:
Then maybe verify that it keeps returning HTTP 200 in the tests? There are frameworks out there (Django REST Framework) that automatically returns a HTTP 204 when the body is empty so a test to specify the behaviour would be nice.

dmitri-d:
* updated dhcp api tests to check for 200 return status on delete record calls.

domcleal:
> updated dhcp api tests to check for 200 return status on delete record calls.

I can't see this change, can you re-check please?

dmitri-d:
Apologies, forgot to push.

ekohl (approved):

domcleal (approved):

domcleal:
Thanks @witlessbird.

PR #530, opened 2017-05-23
Fixes #6945 - use autosign path for puppetca

mmoll:


mmoll:
I'm unsure about the migration and how the test for it should look like, probably I'm missing some nice helpers there?

extra/migrations/20170523000000_migrate_autosign_setting.rb
```diff
@@ -0,0 +1,29 @@
...
+    unless File.exist?(puppetca_config)
+      return
+    end
+
+    to_migrate = YAML.load_file(puppetca_config)
+
+    output = migrate_puppetca_configuration(to_migrate)
+
+    write_yaml(puppetca_config, output)
+  end
```
dmitri-d: I can see two issues with this migration: it doesn't move unchanged config files into the destination directory (an example: https://github.com/theforeman/smart-proxy/blob/develop/extra/migrations/20161209000000_migrate_realm_settings.rb#L21), and it uses wrong directory to save the updated config file (```dst_dir``` points to the root. see https://github.com/theforeman/smart-proxy/blob/develop/extra/migrations/20161209000000_migrate_realm_settings.rb#L41 for an example).

dmitri-d (changes requested):

dmitri-d:
Re: migration testing. A common pattern is to test settings transformation (see https://github.com/theforeman/smart-proxy/blob/develop/test/migrations/realm_migration_test.rb for an example).

mmoll:
should be better now

extra/migrations/20170523000000_migrate_autosign_setting.rb
```diff
@@ -0,0 +1,52 @@
...
+    end
+    migrated
+  end
+
+  def write_to_files(output)
+    output.keys.each do |m|
+      next if output[m].empty? || m == :unknown
+      File.open(path(dst_dir, "settings.d", "#{m}.yml"),'w') do |f|
+        f.write(strip_ruby_symbol_encoding(output[m].to_yaml))
+        if (m == :realm) && !output[:unknown].empty?
```
dmitri-d: I don't think you need this conditional at all -- I don't see remap_parameter() returning :unknown keys.

mmoll: *sigh*, I thought I had eradicated all copypastos from the realm migration... anyway, fixed.

dmitri-d (approved):

dmitri-d:
Merged, thanks @mmoll!

PR #531, opened 2017-05-24
Fixes #19649 - dhcpd provider ignores conflicts with leases when a new record is being created

dmitri-d:


lzap:
Works like charm, thanks for quick patch @witlessbird ! Merging.

NeilHanlon:
Sorry to drag up old code, but I'm running into something similar to this, where a lease exists but for some reason there's a mismatch between the request and the existing record. Specifically, when I turn on debugging, everything in the two hashes (from the `unless similar_records.empty?` block in server.rb) matches, except the 'existing' record contains `:deleteable=>true`  in its hash--which it looks like is due to the lease in the file being marked `dynamic`.

Was this perhaps a code change on how requests are made from Foreman itself?

NeilHanlon:
@witlessbird thoughts on this?

dmitri-d:
@NeilHanlon: try latest stable release and see if the issue persists. If it does, please file a bug report at http://projects.theforeman.org/projects/smart-proxy/issues.

NeilHanlon:
@witlessbird I think I figured it out.. I'll open a bug report

PR #532, opened 2017-05-25
Fixes #19666 - add DHCP record type field

lzap:
There are two changes in this patch. First, we do not show if a record is type of lease or reservation. This can be useful information. But more importantly, this removes unnecessary "name" field from lease record, which is treatened as a hostname in Foreman. This will fix:

http://projects.theforeman.org/issues/19634

I am filing this for #19634 bug but I might end up not use this information, still I would like Smart Proxy to report this.

### DIFF
--- a/modules/dhcp_common/record.rb
+++ b/modules/dhcp_common/record.rb
@@ -5,7 +5,7 @@ module Proxy::DHCP
   # represent a DHCP Record
   class Record
 
-    attr_reader :ip, :mac, :subnet, :options
+    attr_reader :ip, :mac, :subnet, :options, :type
     include Proxy::DHCP
     include Proxy::Log
     include Proxy::Validations
--- a/modules/dhcp_common/record/lease.rb
+++ b/modules/dhcp_common/record/lease.rb
@@ -5,6 +5,7 @@ module Proxy::DHCP
     attr_reader :name, :starts, :ends, :state
 
     def initialize(name, ip_address, mac_address, subnet, starts, ends, state, options = {})
+      @type = "lease"
       @name = name || "lease-#{mac_address.tr(':-','')}"
       @starts = starts
       @ends = ends
@@ -21,7 +22,7 @@ module Proxy::DHCP
     end
 
     def to_json(*opts)
-      Hash[[:name, :ip, :mac, :subnet, :starts, :ends, :state].map{|s| [s, send(s)]}].merge(options).to_json(*opts)
+      Hash[[:name, :ip, :mac, :subnet, :starts, :ends, :state, :type].map{|s| [s, send(s)]}].merge(options).to_json(*opts)
     end
   end
 end
--- a/modules/dhcp_common/record/reservation.rb
+++ b/modules/dhcp_common/record/reservation.rb
@@ -5,6 +5,7 @@ module Proxy::DHCP
     attr_reader :name
 
     def initialize(name, ip_address, mac_address, subnet, options = {})
+      @type = "reservation"
       @name = name
       super(ip_address, mac_address, subnet, options)
     end
@@ -22,7 +23,7 @@ module Proxy::DHCP
     end
 
     def to_json(*opts)
-      Hash[[:name, :ip, :mac, :subnet].map{|s| [s, send(s)]}].merge(options).to_json(*opts)
+      Hash[[:name, :ip, :mac, :subnet, :type].map{|s| [s, send(s)]}].merge(options).to_json(*opts)
     end
   end
 end
--- a/test/dhcp/dhcp_api_test.rb
+++ b/test/dhcp/dhcp_api_test.rb
@@ -144,18 +144,39 @@ class DhcpApiTest < Test::Unit::TestCase
     assert_equal 404, last_response.status
   end
 
-  def test_get_record_by_ip
+  def test_get_reservation_record_by_ip
     @server.expects(:find_records_by_ip).with("192.168.122.0", "192.168.122.1").returns([@reservations.first])
 
     get "/192.168.122.0/ip/192.168.122.1"
 
     assert last_response.ok?, "Last response was not ok: #{last_response.status} #{last_response.body}"
     expected = [{
-      "hostname" =>"test.example.com",
-      "ip"       =>"192.168.122.1",
-      "mac"      =>"00:11:bb:cc:dd:ee",
+      "type"     => "reservation",
+      "hostname" => "test.example.com",
+      "ip"       => "192.168.122.1",
+      "mac"      => "00:11:bb:cc:dd:ee",
       "name"     => 'test.example.com',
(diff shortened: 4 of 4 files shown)


dmitri-d:
Name field may or may not be present depending on dhcp backend. Ms dhcp, for example, assigns names to leases. Ultimately, however, I think both reservation and lease name attributes end up as dhcp host-name option (needs verification), and can be modelled as such (cleaning up name assignments in the process).

dmitri-d:
> This will fix: http://projects.theforeman.org/issues/19634

I don't think it will: the equality method used for detection of conflicts is too dhcpd-centric. Either foreman needs to be aware of what backend it's dealing with (not great), or it needs to delegate conflict detection logic to smart-proxy, or smart-proxy's "create record" contract needs to change and encompass conflicting lease removals when needed/possible.

modules/dhcp_common/record.rb
```diff
@@ -21,6 +21,10 @@ def subnet_address
       @subnet.network
     end
 
+    def type
+      "record"
+    end
```
dmitri-d: Probably not needed on the base class -- these are never valid instances (only used as a base class).

lzap:
I just realized that this only fixed the new endpoint and not the old one!

```
[lzap@lzapx ~]$ curl-proxy /dhcp/192.168.122.0/192.168.122.46
{
    "hostname": "lease-000011223344",
    "ip": "192.168.122.46",
    "mac": "00:00:11:22:33:44"
}
[lzap@lzapx ~]$ curl-proxy /dhcp/192.168.122.0/ip/192.168.122.46
[
    {
        "ip": "192.168.122.46",
        "mac": "00:00:11:22:33:44",
        "subnet": "192.168.122.0/255.255.255.0",
        "starts": "2017-05-25 10:09:37 UTC",
        "ends": "2017-05-25 10:19:37 UTC",
        "state": "active",
        "type": "lease",
        "cltt": "2017-05-25 10:09:37 UTC",
        "next_state": "free",
        "hardware_type": "ethernet"
    }
]
```

dmitri-d:
> I just realized that this only fixed the new endpoint and not the old one!

Please don't make changes to deprecated api.

lzap:
> Please don't make changes to deprecated api.

If I will need to I will, someone made a breaking deprecated API change with:

    @name = name || "*lease*"

That completely changed result from ISC DHCP module:

```
[lzap@lzapx ~]$ curl-proxy /dhcp/192.168.122.0/192.168.122.46
{
    "hostname": null,
    "ip": "192.168.122.46",
    "mac": "00:00:11:22:33:44"
}
```

to

```
[lzap@lzapx ~]$ curl-proxy /dhcp/192.168.122.0/192.168.122.46
{
    "hostname": "*lease*",
    "ip": "192.168.122.46",
    "mac": "00:00:11:22:33:44"
}
```

But I need to investigate more, looks like hostname is not the issue.

dmitri-d:
> If I will need to I will, someone made a breaking deprecated API change with

There's no point in fixing deprecated api: you should switch to the new implementation, the deprecated end-point will be removed immediately after 1.16 release.

dmitri-d:
Also, please see my comments re: name field on lease object. The field should either stay or be converted to a dhcp option (but on both reservation and lease).

lzap:
Ok I somehow got impression that http://projects.theforeman.org/issues/19634#note-17 is a regression, but after extensive testing I come to conclusion that it always worked like that. Therefore no hostname changes in this patch at all.

But our code around hostname (assuming it's a lease) is fragile and does not work anymore perhaps for many releases (I only tested up to 1.14), so this patch introduces the "type" flag which will enable Foreman not to treat leases as conflicts. We can add this as a configuration global setting, or perhaps smart-proxy could expose an API that would tell foreman handling of this.

lzap:
I still see one difference, which might be not relevant to this PR. Hostname is present in the old API and not present in the new API endpoint for leases. Compare:

```
[lzap@lzapx smart-proxy]$ curl-proxy /dhcp/192.168.122.0/ip/192.168.122.46
[
    {
        "ip": "192.168.122.46",
        "mac": "00:00:11:22:33:44",
        "subnet": "192.168.122.0/255.255.255.0",
        "starts": "2017-05-25 11:01:09 UTC",
        "ends": "2017-05-25 11:11:09 UTC",
        "state": "active",
        "type": "lease",
        "cltt": "2017-05-25 11:01:09 UTC",
        "next_state": "free",
        "hardware_type": "ethernet"
    }
]
[lzap@lzapx smart-proxy]$ curl-proxy /dhcp/192.168.122.0/192.168.122.46
{
    "hostname": "lease-000011223344",
    "ip": "192.168.122.46",
    "mac": "00:00:11:22:33:44"
}
```
Shall I open another PR for this?

dmitri-d:
I think adding a "type" field is fine. I think the field should always be present and its appearance not triggered by a setting.

lzap:
Sorry I mean that smart-proxy could expose information to Foreman to treat leases as conflicts or not. MSDHCP would return yes, otherwise no. That was what I meant.

lzap:
Or we could simply add an additional flag "conflicting" and when explicitly set to "yes" we would do this, otherwise (default behavior) lease is not a conflict.

dmitri-d:
> I still see one difference, which might be not relevant to this PR. Hostname is present in the old API and not present in the new API endpoint for leases. 

I can see "name" field returned with leases with HEAD in develop branch.

dmitri-d:
> Sorry I mean that smart-proxy could expose information to Foreman to treat leases as conflicts or not. MSDHCP would return yes, otherwise no. That was what I meant.

I think this might be too narrow. This isn't the only difference between dhcpd and ms dhcp. A better approach would be to delegate all backend-specific logic to smart-proxy. In this case it would mean that smart-proxy's "create record" contract would need to change and encompass conflicting lease removals when needed/possible.

lzap:
Oh of course, I missed the fact I removed it! Amended from the patch, now it looks good:

```
[lzap@lzapx smart-proxy]$ curl-proxy /dhcp/192.168.122.0/192.168.122.14
{
    "hostname": "mac525400e47953.local.lan",
    "ip": "192.168.122.14",
    "mac": "52:54:00:e4:79:53"
}
[lzap@lzapx smart-proxy]$ curl-proxy /dhcp/192.168.122.0/ip/192.168.122.14
[
    {
        "name": "mac525400e47953.local.lan",
        "ip": "192.168.122.14",
        "mac": "52:54:00:e4:79:53",
        "subnet": "192.168.122.0/255.255.255.0",
        "type": "reservation",
        "hostname": "mac525400e47953.local.lan",
        "deleteable": true,
        "hardware_type": "ethernet",
        "filename": "grub/bootx64.efi",
        "nextServer": "192.168.122.1"
    }
]
```

lzap:
Created a ticket for what you suggested: http://projects.theforeman.org/issues/19668

Thanks.

lzap:
@witlessbird I improved the patch, using attribute from record now instead explicit methods which is cleaner, also added extra test for lease type.

dmitri-d:
Merged, thanks @lzap!

PR #534, opened 2017-06-30
Fixes #20173 - removed race condition in Subnet#unused_ip

dmitri-d:
Also fixes #6819

### DIFF
--- a/modules/dhcp_common/subnet.rb
+++ b/modules/dhcp_common/subnet.rb
@@ -28,6 +28,8 @@ module Proxy::DHCP
       @options[:ntp_servers] = options[:ntp_servers] if options[:ntp_servers]
       @options[:interface_mtu] = options[:interface_mtu].to_i if options[:interface_mtu]
       @options[:range] = options[:range] if options[:range] && options[:range][0] && options[:range][1] && valid_range(:from => options[:range][0], :to => options[:range][1])
+
+      @m = Monitor.new
     end
 
     def include? ip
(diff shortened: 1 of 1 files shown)


lzap:
Can you amend the commit message to "Fixes #A,#B" form? It will work.

dmitri-d:
- amended the commit message.

modules/dhcp_common/subnet.rb
```diff
@@ -62,28 +64,18 @@ def netmask_to_i
...
-
       # this returns the index in the file
       return @file.readlines.first.to_i rescue 0
     end
 
-    def write_index_and_unlock index
+    def write_index index
       @file.reopen(@filename,'w')
       @file.write index
       @file.close
```
lzap: Clouse could be in `ensure` block here, nitpick.

modules/dhcp_common/subnet.rb
```diff
@@ -62,28 +64,18 @@ def netmask_to_i
...
-      # Loop if the file is locked
-      Timeout::timeout(30) { sleep 0.1 while File.exist? @lockfile }
-
-      # Touch the lock the file
-      File.open(@lockfile, "w") {}
-
       @file = File.new(@filename,'r+') rescue File.new(@filename,'w+')
-
       # this returns the index in the file
       return @file.readlines.first.to_i rescue 0
```
lzap: Possible DoS attach would be constructing huge file as we read all of it, improvement might be to use just read of maximum 5 characters (3 digits + newline). This is just an idea for improvement here.

dmitri-d:
* Moved file closing in #write_index method into an "ensure" block.

modules/dhcp_common/subnet.rb
```diff
@@ -62,28 +64,19 @@ def netmask_to_i
...
     end
 
-    def write_index_and_unlock index
+    def write_index index
       @file.reopen(@filename,'w')
       @file.write index
+    ensure
       @file.close
-      File.delete @lockfile
     end
```
lzap: One change needed tho, since we do not use the file for locking now, there is no point of having both file and filename as member variables, I'd simply rewrite read and write methods using safe block file operations.

lzap (changes requested):
I suggest small refactoring there.

modules/dhcp_common/subnet.rb
dmitri-d: I think this whole thing should be re-written with an in-memory implementation. Additionally, there are issues with re-using the index: the list of free ips changes as hosts get added/deleted. A position in such a list isn't a great index.

dmitri-d:
* rewrote reading and wiring of the index without using any state and using implicit file closures.

lzap (approved):

lzap:
Thanks, I was not able to reproduce the race condition using curl/parallel but it was giving good results.

PR #536, opened 2017-07-20
Fixes #20370 - Add sshfp DNS record support

agx:
as per

    https://tools.ietf.org/html/rfc4255
    https://tools.ietf.org/html/rfc6594

looks like

    $ curl -d 'fqdn=testsshfp.example.com&value=3 2 1235&type=SSHFP' \
        http://localhost:8000/dns
    $ host -t SSHFP testsshfp.example.com
    testsshfp.example.com has SSHFP record 3 2 1235
    $ curl -X DELETE http://localhost:8000/dns/testsshfp.example.com/SSHFP
    $ host -t SSHFP testsshfp.example.com
    Host testsshfp.example.com not found: 3(NXDOMAIN)

Due to the current smart-proxy API this is a bit unsymmetric.  While we
can create individual records we can only delete all of them.

That's fine for my use case: Letting a Foreman plugin handle SSH-keys
via SSHFP records so one gets rid of "host key has changed prompts once
and for all" (making ssh keys part of the VM life cycle) - but we might
want to extend this at a later point.

This needs

    https://github.com/rubysl/rubysl-resolv/pull/1

### DIFF
--- a/modules/dns/dns_api.rb
+++ b/modules/dns/dns_api.rb
@@ -29,6 +29,9 @@ module Proxy::Dns
         when 'CNAME'
           validate_dns_name!(value, type)
           server.create_cname_record(fqdn, value)
+        when 'SSHFP'
+          validate_sshfp_data!(value)
+          server.create_sshfp_record(fqdn, value)
         when 'PTR'
           validate_reverse_dns_name!(value)
           server.create_ptr_record(fqdn, value)
@@ -63,6 +66,8 @@ module Proxy::Dns
           server.remove_aaaa_record(name)
         when 'CNAME'
           server.remove_cname_record(name)
+        when 'SSHFP'
+          server.remove_sshfp_records(name)
         when 'PTR'
           validate_reverse_dns_name!(name)
           server.remove_ptr_record(name)
@@ -107,6 +112,10 @@ module Proxy::Dns
     def validate_reverse_dns_name!(name)
       validate_dns_name!(name, 'PTR')
       raise Proxy::Dns::Error.new("Invalid reverse DNS #{name}") unless name =~ /\.(in-addr|ip6)\.arpa$/
+    end
+
+    def validate_sshfp_data!(data)
+      raise Proxy::Dns::Error.new("Invalid SSHFP data #{data}") unless data =~/^[0-9]\s+[0-9]\s+[0-9a-f]+$/i
     end
   end
 end
--- a/modules/dns_common/dns_common.rb
+++ b/modules/dns_common/dns_common.rb
@@ -1,5 +1,6 @@
 require 'resolv'
 require 'ipaddr'
+require 'dns_common/dns_resources'
 
 module Proxy::Dns
   class Error < RuntimeError; end
@@ -68,6 +69,18 @@ module Proxy::Dns
       end
     end
 
+    def create_sshfp_record(fqdn, value)
+      algorithm, type, fingerprint = value.split()
+      case sshfp_record_conflicts(fqdn, algorithm.to_i, type.to_i, fingerprint) #returns -1, 0, 1
+        when 1 then
+          raise(Proxy::Dns::Collision, "'SSHFP record for #{fqdn}' is already in use")
+        when 0
+          return nil
+        else
+          do_create(fqdn, "#{algorithm} #{type} #{fingerprint}", "SSHFP")
+      end
+    end
+
     def do_create(name, value, type)
       raise(Proxy::Dns::Error, "Creation of #{type} not implemented")
     end
@@ -90,6 +103,10 @@ module Proxy::Dns
 
     def remove_ptr_record(name)
       do_remove(name, "PTR")
+    end
+
+    def remove_sshfp_records(fqdn)
+      do_remove(fqdn, "SSHFP")
     end
 
     def do_remove(name, type)
@@ -176,6 +193,14 @@ module Proxy::Dns
       end
       record_conflicts_name(name, Resolv::DNS::Resource::IN::PTR, content)
     end
+
+    def sshfp_record_conflicts(fqdn, algorithm, type, fingerprint)
+      current = resolver.getresources(fqdn, DnsResources::SSHFP)
+      return -1 if current.empty?
+      return 0 if current.any? {|sshfp| sshfp.type == type && sshfp.algorithm == algorithm && sshfp.fingerprint == fingerprint}
+      1
+    end
+
 
     def to_ipaddress ip
       IPAddr.new(ip) rescue false
--- a/modules/dns_common/dns_resources.rb
+++ b/modules/dns_common/dns_resources.rb
@@ -1,0 +1,35 @@
+# Additional DNS resource types not (yet) in rubysl-resolv
+
+require 'resolv'
+
+module DnsResources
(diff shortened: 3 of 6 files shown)


modules/dns/dns_api.rb
```diff
@@ -29,6 +29,8 @@ class Api < ::Sinatra::Base
         when 'CNAME'
           validate_dns_name!(value)
           server.create_cname_record(fqdn, value)
+        when 'SSHFP'
```
ekohl, line 32: Can you verify the value? It doesn't need exact but at least verify there are (exactly) 3 values if you split it.

ekohl (changes requested):
As I understand it the current Ruby doesn't have the SSHFP type in the resolv library, correct? I'm not sure what's the best way to accept this then. @witlessbird?

agx:
@ekohl Exactly, no support for SSHFP in ruby-core so far. Monkey patch it in until it's added upstream (and sufficiently available in distros)?

dmitri-d:
@agx: Please don't monkey-patch core ruby libraries. You could keep all SSHFP-related calls in a dedicated class that you could keep in dns_common.

agx:
@witlessbird o.k., does this look better?

dmitri-d:
@agx: This looks good, thank you. Could you add tests for ```DnsResources::SSHFP``` class too?

agx:
@witlessbird done now.

ekohl (approved):
I think it's looking very good.

agx:
@witlessbird anything I can do to get this applied?

dmitri-d:
@agx -- sorry, I'm on vacation right now, and have limited ability to test this PR. What provider(s) have you tested these changes with? Any chance you can test with all the core ones (bind, libvirt, and ms dns)? Ideally all our current providers should be tested with these changes. My concern is that it's possible that not all providers support SSHFP records and a "not implemented" (or similar) error should be returned for those.

agx:
It's teste(d with bind (since there we can also dnssec sign the zone). Libvirt hasn't SSHFP, mdns I don't know.
Having it in for bind would be a good start IMHO since people can then have valid SSH host keys as part of their VM life cycle.

@witlessbird  enjoy your vacation.  There's little risk of the patch bitrotting which was my man concern.

ekohl:
Like @witlessbird I currently don't have access to my testing infra but I'll try to add SSHFP to https://github.com/theforeman/forklift/blob/master/bats/fb-proxy-dns.bats and test it with PowerDNS as well. The code implies it should work without any change, which was my idea when I refactored the code to use do_create.

agx:
@witlessbird @ekohl anything I can do to move this forward?

agx:
@witlessbird @ekohl @dLobatog Ping?

dmitri-d:
Thanks for your patience, @agx. Some feedback:

 - I tested both libvirt and dnscmd (on Windows 2012 server) providers. In neither case I was able to create SSHFP records, albeit for different reasons. Since two out of three default providers do not presently support this functionality (if ever), I think the default implementations of create_sshfp_record and remove_sshfp_record should return a "not implemented error" (pls. see    https://github.com/theforeman/smart-proxy/blob/develop/modules/dns_common/dns_common.rb#L68 for an example).

- I'll need to test this PR with infoblox (@ekohl -- could you test it with powerdns please?), but for now I would suggest to keep sshfp-specific code in dns_common.rb. Possibly in a dedicated module? Providers that support sshfp records can include this module in order to override the default (not implemented) behaviour.

- Please rebase your PR.

ekohl:
I verified this works perfectly with PowerDNS.

ekohl:
Can we move this forward? I don't think anything is blocking this.

dmitri-d:
Please see my comments above.

mmoll:
@agx do you plan do continue working on this?

agx:
@mmoll yes, I intend to pick this up again. In fact I've started to work on foreman related stuff the other week after several months of hiatus.

lzap:
I think we should leverage capabilities once merged into develop: https://github.com/theforeman/smart-proxy/pull/625/files

We can expose DHCP implementation capabilities: A, AAAA, PTR, SSHFP and depending on them we can enable/disable the API. We can then start small (only ISC DHCP) and extend later as we test those.

ekohl:
@lzap that was one of the original ideas behind capabilities. The provider can declare them and our default implementation can inspect the capabilities. It'd also be very useful for Foreman to really know which DNS types are really supported.

lzap:
Very well, I am merging the capabilities PR today, please rebase @agx on top of that and make use of caps in this PR at least in a way that if there is no SSHFS capability, let the API raise a nice error like "Current provider does not provide this capability" or something like that.

lzap:
Here is an example how to leverage capabilities and add a new one: https://github.com/theforeman/smart-proxy/pull/636/files

agx:
Just a rebase for the moment.

modules/dns/dns_api.rb
```diff
@@ -108,5 +113,9 @@ def validate_reverse_dns_name!(name)
       validate_dns_name!(name, 'PTR')
       raise Proxy::Dns::Error.new("Invalid reverse DNS #{name}") unless name =~ /\.(in-addr|ip6)\.arpa$/
     end
+
+    def validate_sshfp_data!(data)
+      raise Proxy::Dns::Error.new("Invalid SSHFP data #{data}") unless data =~/^[0-9]\s+[0-9]\s+[0-9a-f]+$/i
+    end
```
lzap, line 119: Nitpick, exclamation mark is usually used when method changes state or data. But it's incorrect also above, so let's keep it consistent.

lzap (commented):
The code looks good although it's not yet finished. Please next time clearly state this is not yet ready for review ;-)

modules/dns/dns_api.rb
ares: I'd say the method with ! simply "can cause surprise", some rails method use it to indicate they raise exception, some ruby methods changes internal state (hence the surprise) but there's no good rule of thumb.

ares:
Thansk for a nice contribution @agx, we appologize it took us time to get this moving. This now needs rebase and there are more things needed in the code as per last comment. Given this didn't change since May, I'd like to ask, if you're interested to finish this great work. SSHFP is definitely a good adition that even remote execution plugin could take advantage of.

At this moment I'm setting a Waiting on contributor flag. If we don't hear from you in next weeks, we'll proceed with closing this PR. If you get time to finish this later, it can always be reopened.

ares:
It's been 2 weeks, I'm sorry this didn't make it in as I think it's a great improvement. If you find a time and would be interested in rebasing and incorporating comments from the review, let us know, it's very easy to reopen. To reflect the current review queue a bit better, we're closing this PR now.

agx:
> Here is an example how to leverage capabilities and add a new one: https://github.com/theforeman/smart-proxy/pull/636/files

seems capabilities are not there yet for DNS so we discussed here at cfgmgmt camp to go without them for the moment.

agx:
Hmm..can we reopen this after discussion at cfgmgmt about the lack of capabilities on develop?

ekohl:
Yes, we should. Would it make sense to open a [RFC](https://community.theforeman.org/c/development/rfcs/18) about this?

PR #543, opened 2017-09-13
Fixes #20474 - return different ips on sequential unused_ip calls

dmitri-d:


modules/dhcp_common/free_ips.rb
```diff
@@ -0,0 +1,133 @@
...
+          break if @allocation_timestamps.first.nil? || @allocation_timestamps.first > time_now
+          @allocation_timestamps.shift
+          ip = @allocated_ips.shift
+          logger.debug("#{ip} marked as free.")
+        end
+      end
+    end
+
+    def start
+      logger.info("Starting allocated ip address maintenance (used by unused_ip call.")
```
timogoebel: This is missing a closing bracket.

modules/dhcp_common/free_ips.rb
```diff
@@ -0,0 +1,133 @@
...
+        rescue Exception => e
+          logger.debug e.message
+        end
+      end
+
+      logger.warn "No free IPs in #{subnet_address}."
+      nil
+    end
+
+    def tcp_pingable? ip
```
timogoebel: I'd suggest extracting `tcp_pingable?` and `icmp_pingable?` to an extra module.

timogoebel (commented):
@witlessbird: Thanks for this. I didn't have time for a full review, yet but have a question:
What happens if I have a pool of 10 IPs and reserve all 10 of them. Would the user then get the last recently used IP or no IP at all?

timogoebel:
One more note: Do we need a possibility to release a IP back to the "free pool"? I'm thinking of when a user clicks on "Suggest new" in core, the IP is in theory free for another host and can be released again.

dmitri-d:
> Do we need a possibility to release a IP back to the "free pool"?

Not sure -- this won't be possible for all backends: don't think anything of the sort is available under Windows dhcpsapi or Infoblox api. I think ip address shortage could be an issue in networks with really small address pools (i.e. small networks?), and could be dealt with by reducing the "blackout" time interval.

dmitri-d:
> if I have a pool of 10 IPs and reserve all 10 of them. Would the user then get the last recently used IP or no IP at all?

The latter.

modules/dhcp_common/free_ips.rb
dmitri-d: May I ask why -- the code isn't being used anywhere else (nor should it be). The way it's currently factored doesn't make it harder to test either.

timogoebel: I just thought it would be nice for plugins to be able to reuse the code. But I have a feeling, you don't actually want that to happen. :-)

dmitri-d: I'm not against, actually. But I am of the school of thought that reuse should be discovered, not anticipated. I would suggest we wait until someone needs this code -- they can do the refactor then.

dmitri-d:
- Added missing bracket on line https://github.com/theforeman/smart-proxy/pull/543/files#diff-570b0e6b8cfda7b673a43600df034738R38

modules/dhcp_common/free_ips.rb
```diff
@@ -0,0 +1,133 @@
...
+      @timer_task.shutdown unless @timer_task.nil?
+    end
+
+    def time_now
+      Time.now.to_i
+    end
+
+    def find_free_ip(subnet_address, possible_ips, records)
+      @m.synchronize do
+        free_ips = possible_ips - @allocated_ips - records.collect{|record| record.ip}
```
lzap: This line creates huge array in memory, it is very inefficient way picking an IP from a range. This test is a A class IP address range and the rake task is killed by OOM unable to finish the test case:

```
  def test_find_free_ip_aclass
    @subnet = Proxy::DHCP::Subnet.new("10.0.0.0", "255.0.0.0")
    @free_ips.stubs(:icmp_pingable?).returns(false)
    @free_ips.stubs(:tcp_pingable?).returns(false)
    r = Proxy::DHCP::Reservation.new('test', "10.0.0.1", "aa:bb:cc:dd:ee:ff", @subnet, :hostname =>'test')

    assert_equal "10.0.0.2", @free_ips.find_free_ip(@subnet.network, @subnet.valid_range(:from => "10.0.0.1", :to => "10.250.250.250"), [r])
  end
```

Although such a ranges are rather rare, this way of finding IPs is way too memory inefficient. When added with Ruby poor memory management and inability to return allocated memory to OS, this is an issue we should no longer accept in the codebase.

modules/dhcp_common/free_ips.rb
```diff
@@ -0,0 +1,133 @@
+require 'concurrent'
+require 'logger'
+
+module Proxy::DHCP
+  class FreeIps
+    include ::Proxy::Log
+
+    DEFAULT_ADDRESS_KEEP_TIME = 30*60 # 30 mins
+    DEFAULT_CLEANUP_INTERVAL = 60 # 1 min
```
lzap: Those two guys must be configurable.

modules/dhcp_common/free_ips.rb
```diff
@@ -0,0 +1,133 @@
...
+    end
+
+    def clean_up_allocated_ips
+      @m.synchronize do
+        loop do
+          logger.debug("Starting allocated ip addresses cleanup pass...")
+          break if @allocation_timestamps.first.nil? || @allocation_timestamps.first > time_now
+          @allocation_timestamps.shift
+          ip = @allocated_ips.shift
+          logger.debug("#{ip} marked as free.")
```
lzap: I'd appreciate compacting the two debug statements into single line (IPs joined by comma).

modules/dhcp_common/free_ips.rb
```diff
@@ -0,0 +1,133 @@
...
+              mark_ip_as_allocated(ip)
+              logger.debug "Found a pingable IP(#{ip}) address which does not have a Proxy::DHCP record."
+            else
+              logger.debug "Found free IP #{ip} out of a total of #{free_ips.size} free IPs."
+              mark_ip_as_allocated(ip)
+              return ip
+            end
+          end
+        rescue Exception => e
+          logger.debug e.message
```
lzap: Never swallow exceptions into debug level, please print with error level. Debug is ideal target for stacktrace.

lzap (changes requested):
Since lot of code is being refactored, I would like to see the memory issue problem solved as well. I think this is doable with a simple loop (trial and error), but I am pretty sure you will come up with something more clever :-)

modules/dhcp_common/free_ips.rb
dmitri-d: Why? Users would need to understand the internals, this is asking for trouble. 
I don't see why anyone would want to change ```DEFAULT_CLEANUP_INTERVAL```. I thought 30 mins is reasonable for ```DEFAULT_ADDRESS_KEEP_TIME```, but can adjust it in either direction.

modules/dhcp_common/free_ips.rb
dmitri-d: They are separate on purpose -- the first one indicates the cleanup started, the second that something has been cleaned up (which may not always be the case).

dmitri-d:
- rewrote free ip selection not to use ruby ranges.

modules/dhcp_common/free_ips.rb
timogoebel: My concern with this would be, that if you have a small subnet a user has no way to see why the IPs are taken and there is no way to "unlock" the IPs without restarting smart-proxy. Feels like trouble to me as well.

dmitri-d: Again, tinkering with the "CLEANUP_INTERVAL" would require understanding of internal logic, something we cannot demand from users. If 30 mins seems unreasonable, let's arrive at something that works better. We cannot make every parameter we unsure of our users' problem.

modules/dhcp_common/dhcp_common.rb
```diff
@@ -32,4 +32,12 @@ class NotImplemented < RuntimeError; end
   def self.ipv4_to_i(ipv4_address)
     ipv4_address.split('.', 4).inject(0) { |i, octet| (i << 8) | octet.to_i }
```
ekohl, line 33: Unrelated to this patch, but isn't this equal: ```IPAddr.new(ipv4_address).to_i```?

modules/dhcp_common/dhcp_common.rb
```diff
@@ -32,4 +32,12 @@ class NotImplemented < RuntimeError; end
...
   end
+
+  def self.i_to_ipv4(ipv4_as_i)
```
ekohl, line 36: I think it's possible to use the IPAddr module for this:
```IPAddr.new(ipv4_as_i, Socket::AF_INET).to_s```

File: modules/dhcp_common/free_ips.rb (hunk 4 above)
Comment (ekohl) [reply modules/dhcp_common/free_ips.rb]: Maybe it should take the free pool size into account? That way you don't require users to tinker with internal mechanics but can still handle pools with just a few IPs.

Comment (timogoebel) [reply modules/dhcp_common/free_ips.rb]: Well, I don't think we should expose these parameters to users. And I don't want to block this PR. However, I think a UI similar to foreman_dhcp_browser would be in order for Foreman core.

File: modules/dhcp_common/dhcp_common.rb (hunk 7 above)
Comment (dmitri-d) [reply modules/dhcp_common/dhcp_common.rb:33]: Probably, but I didn't want to instantiate an IPAddr object just to calculate an integer representation of an ip address. Moreover, this function is used someplace else already (it was originally added as a more efficient way to do this conversion, if I remember correctly), no reason not to use it.

File: modules/dhcp_common/dhcp_common.rb (hunk 8 above)
Comment (dmitri-d) [reply modules/dhcp_common/dhcp_common.rb:36]: Probably, but again, didn't want to instantiate an IPAddr just to do the conversion.

File: modules/dhcp_common/free_ips.rb (hunk 4 above)
Comment (dmitri-d) [reply modules/dhcp_common/free_ips.rb]: @ekohl: pool size size is probably not the metric that should determine the "CLEANUP_INTERVAL" -- it mainly depends on how long it takes a host complete pxe boot + kickstart (or an analogue of it). It's ok if the interval is shorter than the above though, the only side-effect would be an additional delay (due to a failing ping check) when an ip address recommendation is requested.

Comment (ekohl) [reply modules/dhcp_common/free_ips.rb]: @witlessbird I'd expect that once foreman starts to create the host it'll create a DHCP reservation which means it's no longer available. That'd imply you only need to keep it for as long as the user has the host form open, presses save and foreman creates the reservation.

Comment (dmitri-d) [reply modules/dhcp_common/free_ips.rb]: @ekohl: Yes, I'm quite sure this is what happens.

Comment (lzap) [reply modules/dhcp_common/free_ips.rb]: Sorry to be late in the discussion.

I am fine not exposing `DEFAULT_CLEANUP_INTERVAL` which you are right is little bit low level, but what is important to have the ability to change is the other one. Depending on many circumstances, your mileage will vary for `DEFAULT_ADDRESS_KEEP_TIME`. If my understanding is correct, this is how long an IP is "put aside" from free pool until user hits Submit button in the UI.

I work with enterprise customers and they sometimes use our project very different way we can't even think of. I can easily imagine someone needs to make a phone call in order to confirm Create host form and 30 minutes might not be enough. Other customer will be booting 250 hosts at one time and that setting is too much for them.

For completeness, I would like to see both as settings with explanation how to set things up. The default values are I think reasonable for regular foreman upstream user, that's fine.

File: modules/dhcp_common/dhcp_common.rb (hunk 8 above)
Comment (lzap) [reply modules/dhcp_common/dhcp_common.rb:36]: I don't mind any approach, but can you please add few unit tests for this method if you prefer the lower-level implementation? I know it is covered by high-level tests, but just to see regressions when we touch this code in the future.

File: modules/dhcp_common/free_ips.rb [hunk 9]
```diff
@@ -0,0 +1,157 @@
...
+          ip, _ = @allocation_timestamps.shift
+          @allocated_ips.delete(ip)
+          logger.debug("#{ip} marked as free.")
+        end
+      end
+    end
+
+    def start
+      logger.info("Starting allocated ip address maintenance (used by unused_ip call).")
+      @timer_task = Concurrent::TimerTask.new(:execution_interval => DEFAULT_CLEANUP_INTERVAL) { clean_up_allocated_ips }
```
lzap, line 42: Nitpick, in case of exception, will this thread (or whatever this is) survive? Shall we encapsulate this in rescue block putting down an error in logs and caring over?

modules/dhcp_common/free_ips.rb
```diff
@@ -0,0 +1,157 @@
...
+    def random_index(total_number_of_indices)
+      raise ArgumentError.new("Zero or negative number of indices") if total_number_of_indices <= 0
+      rng = Random.new(Time.now.to_i)
+      past_indices = Set.new
+      loop do
+        break if past_indices.size >= total_number_of_indices
+        current_index = rng.rand(total_number_of_indices)
+        next if past_indices.add?(current_index).nil?
+        yield(current_index)
+      end
```
lzap, line 100: Sidenote for reviewers: when reaching it's full capacity, bigger ranges could possibly spend a lot of CPU cycles in this loop. But this is ok, we are theoretically swapping memory efficiency for CPU efficiency (in practice this code is much better because creating huge array was basically a blocker).

Also it is worth nothing we are using PSEUDO RANDOM GENERATOR which is actually better than real random generator. PRNG produces numbers in cycles making sure the whole space is walked in sequence. In Ruby the Mersenne Twister algorithm is used. One nitpick - would be great to ensure the generator is really pseudo, so there won't be regression in future Ruby version. We could achieve this with a simple test going through whole space of let's say hundred of random numbers. Second, pseudo random generators won't block execution when OS don't have enough entropy, which is what we want to avoid.

test/dhcp/free_ips_test.rb
```diff
@@ -0,0 +1,108 @@
...
+    assert_equal "192.168.1.2", @free_ips.find_free_ip("192.168.1.1", "192.168.1.2", [r])
+  end
+
+  def test_find_free_ip_should_return_nil_when_no_addresses_available
+    @free_ips.stubs(:icmp_pingable?).returns(false)
+    @free_ips.stubs(:tcp_pingable?).returns(false)
+    r = Proxy::DHCP::Reservation.new('test', "192.168.1.1", "aa:bb:cc:dd:ee:ff", @subnet, :hostname =>'test')
+
+    assert_nil @free_ips.find_free_ip("192.168.1.1", "192.168.1.1", [r])
+  end
```
lzap: I'd also add a edge case when all addresses are blacklisted, just to make sure. Nitpick: no space in `hostname => 'test'`.

lzap (changes requested):
Ok well done, thanks. There is one nitpick. Please consider extracting the setting(s) but if you insist this must be not configurable, I am fine taking this as-is.

modules/dhcp_common/dhcp_common.rb
ekohl: I'd prefer the built in method because it's less code to write and test.

dmitri-d:
- replaced DEFAULT_ADDRESS_KEEP_TIME constant with a parameter (pls. see :blacklist_duration_minutes config option for dhcp_isc provider)
- added tests for ipv4_to_i and i_to_ipv4 methods
- added a test for unused_ip call when the address pool is exhausted

modules/dhcp_common/free_ips.rb
dmitri-d: TimerTask catches all exceptions and continues/restarts the task should one occur.

timogoebel:
@lzap: Are you going to finish reviewing this?

lzap:
Apologies, absolutely. Got busy downstsream a have a huge backlog on github. Doing right now.

modules/dhcp_common/dhcp_common.rb
lzap: If there is a test down below, I am fine. This is being called many times down there and ruby sucks in memory management at the moment.

modules/dhcp_common/free_ips.rb
```diff
@@ -0,0 +1,158 @@
+require 'concurrent'
+require 'logger'
+require 'set'
+
+module Proxy::DHCP
+  class FreeIps
+    include ::Proxy::Log
+
+    DEFAULT_ADDRESS_KEEP_TIME = 30*60 # 30 mins
```
lzap: I only have one tiny nitpick - this constant does only seem to be used in tests but in the default value you don't use it. Is this intentional? Might be some leftover.

lzap (approved):
Once you rebase/confirm we are good to go. Thanks.

modules/dhcp_common/dhcp_common.rb
dmitri-d: there is: https://github.com/theforeman/smart-proxy/pull/543/files#diff-ea91ddc372683c1cdd90b093548992c2R12

modules/dhcp_common/free_ips.rb
dmitri-d: Removed.

dmitri-d:
* rebased.

lzap:
Sorry for the delay, I have huge backlog. Tests are green locally, not sure what happened on CI. Merging.

@witlessbird what are the chances of getting this into 1.16/1.15? It's quite a change.

PR #545, opened 2017-10-09
Fixes #21249 - Check MS DHCP range versus start/end address 

achevalet:
Tests done:

MS DHCP range: 10.3.28.100-10.3.28.150

Foreman range: 50-99
WARN -- : End of IP range 10.3.28.99 is out of DHCP range (10.3.28.100-10.3.28.150)

Foreman range: 50-100
DEBUG -- : Start of IP range 10.3.28.50 is out of DHCP range (10.3.28.100-10.3.28.150), using 10.3.28.100

Foreman range: 50-150
DEBUG -- : Start of IP range 10.3.28.50 is out of DHCP range (10.3.28.100-10.3.28.150), using 10.3.28.100

Foreman range: 100-200
DEBUG -- : End of IP range 10.3.28.150 is out of DHCP range (10.3.28.100-10.3.28.150), using 10.3.28.150

Foreman range: 50-200
DEBUG -- : Start of IP range 10.3.28.50 is out of DHCP range (10.3.28.100-10.3.28.150), using 10.3.28.100
DEBUG -- : End of IP range 10.3.28.200 is out of DHCP range (10.3.28.100-10.3.28.150), using 10.3.28.150

### DIFF
--- a/modules/dhcp_native_ms/dhcp_native_ms_main.rb
+++ b/modules/dhcp_native_ms/dhcp_native_ms_main.rb
@@ -78,6 +78,36 @@ module Proxy::DHCP::NativeMS
         raise Proxy::DHCP::NotImplemented.new("DhcpsApi::Server#get_free_ip_address is not available on Windows Server 2008 and earlier versions.")
       end
 
+      dhcp_range = dhcpsapi.list_subnet_elements(subnet_address, DhcpsApi::DHCP_SUBNET_ELEMENT_TYPE::DhcpIpRanges) rescue nil
+      if dhcp_range.nil?
+        logger.warn "The specified subnet #{subnet_address} does not exist on the DHCP server"
+        return nil
+      end
+      dhcp_start_addr = dhcp_range.map {|r| r[:element][:start_address]}.first
+      dhcp_end_addr = dhcp_range.map {|r| r[:element][:end_address]}.first
+      if dhcp_start_addr.nil? || dhcp_end_addr.nil?
+        logger.warn "Unable to find IP range on the DHCP server for subnet #{subnet_address}"
+        return nil
+      end
+
+      if from_address.nil? || to_address.nil?
+        from_address = dhcp_start_addr
+        to_address = dhcp_end_addr
+      else
+        foreman_start_addr_ip = IPAddr.new(from_address)
+        foreman_end_addr_ip = IPAddr.new(to_address)
+        dhcp_start_addr_ip = IPAddr.new(dhcp_start_addr)
+        dhcp_end_addr_ip = IPAddr.new(dhcp_end_addr)
+        if not (dhcp_start_addr_ip..dhcp_end_addr_ip).cover? foreman_start_addr_ip
+          logger.warn "Start of IP range #{from_address} is out of DHCP range (#{dhcp_start_addr}-#{dhcp_end_addr})"
+          return nil
+        end
+        if not (dhcp_start_addr_ip..dhcp_end_addr_ip).cover? foreman_end_addr_ip
+          logger.warn "End of IP range #{to_address} is out of DHCP range (#{dhcp_start_addr}-#{dhcp_end_addr})"
+          return nil
+        end
+      end
+
       return dhcpsapi.get_free_ip_address(subnet_address, from_address, to_address).first
     end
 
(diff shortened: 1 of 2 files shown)


ekohl:
I wonder if this should be implemented at a higher level so all DHCP providers can benefit from this validation.

dmitri-d:
@ekohl: There's already code to validate subnet ranges in a general case. This change is ms-dhcp specific.

modules/dhcp_native_ms/dhcp_native_ms_main.rb
```diff
@@ -78,6 +78,41 @@ def unused_ip(subnet_address, mac_address, from_address, to_address)
         raise Proxy::DHCP::NotImplemented.new("DhcpsApi::Server#get_free_ip_address is not available on Windows Server 2008 and earlier versions.")
       end
 
+      dhcp_range = dhcpsapi.list_subnet_elements(subnet_address, DhcpsApi::DHCP_SUBNET_ELEMENT_TYPE::DhcpIpRanges)
+      dhcp_start_addr = dhcp_range.map {|r| r[:element][:start_address]}.first
```
dmitri-d: It's possible to create a subnet without a subnet range, so both dhcp_start_addr and dhcp_end_addr can come back as nils.

modules/dhcp_native_ms/dhcp_native_ms_main.rb
```diff
@@ -78,6 +78,41 @@ def unused_ip(subnet_address, mac_address, from_address, to_address)
...
+      dhcp_end_addr = dhcp_range.map {|r| r[:element][:end_address]}.first
+
+      if from_address.to_s.empty?
```
dmitri-d: from_address and to_address will be nils if the caller skipped these parameters. ```from_address.nil?``` is better check in this case.

modules/dhcp_native_ms/dhcp_native_ms_main.rb
```diff
@@ -78,6 +78,41 @@ def unused_ip(subnet_address, mac_address, from_address, to_address)
...
+        dhcp_start_addr_ip = IPAddr.new(dhcp_start_addr, Socket::AF_INET)
+      end
+      if to_address.to_s.empty?
+        to_address = dhcp_end_addr
+      else
+        foreman_end_addr_ip = IPAddr.new(to_address, Socket::AF_INET)
+        dhcp_end_addr_ip = IPAddr.new(dhcp_end_addr, Socket::AF_INET)
+      end
+
+      if foreman_start_addr_ip.to_i > dhcp_end_addr_ip.to_i
```
dmitri-d: There's a PR open (which should be merged in soon) that introduces a dedicated validator for range checks (https://github.com/theforeman/smart-proxy/pull/543/files#diff-42cb95eb87ed37db7fd2fc99a3064f06R67). I would suggest using it instead.

dmitri-d (changes requested):
Could you also add test(s) to verify added functionality please?

modules/dhcp_native_ms/dhcp_native_ms_main.rb
achevalet: Well, this is the range on the MS DHCP, afaik we can't create a dhcp scope without a range?

modules/dhcp_native_ms/dhcp_native_ms_main.rb
achevalet: Right, I will update it.

modules/dhcp_native_ms/dhcp_native_ms_main.rb
achevalet: There are 2 different use cases here: 
- foreman range is out of dhcp scope = warning message and exit
- foreman range is out of dhcp scope but dhcp scope is inside foreman range, in that case we just update the from/to to match the dhcp scope and we continue.
Not sure how I can use the validate_subnet_range for that cases.

achevalet:
I have no clue how tests work, any help would be much appreciated!

dmitri-d:
Please see my comment re: tests in your other PR.

modules/dhcp_native_ms/dhcp_native_ms_main.rb
dmitri-d: I think it's possible when api is used to create a subnet.

modules/dhcp_native_ms/dhcp_native_ms_main.rb
dmitri-d: Now that I think of it, I'd say treat #2 as an error and let the user to correct the data/scope (as it may have impact on other parts of Foreman) instead of massaging it.

achevalet:
@witlessbird I've amended the tests, please let me know if that's ok.

dmitri-d:
Merged, thanks @achevalet!

PR #548, opened 2017-10-25
fixes #21463 - new reservations are deletable

timogoebel:
This should make smart proxy detect that a record already exists when the record to be created does not have `options[:deleteable]` set.

### DIFF
--- a/modules/dhcp_common/record/reservation.rb
+++ b/modules/dhcp_common/record/reservation.rb
@@ -7,7 +7,7 @@ module Proxy::DHCP
     def initialize(name, ip_address, mac_address, subnet, options = {})
       @type = "reservation"
       @name = name
-      super(ip_address, mac_address, subnet, options)
+      super(ip_address, mac_address, subnet, {:deleteable => true}.merge(options))
     end
 
     def to_s
--- a/modules/dhcp_common/isc/subnet_service_initialization.rb
+++ b/modules/dhcp_common/isc/subnet_service_initialization.rb
@@ -93,7 +93,12 @@ module Proxy::DHCP::CommonISC
         all
       end
 
-      options_and_attributes = parsed_host.dhcp_options.inject({:hostname => name}.merge(attributes_only)) do |all, current|
+      default_attributes = {
+        :hostname => name,
+        :deleteable => false
+      }
+
+      options_and_attributes = parsed_host.dhcp_options.inject(default_attributes.merge(attributes_only)) do |all, current|
         key, values = process_dhcpd_option(current[0], current[1])
         all[key] = values
         all
--- a/modules/dhcp_common/record/lease.rb
+++ b/modules/dhcp_common/record/lease.rb
@@ -13,7 +13,7 @@ module Proxy::DHCP
       super(ip_address, mac_address, subnet, options)
     end
 
-    def deletable?
+    def deleteable?
       false
     end
 
--- a/test/dhcp/dhcp_api_test.rb
+++ b/test/dhcp/dhcp_api_test.rb
@@ -151,12 +151,13 @@ class DhcpApiTest < Test::Unit::TestCase
 
     assert last_response.ok?, "Last response was not ok: #{last_response.status} #{last_response.body}"
     expected = [{
-      "type"     => "reservation",
-      "hostname" => "test.example.com",
-      "ip"       => "192.168.122.1",
-      "mac"      => "00:11:bb:cc:dd:ee",
-      "name"     => 'test.example.com',
-      "subnet"   => "192.168.122.0/255.255.255.0"
+      "deleteable" => true,
+      "type"       => "reservation",
+      "hostname"   => "test.example.com",
+      "ip"         => "192.168.122.1",
+      "mac"        => "00:11:bb:cc:dd:ee",
+      "name"       => 'test.example.com',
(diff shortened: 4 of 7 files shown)


modules/dhcp_common/record.rb
```diff
@@ -45,7 +45,7 @@ def eql?(other)
     end
 
     def ==(other)
-      !other.nil? && self.class == other.class && ip == other.ip && mac == other.mac && subnet == other.subnet && deleteable? == other.deleteable? && options == other.options
+      !other.nil? && self.class == other.class && ip == other.ip && mac == other.mac && subnet == other.subnet && deleteable? == other.deleteable? && options.tap { |option| option.delete(:deleteable) } == other.options.tap { |option| option.delete(:deleteable) }
```
dmitri-d: This mutates both records, which it shouldn't (options hash will not contain :deleteable key if it was there before the comparison). Any reason you didn't use ```options[:deleteable] == other.options[:deleteable]``` instead?

This raises another question -- what options should be taken into account in equality comparison? For example, filename, nextserver, hostname, and a bunch of vendor-specfic options are used during record creation and were used for equality comparisons before this change.

modules/dhcp_common/server.rb
```diff
@@ -173,7 +173,10 @@ def clean_up_add_record_parameters(in_options)
       to_return.delete("subnet") # Not a valid key; remove it to prevent conflict with :subnet
       subnet = to_return.delete("network")
 
-      [name || hostname, ip, mac, subnet, to_return.merge!(:hostname => hostname || name)]
+      deleteable = to_return.delete("deleteable")
```
dmitri-d: I think a better approach would be to initialize ::Proxy::DHCP::Reservation with options hash containing ```:deleteable => true```.

modules/dhcp_common/record.rb
dmitri-d: Considering that we assume that reservations are always deleteable, and if we make them so by default (pls. see my comment below), I don't think this change is needed at all.

dmitri-d (changes requested):

timogoebel:
@witlessbird: Thanks. I think I addressed all your comments. All reservations now default to being deletable.

modules/dhcp_common/record/reservation.rb
```diff
@@ -7,6 +7,7 @@ class Reservation < Record
     def initialize(name, ip_address, mac_address, subnet, options = {})
       @type = "reservation"
       @name = name
+      options[:deleteable] = true unless options.key?(:deleteable)
       super(ip_address, mac_address, subnet, options)
```
dmitri-d: You could avoid modifying the options parameter by doing something like:
```
super(ip_address, mac_address, subnet, {:deleteable => true}.merge(options))
```

dmitri-d:
There's an issue specific to isc dhcpd: all reservations are now deletable by default, while in actuality only "dynamic" ones are. 

One of the ways to handle this is to change how Proxy::DHCP::Reservation options are initialized. Adding :deleteable => false to the initial hash here: https://github.com/theforeman/smart-proxy/blob/develop/modules/dhcp_common/isc/subnet_service_initialization.rb#L96 would resolve this -- dynamic reservations have :deleteable key set to true.

timogoebel:
@witlessbird: Thanks for the hint. Added the missing default parameter and some tests to cover that.

modules/dhcp_common/record/reservation.rb
dmitri-d: Not sure if you missed this -- could you change how "deleteable" option is initialized?

dmitri-d:
Could fix the tests please?

timogoebel:
@witlessbird: Sorry for the delay. Fixed the tests.

dmitri-d:
[test]

dmitri-d:
Could you update the constructor here: https://github.com/theforeman/smart-proxy/pull/548/files#diff-86d12e513b5cb76326d7322cd5aa0f35R10 to not modify the parameters that passed into it?

dmitri-d:
[test]

timogoebel:
@witlessbird: Omg, what am I doing here? Addressed your comment, sorry. I missed that. Twice. :-/

dmitri-d:
Merged, thanks @timogoebel!

timogoebel:
@witlessbird: Thanks for your help with this.
Do you want to put this on 1.16 or 1.17?

dmitri-d:
@timogoebel: This has a potential to break stuff; unless someone specifically asked for this fix in the stable branch, I'd like to keep it out of it.

modules/dhcp_common/record/reservation.rb
```diff
@@ -7,7 +7,7 @@ class Reservation < Record
     def initialize(name, ip_address, mac_address, subnet, options = {})
       @type = "reservation"
       @name = name
-      super(ip_address, mac_address, subnet, options)
+      super(ip_address, mac_address, subnet, {:deleteable => true}.merge(options))
```
lzap, line 10: I don't know why but this line created regression from 1.16 which I reported here: http://projects.theforeman.org/issues/21975

lzap:
Indeed, in develop "filename" options are no longer stored and completely ignored. Not sure why, can you guys take a look please?

PR #549, opened 2017-11-06
fixes #21581 - tests run without bundler development group

timogoebel:


bundler.d/test.rb
```diff
@@ -7,6 +7,7 @@
   gem 'test-unit' if RUBY_VERSION > "1.8.7"
   gem 'addressable', '~> 2.3.8' if RUBY_VERSION == '1.8.7' # 2.4.0 drops support for ruby 1.8.7
   gem 'public_suffix', '< 3' if RUBY_VERSION < '2.1'
+  gem 'benchmark-ips'
```
dmitri-d, line 10: I think in this case ruby-prof (used together with benchmark-ips) and rubocop should be moved into the "test" group. Alternatively, why not use both "development" and "test" groups for jenkins CI runs?

timogoebel: Sure, that'd be fine with me as well. Actually, I'm wondering why this is necessary at all. The `bundle install --without=development` has been there for ages on Jenkins. Any idea?

timogoebel: Ok, I see it now. Makes sense.
To be honest, I think I prefer this solution. I'll move the the two gems in a sec.

dmitri-d: ok, sounds good.

timogoebel:
Started the [CI tests manually](http://ci.theforeman.org/job/test_proxy_develop_pull_request/1584/), tests are 💚

dmitri-d:
Merged, thanks @timogoebel!

PR #550, opened 2017-11-08
Fixes #21605 - make authentication extendable

ares:


ekohl:
Is there a risk that ```@authenticated``` is reused on subsequent requests from other clients?

ares:
It seemed it's nil for first auth method with every request, the before block should be called in fresh context every time I'd say. Or am I missing something?

dmitri-d:
I need to double check, but chances are the blocks passed into the ```before``` are executed in the context of the controller instance.

Having said that:
 - I don't think introducing a state into helpers is a good idea
 - This change breaks current authorization logic: for example puppet_proxy controller, when accessed over https requires that both ssl cert to be present *and* client recognized. With this change passing either check is sufficient.

It's not clear to me why this change is required, could you give more details please, @ares? Could you implement your own authentication helpers in REX plugin?

ares:
> I don't think introducing a state into helpers is a good idea

it's not, but with current code it's the only easy option I see, these methods should not call `before` in the first place, there should be one method driving the authentication flow that would be executed in context of existing request so it could decide what authentication methods are applicable

> This change breaks current authorization logic: for example puppet_proxy controller, when accessed over https requires that both ssl cert to be present and client recognized. With this change passing either check is sufficient.

that would be a showstopper for this change I guess, is this the only controller? maybe that should be a different authentication method in such case

> It's not clear to me why this change is required, could you give more details please, @ares? Could you implement your own authentication helpers in REX plugin?

The problem is that if we want to reuse existing authentication methods from smart-proxy and avoid copying the code, every SSL request will fail if there's no client certificate specified. See https://github.com/theforeman/smart_proxy_dynflow/pull/28 for details. The smart proxy dynflow core can either run as part of smart proxy, then we want to use proxy built-in authentication methods and also allow token based auth. If we want to do token one through SSL, it would always fail on missing client cert. When dynflow core is running as a separate service, we use our own authentication method which is slightly different from proxy one)

lib/sinatra/authorization.rb
```diff
@@ -18,7 +19,9 @@ def authorize_with_trusted_hosts
           end
           fqdn = fqdn.downcase
 
-          unless Proxy::SETTINGS.trusted_hosts.include?(fqdn)
+          if Proxy::SETTINGS.trusted_hosts.include?(fqdn)
```
ekohl, line 22: Looks like this can already use the local variable ```trusted_hosts``` that's set at line 10.

ekohl (commented):
I think that splitting the actual checks from the configuration into pure functions that return booleans instead of ```log_halt``` is a better approach. Then you can introduce settings to choose the authentication methods.

dmitri-d:
> that would be a showstopper for this change I guess, is this the only controller? maybe that should be a different authentication method in such case

These are used in a quite a few controllers (majority in the core?).

I'd agree with @ekohl -- extracting authentication logic into methods and modifying existing helpers to use newly extracted methods is probably the way to go. You can then reuse authentication logic as needed.

ares:
Just to set clear expectations, while I agree this is suboptimal, I don't have capacity to work on suggested improvement. I think this allows us to fix the issue described in https://github.com/theforeman/smart_proxy_dynflow/pull/28 but I don't insist, if there's a strong opinion it shouldn't be merged, please close.

lzap:
Agreed that this part needs revamp, this smells bad.

PR #552, opened 2017-11-21
Fixes #21728 - pinned rack and rack-test for rubies < 2.2

dmitri-d:


iNecas (approved):

iNecas:
Makes sense, thanks

iNecas:
I guess this should be cherrypicjed to 1.15 and 1.16 as well

PR #554, opened 2017-11-23
Fixes #15535 - http ports used by plugins are being logged now.

dmitri-d:


lib/proxy/plugin_initializer.rb
```diff
@@ -290,6 +301,16 @@ def log_used_settings(settings)
     sorted_keys = settings.keys.map(&:to_s).sort.map(&:to_sym) # ruby 1.8.7 doesn't support sorting of symbols
     to_log = sorted_keys.map {|k| "'%s': %s%s" % [k, settings[k], (default_settings.include?(k) && default_settings[k] == settings[k]) ? " (default)" : ""] }.join(", ")
     logger.debug "'%s' settings: %s" % [plugin.plugin_name, to_log]
+    logger.debug("'%s' ports: 'http': %s, 'https': %s" % [plugin.plugin_name,
```
iNecas: What's the added value of logging this? We already have the enabled option logged in the logs themselves. Or am I missing something here?

dmitri-d: I had people ask me about this. If anything, this is more of a clarification.

lib/proxy/plugin_initializer.rb
```diff
@@ -290,6 +301,16 @@ def log_used_settings(settings)
...
+                                                          ::Proxy::PluginGroup::HTTP_ENABLED.include?(settings[:enabled]),
+                                                          ::Proxy::PluginGroup::HTTPS_ENABLED.include?(settings[:enabled])])
+  end
+
+  def log_provider_settings(settings)
+    default_settings = plugin.plugin_default_settings
+    sorted_keys = settings.keys.map(&:to_s).sort.map(&:to_sym) # ruby 1.8.7 doesn't support sorting of symbols
+    to_log = sorted_keys.map {|k| "'%s': %s%s" % [k, settings[k], (default_settings.include?(k) && default_settings[k] == settings[k]) ? " (default)" : ""] }.join(", ")
```
iNecas: It looks like there is a lot of code duplicate between this method and `log_used_settings`, seems like a good candidate to extract the code to some common place

dmitri-d:
- Removed duplication in settings logging code.

lib/proxy/plugin_initializer.rb
```diff
@@ -238,7 +241,15 @@ def start_services(services, container)
 end
 
 module ::Proxy::DefaultSettingsLoader
-  def load_settings(main_plugin_settings = {})
+  def load_plugin_settings
+    load_settings({}, :log_used_settings)
```
iNecas: What about naming `log_plugin_settings` vs. `log_user_settings`.

lib/proxy/plugin_initializer.rb
```diff
@@ -254,7 +265,7 @@ def load_settings(main_plugin_settings = {})
 
     plugin.settings = ::Proxy::Settings::Plugin.new({}, settings)
 
-    log_used_settings(settings)
+    send(how_to_log, settings)
```
iNecas: Wouldn't `yield if block_given` and mapping the log statements via `block` be better here?

iNecas:
Tested and work well. Last to suggestions inline

lib/proxy/plugin_initializer.rb
dmitri-d: The need for slightly different logging  reflects one of the differences between plugins and providers -- the latter do not provide public api. Not sure what the rationale behind `log_user_settings` name is...

lib/proxy/plugin_initializer.rb
dmitri-d: Replaced with a block.

lib/proxy/plugin_initializer.rb
iNecas: Sry, typo, I was suggesting renaming `log_used_settings` to `log_plugin_settings`

iNecas: This would mirror the `load_plugin_settings`/`load_provider_settings` convention

dmitri-d: `log_used_settings` is/can be overridden by modules/provides, I'd like to keep the name as is.

iNecas (approved):

iNecas:
Thanks @witlessbird

PR #555, opened 2017-12-14
Fixes #21975 - filename dhcp option is accepted now

dmitri-d:


modules/dhcp_common/server.rb
```diff
@@ -166,21 +167,24 @@ def find_similar_records(subnet_address, ip_address, mac_address)
...
 
-      name = to_return.delete("name")
-      hostname = to_return.delete("hostname")
+      options.delete("subnet") # Not a valid key; remove it to prevent conflict with :subnet
+      subnet = options.delete("network")
 
-      to_return.delete("subnet") # Not a valid key; remove it to prevent conflict with :subnet
-      subnet = to_return.delete("network")
+      # convert string keys to symbols
+      options = options.inject({}){|all, current| all[current[0].to_sym] = current[1]; all}
```
timogoebel, line 185: If I don't misread the diff, this line is the only relevant change.
Wouldn't it be cleaner to first convert all the keys to symbols and then extract the relevant data (ip, mac, name, hostname, ...)?

The ticket states, this was related to #548 - is it?

dmitri-d: > Wouldn't it be cleaner to first convert all the keys to symbols and then extract the relevant data (ip, mac, name, hostname, ...)?

I don't think it makes any difference, tbh.

> The ticket states, this was related to #548 - is it?

I don't think so.

dmitri-d:
* fixed rubocop warnings.

modules/dhcp_common/server.rb
lzap: Well, I was doing git bisect and it pointed me on that commit. I *think* the issue was that it made additional copy (merge call) instead passing in as reference.

Anyway, I think we should be defensive in this method and accept keys both as strings and symbols - I'd also prefer "convert first, use symbols everywhere" approach to be honest, possibly can be slower during DoS but that should not be huge problem.

lzap (changes requested):
Thanks for looking into this, I only have one concern - we don't have a test for "filename" option, can you add one to make sure it won't get cleared out in the future?

dmitri-d:
> I only have one concern - we don't have a test for "filename" option

We have: https://github.com/theforeman/smart-proxy/blob/develop/test/dhcp/isc_omapi_provider_test.rb#L45

The tests I added verify that `clean_up_add_record_parameters` works as expected, which includes conversion of string keys to symbols -- the bug affected *all* options, "filename" isn't special in this regard.

dLobatog (commented):
@lzap @witlessbird Any reason this could not go in 1.16.1 directly? It ties in with https://github.com/theforeman/foreman/pull/5085 pretty well.

lzap:
@dLobatog 1.16 is not affected by this, only develop.

Well, I see the test now thanks, I am merging.

lzap (approved):

PR #556, opened 2017-12-19
Fixes #19053 - rotate logs on SIGHUP

lzap:
This is an anternative solution to the SELinux logrotate problem. SIGHUP is allowed by the default policy, so I am making smart-proxy to respond to that signal as expected and leaving SIGUSR1 behavior untouched for backward compatibility. This allows me to change the logrotate script to send HUP instead SIG1 which will just work.

Replaces https://github.com/theforeman/smart-proxy/pull/544

As part of this effort I will also add `delaycompress` to the logrotate script to solve missing lines during rotation.

Edit: Moved the issue to the project.

### DIFF
--- a/lib/proxy/signal_handler.rb
+++ b/lib/proxy/signal_handler.rb
@@ -8,7 +8,10 @@ class Proxy::SignalHandler
     handler.install_ttin_trap unless RUBY_PLATFORM =~ /mingw/
     handler.install_int_trap
     handler.install_term_trap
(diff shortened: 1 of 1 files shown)


ekohl (approved):
Looks like the easiest implementation and there's no specified behavior on what should happen on SIGHUP so this falls within the spec.

lzap:
Hmmm after chat with SELinux team, it looks like they confirmed a bug in RHEL policy and we might get proper fix and workaround for Foreman: https://bugzilla.redhat.com/show_bug.cgi?id=1527522

PR #557, opened 2017-12-29
fixes #4917 - add active directory realm provider

timogoebel:
This is an attempt to add active directory as a core realm provider to smart-proxy.

This is a port of [smart_proxy_realm_ad_plugin](https://github.com/martencassel/smart_proxy_realm_ad_plugin).

To interact with active directory, this used the [radcli gem](https://github.com/martencassel/radcli). The gem is not available on rubygems for now and should probably be reviewed as well. I have no real knowledge of the C programming language, so I can't say anything about it. We might want to consider moving it to the foreman org and maintaining it there. But that's just a bold suggestion :-)

I'm opening this for discussion. Please let me know your thoughts.

cc: @martencassel

### DIFF
--- a/bundler.d/realm_ad.rb
+++ b/bundler.d/realm_ad.rb
@@ -1,0 +1,3 @@
+group :realm_ad do
+  gem 'radcli'
+end
--- a/config/settings.d/realm.yml.example
+++ b/config/settings.d/realm.yml.example
@@ -4,4 +4,5 @@
 
 # Available providers:
 #   realm_freeipa
+#   realm_ad
 :use_provider: realm_freeipa
--- a/config/settings.d/realm_ad.yml.example
+++ b/config/settings.d/realm_ad.yml.example
@@ -1,0 +1,24 @@
+---
+# Authentication for Kerberos-based Realms
+:realm: EXAMPLE.COM
+
+# Kerberos pricipal used to authenticate against Active Directory
+:principal: realm-proxy@EXAMPLE.COM
+
+# Path to the keytab used to authenticate against Active Directory
+:keytab_path:  /etc/foreman-proxy/realm_ad.keytab
+
+# FQDN of the Domain Controller
+:domain_controller: dc.example.com
+
+# Optional: OU where the machine account shall be placed
+#:ou: OU=Linux,OU=Servers,DC=example,DC=com
+
+# Optional: Prefix for the computername
+#:computername_prefix: ''
+
+# Optional: Generate the computername by calculating the SHA256 hexdigest of the hostname
+#:computername_hash: false
+
+# Optional:  use the fqdn of the host to generate the computername
+#:computername_use_fqdn: false
--- a/lib/smart_proxy_main.rb
+++ b/lib/smart_proxy_main.rb
@@ -79,6 +79,7 @@ module Proxy
   require 'bmc/bmc'
   require 'realm/realm'
   require 'realm_freeipa/realm_freeipa'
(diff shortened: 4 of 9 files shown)


martencassel:
The plugin was first started as a pull request: https://github.com/theforeman/smart-proxy/pull/480
but i've moved it out into a plugin project.
 
It was decided at that time to have this realm provider as a plugin outside smart-proxy. You have to maintain this code and the dependent radcli ruby extension.

 cc: @witlessbird

lzap:
We already took ownership of a rubygem with C extension (https://github.com/theforeman/ruby-newt) so helping with maintaining is not bold suggestion at all, that is fine. What's actually bold is making this a core feature - the flexibility of deciding not to install C code without active maintainer is key. We had couple of ugly memory leaks caused by C extension (openscap), the same happened with newt (not a smart-proxy feature tho but same problems). There are not many of us with deep C knowledge and it was a pain to troubleshoot those, so ability to decide not to ship a plugin is important.

I would like to suggest this to be a plugin for better flexibility. This is highly requested feature and I am pretty sure that the plugin will shortly become a core plugin, thus I suggest creating theforeman github repos for both plugin and the dependency. I have to say that this is more downstream discussion, but we are transparent - ability to ship one another downstream release without this plugin until we are sure it is stable enough would be great - @ohadlevy @ares @iNecas

timogoebel:
> I suggest creating theforeman github repos for both plugin and the dependency

Thanks for your comment, @lzap.
This approach would be fine for me as well. Just to clarify: I don't insist on this plugin being in smart-proxy core, I just want to move things forward and make it easy for users to actually use this (e.g. packaging, CI tests, puppet module support, the whole package). This wasn't trivial to find or use, which is a pity as @martencassel probably invested a lot of time in this and this is a really useful feature if you have AD around.

martencassel:
I agree, its very hard to find and start using this plugin solution.
 I'm currently working to improve this and for a starter i've published the radcli library and the plugin as a rubygem.  We need a guide on how to set this up.

dmitri-d:
Please keep this module (this also applies to all new modules) out of smart-proxy core. 

@martencassel: if you'd like we can chat about moving your provider into theforeman github organization. Regardless of whether AD provider is or isn't under theforeman org, it should be listed on foreman wiki (if it isn't already). I'll update the page tomorrow.

martencassel:
@witlessbird sounds good, we can do that.

timogoebel:
> Please keep this module (this also applies to all new modules) out of smart-proxy core.

@witlessbird: I do agree, that this policy makes sense. Should we add this to the README in this repo so it's easier to find?

> for a starter i've published the radcli library and the plugin as a rubygem

This is a great start. I'll open rpm packaging PRs tomorrow and a PR for the puppet module. I basically have the code ready.

timogoebel:
I opened pull requests for packaging and the puppet modules (https://github.com/theforeman/puppet-foreman_proxy/pull/396).

ekohl:
I agree this doesn't have to be in core but the packaging and installer changes are great. I also wouldn't mind moving the plugin to the foreman namespace but that's up to @martencassel.

martencassel:
It's ok with me to move my plugin repo into the foreman namespace.

dmitri-d:
@martencassel: I'm in PST timezone now, will try to catch you online tomorrow morning.

dmitri-d:
Closing the PR, as the smart_proxy_realm_ad_plugin repository has been moved under theforeman github organization.

PR #561, opened 2018-01-16
Fixes #22292 - Config option for tftp timeout

parthaa:
Smart proxy TFTP plugin uses wget to download initrd, vmlinuz,
squash.img etc. It by default uses a connection/dns and read timeout of 10 seconds.

This commit enables one to the set the individual time outs via config
options. If these options are not set the code assumes the default 10s.

### DIFF
--- a/config/settings.d/tftp.yml.example
+++ b/config/settings.d/tftp.yml.example
@@ -5,3 +5,15 @@
 #:tftproot: /var/lib/tftpboot
 # Defines the TFTP Servername to use, overrides the name in the subnet declaration
 #:tftp_servername: tftp.domain.com
+
+# Defines the default read timeout in seconds needed to download tftp artifacts
+# like initrd and vmlinuz. Default value 60 seconds
+#:tftp_read_timeout: 60
+
+# Defines the default connection timeout in seconds needed to download tftp artifacts
+# like initrd and vmlinuz. Default value 10 seconds
+#:tftp_connect_timeout: 10
+
+# Defines the default dns timeout in seconds needed to download tftp artifacts
+# like initrd and vmlinuz. Default value 10 seconds
+#:tftp_dns_timeout: 10
--- a/lib/proxy/http_download.rb
+++ b/lib/proxy/http_download.rb
@@ -3,11 +3,19 @@ require 'proxy/file_lock'
 module Proxy
   class HttpDownload < Proxy::Util::CommandTask
     include Util
+    DEFAULT_READ_TIMEOUT = 60
+    DEFAULT_CONNECT_TIMEOUT = 10
+    DEFAULT_DNS_TIMEOUT = 10
 
-    def initialize(src, dst)
+    def initialize(src, dst, read_timeout = nil, connect_timeout = nil, dns_timeout = nil)
       @dst = dst
       wget = which("wget")
-      super("#{wget} --timeout=10 --tries=3 --no-check-certificate -nv -c \"#{escape_for_shell(src.to_s)}\" -O \"#{escape_for_shell(dst.to_s)}\"")
+      read_timeout ||= DEFAULT_READ_TIMEOUT
+      dns_timeout ||= DEFAULT_CONNECT_TIMEOUT
+      connect_timeout ||= DEFAULT_DNS_TIMEOUT
+
+      super("#{wget} --connect-timeout=#{connect_timeout} --dns-timeout=#{dns_timeout} --read-timeout=#{read_timeout}"\
+            " --tries=3 --no-check-certificate -nv -c \"#{escape_for_shell(src.to_s)}\" -O \"#{escape_for_shell(dst.to_s)}\"")
     end
 
     def start
--- a/modules/tftp/server.rb
+++ b/modules/tftp/server.rb
@@ -153,7 +153,11 @@ module Proxy::TFTP
   def self.choose_protocol_and_fetch(src, destination)
     case URI(src).scheme
     when 'http', 'https', 'ftp'
-      ::Proxy::HttpDownload.new(src.to_s, destination.to_s).start
+      ::Proxy::HttpDownload.new(src.to_s,
+                                destination.to_s,
+                                Proxy::TFTP::Plugin.settings.tftp_read_timeout,
+                                Proxy::TFTP::Plugin.settings.tftp_connect_timeout,
+                                Proxy::TFTP::Plugin.settings.tftp_dns_timeout).start
     when 'nfs'
       logger.debug "NFS as a protocol for installation medium detected."
     else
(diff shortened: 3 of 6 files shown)


lib/proxy/http_download.rb
```diff
@@ -4,10 +4,14 @@ module Proxy
   class HttpDownload < Proxy::Util::CommandTask
     include Util
 
-    def initialize(src, dst)
+    def initialize(src, dst, options = {})
       @dst = dst
       wget = which("wget")
-      super("#{wget} --timeout=10 --tries=3 --no-check-certificate -nv -c \"#{escape_for_shell(src.to_s)}\" -O \"#{escape_for_shell(dst.to_s)}\"")
+      read_timeout = options[:read_timeout] || 10
```
dmitri-d: It's only three more parameters, probably no need for using a hash, esp. when the other two parameters are explicit.

modules/tftp/server.rb
```diff
@@ -153,7 +153,12 @@ def self.fetch_boot_file dst, src
   def self.choose_protocol_and_fetch(src, destination)
     case URI(src).scheme
     when 'http', 'https', 'ftp'
-      ::Proxy::HttpDownload.new(src.to_s, destination.to_s).start
+      options = {
+        :read_timeout => Proxy::TFTP::Plugin.settings.read_timeout || 10,
```
dmitri-d: Default values are defined in ```HttpDownload``` already, no need to define them here too?

modules/tftp/server.rb
```diff
@@ -153,7 +153,12 @@ def self.fetch_boot_file dst, src
   def self.choose_protocol_and_fetch(src, destination)
     case URI(src).scheme
     when 'http', 'https', 'ftp'
-      ::Proxy::HttpDownload.new(src.to_s, destination.to_s).start
+      options = {
+        :read_timeout => Proxy::TFTP::Plugin.settings.tftp_read_timeout || 10,
```
dmitri-d: Default values are defined in HttpDownload already, no need to define them here too?

parthaa: hmm I thought just in case the "defaults" are different for tftp vs general this would be a good spot. I can remove em if you dont think they are necessary.

dmitri-d: Nah, I think one set of defaults is enough. Moreover, ```HttpDownload``` isn't used anywhere else, it really ought to live in the tftp module.

lib/proxy/http_download.rb
parthaa: I had it as expicit first and then thought that it might not be the brightest  idea to be reliant on the ordering of the input params. For example in my tests I switched the order of  read_timeout and connect_timeout without realizing...

dmitri-d: Please either use explicit parameters here, or switch to named parameters (although I don't know if this has implications downstream, as these aren't available on rubies < 2.0).

lzap:
Other than that, LGTM. Please edit tftp.yaml.example file at least.

test/tftp/tftp_test.rb
```diff
@@ -57,6 +57,40 @@ def test_choose_protocol_and_fetch_wget
...
   end
 
+  def test_choose_protocol_and_fetch_wget_with_timeouts
+    src = "https://proxy.test"
+    dst = "/destination"
+    tftp_read_timeout = "1000"
+    tftp_connect_timeout = "40"
+    tftp_dns_timeout = "14300"
+
+    Proxy::TFTP::Plugin.settings.expects(:tftp_read_timeout).returns(tftp_read_timeout)
```
dmitri-d: You should be able to use ```Proxy::TFTP::Plugin. load_test_settings({...})``` instead of creating mocks for individual settings.

test/tftp/tftp_test.rb
```diff
@@ -57,6 +57,40 @@ def test_choose_protocol_and_fetch_wget
...
+    Proxy::TFTP::Plugin.settings.expects(:tftp_connect_timeout).returns(tftp_connect_timeout)
+    Proxy::TFTP::Plugin.settings.expects(:tftp_dns_timeout).returns(tftp_dns_timeout)
+
+    start = stub(:start => true)
```
dmitri-d: This is a bit weird -- the return of #start isn't being used in the code or test, I think it can be safely ignored altogether.

test/tftp/tftp_test.rb
```diff
@@ -57,6 +57,40 @@ def test_choose_protocol_and_fetch_wget
...
+  end
+
+  def test_choose_protocol_and_fetch_wget_with_read_timeout
+    src = "https://proxy.test"
+    dst = "/destination"
+    tftp_read_timeout = "1000"
+
+    Proxy::TFTP::Plugin.settings.expects(:tftp_read_timeout).returns(tftp_read_timeout)
+
+    start = stub(:start => true)
```
dmitri-d: ditto for both setting expectation and the stub immediately above.

dmitri-d:
@parthaa: other than the couple of comments tests look good!

test/tftp/tftp_test.rb
parthaa: That mock is to avoid
```
Error: test_choose_protocol_and_fetch_wget_with_read_timeout(TftpTest): NoMethodError: undefined method `start' for nil:NilClass
```

test/tftp/tftp_test.rb
```diff
@@ -57,6 +57,37 @@ def test_choose_protocol_and_fetch_wget
...
+
+    Proxy::TFTP.choose_protocol_and_fetch src, dst
+  end
+
+  def test_choose_protocol_and_fetch_wget_with_read_timeout
+    src = "https://proxy.test"
+    dst = "/destination"
+    tftp_read_timeout = "1000"
+
+    Proxy::TFTP::Plugin.settings.expects(:tftp_read_timeout).returns(tftp_read_timeout)
```
dmitri-d: Could you replace this with ```Proxy::TFTP::Plugin.load_test_settings``` too?

config/settings.d/tftp.yml.example
```diff
@@ -5,3 +5,15 @@
 #:tftproot: /var/lib/tftpboot
 # Defines the TFTP Servername to use, overrides the name in the subnet declaration
 #:tftp_servername: tftp.domain.com
+
+# Defines the default read timeout in seconds needed to download tftp artifacts
+# like initrd and vmlinuz. Default value 60 seconds
+#:tftp_read_timeout: 60
```
dmitri-d, line 11: Could you add timeout settings and their values to the list of default settings here: https://github.com/theforeman/smart-proxy/blob/develop/modules/tftp/tftp_plugin.rb#L8

parthaa: @witlessbird done

modules/tftp/tftp_plugin.rb
```diff
@@ -5,6 +5,10 @@ class Plugin < ::Proxy::Plugin
     http_rackup_path File.expand_path("http_config.ru", File.expand_path("../", __FILE__))
     https_rackup_path File.expand_path("http_config.ru", File.expand_path("../", __FILE__))
 
-    default_settings :tftproot => '/var/lib/tftpboot'
+    default_settings(:tftproot => '/var/lib/tftpboot',
```
dmitri-d: Nitpick: Could you remove the parentheses please?

dmitri-d:
Merged, thanks, @parthaa!

PR #562, opened 2018-01-17
Fixes #22304 - Allow Puppet 5.x in bundle

mmoll:


mmoll:
in another PR I'll change the test matrix for develop to be the same as 1.17, after this got in.

dmitri-d:
Thanks, @mmoll, merged!

PR #565, opened 2018-01-27
Fixes #22435 - Switch to to_json for FreeIPA

ehelms:
On current Puppet 4, pretty_generate calls fail for types like string
and boolean. This is in part due to Puppet 4 requiring json_pure
~> 1.8.

### DIFF
--- a/modules/realm_freeipa/provider.rb
+++ b/modules/realm_freeipa/provider.rb
@@ -76,10 +76,10 @@ module Proxy::FreeIPARealm
         result = do_host_modify(hostname, setattr)
       end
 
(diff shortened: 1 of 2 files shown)


ehelms:
The current CI tests due not show this error because they are testing Puppet 4.4.0. At the time, Puppet 4.4.0 required ANY json_pure. Newer versions of Puppet 4 lock down json_pure to ~> 1.8.

dmitri-d:
[test]

ehelms:
Closing as this was in my environment, I had locked the json gem to 1.8.6 which does lead to a breakage but in reality this dependency is free to roam.

PR #567, opened 2018-02-13
Fixes #22496 - use common free_ips code in ms dhcp provider

dmitri-d:


lzap (approved):
I am not able to test but looks good!

dmitri-d:
* added a default value for ```blacklist_duration_minutes``` setting

lzap:
Mind adding it to example yaml config file?

dmitri-d:
We don't advertise this setting for dhcp_isc modules -- it was introduced "just in case", with the idea that shouldn't need to touch it. I think the same reasoning applies here as well.

lzap:
Ok, I don't want to merge untested patches, but ping me if you need me to hit the button!

dmitri-d:
@lzap: fwiw, I tested it in my windows VM -- seems to be working ok.

lzap:
Good test coverage, I am pretty confident here. Thanks.

PR #568, opened 2018-02-22
Fixes #22651 - extend dhcpd config parser support for "include"'s

dmitri-d:
"include" statements inside "shared_network" blocks are supported now.


lzap (approved):
WFM

PR #569, opened 2018-02-27
Fixes #22704 - handle dhcpd's "never" expiring leases

dmitri-d:


lzap:
Hehe, 2038-01-19 03:14:07 UTC is far enough :-) Thanks.

PR #571, opened 2018-03-27
Fixes #23031 - parse server duid's with nested double quotes

dmitri-d:


ekohl (approved):

modules/dhcp_common/isc/configuration_parser.rb
```diff
@@ -209,10 +209,11 @@ def server_duid
           Rsec::Fail.reset
           keyword = word('server-duid').fail 'keyword_server_duid'
           anything = /[^;,{}\s]+/.r
+          literal = /"([^"]|\")*"/.r
```
lzap, line 212: It looks like ISC configuration parser treats all strings like that, isn't cleaner to use this in `LITERAL` constant?

https://github.com/42wim/isc-dhcp/blob/f54a146c7fe88889d60f0c1aa8e6f04707f95223/common/conflex.c#L512

lzap:
If it won't break our tests, I think we should go with:

    LITERAL = /"([^"]|\")*"/.r

Single quotes do not appear to be supported by the parser, if I am looking into the correct place in the codebase. https://lists.isc.org/pipermail/dhcp-users/2011-February/013135.html

dmitri-d:
I think the ultimate goal  should be parsing strings the same way as dhcpd's parser does. Unfortunately, I don't think it is doable using a non-monstrous regex (if at all). The reason I didn't reuse the regex the way you are suggesting is that it will also match comma-separated strings as one string, which breaks the parser.

DUID is an outlier, as majority of the strings are host or option names and values, I think we are pretty safe with the fix in the PR. I'll add extending rsec parser with a custom string parser as a feature to smart-proxy's backlog.

lzap (approved):
Ok that's good. Thanks.

mark-wagner:
I think a similar change will be needed for the uid within a lease.

lzap:
I also think doing this for all strings won't hurt, the ISC parser does seem to work that way.

alvinstarr:
I believe that  literal should be  `literal = /"([^"]|\\")*"/.r`
The patch you suggested is in the version of 1.17 that I have installed but I was still getting duid errors until I added the second \.
I am really unfamiliar with ruby so I may well be wrong.

mark-wagner:
@alvinstarr that regexp doesn't work for. Here is the uid that was tripping me up:

`uid "\001X\357h}g\""`

http://rubular.com/r/gqxRXXs1YZ

The regexp in https://stackoverflow.com/questions/249791/regex-for-quoted-string-with-escaping-quotes seems to work but may have other problems.

http://rubular.com/r/vRvdrzFYxe

dmitri-d:
Please see my comment above. Feel free to experiment, but please run tests before proposing a fix. Please also note complicated regular expressions will be rejected.

alvinstarr:
This may be better 
`"([^\\"]|\\.)*"`
I believe the string consists of:
char - a single character
\### - 3 digits 0-7 for octal characters
\char - an escaped character
This of course only works for ASCII characters if my memories of C strings is correct.

dmitri-d:
I am locking this conversation. If you are seeing this issue, please update the redmine ticket referenced in this PR, or create a new one.

lzap:
I am not sure if locking was necessary here, "this might be better" is appropriate comment although I do agree some of them can be useless when not properly tried and tested. However, I think leaving those comments "as-is" is probably preferred than locking discussion.

PR #572, opened 2018-04-02
Fixes #23236 - fix for choria puppetrun on ubuntu

matt-cahill:
Choria's (https://github.com/choria-io/mcollective-choria)
mcollective ssl auth plugin looks for authentication keys
in the running user's home directory.

When executed via `sudo -u` '~' is expanded to
the running user's home directory rather than
the impersonated user causing an authentication error.

e.g.
```
sysadmin@puppet:/$ sudo -u foreman-proxy mco find

The find application failed to run, use -v for full 
error backtrace details: No such file or directory 
@ rb_sysopen - 
~/.puppetlabs/etc/puppet/ssl/certs/foreman-proxy.mcollective.pem
```

### DIFF
--- a/modules/puppet_proxy_mcollective/mcollective_main.rb
+++ b/modules/puppet_proxy_mcollective/mcollective_main.rb
@@ -21,7 +21,7 @@ class Proxy::PuppetMCollective::Runner < Proxy::Puppet::Runner
     # Defaults:foreman-proxy !requiretty
     # foreman-proxy ALL=(peadmin) NOPASSWD: /opt/puppet/bin/mco *',
     if user
-      cmd.push("-u", user)
+      cmd.push("-Hu", user)
     end
 
     mco_path = which("mco", ["/opt/puppet/bin", "/opt/puppetlabs/bin"])
--- a/test/puppet/mcollective_test.rb
+++ b/test/puppet/mcollective_test.rb
@@ -20,7 +20,7 @@ class MCollectiveTest < Test::Unit::TestCase
     @mcollective = Proxy::PuppetMCollective::Runner.new("peadmin")
     @mcollective.stubs(:which).with("sudo", anything).returns("/usr/bin/sudo")
     @mcollective.stubs(:which).with("mco", anything).returns("/usr/bin/mco")
-    @mcollective.expects(:shell_command).with(["/usr/bin/sudo", "-u", "peadmin", "/usr/bin/mco", "puppet", "runonce", "-I", "host1", "host2"]).returns(true)
+    @mcollective.expects(:shell_command).with(["/usr/bin/sudo", "-Hu", "peadmin", "/usr/bin/mco", "puppet", "runonce", "-I", "host1", "host2"]).returns(true)
     assert @mcollective.run(["host1", "host2"])
   end
 


modules/puppet_proxy_mcollective/mcollective_main.rb
```diff
@@ -21,7 +21,8 @@ def run(nodes)
     # Defaults:foreman-proxy !requiretty
     # foreman-proxy ALL=(peadmin) NOPASSWD: /opt/puppet/bin/mco *',
     if user
-      cmd.push("-u", user)
+      homedir = File.expand_path("~#{user}")
```
ekohl: Would it be good to verify ```HOME``` exists and prefer that, but use this as a fallback?

matt-cahill: I think `HOME` does exist it's just the value contains the home directory for the running user not the impersonated user. I'll do some tests to confirm though.

ekohl: Oh, that's a subtlety I missed. For me it does set the HOME var:
```
$ cat test.py
#!/usr/bin/env python3
import os
print(os.environ['HOME'])
$ ./test.py 
/home/ekohl
$ sudo -u root ./test.py
/root
```

There could be a different interaction in a non-interactive shell though.

matt-cahill: Ok I think this is actually an ubuntu vs other distros default behaviour thing, sorry about this

```
sysadmin:~$ ruby -e 'puts File.expand_path("~"); puts ENV.fetch("HOME")'
/home/sysadmin
/home/sysadmin
sysadmin:~$ sudo -u root ruby -e 'puts File.expand_path("~"); puts ENV.fetch("HOME")'
/home/sysadmin
/home/sysadmin
```
(same in python)

However setting `Defaults        env_keep -= "HOME"` in `/etc/sudoers` gives the behaviour you are seeing so this PR isn't needed I don't think. Maybe just a note in the docs?

ekohl: I figured it would be something like that. My sudoers (Fedora 27) has `Defaults env_reset` and then doesn't list `HOME` in `env_keep`.

Which version of Ubuntu are you on?
Is this the default config? If so, it'd be nice to implement a solution so it works out of the box.

matt-cahill: They're stock Ubuntu 16.04.2 LTS with basically just the foreman puppet::server and foreman-proxy puppet modules applied. Perhaps adding another section to the foreman-proxy sudoers.d file (or a separate file) is better?
It would also mean foreman-proxy could run `mco puppet runonce` out of the box too.

ekohl:
I just saw on centos-devel that someone referred to `sudo -H` or `sudo --set-home`:
> Request that the security policy set the HOME environment variable to the home directory specified by the target user's password database entry.  Depending on the policy, this may be the default behavior.

I think this is exactly what we want.

matt-cahill:
Yeah that looks good, I'll edit the PR.
```
sudo -Hu root ruby -e 'puts File.expand_path("~"); puts ENV.fetch("HOME")'
/root
/root
```

mmoll:
[test default]

mmoll:
[test]

mmoll:
@matt-cahill please check the test failure and update the expectation accordingly :bowtie:

matt-cahill:
sorry about that @mmoll, that should be in spec now

mmoll:
@matt-cahill I don't see the needed change, which would be in `test/puppet/mcollective_test.rb` line 23

matt-cahill:
sorry again, I was only seeing the prprocessor failures, I've made the test change too now.

mmoll:
[test]

ekohl:
ok to test

mmoll:
:green_heart:

ekohl (approved):

mmoll:
[test default]

mmoll:
anyway, this was :green_heart:

mmoll:
merged, thanks @matt-cahill!

matt-cahill:
Thanks @mmoll :)

PR #574, opened 2018-04-08
Update the README

ekohl:
* Switch URLs to https
* Move from mailing lists to discourse
* Remove Issue stats since it's broken and we don't use GH issues


dmitri-d:
Merged, thanks @ekohl!

PR #575, opened 2018-04-09
Fixes #23166 - MAC DHCP CRUD not case sensitive

lzap:
Problem for Hyper-V DHCP as reported by our user.

Now, the question is - shall I also do the same for our deprecated endpoints? It looks like a bug, maybe.

### DIFF
--- a/modules/dhcp/dhcp_api.rb
+++ b/modules/dhcp/dhcp_api.rb
@@ -76,13 +76,15 @@ class Proxy::DhcpApi < ::Sinatra::Base
 
   # returns a record for a mac address
   get "/:network/mac/:mac_address" do
+    param_network = params[:network]
+    param_mac = params[:mac_address].downcase unless params[:mac_address].nil?
     begin
       content_type :json
-      record = server.find_record_by_mac(params[:network], params[:mac_address])
-      log_halt 404, "No DHCP record for MAC #{params[:network]}/#{params[:mac_address]} found" unless record
+      record = server.find_record_by_mac(param_network, param_mac)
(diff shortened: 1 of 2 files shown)


ekohl (commented):
Should `server.*mac` functions be made case insensitive? MAC addresses are case insensitive so every comparison should be case insensitive. I don't know if any backend will provide us with uppercase MACs.

dmitri-d:
@lzap: Thanks, merged.
@ekohl: apologies. I noticed your comment only after I merged the PR. Not all providers rely on implementations provided by ```Proxy::DHCP::Server``` base class, which would mean dealing with the issue for each provider separately (and an opportunity for the bug to resurface in new providers). I figured dealing with the bug in the controller is probably the least amount of pain now and in the future. Let me know if you aren't happy with this, and I'll re-open the issue.

ekohl:
@witlessbird I don't know if any provider ever returned something uppercase and we silently relied on this behavior. The `SubnetService` maintains reservations and leases by mac in a hash where the mac is used as a key. That suggests to me that we should also normalize reading any mac from providers.

dmitri-d:
@ekohl: I think you right re: mac address normalization across the board. I also think this is a bigger scope than this particular issue. Would you be ok with me creating a dedicated ticket for it?

ekohl:
@witlessbird I think that's fine given the scope is wider than the initial issue. Would be good to relate them in Redmine.

dmitri-d:
@ekohl: Added a ticket re: mac address normalization: http://projects.theforeman.org/issues/23187

lzap:
Thanks guys, I just wanted to help the guy from our community to get moving, I won't be working on bigger refactoring for the moment.

PR #576, opened 2018-04-18
[WIP] Fixes #23211 - Token based PuppetCA autosigning

juliantodt:
Removes old autosigning endpoints and adds new ones
that take a incoming CSR from puppet, extract the token
and forward it to foreman for verfication.

See also PRs in [foreman-core](https://github.com/theforeman/foreman/pull/5465), [community-templates](https://github.com/theforeman/community-templates/pull/475), [puppet-foreman_proxy](https://github.com/theforeman/puppet-foreman_proxy/pull/425) & [puppet-puppet](https://github.com/theforeman/puppet-puppet/pull/591).

We would like to see some feedback early on while we are testing this thoroughly.

### DIFF
--- a/lib/proxy/request.rb
+++ b/lib/proxy/request.rb
@@ -20,6 +20,12 @@ module Proxy::HttpRequest
       req
     end
 
+    def create_delete(path, query={}, headers={})
+      req = Net::HTTP::Delete.new(uri(path).path + "/" + query[:id])
+      req = add_headers(req, headers)
+      req
+    end
+
     def uri(path)
       URI.join(@base_uri.to_s, path)
     end
--- a/modules/puppetca/puppetca_api.rb
+++ b/modules/puppetca/puppetca_api.rb
@@ -1,7 +1,8 @@
-require 'puppetca/puppetca_main'
-
 module Proxy::PuppetCa
   class Api < ::Sinatra::Base
+    extend Proxy::PuppetCa::DependencyInjection
+    inject_attr :cert_manager, :cert_manager
+
     helpers ::Proxy::Helpers
     authorize_with_trusted_hosts
     authorize_with_ssl_client
@@ -9,40 +10,18 @@ module Proxy::PuppetCa
     get "/?" do
       content_type :json
       begin
-        Proxy::PuppetCa.list.to_json
+        cert_manager.list.to_json
       rescue => e
         log_halt 406, "Failed to list certificates: #{e}"
       end
     end
 
-    get "/autosign" do
-      content_type :json
+    post "/autosign" do
+      request.body.rewind
       begin
-        Proxy::PuppetCa.autosign_list.to_json
+        cert_manager.autosign(request.body.read) ? 200 : 404
       rescue => e
-        log_halt 406, "Failed to list autosign entries: #{e}"
-      end
-    end
-
-    post "/autosign/:certname" do
-      content_type :json
-      certname = params[:certname]
-      begin
-        Proxy::PuppetCa.autosign(certname)
-      rescue => e
-        log_halt 406, "Failed to enable autosign for #{certname}: #{e}"
-      end
-    end
-
-    delete "/autosign/:certname" do
-      content_type :json
-      certname = params[:certname]
-      begin
-        Proxy::PuppetCa.disable(certname)
-      rescue Proxy::PuppetCa::NotPresent => e
-        log_halt 404, e.to_s
-      rescue => e
-        log_halt 406, "Failed to remove autosign for #{certname}: #{e}"
+        log_halt 406, "Failed to check autosigning for CSR: #{e}"
       end
     end
 
@@ -50,7 +29,7 @@ module Proxy::PuppetCa
       content_type :json
       certname = params[:certname]
       begin
-        Proxy::PuppetCa.sign(certname)
+        cert_manager.sign(certname)
       rescue => e
         log_halt 406, "Failed to sign certificate(s) for #{certname}: #{e}"
       end
@@ -60,7 +39,7 @@ module Proxy::PuppetCa
       begin
         content_type :json
         certname = params[:certname]
-        Proxy::PuppetCa.clean(certname)
+        cert_manager.clean(certname)
       rescue Proxy::PuppetCa::NotPresent => e
         log_halt 404, e.to_s
       rescue => e
--- a/modules/puppetca/puppetca_certmanager.rb
+++ b/modules/puppetca/puppetca_certmanager.rb
@@ -1,13 +1,16 @@
 require 'openssl'
 require 'set'
+require 'puppetca/puppetca_csr'
+require 'proxy/request'
 
-module Proxy::PuppetCa
-  extend ::Proxy::Log
-  extend ::Proxy::Util
+module ::Proxy::PuppetCa
 
   class NotPresent < RuntimeError; end
 
-  class << self
+  class Certmanager
+    include ::Proxy::Log
+    include ::Proxy::Util
+
     def sign certname
       puppetca("sign", certname)
     end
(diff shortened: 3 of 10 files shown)


dmitri-d:
Please do not remove old api endpoints, log a warning about them being deprecated instead. We usually keep api for one or two releases after the deprecation announcement.

modules/puppetca/puppetca_api.rb
```diff
@@ -15,34 +16,12 @@ class Api < ::Sinatra::Base
       end
     end
 
-    get "/autosign" do
-      content_type :json
+    post "/autosign" do
+      params = JSON.parse(request.env["rack.input"].read)
```
dmitri-d: Since the api is being changed, why not change it to something like:

```post "/autosign/:csr"``` instead of JSON encoding and then decoding a single parameter.

File: modules/puppetca/puppetca_csr.rb [hunk 2]
```diff
@@ -0,0 +1,25 @@
...
+        ['challengePassword', '1.2.840.113549.1.9.7'].include?(attr[:oid])
+      end
+      attribute ? attribute[:value] : nil
+    end
+
+    def custom_attributes
+      @csr.attributes.map do |attr|
+        {
+          oid: attr.oid,
+          value: attr.value.value.first.value
```
dmitri-d, line 20: This looks fragile: are all of the intermediate calls guaranteed to not return nils?

modules/puppetca/puppetca_main.rb
```diff
@@ -16,51 +18,48 @@ def clean certname
...
-        logger.debug "Attempt to remove nonexistent client autosign for #{certname}"
-        raise NotPresent, "Attempt to remove nonexistent client autosign for #{certname}"
+    # decide if csr should be autosigned
+    # parameter is csr to use
+    def autosign csr
+      if csr.nil?
+        logger.debug "Request did not include a CSR"
+        return false
+      end
+      return true if signall
```
dmitri-d: Is ```signall``` needed? If so, please put warnings both when it's triggered and during initialization.

modules/puppetca/puppetca_main.rb
```diff
@@ -16,51 +18,48 @@ def clean certname
...
     end
 
-    # list of hosts which are now allowed to be installed via autosign
-    def autosign_list
-      return [] unless File.exist?(autosign_file)
-      File.read(autosign_file).split("\n").reject do |v|
-        v =~ /^\s*#.*|^$/ ## Remove comments and empty lines
-      end.map do |v|
-        v.chomp ## Strip trailing spaces
+    def foreman_callback token
```
dmitri-d: Could you use something more descriptive for the method name, something like ```csr_recognised?```

modules/puppetca/puppetca_main.rb
```diff
@@ -16,51 +18,48 @@ def clean certname
...
+      logger.debug "Calling foreman with token '#{token}'"
+      fr = Proxy::HttpRequest::ForemanRequest.new
+      req = fr.request_factory.create_delete("api/puppetca_token", :id => token)
```
dmitri-d: I would suggested splitting the check and cleanup into two separate actions: not only removal of data on check not something that is expected, but it might make sense to keep the token until the host has been successfully created.

modules/puppetca/puppetca_main.rb
```diff
@@ -16,51 +18,48 @@ def clean certname
...
+      begin
+        response = fr.send_request(req)
+        if response.code.to_i == 204
+          logger.debug "Autosigning CSR with token " + token
+          return true
+        end
+      rescue => e
+        logger.debug "Error while calling foreman: " + e.to_s
```
dmitri-d: Please log this at the error level.

dmitri-d:
I would also suggest replacing all class-methods in ```Proxy::PuppetCa``` with instance ones in a dedicated class, and use dependency injection to configure the module/api controller (please see dhcp, dns, or other modules for examples. Also see http://projects.theforeman.org/projects/foreman/wiki/How_to_Create_a_Smart-Proxy_Plugin#How-to-Load-Dependencies).

lib/proxy/request.rb
```diff
@@ -20,6 +20,12 @@ def create_get(path, query={}, headers={})
       req
     end
 
+    def create_delete(path, query={}, headers={})
+      req = Net::HTTP::Delete.new(uri(path).path + "/" + query[:id])
```
timogoebel, line 24: `tap` might come in handy here.

modules/puppetca/puppetca_csr.rb
timogoebel: I think this should be fine. Openssl is not known for its nice code.

modules/puppetca/puppetca_main.rb
```diff
@@ -16,51 +18,48 @@ def clean certname
...
-    def autosign certname
-      FileUtils.touch(autosign_file) unless File.exist?(autosign_file)
-
-      open(autosign_file, File::RDWR) do |autosign|
-        # Check that we don't have that host already
-        found = autosign.readlines.find { |line| line.chomp == certname }
-        autosign.puts certname unless found
-      end
-      logger.debug "Added #{certname} to autosign"
+    def signall
```
timogoebel: Can we rename this to `sign_all`? I always read it as `signal` and start wondering where the second `l` comes from.

modules/puppetca/puppetca_main.rb
timogoebel: Removing the token is part of the design. For security reasons, it should just be able to use the token once.

modules/puppetca/puppetca_api.rb
juliantodt: I'm pretty sure that CSRs are not URL-safe, as they include slashes and line-breaks.

modules/puppetca/puppetca_csr.rb
juliantodt: Yeah, would probably be better to catch possible errors from this.

modules/puppetca/puppetca_main.rb
juliantodt: signall is the equivalent of putting `*` in your autosign.conf at the moment. As there are currently people using this it was a requested feature, so we're adding this as a setting. I don't think we need warnings for this as it is disabled by default and only is used if you manually update the settings, so you need to know what you're doing.

modules/puppetca/puppetca_main.rb
juliantodt: Will do.

lib/proxy/request.rb
juliantodt: It's the same as in all the other functions there...

modules/puppetca/puppetca_main.rb
juliantodt: Will do.

modules/puppetca/puppetca_main.rb
juliantodt: In what scenario do you think would it make sense to still have the token after the CSR was verified with it? If anything goes wrong during the verification, the token will not be deleted, so we just deleted it when the host has his cert signed, when we don't need the token anymore.

modules/puppetca/puppetca_main.rb
juliantodt: Will change that. Do you think that the other logs are correct on debug? Considering you normally don't run it on debug-loglevel, you won't be able to trace back why a CSR wasn't verified if you wanted.

juliantodt:
@witlessbird 
_Removing old API-Endpoints:_ Puppet use the autosign.conf XOR the autosign-script. When we change to the script, the autosign.conf will be useless and with it all functions editing it. Sure we could keep the API endpoints and log warnings but they will have no effect aswell.

_Outsourcing Methods:_ I will take a look at this.

Also some comments inline.
Could you take a look at the test? I don't think the fails are related to my changes, are they?

timogoebel:
> Could you take a look at the test? I don't think the fails are related to my changes, are they?

Don't worry, [test] failures are unrelated.

modules/puppetca/puppetca_api.rb
dmitri-d: You could UR-encode the parameter? Alternatively, instead of putting the parameter in the url, you could put it in the body of the request (would still need to be url-encoded though), which is a bit cleaner.

modules/puppetca/puppetca_main.rb
dmitri-d: I think a warning would be a good thing -- people forget to switch things back all the time. This flag completely circumvents token authorization, a warning would be a nice reminder (if verbose).

modules/puppetca/puppetca_main.rb
dmitri-d: It felt to me that this is a call with a (big) side-effect, which also broke the pattern of how we use "delete" method in smart-proxy.

If you renamed ```foreman_callback``` to something like ```delete_token```, this would resolve the whole side-effect issue.

modules/puppetca/puppetca_main.rb
dmitri-d: I'm ok either way, but perhaps a warning would be a bit more helpful.

modules/puppetca/puppetca_main.rb
```diff
@@ -16,51 +18,48 @@ def clean certname
...
-        v =~ /^\s*#.*|^$/ ## Remove comments and empty lines
-      end.map do |v|
-        v.chomp ## Strip trailing spaces
+    def foreman_callback token
+      logger.debug "Calling foreman with token '#{token}'"
+      fr = Proxy::HttpRequest::ForemanRequest.new
+      req = fr.request_factory.create_delete("api/puppetca_token", :id => token)
+      begin
+        response = fr.send_request(req)
+        if response.code.to_i == 204
```
dmitri-d: I'd say this is too specific, any 2xx would be ok?

juliantodt: Since we write the Foreman-API at the same time, why not be specific?

dmitri-d: Foreman side of things can change which would require this code to be updated. Since we don't care about particulars of the response as long as it is an "ok" (aka 2xx) one, making it too specific makes it more fragile.

modules/puppetca/puppetca_certmanager.rb
```diff
@@ -16,51 +18,52 @@ def clean certname
...
-        end
-        logger.debug "Removed #{certname} from autosign"
-      else
-        logger.debug "Attempt to remove nonexistent client autosign for #{certname}"
-        raise NotPresent, "Attempt to remove nonexistent client autosign for #{certname}"
+    # decide if csr should be autosigned
+    # parameter is csr to use
+    def autosign csr
+      if csr.nil?
+        logger.warn "Request did not include a CSR"
```
timogoebel: Nit: missing `.`

modules/puppetca/puppetca_csr.rb
```diff
@@ -0,0 +1,25 @@
+module Proxy::PuppetCa
+  class CertificateRequest
```
timogoebel: The class name should match the name of the file.

timogoebel (commented):
@juliantodt: Thanks. Some more comments that came up when I read the code. The DI stuff looks good now.

dmitri-d:
> Removing old API-Endpoints: Puppet use the autosign.conf XOR the autosign-script. When we change to the script, the autosign.conf will be useless and with it all functions editing it. Sure we could keep the API endpoints and log warnings but they will have no effect aswell.

If old behavior cannot be preserved at all, then all of these changes must reside in a new module/provider.

timogoebel:
> If old behavior cannot be preserved at all, then all of these changes must reside in a new module/provider.

@witlessbird: Why is that? Do we want to continue to support the old behavior so that users can use a newer smart-proxy version with an older Foreman version? Or is this meant for all the standalone users who use smart-proxy without Foreman?

juliantodt:
Took care of all the comments as well as the test failures.

@witlessbird I can see that removing the old functionality might seem problematic, but it will be same with the updates to the foreman-core and I don't think we need to support older foreman versions with newer smart-proxy versions. Also allowing both signing options to be chosen by the user (-> will need to allow this in the foreman-core as well), will just cause problems like foreman & smart-proxy using tokens but puppet is configured to use the autosign.conf. When we are strict with these new changes and remove old functions we get obvious error messages at misconfigurations and the solution will be to just update to the latest version. I know it's not perfect to make these changes that affect multiple components that are backwards-incompatible, but I prefer it over the alternatives.

ekohl:
I agree with @witlessbird that the API shouldn't change behavior of endpoints. If it changes, then it should use other URLs so Foreman gets a clear error. There are good reasons for this: until now we've always been flexible with versions. Running a 1.17 proxy with a 1.16 foreman worked, but also a 1.16 proxy with a 1.17 foreman. We know there are users with many proxies and upgrading them all at once is not feasible.

That said, I think there's still room for a compatible API. The one that's currently changed is `GET /autosign`. It could return an empty list.

dmitri-d:
What @ekohl said. There is also a number of people who use smart-proxy api for integration with their own software -- they'll be affected by such a break too.

juliantodt:
@witlessbird @ekohl So what you're saying is that you don't necessarily have a problem with removing old endpoints in favour of new ones but with keeping some and changing how they work. So change the name of the /autosign endpoint and done?

dmitri-d:
@juliantodt: not quite. Old endpoints have to keep names and preserve functionality for a couple releases after being declared as deprecated. 

If new functionality is being added to an existing end-point, the old behaviour must also be preserved.

timogoebel:
@juliantodt: It looks like there is a lot of resistance in getting this in, which is unfortunate as we did ask if there were any concerns before starting the implementation. I'd suggest to split this up and try to get the refactorings in (moving the code to a provider) so the puppetca api is pluggable. Maybe we can then either support both providers or move this to a plugin.

juliantodt:
Closing this in favour of the pluggable, backwards compatible approach #586.

PR #577, opened 2018-04-24
add support for srv dns records

pronix:


modules/dns/dns_api.rb
```diff
@@ -73,6 +78,14 @@ class Api < ::Sinatra::Base
       end
     end
 
+    def validate_srv_value!(value)
+      priority,weight,port,target = value.split(' ')
+      validate_dns_name!(target)
+      raise Proxy::Dns::Error.new("Invalid DNS SRV value #{value}") unless priority.to_i >= 0 and
+        weight.to_i >= 0 and
```
ekohl: Both weight and priority have a rage of 0-65535 according to [RFC2782](https://tools.ietf.org/html/rfc2782)

modules/dns/dns_api.rb
```diff
@@ -17,7 +17,7 @@ class Api < ::Sinatra::Base
       log_halt(400, "'create' requires fqdn, value, and type parameters") if fqdn.nil? || value.nil? || type.nil?
 
       begin
-        validate_dns_name!(fqdn)
+        validate_dns_name!(fqdn) if type != 'SRV'
```
ekohl: If it's SRV, we can still validate it if we allow underscores. It'd be good to have this validation both here and below.

ekohl (changes requested):
This makes sense.

modules/dns/dns_api.rb
pronix: but underscore is not allowed for fqdn.

ekohl: I meant that you should split the validation. Roughly like this:
```ruby
if type == 'SRV'
  validate_srv_name!(fqdn)
else
  validate_dns_name!(fqdn)
end
```

pronix:
@ekohl fixes pushed.

modules/dns/dns_api.rb
```diff
@@ -73,6 +86,18 @@ class Api < ::Sinatra::Base
       end
     end
 
+    def validate_srv_value!(value)
+      priority,weight,port,target = value.split(' ')
```
ekohl: Not an expert on ruby, but you probably need to ensure that it also handles values where it isn't 4. Could you write tests for this function?

pronix:
@ekohl  added check for extra data and test for it.

ekohl:
@pronix thanks! Could you also add an example with too little values or incorrect types like a letter for a priority instead of a number?

pronix:
@ekohl added more tests and more checks, letter for priority was missed.

modules/dns/dns_api.rb
```diff
@@ -17,7 +17,11 @@ class Api < ::Sinatra::Base
       log_halt(400, "'create' requires fqdn, value, and type parameters") if fqdn.nil? || value.nil? || type.nil?
 
       begin
-        validate_dns_name!(fqdn)
+        if type == 'SRV'
```
dmitri-d, line 20: To reduce duplication, could you move the conditional inside ```validate_dns_name!```, but perhaps keep type-specific validation in dedicated methods?

test/dns/dns_api_test.rb
```diff
@@ -54,6 +57,29 @@ def app
...
+    assert_equal '0 5 5060 sipserver.example.com.', @server.ip
+    assert_equal 'SRV', @server.type
+  end
+
+  def test_create_srv_record_returns_error_if_value_is_wrong
+    post '/', :fqdn => '_sip._tcp.example.com.', :value => '0 5 5060 sipserver.example.com. 1 2 3', :type => 'SRV'
+    assert_equal 400, last_response.status
+  end
+
+  def test_create_srv_record_returns_error_if_value_is_wrong2
```
dmitri-d, line 73: Could you convert this test and the other two below to unit-tests that exercise the validator directly? They'd be clearer/easier to follow and more helpful in troubleshooting.

test/dns/dns_api_test.rb
```diff
@@ -54,6 +57,29 @@ def app
...
+    post '/', :fqdn => '_sip._tcp.example.com.', :value => 'a 5 5060 sipserver.example.com.', :type => 'SRV'
+    assert_equal 400, last_response.status
+  end
+
+  def test_create_srv_record_returns_error_if_value_is_wrong3
```
dmitri-d, line 78: ditto

modules/dns/dns_api.rb
```diff
@@ -73,6 +86,21 @@ class Api < ::Sinatra::Base
       end
     end
 
+    def validate_srv_value!(value)
+      priority,weight,port,target,nillval = value.split(' ')
+      validate_dns_name!(target)
+      raise Proxy::Dns::Error.new("Invalid DNS SRV value #{value}") unless priority.scan(/\D/).empty? and
```
dmitri-d, line 92: Could you replace ```and``` with ```&&``` and convert this expression to an ```if ... end``` form.

dmitri-d: What's the purpose of ```nillval```? To verify that the input string has been split into a correct number parts? If so, I'd do something like this instead:

```
parts = value.split(' ')
raise "!!!" if parts.size != 4
```

As it's clearer/more explicit.

modules/dns/dns_api.rb
```diff
@@ -73,6 +86,21 @@ class Api < ::Sinatra::Base
...
+        weight.scan(/\D/).empty? and
```
dmitri-d, line 93: Instead of doing a regex check followed by a conversion, you can do ```Integer(weight)```, which will do both. You still need to do a range check however.

dmitri-d (changes requested):

eugen-berlin:
I'm on it.

eugen-berlin:
See #582.

PR #579, opened 2018-04-24
Pin gems for Ruby <2.2 (1.16-stable)

mmoll:


bundler.d/test.rb
```diff
@@ -16,4 +14,16 @@
...
     gem 'webmock'
   end
+
+  if RUBY_VERSION < '2.2.2'
+    gem 'rack-test', '~> 0.7.0'
+  else
+    gem 'rack-test'
+  end
+
+  if RUBY_VERSION < '2.2'
```
dmitri-d, line 24: This duplicates the version spec in development.rb; this one can be removed, no?

mmoll: rdoc was part of both files before this change also, I guess when one does exclude one of the two bundler groups, the problem still could happen when it's not in both.

dmitri-d: Nods (on both accounts). Test probably won't be used without development though, but I'll leave it up to you.

dmitri-d: Looks good otherwise!

mmoll: I'll leave it as-is here, then. If this should get tackled, it would need to happen in develop anyway. :hammer_and_wrench:

dmitri-d:
Merged, thanks @mmoll

PR #581, opened 2018-04-25
fixes #23399 - Fix puppet kick instructions

swadeley:
Changes must be made in the sudo file, not in the YAML file where the
 instructions are found.
I think  "'," at end of line is not required.
As puppetrun was in Puppet 2.7 and puppet kick is not supported in
Puppet 4, lets only mention Puppet 3 example here.


dmitri-d:
Thanks for the PR. Could you add a ticket here: http://projects.theforeman.org/projects/smart-proxy/issues, and then update the commit message to include the ticket number (pls. see theforeman-bot comment above for details).

dmitri-d:
Merged, thanks @swadeley!

PR #582, opened 2018-05-02
add support for srv dns records

eugen-berlin:
Here are original changes made by @pronix from #577 + fixes suggested by @witlessbird. Except I didn't change API tests to unit tests - I didn't understand how it should look like in the end. Do you mean to just test methods that validate records? Because we have no separate validator there.

### DIFF
--- a/modules/dns/dns_api.rb
+++ b/modules/dns/dns_api.rb
@@ -17,7 +17,7 @@ module Proxy::Dns
       log_halt(400, "'create' requires fqdn, value, and type parameters") if fqdn.nil? || value.nil? || type.nil?
 
       begin
-        validate_dns_name!(fqdn)
+        validate_dns_name!(fqdn, type)
 
         case type
         when 'A'
@@ -27,11 +27,14 @@ module Proxy::Dns
           ip = IPAddr.new(value, Socket::AF_INET6).to_s
           server.create_aaaa_record(fqdn, ip)
         when 'CNAME'
-          validate_dns_name!(value)
+          validate_dns_name!(value, type)
           server.create_cname_record(fqdn, value)
         when 'PTR'
           validate_reverse_dns_name!(value)
           server.create_ptr_record(fqdn, value)
+        when 'SRV'
+          validate_srv_value!(value)
+          server.create_srv_record(fqdn, value)
         else
           log_halt(400, "unrecognized 'type' parameter: #{type}")
         end
@@ -51,7 +54,7 @@ module Proxy::Dns
       end
 
       begin
-        validate_dns_name!(name)
+        validate_dns_name!(name, type)
 
         case type
         when 'A'
@@ -63,6 +66,8 @@ module Proxy::Dns
         when 'PTR'
           validate_reverse_dns_name!(name)
           server.remove_ptr_record(name)
+        when 'SRV'
+          server.remove_srv_record(name)
         else
           log_halt(400, "unrecognized 'type' parameter: #{type}")
         end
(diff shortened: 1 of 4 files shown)


dmitri-d:
@jewgeni-prokhorenko cheers for the PR.

> Do you mean to just test methods that validate records? Because we have no separate validator there.

You could invoke ```validate_srv_value!``` and ```validate_srv_name!``` directly in tests, as opposed to calling the top-level methods in the api controller. This would result in clearer tests and would help with maintenance down the road (tests written in this fashion tend to break less).

eugen-berlin:
@witlessbird please review - I added a commit that replaces the current way of testing with unit tests.

modules/dns/dns_api.rb
```diff
@@ -73,12 +78,34 @@ class Api < ::Sinatra::Base
...
+      end
+    rescue ArgumentError, RuntimeError # if any of the three tests on integer above fails or the condition was not met
+      raise Proxy::Dns::Error.new("Invalid DNS SRV value #{value}")
+    end
+
+    def validate_srv_name!(name)
+      raise Proxy::Dns::Error.new("Invalid DNS srv name #{name}") unless name =~ /^([a-zA-Z0-9_]([-a-zA-Z0-9_]+)?\.?)+$/
+    end
+
+    def validate_dns_name!(name, type)
```
ekohl: The function remains compatible if you use `type=nil`. Not sure if plugins use this function.

test/dns/dns_api_test.rb
```diff
@@ -223,4 +234,33 @@ def test_delete_returns_error_on_invalid_type
...
+    end
+    assert_raise Proxy::Dns::Error do
+      app.validate_srv_value!('a 5 5060 sipserver.example.com.')
+    end
+    assert_raise Proxy::Dns::Error do
+      app.validate_srv_value!('0 5 sipserver.example.com.')
+    end
+  end
+
+  def test_validate_srv_name!
```
ekohl: Could you add a test for `_sip._tcp.example.com`?

ekohl (changes requested):

eugen-berlin:
added test and made `validate_dns_name!` backwards compatible

ekohl:
ok to test

test/dns/dns_api_test.rb
```diff
@@ -223,4 +234,36 @@ def test_delete_returns_error_on_invalid_type
...
+    assert_raise Proxy::Dns::Error do
+      app.validate_srv_value!('0 5 5060 sipserver.example.com. 1 2 3')
+    end
+    assert_raise Proxy::Dns::Error do
+      app.validate_srv_value!('a 5 5060 sipserver.example.com.')
+    end
+    assert_raise Proxy::Dns::Error do
+      app.validate_srv_value!('0 5 sipserver.example.com.')
+    end
+  end
```
dmitri-d: Coud you either split this test into three or add comments about why exceptions are expected to be raised? The former is a better way to go about it.

dmitri-d:
@jewgeni-prokhorenko, thanks for the changes, looks good other than one nitpick I mentioned.

eugen-berlin:
@witlessbird thank you very much for your suggestions

dmitri-d:
These changes look good, but I originally meant the test here: https://github.com/theforeman/smart-proxy/pull/582/files#diff-cdb8479e65c28a30d109362f98c679baR238. 

FWIW the reason I'm being picky is that one assertion per test-type tests are easier to live with down the line. When they fail it's immediately clear why, and a good description of the test helps to understand the root-cause.

eugen-berlin:
@witlessbird understandable - please see the new commit

dmitri-d:
[test]

dmitri-d:
Rubocop isn't happy.

here: https://github.com/theforeman/smart-proxy/pull/582/files#diff-c603a2778d38b35e4c4c3f5a1edc6d35R86 -- could you either disable the rubocop check, or replace ```===``` with ```cover?```.
On the same lines, could you separate each three orders of magnitude in all integer constants with a ```_```, e.g. ```123456789``` becomes ```123_456_789```.

eugen-berlin:
fixed

dmitri-d:
[test]

eugen-berlin:
Hello colleagues. Can this pull request be merged now? Tests and code style checks are in order.

dmitri-d:
I'm getting 

```
D, [2018-05-09T11:25:16.350171 ] DEBUG -- : undefined local variable or method `fqdn' for #<Proxy::Dns::Nsupdate::Record:0x00000004640ab0> (NameError)
/home/wb/sandbox/smart-proxy/modules/dns_common/dns_common.rb:80:in `remove_srv_record'
/home/wb/sandbox/smart-proxy/modules/dns/dns_api.rb:71:in `block in <class:Api>'
```

when attempting to delete an ```SRV``` record: ```Proxy::Dns::Record#remove_srv_record``` method doesn't have ```fqdn``` parameter. 

Also, could you add test(s) for ```SRV``` record removal please?

dmitri-d:
[test]

eugen-berlin:
@witlessbird thank you

"The elephant was what I didn't notice." -- Ivan Krylov :)

dmitri-d:
@jewgeni-prokhorenko: could you rebase and squash commits please?

eugen-berlin:
@witlessbird rebased & squashed

dmitri-d:
[test]

dmitri-d:
There are still two commits here, could you squash it to one please? Otherwise this can potentially cause problems during cherry-picking and such.

eugen-berlin:
@witlessbird squashed

mmoll:
[test]

dmitri-d:
Merged, thanks @jewgeni-prokhorenko, @pronix.

PR #583, opened 2018-05-02
Fixes #23433 - lease uids are now parsed correctly

dmitri-d:


dmitri-d:
@lzap -- this is a custom string parser, very roughly based on the one used by dhcpd.

dmitri-d:
* fixed rubocop's conflicts

modules/dhcp_common/isc/configuration_parser.rb
```diff
@@ -396,7 +427,7 @@ def visit_parse_tree_nodes(parse_tree, all_hosts, all_subnets, all_ignored, root
         end
 
         def parse_file(a_path)
-          File.open(a_path, 'r') {|f| conf.parse!(f.read, a_path)}
+          File.open(a_path, 'r:ASCII-8BIT') {|f| conf.parse!(f.read, a_path)}
```
lzap, line 430: This is new to me, where can I find this in ruby docs?

modules/dhcp_common/isc/configuration_parser.rb
```diff
@@ -168,12 +168,44 @@ def visit(_, _, _, parent)
         end
         # rubocop:enable Style/StructInheritance
 
+        class Literal < Rsec::Binary
+          def _parse ctx
+            buffer = StringIO.new
+            bs = false
```
lzap, line 174: I think this could be named simply `backslash`.

modules/dhcp_common/isc/configuration_parser.rb
```diff
@@ -168,12 +168,44 @@ def visit(_, _, _, parent)
...
+            end_of_string = false
+            double_quote_counter = 0
+
+            return Rsec::INVALID if ctx.peek(1) != '"'
+            until end_of_string
+              return Rsec::INVALID if ctx.eos?
+              c = ctx.get_byte
+              if bs
+                buffer << sprintf("\\%c", c)
```
lzap, line 183: Can you point me on tests which runs through this branch? I only see regular strings in tests, we might need to add couple more like `'test escape: \\"'` and similar.

lzap (changes requested):
Nice, minor comments.

modules/dhcp_common/isc/configuration_parser.rb
dmitri-d: https://ruby-doc.org/core-2.0.0/Encoding.html, search for ascii-8bit

modules/dhcp_common/isc/configuration_parser.rb
dmitri-d: https://github.com/witlessbird/smart-proxy/blob/a15c8cd8fa6ba218d1c8196ed580ff99a3fa4d47/test/dhcp/conf_parser_test.rb#L439

lzap (approved):

lzap:
Thanks!

PR #585, opened 2018-06-02
[WIP] Introduce capabilities

ekohl:
This is still very early but I wanted some feedback before continuing too far.

The idea is that modules can have various capabilities. In the example code I added various DNS record types, but I want to expand this to managing domains themselves. Not every provider can actually do this due to various limitations, but some can. Foreman can query the capabilities and decide to expose functionalities or hide them rather than try-and-fail.

### DIFF
--- a/lib/proxy/plugin_initializer.rb
+++ b/lib/proxy/plugin_initializer.rb
@@ -33,6 +33,11 @@ class ::Proxy::PluginGroup
     [members.map { |m| m.capabilities }.compact].flatten.uniq.sort
   end
 
+  def settings
+    exposed_settings = [members.map { |m| m.exposed_settings }.compact].flatten.uniq.sort
+    Hash[exposed_settings.map { |setting| [setting, @plugin.settings[setting]] }]
+  end
+
   def resolve_providers(all_plugins_and_providers)
     return if inactive?
     return unless @plugin.uses_provider?
@@ -193,6 +198,7 @@ class ::Proxy::PluginInitializer
       updated[:http_enabled] = group.http_enabled?
       updated[:https_enabled] = group.https_enabled?
       updated[:capabilities] = group.capabilities
+      updated[:settings] = group.settings
       group.providers.each do |group_member|
         updated = to_update.find {|loaded_plugin| loaded_plugin[:name] == group_member.plugin_name}
         updated[:di_container] = group.di_container
--- a/lib/proxy/plugin_initializer.rb
+++ b/lib/proxy/plugin_initializer.rb
@@ -29,6 +29,10 @@ class ::Proxy::PluginGroup
     @https_enabled
   end
 
+  def capabilities
+    [members.map { |m| m.capabilities }.compact].flatten.uniq.sort
+  end
+
   def resolve_providers(all_plugins_and_providers)
     return if inactive?
     return unless @plugin.uses_provider?
@@ -188,6 +192,7 @@ class ::Proxy::PluginInitializer
       updated[:state] = group.state
       updated[:http_enabled] = group.http_enabled?
       updated[:https_enabled] = group.https_enabled?
+      updated[:capabilities] = group.capabilities
       group.providers.each do |group_member|
         updated = to_update.find {|loaded_plugin| loaded_plugin[:name] == group_member.plugin_name}
         updated[:di_container] = group.di_container
--- a/modules/dns/dns_plugin.rb
+++ b/modules/dns/dns_plugin.rb
@@ -6,6 +6,11 @@ module Proxy::Dns
     uses_provider
     default_settings :use_provider => 'dns_nsupdate', :dns_ttl => 86_400
     plugin :dns, ::Proxy::VERSION
+    capability :TYPE_A
+    capability :TYPE_AAAA
+    capability :TYPE_CNAME
+    capability :TYPE_PTR
+    capability :TYPE_SRV
 
     load_classes ::Proxy::Dns::ConfigurationLoader
   end
--- a/modules/root/root_api.rb
+++ b/modules/root/root_api.rb
@@ -20,15 +20,10 @@ class Proxy::RootApi < Sinatra::Base
           p[:class].ancestors.include?(::Proxy::Plugin)
       end
       content_type :json
-      enabled_plugins.sort { |x, y| x[:name] <=> y[:name] }.map do |p|
-        {
-          'name' => p[:name],
-          'capabilities' => p[:capabilities],
-          'http_enabled' => p[:http_enabled],
-          'https_enabled' => p[:https_enabled],
-          'settings' => p[:settings],
-        }
-      end.to_json
+
+      attributes = %i[capabilities http_enabled https_enabled settings]
+
+      Hash[enabled_plugins.collect { |p| [p[:name], Hash[attributes.map { |a| [a, p[a]] }]] }].to_json
     rescue => e
       log_halt 400, e
     end
--- a/modules/root/root_api.rb
+++ b/modules/root/root_api.rb
@@ -12,6 +12,27 @@ class Proxy::RootApi < Sinatra::Base
     end
   end
 
+  get "/v2/features" do
+    begin
+      enabled_plugins = ::Proxy::Plugins.instance.select do |p|
+        p[:name] != :foreman_proxy && \
+          p[:state] == :running && \
+          p[:class].ancestors.include?(::Proxy::Plugin)
+      end
+      content_type :json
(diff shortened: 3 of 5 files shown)


modules/dns/dns_plugin.rb
```diff
@@ -6,6 +6,13 @@ class Plugin < ::Proxy::Plugin
     uses_provider
     default_settings :use_provider => 'dns_nsupdate', :dns_ttl => 86_400
     plugin :dns, ::Proxy::VERSION
+    capabilities [
+      :TYPE_A,
+      :TYPE_AAAA,
+      :TYPE_CNAME,
+      :TYPE_PTR,
+      :TYPE_SRV,
+    ]
```
lzap: Nicer DSL could be:

```
capability :DNS_TYPE_A
```

I'd suggest to prefix capabilities with generic module/api name so when there is "TYPE_A" added into DHCP at some point we don't get a name clash.

Lowercase might look better when presented together with list of features, just a thought.

modules/root/root_api.rb
```diff
@@ -12,6 +12,27 @@ class Proxy::RootApi < Sinatra::Base
...
+          'capabilities' => p[:capabilities],
+          'http_enabled' => p[:http_enabled],
+          'https_enabled' => p[:https_enabled],
+        }
+      end.to_json
+    rescue => e
+      log_halt 400, e
+    end
+  end
+
```
lzap, line 31: Oh there you have it. I vote for lowercasing them - they are symbols anyway, they read better lowercase.

lzap (commented):
Nice, we should definitely expose these via `/features` and present in the UI/CLI.

How each particular DNS implementation (provider) would advertise capabilities then?

modules/dns/dns_plugin.rb
ekohl: Perhaps. I've also considered something like disabling capabilities from a provider. For example the libvirt provider doesn't support some DNS types that we have in core. Currently it just silently drops them which is not nice.

lib/proxy/plugin_initializer.rb
```diff
@@ -29,6 +29,10 @@ def https_enabled?
     @https_enabled
   end
 
+  def capabilities
+    [members.map { |m| m.get_capabilities }.reject { |c| c.nil? }].flatten.uniq
```
timogoebel: `compact` could be of help here

modules/root/root_api.rb
```diff
@@ -12,6 +12,27 @@ class Proxy::RootApi < Sinatra::Base
     end
   end
 
+  get "/features/v2" do
```
ehelms: Just a thought if this should be `/v2/features` if addtional APIs ever changed or came online you'd either get a pattern of:

```
/v2/blah
/v2/foo
```

or

```
/foo/v2
/bar/v2
```

modules/root/root_api.rb
```diff
@@ -12,6 +12,27 @@ class Proxy::RootApi < Sinatra::Base
...
+      enabled_plugins = ::Proxy::Plugins.instance.select do |p|
+        p[:name] != :foreman_proxy && \
+          p[:state] == :running && \
+          p[:class].ancestors.include?(::Proxy::Plugin)
+      end
+      content_type :json
+      enabled_plugins.sort { |x, y| x[:name] <=> y[:name] }.map do |p|
+        {
+          'name' => p[:name],
+          'capabilities' => p[:capabilities],
```
ehelms: I assume this could be anything you want data structure wise in which case YES please. This essentially what I wanted with (https://github.com/theforeman/smart_proxy_pulp/pull/11) without going full out implementing it. I'd rather have this than the one off solution.

In my case, I am not necessarily wanting to store "capabilities" so much as settings or properties. Do we forsee a conflict with concept naming there?

ehelms: Maybe we need both concepts? In the Pulp case I could see:

Setting: pulp URL, Pulp version
Capability: RPM, File

But that may be unncessary specification and a single "settings" type object could give freedom to specify whatever you want

ekohl: I think both concepts could be good. Perhaps it needs a generic `options` or `data` hash that every module can fill as they see fit?

modules/root/root_api.rb
ekohl: I see them a bit like constants and constants are typically uppercase. I do think that we should upper case them in the REST API.

ekohl:
I've updated it a bit.
* Moved to `/v2/features`
* Changed the DSL to `capability` which can be added multiple times.
* Added a DSL to expose settings. This is for smart_proxy_pulp's use case but TFTP also exposes a setting via a separate endpoint.

I'm thinking of a `disable_capability` DSL - for example in the virsh DNS provider which doesn't support some types. I also want to look at the DNS API and reuse capabilities there for the DNS types. This would allow us to expose the SSHFP type for those providers that support it (https://github.com/theforeman/smart-proxy/pull/536) with a lot less code.

ehelms:
How do you see the capabilities being used externally? For example, whats a workflow in Foreman to use them? Since I am less familiar with these types it would help me to think about the use case.

ekohl:
> How do you see the capabilities being used externally? For example, whats a workflow in Foreman to use them? Since I am less familiar with these types it would help me to think about the use case.

I think that the Foreman should be modified so it first probes the v2 endpoint. If that's available then use it. We'd modify the smart proxy model to store the capabilities and exposed settings on registration. Once that's in place, the hard part starts. Finding out where capabilities might make sense.

It is up to Foreman how it interprets capabilities. In the features API they are per module and Foreman should probably have something like `smart_proxy.has_module_capability?(module, capability)` to test for this.

My first goal is to start a DNS v2 API that exposes domains CRUD as well. On the Foreman side this would need client code. A capability would be added by providers that implement this API. When you create a domain in Foreman and the domain capability is present then Foreman creates the domain for you. Otherwise you get a UI warning the Proxy lacks functionality and you need to ensure it yourself. My initial goal is PowerDNS because it has a nice REST API we can interact with but ideally any provider can implement it.

Ideally we would also get it to work with ISC bind and then we can remove it from the installer. This should also help in a containerized world. A similar trick can be applied to DHCP subnets.

ehelms:
@ekohl Thanks for the example. What do you need to move this forward? I'd rather use this than my PR to smart_proxy_pulp but don't want things to linger either.

ekohl:
Last Thursday I played around with the Foreman side of it. https://github.com/theforeman/foreman/compare/develop...ekohl:capabilities already has the new database structure to store this. What needs to be done:
* Write the code to actually store the capabilities and settings in the table
* Test it on various databases. I try to use native table tables where possible but possibly that's a bad idea.

I'll update https://community.theforeman.org/t/exposing-capabilities-in-the-smart-proxy/9860 with a more concrete proposal.

lzap:
Have you considered keeping the current V1 JSON message and adding new element to the hash called "capabilities" which can hold additional information? Older Foreman would report this as a "feature" while newer could fetch additional data from it. I know that on Proxy Detail page we print module versions, that might break tho.

ekohl:
> Have you considered keeping the current V1 JSON message and adding new element to the hash called "capabilities" which can hold additional information? Older Foreman would report this as a "feature" while newer could fetch additional data from it. I know that on Proxy Detail page we print module versions, that might break tho.

IMHO this doesn't work because I intend capabilities to be per module. It might also break the current registration code which assumes it's all strings (though perhaps the mapping check already catches this). Still, treating one element as a special case doesn't sound like a very clean API.

Other than that I'm also exposing a settings hash per module. For that you'd need yet another odd special case. Note that this currently happens with various end points in at least 2 modules (tftp & pulp) so it's not a theoretical need.

jlsherrill:
@ekohl curious if you plan on revising this, or if there is something i can do to further test.  I'm going to need to do very similar things as https://github.com/theforeman/smart_proxy_pulp/pull/11 for pulp 3 (in addition to being able to query what content types are available in pulp 3).

ekohl:
@jlsherrill I'd love to revisit this, but a major issue is time. I'd be happy if someone else would take over. All my current code is on Github (see https://github.com/theforeman/foreman/pull/5954 as well).

The biggest change that should be made here is that it probably needs to be behind authentication if we're exposing settings. I've also considered changing the exposed module structure a bit to expose failed modules as well so you can detect misconfigurations from Foreman during registration. That'd probably mean it could be exposed as `/modules` or similar.

jlsherrill:
I've taken over this PR here: https://github.com/theforeman/smart-proxy/pull/625

ekohl:
Thanks!

PR #586, opened 2018-06-07
Fixes #23799 - Refactor: Make PuppetCa pluggable

juliantodt:
Since we are planning to offer alternative options for autosinging in the future (token based instead of hostname based) (see [redmine issue](http://projects.theforeman.org/issues/23211)) and want to allow users to choose their autosigning variant (primarily to allow compatibility with old versions, old foreman versions etc) we need to make the PuppetCA module pluggable, which means moving the autosinging functionality to a provider, which can then be swapped using the SmartProxy settings. We also want to clean up the module a bit and use dependency injections etc. The common logic for listing/signing/cleaning certificates (not autosign-entries) that uses the `puppet cert` command will remain in the puppetca-module.

This is btw an alternative approach to https://github.com/theforeman/smart-proxy/pull/576 which should allow for backwards compatibility and a compatible API between the autosigning variants.

### DIFF
--- a/config/settings.d/puppetca.yml.example
+++ b/config/settings.d/puppetca.yml.example
@@ -2,7 +2,9 @@
 # Can be true, false, or http/https to enable just one of the protocols
 :enabled: false
 
+# valid providers:
+#   - puppetca_hostname_whitelisting (verify CSRs based on a hostname whitelist)
+#:use_provider: puppetca_hostname_whitelisting
 #:ssldir: /var/lib/puppet/ssl
-#:autosignfile: /etc/puppet/autosign.conf
 #:puppetca_use_sudo: true
 #:sudo_command: /usr/bin/sudo
--- a/modules/puppetca/puppetca_api.rb
+++ b/modules/puppetca/puppetca_api.rb
@@ -1,7 +1,9 @@
-require 'puppetca/puppetca_main'
-
 module Proxy::PuppetCa
   class Api < ::Sinatra::Base
+    extend Proxy::PuppetCa::DependencyInjection
+    inject_attr :puppet_cert, :puppet_cert
+    inject_attr :autosigner, :autosigner
+
     helpers ::Proxy::Helpers
     authorize_with_trusted_hosts
     authorize_with_ssl_client
@@ -9,7 +11,7 @@ module Proxy::PuppetCa
     get "/?" do
       content_type :json
       begin
-        Proxy::PuppetCa.list.to_json
+        puppet_cert.list.to_json
       rescue => e
         log_halt 406, "Failed to list certificates: #{e}"
       end
@@ -18,7 +20,7 @@ module Proxy::PuppetCa
     get "/autosign" do
       content_type :json
       begin
-        Proxy::PuppetCa.autosign_list.to_json
+        autosigner.autosign_list.to_json
       rescue => e
         log_halt 406, "Failed to list autosign entries: #{e}"
       end
@@ -28,7 +30,7 @@ module Proxy::PuppetCa
       content_type :json
       certname = params[:certname]
       begin
-        Proxy::PuppetCa.autosign(certname)
+        autosigner.autosign(certname)
       rescue => e
         log_halt 406, "Failed to enable autosign for #{certname}: #{e}"
       end
@@ -38,7 +40,7 @@ module Proxy::PuppetCa
       content_type :json
       certname = params[:certname]
       begin
-        Proxy::PuppetCa.disable(certname)
+        autosigner.disable(certname)
       rescue Proxy::PuppetCa::NotPresent => e
         log_halt 404, e.to_s
       rescue => e
@@ -50,7 +52,7 @@ module Proxy::PuppetCa
       content_type :json
       certname = params[:certname]
       begin
-        Proxy::PuppetCa.sign(certname)
+        puppet_cert.sign(certname)
       rescue => e
         log_halt 406, "Failed to sign certificate(s) for #{certname}: #{e}"
       end
@@ -60,7 +62,7 @@ module Proxy::PuppetCa
       begin
         content_type :json
         certname = params[:certname]
-        Proxy::PuppetCa.clean(certname)
+        puppet_cert.clean(certname)
       rescue Proxy::PuppetCa::NotPresent => e
         log_halt 404, e.to_s
       rescue => e
--- a/config/settings.d/puppetca_hostname_whitelisting.yml.example
+++ b/config/settings.d/puppetca_hostname_whitelisting.yml.example
@@ -1,0 +1,6 @@
+---
+#
+# Configuration of the PuppetCA hostname_whitelisting provider
+#
+
+#:autosignfile: /etc/puppet/autosign.conf
--- a/extra/migrations/2018062000000_migrate_puppetca_settings.rb
+++ b/extra/migrations/2018062000000_migrate_puppetca_settings.rb
@@ -1,0 +1,33 @@
+require 'yaml'
+
+class MigratePuppetCaSettings < ::Proxy::Migration
+  def migrate
+    copy_original_configuration_except(path('settings.d', 'puppetca.yml'),
+                                       path('settings.d', 'puppetca_hostname_whitelisting.yml.example'))
+
+    module_settings   = YAML.load_file(path(src_dir, 'settings.d', 'puppetca.yml'))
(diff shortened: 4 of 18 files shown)


config/settings.d/puppetca.yml.example
```diff
@@ -2,7 +2,9 @@
 # Can be true, false, or http/https to enable just one of the protocols
 :enabled: false
 
+# valid providers:
+#   - puppetca_hostverify (verify CSRs based on hostnames)
```
timogoebel: Puppet calls this [basic autosigning](https://puppet.com/docs/puppet/5.5/ssl_autosign.html#enabling-basic-autosigning). I think we should stick with the term or use something like `autosignfile`.

modules/puppetca/puppetca_certmanager.rb
```diff
@@ -2,12 +2,13 @@
 require 'set'
 
 module Proxy::PuppetCa
-  extend ::Proxy::Log
-  extend ::Proxy::Util
 
   class NotPresent < RuntimeError; end
 
-  class << self
+  class Certmanager
```
timogoebel: Can we find a better name for this? It's not clear what this is doing. Maybe something like `CaInterface` or `CertificateAuthority`?

modules/puppetca_hostverify/puppetca_hostverify_autosigner.rb
```diff
@@ -0,0 +1,57 @@
+module ::Proxy::PuppetCa::Hostverify
+  class Autosigner
```
timogoebel: Sorry to be so picky about the names, but what about `FileBasedAutosign`.

modules/puppetca/puppetca_api.rb
```diff
@@ -2,14 +2,18 @@
 
 module Proxy::PuppetCa
   class Api < ::Sinatra::Base
+    extend Proxy::PuppetCa::DependencyInjection
+    inject_attr :cert_manager, :cert_manager
```
timogoebel: I'd call this `ca` or `ca_interface` or something similar.

E.g. `inject_attr :ca_provider, :ca` reads a lot cleaner in my opinion.

modules/puppetca/puppetca_api.rb
```diff
@@ -2,14 +2,18 @@
...
+    inject_attr :autosigner, :autosigner
```
timogoebel: How about just `autosign`?

test/puppetca_hostverify/puppetca_hostverify_config_test.rb
```diff
@@ -0,0 +1,11 @@
+require 'test_helper'
+require 'puppetca/puppetca'
+require 'puppetca_hostverify/puppetca_hostverify'
+require 'puppetca_hostverify/puppetca_hostverify_plugin'
+
+class PuppetCAHostverifyConfigTest < Test::Unit::TestCase
```
timogoebel: Nit: Please stick either to `Ca` or `CA`. I'd prefer the former.

timogoebel (changes requested):
@juliantodt: Thanks. Looks like solid work. I just left some small comments regarding naming. If they are resolved this looks ready to 🚢.

config/settings.d/puppetca.yml.example
juliantodt: I don't like `basic autosigning` because it says nothing about how it works, `autosignfile` is better, but because the token based approach will also use a file it is not clear which is which. The thing that will distinguish the providers is the attribute which is compared during autosigning which is `certname` vs. `token`.
So imho the best name would be `puppetca_hostnamebasedautosigning` (or `certnamebased`) but this is a pain to read.

modules/puppetca/puppetca_api.rb
juliantodt: Well `ca` is the whole thing, which consists of managing certificates and autosign-entries. Since the `cert_manager` basically uses the `puppet cert` command for everything it does, I thought this would be matching name.

modules/puppetca/puppetca_api.rb
juliantodt: I like to refer to objects with nouns, makes it better to read imho ^^

modules/puppetca_hostverify/puppetca_hostverify_autosigner.rb
juliantodt: The token-based-autosigning-approach will also use a file, so that would not work. See above.

juliantodt (commented):
Thanks! Left some thoughts about the naming inline @timogoebel

config/settings.d/puppetca.yml.example
timogoebel: How about `HostnameWhitelisting`?

modules/puppetca/puppetca_api.rb
timogoebel: I don't see autosigning as part of the CA. Basically you have two separate strings:

1. Interface to the CA (ask the ca to issue/remove/... a certificate, ...)
2. Interface to an autosigning provider (list whitelisted hosts, whitelist a host, ...). This is typically not part of the CA, but of the RA.

Certmanager is definitely misleading. Please use something else.

modules/puppetca/puppetca_api.rb
timogoebel: Ok, works for me.

config/settings.d/puppetca.yml.example
juliantodt: I like that, but wouldn't we want to use snake_case?

modules/puppetca/puppetca_api.rb
juliantodt: Well, its part of the puppetCA module and under the `puppet/ca/` api endpoint...
But I can understand that that may be confusing, how about just `puppet_cert` so its obvious it uses the `puppet cert` command?

config/settings.d/puppetca.yml.example
timogoebel: Yep, 🐍.

modules/puppetca/puppetca_api.rb
timogoebel: Ok, works for me.

juliantodt:
Updated the PR with the requested naming changes.

modules/puppetca/puppetca_api.rb
```diff
@@ -2,14 +2,18 @@
```
dgoetz: The line above has to be changed to `require 'puppetca/puppetca_puppet_cert'` as modules/puppetca/puppetca_main.rb is renamed to modules/puppetca/puppetca_puppet_cert.rb so it can load the feature.

juliantodt: `puppetca/puppetca_puppet_cert` is loaded via `plugin_configuration::load_classes`, so we don't need it here. This require can therefore be deleted.

dgoetz: You are right, deleting it works also.

juliantodt:
Fixed @dgoetz. Added a migration for the settings, so that users who had set the `autosignfile` setting will not experience problems.

dgoetz (approved):
Tested the latest commit including the migration and it works.

timogoebel (approved):

timogoebel:
Thanks, @juliantodt.

tbrisker:
Looks like nightly bats tests are now failing with an error regarding PuppetCA, could it be related to this change? http://ci.theforeman.org/blue/rest/organizations/jenkins/pipelines/foreman-nightly-release/runs/248/nodes/34/steps/41/log/?start=0

dgoetz:
@tbrisker: Yes, it is probably related. The proxy logs could be helpful.
@timogoebel, @juliantodt: Puppet code needs to be adjusted as it adds still autosign file to puppetca after installation and migration.

timogoebel:
We already have a PR open to fix the puppet modules, see https://github.com/theforeman/puppet-foreman_proxy/pull/433.
@juliantodt is going to address the comments today.

tbrisker:
Thanks @juliantodt and @timogoebel! in the future, when something requires changes in multiple repos, please make sure all PRs are ready to merge first and are merged in coordination to avoid breakages such as this.

tbrisker:
@juliantodt @timogoebel  Care to update the manuals for 1.19 forward with this change? Looks like some users are stumbling on it - https://community.theforeman.org/t/puppetca-foreman-proxy-disabled-beacuse-of-autosign-issue/12290

PR #591, opened 2018-06-20
Fixes #24016 - Allow building a tarball without bundle

ekohl:
This allows generation of the source tarball without the whole bundle installed. It also prints the generated filename.


lzap (approved):
TY

PR #592, opened 2018-06-21
Fixes #23211 - Add PuppetCa TokenWhitelisting provider

juliantodt:
This adds an alternative PuppetCa provider.
Instead of signing CSRs based on whitelist of hosts (directly via puppet), we will create a signed JWT and return it to foreman that will provision the token on the host so that it is included in the CSR the hosts sends to the PuppetCa. The PuppetCa will then use a script (rather than the textfile) (see puppet-foreman_proxy) that calls the SmartProxy with the CSR that will then validate the token inside it. This should be much more secure and easier to debug than the current hostname_whitelisting provider.
Note: For this to work, you will need #[5730](https://github.com/theforeman/foreman/pull/5730) in your foreman instance.


modules/puppetca_token_whitelisting/puppetca_token_whitelisting_autosigner.rb
```diff
@@ -0,0 +1,96 @@
...
+    def disable certname
+      hosts = autosign_list
+      hosts.delete(certname)
+      File.write hosts_file, hosts.to_yaml
+    end
+
+    # Create a new token for a certname
+    def autosign certname
+      ensure_hostsfile
+      payload = { certname: certname, exp: Time.now.to_i + 10 * 60 * 60 }
```
ekohl: Perhaps this time should be configurable. IMHO at least defined as a constant so you know what this magic value is.

modules/puppetca_token_whitelisting/puppetca_token_whitelisting_autosigner.rb
```diff
@@ -0,0 +1,96 @@
...
+      token = JWT.encode payload, smartproxy_cert, 'RS512'
+      File.write hosts_file, autosign_list.push(certname).to_yaml
+      { generated_token: token }.to_json
```
ekohl: Should the JSON representation be done in this function? IMHO it's something that you do in the API class

modules/puppetca_token_whitelisting/puppetca_token_whitelisting_autosigner.rb
```diff
@@ -0,0 +1,96 @@
...
+      Proxy::PuppetCa::TokenWhitelisting::Plugin.settings.hosts_file
+    end
+
+    def smartproxy_cert
+      OpenSSL::PKey::RSA.new File.read Proxy::SETTINGS.ssl_private_key.to_s
+    end
+
+    def ensure_hostsfile
+      unless File.exist?(hosts_file)
+        FileUtils.mkdir_p hosts_file.rpartition('/').first
```
ekohl: Isn't the `rpartition('/').first` equivalent to `dirname()`?

modules/puppetca_token_whitelisting/puppetca_token_whitelisting_plugin.rb
```diff
@@ -0,0 +1,11 @@
+module ::Proxy::PuppetCa::TokenWhitelisting
+  class Plugin < ::Proxy::Provider
+    plugin :puppetca_token_whitelisting, ::Proxy::VERSION
+
+    requires :puppetca, ::Proxy::VERSION
+    default_settings :sign_all => false, :hosts_file => '/tmp/foreman-proxy/hosts.yml'
```
ekohl: I don't like this default because `/tmp` is a shared directory. Another user on the system can create `/tmp/foreman-proxy` prior to startup and have control. That makes it a security risk. We do have `/var/lib/foreman-proxy` in some packages (notably REX/Ansible I think) and perhaps we should standardize this. An alternative is doing it in `/etc/foreman-proxy` but I don't like it because it gives the impression users can/should modify it.

modules/puppetca_token_whitelisting/puppetca_token_whitelisting_autosigner.rb
```diff
@@ -0,0 +1,96 @@
+require 'jwt'
+require 'openssl'
+
+module ::Proxy::PuppetCa::TokenWhitelisting
+  class Autosigner
+    include ::Proxy::Log
+    include ::Proxy::Util
+
+    def sign_all
+      Proxy::PuppetCa::TokenWhitelisting::Plugin.settings.sign_all
```
ekohl: I am still wondering if it should allow reading a whitelist from the actual autosign file but this might be enough.

modules/puppetca_token_whitelisting/puppetca_token_whitelisting_autosigner.rb
```diff
@@ -0,0 +1,96 @@
...
+      hosts = autosign_list
+      hosts.delete(certname)
+      File.write hosts_file, hosts.to_yaml
+    end
+
+    # Create a new token for a certname
+    def autosign certname
+      ensure_hostsfile
+      payload = { certname: certname, exp: Time.now.to_i + 10 * 60 * 60 }
+      token = JWT.encode payload, smartproxy_cert, 'RS512'
```
ekohl: I think the `RS512` algorithm should be stored in a constant.

modules/puppetca_token_whitelisting/puppetca_token_whitelisting_autosigner.rb
```diff
@@ -0,0 +1,96 @@
...
+
+    def smartproxy_cert
+      OpenSSL::PKey::RSA.new File.read Proxy::SETTINGS.ssl_private_key.to_s
+    end
+
+    def ensure_hostsfile
+      unless File.exist?(hosts_file)
+        FileUtils.mkdir_p hosts_file.rpartition('/').first
+        FileUtils.touch hosts_file
+        File.write(hosts_file, [].to_yaml)
```
ekohl: I'd move writing to this file to a (private) helper function rather than duplicate it in 3 places.

modules/puppetca_token_whitelisting/puppetca_token_whitelisting_autosigner.rb
```diff
@@ -0,0 +1,96 @@
...
+      payload = { certname: certname, exp: Time.now.to_i + 10 * 60 * 60 }
+      token = JWT.encode payload, smartproxy_cert, 'RS512'
+      File.write hosts_file, autosign_list.push(certname).to_yaml
+      { generated_token: token }.to_json
+    end
+
+    # List the hosts that are currently valid
+    def autosign_list
+      ensure_hostsfile
+      YAML.load_file(hosts_file).to_a
```
ekohl: Should `safe_load` be used? Especially if the default location is in `/tmp` there's a risk of executing arbitrary code as the `foreman-proxy` user as long as you can write anything in `/tmp`. Because of it I'd actually consider JSON rather than YAML. Since no user has to actually modify the file, it doesn't matter that it's not very human friendly.

ekohl (changes requested):
How do we prevent concurrency issues with writing to the hosts file?

modules/puppetca_token_whitelisting/puppetca_token_whitelisting_autosigner.rb
juliantodt: Yes, I agree. Do you think 10 hours is a good default?

modules/puppetca_token_whitelisting/puppetca_token_whitelisting_autosigner.rb
juliantodt: Will move.

modules/puppetca_token_whitelisting/puppetca_token_whitelisting_autosigner.rb
juliantodt: Yeah, sometimes I just remember the complicated ways to do thinks. Will change this.

modules/puppetca_token_whitelisting/puppetca_token_whitelisting_autosigner.rb
timogoebel: Currently we allow 6 hours in foreman core as a default (for build tokens). I'd suggest do use the same.

modules/puppetca/puppetca_api.rb
```diff
@@ -36,6 +36,19 @@ class Api < ::Sinatra::Base
       end
     end
 
+    post "/validate" do
+      content_type :json
+      unless autosigner.respond_to?('validate_csr')
+        log_halt 401, "Provider only supports trivial autosigning"
```
timogoebel: 401 is unauthorized, how about 406 or 412?

modules/puppetca_token_whitelisting/puppetca_token_whitelisting_autosigner.rb
ekohl: If Foreman has a setting for this, should we have it as an API parameter instead? You could still apply a default there but that way you would have a single setting.

modules/puppetca_token_whitelisting/puppetca_token_whitelisting_autosigner.rb
```diff
@@ -0,0 +1,96 @@
...
+
+    # Check whether a csr is valid and should be signed
+    # by checking its token if it exists
+    def validate_csr csr
+      if csr.nil?
+        logger.warn "Request did not include a CSR."
+        return false
+      end
+      if sign_all
+        logger.warn "Signing CSR without token verification."
```
timogoebel, line 84: Should be info I think.

modules/puppetca_token_whitelisting/puppetca_token_whitelisting_plugin.rb
juliantodt: We also weren't sure where to put the file. Its not code, it's not config, it does not belong in the puppet dir. I settled for this as it is a temporary file, but am not quite happy, so I'm open to suggestions. You think `/var/lib/foreman-proxy/` would be the best idea?

modules/puppetca_token_whitelisting/puppetca_token_whitelisting_autosigner.rb
juliantodt: I think falling back to the `autosign.conf` could be confusing. Allowing this for people who really really want to put the `*` in the file should be fine.

modules/puppetca_token_whitelisting/puppetca_token_whitelisting_autosigner.rb
```diff
@@ -0,0 +1,96 @@
...
+        FileUtils.touch hosts_file
+        File.write(hosts_file, [].to_yaml)
+      end
+    end
+
+    # Invalidate a token based on the certname
+    def disable certname
+      hosts = autosign_list
+      hosts.delete(certname)
+      File.write hosts_file, hosts.to_yaml
```
timogoebel: I think we should extract this to a dedicated storage class.

modules/puppetca_token_whitelisting/puppetca_token_whitelisting_autosigner.rb
```diff
@@ -0,0 +1,96 @@
...
+    def sign_all
+      Proxy::PuppetCa::TokenWhitelisting::Plugin.settings.sign_all
+    end
+
+    def hosts_file
+      Proxy::PuppetCa::TokenWhitelisting::Plugin.settings.hosts_file
+    end
+
+    def smartproxy_cert
+      OpenSSL::PKey::RSA.new File.read Proxy::SETTINGS.ssl_private_key.to_s
```
timogoebel: What happens if the smart proxy is run without ssl?

modules/puppetca_token_whitelisting/puppetca_token_whitelisting_autosigner.rb
```diff
@@ -0,0 +1,96 @@
...
+      unless File.exist?(hosts_file)
+        FileUtils.mkdir_p hosts_file.rpartition('/').first
+        FileUtils.touch hosts_file
+        File.write(hosts_file, [].to_yaml)
+      end
+    end
+
+    # Invalidate a token based on the certname
+    def disable certname
+      hosts = autosign_list
```
timogoebel: I believe we should store the (in)valid tokens here instead of the certnames. If a token is lost, a user has no way of actually revoking it for good.

timogoebel (changes requested):
@juliantodt: Thanks, I left some comments inline.

modules/puppetca/puppetca_api.rb
ekohl: 406 would be incorrect because that's on `Accept` headers. I'd consider HTTP 501 Not Implemented.

modules/puppetca_token_whitelisting/puppetca_token_whitelisting_plugin.rb
ekohl: `/tmp` is wiped after a reboot. Losing that kind of state doesn't sound right for this file. `/var/tmp` would be better but to me it sounds like this is actually a state file. For that `/var/lib` is generally used.

modules/puppetca/puppetca_api.rb
juliantodt: I was thinking `Bad Request` but that would have been 400. 501 would be fine for me.

modules/puppetca_token_whitelisting/puppetca_token_whitelisting_autosigner.rb
juliantodt: We had that when we talked about the first SmartProxyPR #576 where witlessbird suggested warnings.

modules/puppetca_token_whitelisting/puppetca_token_whitelisting_autosigner.rb
juliantodt: Well, that would be a problem. I guess we would have to ensure that there is a RSA certificate there if used for SSL or not...

modules/puppetca_token_whitelisting/puppetca_token_whitelisting_autosigner.rb
juliantodt: Agreed, I think storing valid tokens is the way to go. It allows us to do everything we do now (since we can always extract the certname from the tokens) while we still don't have to clean up periodically because the list would auto-clean itself + additionally we can revoke specific tokens manually.

modules/puppetca_token_whitelisting/puppetca_token_whitelisting_autosigner.rb
juliantodt: That would mean that the API would be different for the providers, right?

ekohl: I think an optional API parameter for the TTL would still be compatible enough. You may want to raise a 400 Bad Request on the old provider if a TTL is provided.

test/puppetca_token_whitelisting/puppetca_token_whitelisting_autosigner_test.rb
```diff
@@ -0,0 +1,96 @@
...
+      ## Setup
+      FileUtils.cp './test/fixtures/puppetca/hosts.yml', @file.path
+    rescue
+      @file.close
+      @file.unlink
+      @file = nil
+    end
+    @autosigner = Proxy::PuppetCa::TokenWhitelisting::Autosigner.new
+    @autosigner.stubs(:hosts_file).returns(@file.path)
+    rsa_cert = OpenSSL::PKey::RSA.generate 2048
```
ekohl: Does this mean a RSA cert is generated for every test in this file? Isn't that somewhat slow?

modules/puppetca_token_whitelisting/puppetca_token_whitelisting_autosigner.rb
juliantodt: Raising an 400 on the old provider isn't a good idea, as foreman doesnt know which provider is used and will (if we add that) always include the TTL, meaning the old provider would not work. Will add the optional parameter.

test/puppetca_token_whitelisting/puppetca_token_whitelisting_autosigner_test.rb
juliantodt: Not noticeable (at least for me). But since we decided to change the implementation to save the tokens instead of certnames, we will need to supply a fixtured cert anyway.

modules/puppetca_token_whitelisting/puppetca_token_whitelisting_autosigner.rb
ekohl: That's a fair point. https://github.com/theforeman/smart-proxy/pull/585 is my take at solving this but I don't see this being integrated any time soon.

modules/puppetca_token_whitelisting/puppetca_token_whitelisting_autosigner.rb
juliantodt: Moving the `.to_json` to the API class would mean that we might change the behaviour of the old provider.

juliantodt:
Updated:
* Saving tokens instead of certnames
* Moving TokenStorage to external class
* Locking file during read/writing
* Configurable TTL
* Moved default location of tokens file

TODO:
* Add tests for storage class
* Handle case where SmartProxy doesnt use SSL
* Fix tests, rubocop ^^

juliantodt:
Updated. This should now satisfy all your concerns. Thanks @ekohl @timogoebel! Tests are now green as well.

Question: If we decide to add the token ttl to the foreman api call, we would break compatibility with old SmartProxy versions as they would not recognise the url, right?

modules/puppetca/puppetca_api.rb
```diff
@@ -26,16 +26,30 @@ class Api < ::Sinatra::Base
       end
     end
 
-    post "/autosign/:certname" do
+    post /\/autosign\/([^\/\s]+)(\/([\d]*))?/ do
       content_type :json
-      certname = params[:certname]
+      certname = params['captures'][0]
+      ttl = params['captures'][2]
```
ekohl: Isn't the TTL more of a body parameter in REST?

timogoebel:
> Question: If we decide to add the token ttl to the foreman api call, we would break compatibility with old SmartProxy versions as they would not recognise the url, right?

I think we can live with that. We could check the smart proxy version in foreman and add a switch for that. Or just try again if we get a 404. @ekohl: WDYT?

modules/puppetca_token_whitelisting/puppetca_token_whitelisting_autosigner.rb
```diff
@@ -0,0 +1,123 @@
...
+    # List the hosts that are currently valid
+    def autosign_list
+      storage.read.collect do |token|
+        begin
+          decoded = JWT.decode(token, smartproxy_cert.public_key, true, algorithm: JWT_ALGORITHM)
+          decoded.first['certname']
+        rescue JWT::ExpiredSignature
+          nil
+        end
+      end.reject { |certname| certname.nil? }
```
ekohl: Isn't this equal reject equal to `compact`?

bundler.d/puppetca_token_whitelisting.rb
```diff
@@ -0,0 +1,7 @@
+group :puppetca_token_whitelisting do
+  if RUBY_VERSION < '2.1'
+    gem 'jwt', '1.5.6'
```
timogoebel: Can we use `~> 1.5.6`?

modules/puppetca/puppetca_api.rb
```diff
@@ -26,16 +26,30 @@ class Api < ::Sinatra::Base
...
-        autosigner.autosign(certname)
+        autosigner.autosign(certname, ttl)
       rescue => e
         log_halt 406, "Failed to enable autosign for #{certname}: #{e}"
       end
     end
 
+    post "/validate" do
+      content_type :json
+      unless autosigner.respond_to?('validate_csr')
```
timogoebel: Nit: I'd us a symbol here: `respond_to?(:validate_csr)`

modules/puppetca_token_whitelisting/puppetca_token_whitelisting_token_storage.rb
```diff
@@ -0,0 +1,56 @@
...
+    include ::Proxy::Log
+    include ::Proxy::Util
+
+    def initialize tokens_file
+      @tokens_file = tokens_file
+      ensure_file
+    end
+
+    def ensure_file
+      unless File.exist?(@tokens_file)
```
timogoebel: you can use a guard clause here, `return of File.exists?(@token_file)`

modules/puppetca_token_whitelisting/puppetca_token_whitelisting_token_storage.rb
```diff
@@ -0,0 +1,56 @@
...
+      end
+    end
+
+    def write content
+      lock do
+        locked_write content
+      end
+    end
+
+    def locked_write content
```
timogoebel: I think the method name is misleading, as the write is not actually locked here.

timogoebel (changes requested):
@juliantodt: Yet another round of comments. :-)

modules/puppetca/puppetca_api.rb
timogoebel: Yeah, we could make this a parameter and not part of the URL. That would make this even more backward compatible.

juliantodt:
Updated based on your comments, thank you!
Token TTL is now a body parameter which should make the foreman implementation backwards-compatible.

modules/puppetca_token_whitelisting/puppetca_token_whitelisting_autosigner.rb
```diff
@@ -0,0 +1,123 @@
...
+    def token_ttl
+      Proxy::PuppetCa::TokenWhitelisting::Plugin.settings.token_ttl
+    end
+
+    def cert_file
+      return Proxy::SETTINGS.ssl_private_key.to_s if Proxy::PuppetCa::TokenWhitelisting::Plugin.settings.certificate.nil?
+
+      file = Proxy::PuppetCa::TokenWhitelisting::Plugin.settings.certificate
+      unless File.exist?(file)
+        File.write file, OpenSSL::PKey::RSA.generate(2048)
```
ekohl: It's probably good to do this with a umask so it's not world readable (or some other way). The number of bits should also be at least a constant IMHO.

modules/puppetca_token_whitelisting/puppetca_token_whitelisting_token_storage.rb
```diff
@@ -0,0 +1,55 @@
...
+      end
+    end
+
+    def write content
+      lock do
+        unsave_write content
+      end
+    end
+
+    def unsave_write content
```
ekohl: unsafe?

modules/puppetca_token_whitelisting/puppetca_token_whitelisting_autosigner.rb
```diff
@@ -0,0 +1,123 @@
...
+        tokens = storage.read
+        tokens.delete_if do |token|
+          begin
+            decoded = JWT.decode(token, smartproxy_cert.public_key, true, algorithm: JWT_ALGORITHM)
+            decoded.first['certname'] == certname
+          rescue JWT::ExpiredSignature
+            true
+          end
+        end
+        storage.unsave_write tokens
```
ekohl: Any reason this should be unsafe?

modules/puppetca_token_whitelisting/puppetca_token_whitelisting_token_storage.rb
juliantodt: -.-

modules/puppetca_token_whitelisting/puppetca_token_whitelisting_autosigner.rb
juliantodt: The whole block is already locked so nothing changes the file between read and write. You would not be able to safely write here, as that would try to lock the file again which is already locked.

modules/puppetca_token_whitelisting/puppetca_token_whitelisting_autosigner.rb
juliantodt: Added.

modules/puppetca_token_whitelisting/puppetca_token_whitelisting_autosigner.rb
ekohl: I wonder if the storage should have an operation where you pass in a block to delete tokens. It feels a bit like you're doing storage operations in the signer here.

juliantodt: Added a `TokenStorage::remove_if` method, take a look.

juliantodt:
As requested, I moved the puppet_sign script from the [Puppet-Foreman_Proxy PR](https://github.com/theforeman/puppet-foreman_proxy/pull/425) to here and added a couple of improvements.
@ekohl Is `extra/puppet_sign.rb` the correct place for it?

timogoebel (approved):

timogoebel:
@ekohl: Any further comments?

extra/puppet_sign.rb
```diff
@@ -0,0 +1,56 @@
...
+csr = STDIN.read
+
+if RbConfig::CONFIG['host_os'] =~ /freebsd|dragonfly/i
+  settings_file = '/usr/local/etc/foreman-proxy/settings.yml'
+else
+  settings_file = '/etc/foreman-proxy/settings.yml'
+end
+SETTINGS         = YAML.load_file(settings_file)
+PUPPETCA         = YAML.load_file(SETTINGS[:settings_directory] + '/puppetca.yml')
+protocol         = PUPPETCA[:enabled] == true ? 'https' : PUPPETCA[:enabled]
```
ekohl: Probably good if this also handles the value `false` and print a proper error

modules/puppetca/puppetca_api.rb
```diff
@@ -29,13 +29,27 @@ class Api < ::Sinatra::Base
...
     end
 
+    post "/validate" do
+      content_type :json
+      unless autosigner.respond_to?(:validate_csr)
+        log_halt 501, "Provider only supports trivial autosigning"
+      end
+      begin
+        request.body.rewind
+        autosigner.validate_csr(request.body.read) ? 200 : 404
```
ekohl, line 47: I'm wondering about the 404 code; will there be a need to distinguish between a proxy without this feature and one with? Probably not but it might be worth to consider 400 Bad Request for invalid CSRs.

ekohl (commented):
Overall it looks good and I'd be OK with merging. Just small edge cases you might want to look at.

modules/puppetca/puppetca_api.rb
juliantodt: Well the 404 here only gets returned when the CSR is valid but the token is not. A proxy without this feature returns 501 (line 43) and an invalid CSR raises an Exception leading to a 406 (line 49). Does that work for you?

extra/puppet_sign.rb
juliantodt: Added.

timogoebel:
@ekohl: Final ACK? You still "request changes" on this PR. :-)

ekohl (approved):

timogoebel:
Thanks, @juliantodt.

PR #593, opened 2018-07-04
Fixes for #23353 - Issues with default directory for included DHCP files

slueder:


lzap:
Hello, thanks for the contribution. The commit must be in the format

    Fixes #redmine_number - brief description

Please drop the `for` word otherwise our bot won't associate your ticket with redmine.

lzap:
You know what let me [test] this and I can change the message on merge.

modules/dhcp_common/isc/configuration_parser.rb
```diff
@@ -427,11 +427,13 @@ def visit_parse_tree_nodes(parse_tree, all_hosts, all_subnets, all_ignored, root
         end
 
         def parse_file(a_path)
+          a_path = @@config_path + '/' + a_path if File.dirname(a_path) == '.'
```
lzap: This should be:

    a_path = File.join(@config_path, a_path) if File.dirname(a_path) == '.'

Also I am not so sure about the test you perform here. Dot is not the only case, it can be also path like `../dhcp/some_file` which will lead to `..` that does not match. I suggest to convert the path to absolute path: https://ruby-doc.org/core-2.1.5/File.html#method-c-absolute_path

This would be (untested):

    absolute_path = File.absolute_path(a_path, @config_path)

This will resolve correctly for both absolute paths and relative paths.

modules/dhcp_common/isc/configuration_parser.rb
```diff
@@ -427,11 +427,13 @@ def visit_parse_tree_nodes(parse_tree, all_hosts, all_subnets, all_ignored, root
...
           File.open(a_path, 'r:ASCII-8BIT') {|f| conf.parse!(f.read, a_path)}
         end
 
         # returns all_subnets, all_hosts, root_group
         def subnets_hosts_and_leases(conf_as_string, filename)
+          @@config_path = File.dirname(filename)
```
lzap, line 438: Use `@cofig_path` - instance variable.

test/dhcp/conf_parser_test.rb
```diff
@@ -473,6 +473,11 @@ def test_include
                  included.select {|node| node.class == Proxy::DHCP::CommonISC::ConfigurationParser::IpV4SubnetNode}.map(&:subnet_address)
     assert_equal ['test.example.com'],
                  included.select {|node| node.class == Proxy::DHCP::CommonISC::ConfigurationParser::HostNode}.map(&:fqdn)
+    included = Proxy::DHCP::CommonISC::ConfigurationParser.new.subnets_hosts_and_leases(File.read('test/fixtures/dhcp/dhcp2.conf'), 'test/fixtures/dhcp/dhcp2.conf').flatten
+    assert_equal ['192.168.122.0', '192.168.123.0', '192.168.124.0', '192.168.1.0'],
+                 included.select {|node| node.class == Proxy::DHCP::CommonISC::ConfigurationParser::Ipv4Subnet}.map(&:subnet_address)
+    assert_equal ['test.example.com'],
+                 included.select {|node| node.class == Proxy::DHCP::CommonISC::ConfigurationParser::Host}.map(&:name)
   end
```
lzap: Please create separate test for this named `test_include_relative`. Test two cases:

* include "dhcp_subnets.conf"
* include "../dhcp/dhcp_subnets.conf"

Only assert one thing in the file, for example `assert_equal ['test.example.com'], ...` would be a good candidate. Others are not relevant.

test/fixtures/dhcp/dhcp2.conf
```diff
@@ -0,0 +1,3 @@
+# Test that the ISC DHCP config file parser includes files without an absolute or more specific relative path
+# If none is given, the file to be included will be looked up in the directory where the dhcp configuration file is located
+include "dhcp_subnets.conf";
```
lzap: Please give this file nicer name like `dhcp_include_relative.conf` instead `dhcpd2.conf` which is quite confusing.

lzap (changes requested):
Good direction, although I suggest to refactor it a bit. Thanks, keep going!

lzap:
Once changes are done, simply git commit --amend and push again forcing the changes. Also drop `for` from the commit message so the bot can associate it.

slueder:
All,

due to some internal hick-ups in using GIT i had to re-submit a new Pull request (https://github.com/theforeman/smart-proxy/pull/600) about this topic. So this one can be closed.

PR #596, opened 2018-07-24
Fixes #24364 - non-ascii logs are readable now

lzap:
Logs API is unable to return binary/malformed UTF strings which can end up in logs when request line is incorrect.

1) Start a proxy (upstream develop branch is affected as well):
2) Have "logs" module enabled
3) Send arbitrary binary data via TCP plain connection to HTTP endpoint: nc localhost 8448 < /usr/bin/bash
4) Call logs API: curl-proxy /logs

```
D, [2018-07-24T09:58:14.814774 ] DEBUG -- : "\xB0" from ASCII-8BIT to UTF-8 (Encoding::UndefinedConversionError)
/home/lzap/work/smart-proxy/modules/logs/logs_api.rb:15:in `encode'
/home/lzap/work/smart-proxy/modules/logs/logs_api.rb:15:in `to_json'
/home/lzap/work/smart-proxy/modules/logs/logs_api.rb:15:in `block in <class:LogsApi>'
```

The fix is simple - we convert into UTF-8 earlier replacing all invalid characters with `?`.

### DIFF
--- a/lib/proxy/log_buffer/decorator.rb
+++ b/lib/proxy/log_buffer/decorator.rb
@@ -46,7 +46,7 @@ module Proxy::LogBuffer
           # we accept backtrace, exception and simple string
           backtrace = backtrace.is_a?(Exception) ? backtrace.backtrace : backtrace
           backtrace = backtrace.respond_to?(:join) ? backtrace.join("\n") : backtrace
-          rec = Proxy::LogBuffer::LogRecord.new(nil, severity, message, backtrace, request_id)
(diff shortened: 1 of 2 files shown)


adamruzicka (changes requested):
Fixes the issue for me, but it seems test failures are related

lzap:
Corrected, thanks.

adamruzicka (approved):
Test failure unrelated, ACK

lzap:
Thanks [test] it was an issue on Jenkins perhaps.

dLobatog (approved):
Thanks @lzap & @adamruzicka !

dLobatog:
Should this make 1.18.1/1.19? I'd say yes

lzap:
Yes thanks, CPs are all pending, I set RM FIV as well.

PR #601, opened 2018-08-08
Fixes #24553 - Fix PuppetCa migrations

juliantodt:
The PuppetCA migration failed when the `hostname_whitelisting` settings file did not exist. This fixes this so that the migration succeeds anyway.


lzap:
Thank you very much for your contribution, this is now merged.

PR #602, opened 2018-08-09
Fixes #2720 - nsupdate master zone update setting

lzap:
This patch allows empty strings for default values. Before the patch, when you attempted to create an empty value for any setting which had a default value via

    default_settings :dns_server => 'localhost'

it led to error "Parameter dns_server is expected to have a non-empty value" and this was too strict.

Empty value for dns_server nsupdate setting now means that server is not set via nsupdate which leads to updating master server rather than specific one. See ticket description for more details.

### DIFF
--- a/config/settings.d/dns_nsupdate.yml.example
+++ b/config/settings.d/dns_nsupdate.yml.example
@@ -4,5 +4,8 @@
 #
 
 #:dns_key: /etc/rndc.key
-# use this setting if you are managing a dns server which is not localhost though this proxy
-#:dns_server: dns.domain.com
+
+# DNS server to talk to, usually localhost but can be an arbitrary (remote) host.
+# This can be also set to a blank value to let nsupdate to detect the master
+# server of a zone.
+#:dns_server: 127.0.0.1
--- a/lib/proxy/plugin_initializer.rb
+++ b/lib/proxy/plugin_initializer.rb
@@ -338,7 +338,7 @@ module ::Proxy::DefaultSettingsLoader
   end
 
   def validate_settings(plugin, config)
-    result = execute_validators(plugin.plugin_default_settings.keys.map { |k| {:name => :presence, :setting => k} }, config)
+    result = execute_validators(plugin.plugin_default_settings.keys.map { |k| {:name => :presence_allow_empty, :setting => k} }, config)
     result + execute_validators(plugin.validations, config)
   end
 
--- a/lib/proxy/plugin_validators.rb
+++ b/lib/proxy/plugin_validators.rb
@@ -42,4 +42,11 @@ module ::Proxy::PluginValidators
       true
     end
   end
+
+  class PresenceAllowEmpty < Base
+    def validate!(settings)
+      raise ::Proxy::Error::ConfigurationError, "Parameter '#{@setting_name}' is expected to be present" if settings[@setting_name].nil?
+      true
+    end
+  end
 end
--- a/modules/dns_nsupdate/dns_nsupdate_main.rb
+++ b/modules/dns_nsupdate/dns_nsupdate_main.rb
@@ -39,8 +39,10 @@ module Proxy::Dns::Nsupdate
       nsupdate_cmd = "#{@nsupdate} #{nsupdate_args}"
       logger.debug "running #{nsupdate_cmd}"
       @om = IO.popen(nsupdate_cmd, "r+")
-      logger.debug "nsupdate: executed - server #{@server}"
-      @om.puts "server #{@server}"
+      unless @server.nil? || @server.empty?
+        logger.debug "nsupdate: executed - server #{@server}"
+        @om.puts "server #{@server}"
+      end
     end
 
     def nsupdate_disconnect
--- a/lib/proxy/default_plugin_validators.rb
+++ b/lib/proxy/default_plugin_validators.rb
@@ -1,5 +1,9 @@
 class Proxy::DefaultPluginValidators
   def self.validators
-    {:file_readable => ::Proxy::PluginValidators::FileReadable, :presence => ::Proxy::PluginValidators::Presence}
+    {
+      :file_readable => ::Proxy::PluginValidators::FileReadable,
+      :presence => ::Proxy::PluginValidators::Presence,
+      :presence_allow_empty => ::Proxy::PluginValidators::PresenceAllowEmpty,
+    }
   end
 end
--- a/test/dns_nsupdate/dns_nsupdate_config_test.rb
+++ b/test/dns_nsupdate/dns_nsupdate_config_test.rb
@@ -13,6 +13,12 @@ class DnsNsupdateConfigTest < Test::Unit::TestCase
     assert_nil Proxy::Dns::Nsupdate::Plugin.settings.dns_key
   end
 
+  def test_nsupdate_emty_dnsserver_setting
+    Proxy::Dns::Nsupdate::Plugin.load_test_settings(dns_server: '')
+
+    assert_equal "", Proxy::Dns::Nsupdate::Plugin.settings.dns_server
+  end
+
   def test_nsupdate_gss_default_settings
     Proxy::Dns::NsupdateGSS::Plugin.load_test_settings({})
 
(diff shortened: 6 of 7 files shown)


config/settings.d/dns_nsupdate.yml.example
```diff
@@ -4,5 +4,7 @@
 #
 
 #:dns_key: /etc/rndc.key
-# use this setting if you are managing a dns server which is not localhost though this proxy
-#:dns_server: dns.domain.com
+
+# DNS server to talk to, usually localhost but can be arbitrary (remote) host.
+# This can be also set to blank value to send updates to the master server of the correct zone.
```
ekohl: Perhaps this could use a note that it uses nsupdate's built in support to automatically determine a master for a zone.

lib/proxy/plugin_validators.rb
```diff
@@ -42,4 +42,15 @@ def validate!(settings)
...
     end
   end
+
+  class PresenceAllowEmpty < Base
+    def validate!(settings)
+      setting_value = settings[@setting_name]
+
+      empty_value = setting_value.nil?
+
+      raise ::Proxy::Error::ConfigurationError, "Parameter '#{@setting_name}' is expected to be present" if empty_value
```
ekohl: I think the name `empty_value` here is misleading because the class name states it allows empty but the test fails if it's empty.

lib/proxy/plugin_initializer.rb
```diff
@@ -320,7 +320,7 @@ def load_classes
   end
 
   def validate_settings(plugin, config)
-    result = execute_validators(plugin.plugin_default_settings.keys.map {|k| {:name => :presence, :setting => k}}, config)
+    result = execute_validators(plugin.plugin_default_settings.keys.map {|k| {:name => :presence_allow_empty, :setting => k}}, config)
```
ekohl: I can't quite see the implications of this change. @witlessbird?

modules/dns_nsupdate/dns_nsupdate_main.rb
```diff
@@ -39,8 +39,10 @@ def nsupdate_connect
       nsupdate_cmd = "#{@nsupdate} #{nsupdate_args}"
       logger.debug "running #{nsupdate_cmd}"
       @om = IO.popen(nsupdate_cmd, "r+")
-      logger.debug "nsupdate: executed - server #{@server}"
-      @om.puts "server #{@server}"
+      unless @server.nil? && @server.empty?
```
ekohl: I think you meant `||` here

ekohl (changes requested):
Currently we don't have any tests for `nsupdate_connnect` but I think it would be good to add those. Especially since you already had a logic bug in there.

lzap:
Amended the changes.

lzap:
Incorporated changes from this and the other PR (puppet):

https://github.com/theforeman/puppet-foreman_proxy/pull/444/comment#issuecomment-412047283

lib/proxy/plugin_initializer.rb
ekohl: I do think that changing the default here for every plugin default can open us up to unexpected errors. Not all code might be able to deal with it.

lzap: It only allows empty string, not `nil`.

ekohl: Still, changing this for every plugin sounds incorrect. Changes should be as small as possible and I can't tell exactly what this change will impact.

ekohl (changes requested):
Could you add a test to `test/dns_nsupdate/dns_nsupdate_config_test.rb` that verifies an empty input is accepted? The gss_nsupdate can use a similar change.

lib/proxy/plugin_initializer.rb
lzap: I am not sure how code needs to be ready to deal with empty string, you can set it to "x" or some UTF-8 emoticon character today and the same code would need to work it out. Allowing nil that's another story, but that's not what I do in this PR.

lzap:
Good point, amended the commit with the test.

lzap:
@ekohl when you have time, thanks!

mmoll:
@lzap please rebase and fix the rubocop error

lzap:
Done thanks!

lib/proxy/plugin_initializer.rb
ekohl: Yet, without this change the test pass which makes me think we shouldn't change this. Could we remove the PresenceAllowEmpty part of this PR?

config/settings.d/dns_nsupdate.yml.example
```diff
@@ -4,5 +4,7 @@
 #
 
 #:dns_key: /etc/rndc.key
-# use this setting if you are managing a dns server which is not localhost though this proxy
-#:dns_server: dns.domain.com
+
+# DNS server to talk to, usually localhost but can be an arbitrary (remote) host.
+# This can be also set to a blank value to let nsupdate to detect master server of a zone.
```
ekohl: detect *the* master server

ekohl (changes requested):

config/settings.d/dns_nsupdate.yml.example
lzap: Fixed.

lzap:
>Yet, without this change the test pass which makes me think we shouldn't change this. Could we remove the PresenceAllowEmpty part of this PR?

Tests do not pass with this line reverted. The PresenceAllowEmpty part of this PR is actually the reason why I am doing this so I can't remove this as the purpose of this PR is to actually allow user to pass blank server name to let `nsupdate` to decide hostname to connect to via DNS.

Please elaborate your argument, the `presence` validator is simply too strict, blank values should be allowed. I can rewrite the patch via some explicit value something like 'DONOTSEND' or whatever in the worst case.

lzap:
This should be now green.

mmoll:
test failure unrelated :green_heart:

ekohl:
I'm still not entirely sure if this won't break anything else but I don't have time to properly test this right now.

lzap:
Hello, if you start a review please finish it ;-)

tbrisker:
please rebase @lzap

lzap:
Rebased.

lib/proxy/plugin_initializer.rb
```diff
@@ -338,7 +338,7 @@ def load_classes
   end
 
   def validate_settings(plugin, config)
-    result = execute_validators(plugin.plugin_default_settings.keys.map { |k| {:name => :presence, :setting => k} }, config)
+    result = execute_validators(plugin.plugin_default_settings.keys.map { |k| {:name => :presence_allow_empty, :setting => k} }, config)
```
ezr-ondrej, line 341: I read this, that this is only changing the validation for the default settings.

ezr-ondrej (approved):
Even though it doesn't make necessarily sense for every default setting, I'm for letting it in as it needs to be set explicitly to empty string by user, what is ok with me.

ezr-ondrej:
@ekohl if I read correctly, other validations, than for default settings are executed on https://github.com/theforeman/smart-proxy/blob/588f810e3b0812d2ab2c3d04cff80e825f2cc63e/lib/proxy/plugin_initializer.rb#L342 so this is changing really only the default settings validations.

lib/proxy/plugin_initializer.rb
ekohl, line 341: This line is my concern. My reading of this is that it's changing the validator for every setting with a specified validator from `presence` to `presence_allow_empty`. That means suddenly **all** settings with a default are allowed to be empty strings. The alternative would be to remove the default value just for `dns_server` and rely on the installer to set the correct default. That would be a much more minimal patch since the `PresenceAllowEmpty` validator is no longer needed.

ekohl:
> @ekohl if I read correctly, other validations, than for default settings are executed on
> 
> https://github.com/theforeman/smart-proxy/blob/588f810e3b0812d2ab2c3d04cff80e825f2cc63e/lib/proxy/plugin_initializer.rb#L342

This is incorrect. It is applying the validator for all settings with a default, regardless of the actual value. In short: if there is a default value, a value **must** be provided and empty values in YAML are disallowed. Dropping this allows users to misconfigure the smart proxy.

lzap:
Enough time spent, thanks all.

PR #607, opened 2018-09-14
refs #23211 - make PuppetCA token provider more resilient

timogoebel:


timogoebel:
This fixes three issues I found with the current implementation:

* The helper script calls the wronmg URL. Bummer.
* The autosigner implementation for `#disable` needs to return `nil`.
* The `certificate` setting is `nil` by default. If we specify this explicitly smart-proxy treats it as a required parameter.

timogoebel:
@ekohl: Can you review and merge?

extra/puppet_sign.rb
```diff
@@ -32,8 +32,8 @@ def log(message)
 protocol         = PUPPETCA[:enabled] == true ? 'https' : PUPPETCA[:enabled]
 port             = protocol == 'https' ? SETTINGS[:https_port] : SETTINGS[:http_port]
 fqdn             = Socket.gethostbyname(Socket.gethostname).first
-# e.g. https://hostname.localdomain.com:8443/puppet/ca/autosign
-uri              = URI.parse("#{protocol}://#{fqdn}:#{port}/puppet/ca/autosign")
+# e.g. POST https://hostname.localdomain.com:8443/puppet/ca/verify
+uri              = URI.parse("#{protocol}://#{fqdn}:#{port}/puppet/ca/verify")
```
juliantodt: This should be `/puppet/ca/validate`. See https://github.com/theforeman/smart-proxy/blob/develop/modules/puppetca/puppetca_api.rb#L40

juliantodt (changes requested):
Nice catches! Thanks for the fixes. While you're at it, could you also change https://github.com/theforeman/smart-proxy/blob/develop/modules/puppetca_token_whitelisting/puppetca_token_whitelisting_autosigner.rb#L10 to an Integer as this currently breaks the certificate generation?

timogoebel:
@juliantodt: Fixed your comments. Can you ack now?

juliantodt (approved):

timogoebel:
@lzap: Do you mind doing a quick merge? This fixes some issues we found when testing this in a production deployment.

extra/puppet_sign.rb
```diff
@@ -26,14 +26,14 @@ def log(message)
...
+  log('PuppetCA smart-proxy module is not enabled!')
   exit 1
 end
 protocol         = PUPPETCA[:enabled] == true ? 'https' : PUPPETCA[:enabled]
 port             = protocol == 'https' ? SETTINGS[:https_port] : SETTINGS[:http_port]
 fqdn             = Socket.gethostbyname(Socket.gethostname).first
-# e.g. https://hostname.localdomain.com:8443/puppet/ca/autosign
-uri              = URI.parse("#{protocol}://#{fqdn}:#{port}/puppet/ca/autosign")
+# e.g. POST https://hostname.localdomain.com:8443/puppet/ca/validate
+uri              = URI.parse("#{protocol}://#{fqdn}:#{port}/puppet/ca/validate")
```
lzap, line 36: This is quite a change I can hardly comment but this looks like some `extra` script.

lzap (approved):
I am good thanks.

lzap:
There you have it! Thanks.

PR #609, opened 2018-09-24
Fixes #24978 - support for DDNS lease entries

lzap:
We simply need to ignore these lines, parser was going crazy about incorrect literals.

### DIFF
--- a/modules/dhcp_common/isc/configuration_parser.rb
+++ b/modules/dhcp_common/isc/configuration_parser.rb
@@ -269,10 +269,16 @@ module Proxy
         def option
           Rsec::Fail.reset
           keyword_option = word('option').fail 'keyword_option'
-          keyword_supersede = word('supersede').fail 'keyword_option'
+          keyword_supersede = word('supersede').fail 'keyword_supersede'
           option_name = /[\w\.-]+/.r
           seq_(keyword_option | keyword_supersede, option_name, '='.r._?, option_values, EOSTMT) {|maybe_supersede, name, _, values, _| OptionNode[maybe_supersede == 'supersede', name, values]} |
               vendor_option_space | filename | next_server
+        end
+
+        def set
+          Rsec::Fail.reset
+          keyword_set = word('set').fail 'keyword_set'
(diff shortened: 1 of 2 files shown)


ifreislich:
With this patch the parser hasn't choked (yet) on my dhcpd.leases file.  Some of the originally offending data is in the lease file.

lzap:
Want to take a look @ik_5? See the issue for the whole story.

ifreislich:
Been running with this patch for 5 days.  No issues so far.

modules/dhcp_common/isc/configuration_parser.rb
```diff
@@ -269,12 +269,18 @@ def vendor_option_space
         def option
           Rsec::Fail.reset
           keyword_option = word('option').fail 'keyword_option'
-          keyword_supersede = word('supersede').fail 'keyword_option'
+          keyword_supersede = word('supersede').fail 'keyword_supersede'
```
ik5, line 272: nice catch

ik5 (approved):
LGTM and tests are working as well.

mmoll (approved):
looks good to me

mmoll:
merged, díky @lzap!

modules/dhcp_common/isc/configuration_parser.rb
lzap: Now the question is what this is for? ;-) I have no idea.

ik5: Based on: http://rsec.herokuapp.com/ref

>word(pattern) helper
>
>A word is a token wrapped with word boundaries
>
>example:
>
>```ruby
>assert_equal ['yes', '3'], seq('yes', '3').parse('yes3')
>assert_equal INVALID, seq(word('yes'), '3').parse('yes3')
>```

and
>
>fail(*tokens)
>
>When parsing fails on the preceding term, show "expect tokens" error for that tokens value.

lzap: I felt like this is something like that :-)

ik5:
I went and checked it out for this PR :)

modules/dhcp_common/isc/configuration_parser.rb
ik5: I checked it out prior of me approving it :P

PR #610, opened 2018-09-28
PuppetCA: Add Puppet-6 support

quadriq:
fixes #25065

#### Motivation
with puppet-6 all commands like `puppet cert *` are depricateded. Instead one should use `puppetserver ca *` commands for certificates.

#### Possible Solution
This PR adding a "puppet6" functionality for Puppet-CA. It can be activated by defining `:puppet_6: true` parameter in `puppetca.yml` settings file of the proxy

Please review.

### DIFF
--- a/modules/puppetca/puppetca_puppet_cert.rb
+++ b/modules/puppetca/puppetca_puppet_cert.rb
@@ -19,6 +19,14 @@ module Proxy::PuppetCa
 
     # list of all certificates and their state/fingerprint
     def list
+      if to_bool(::Proxy::PuppetCa::Plugin.settings.puppet_6, false)
+        list_puppet_6
+      else
+        list_legacy
+      end
+    end
+
+    def list_legacy
       find_puppetca
       command = "#{@sudo} #{@puppetca} --list --all"
       logger.debug "Executing #{command}"
@@ -36,6 +44,44 @@ module Proxy::PuppetCa
       # note that this ignores certificates which were revoked multiple times, displaying only the last
       # revocation state
       # additionally, we don't merge revocation info if the host has a pending certificate request
+      hash.merge(ca_inventory) {|key, h1, h2| h1[:state] == "pending" ?  h1 : h1.merge(h2)}
+    end
+
+    def list_puppet_6
+      find_puppetca_6
+      command = "#{@sudo} #{@puppetca} list --all"
+      logger.debug "Executing #{command}"
+      response = `#{command}`
+      unless $? == 0
+        logger.warn "Failed to run puppetca: #{response}"
+        raise "Execution of puppetca failed, check log files"
+      end
+
+      hash_name = "unknown"
+      hash = {}
+      valid = {}
+      revoked = {}
+      pending = {}
+
+      response.split("\n").each do |line|
+        if line.include? 'Signed Certificates:'
+          hash_name = 'valid'
+          next
+        elsif line.include? 'Requested Certificates:'
+          hash_name = 'pending'
+          next
+        elsif line.include? 'Revoked Certificates:'
+          hash_name = 'revoked'
+          next
+        end
+
+        (eval "#{hash_name}").merge! certificate_puppet_6(line, hash_name) rescue logger.warn("Failed to parse line: #{line}")
+      end
+
+      hash.merge! valid
+      hash.merge! revoked
+      hash.merge! pending
+
       hash.merge(ca_inventory) {|key, h1, h2| h1[:state] == "pending" ?  h1 : h1.merge(h2)}
     end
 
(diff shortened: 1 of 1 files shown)


ekohl:
In the puppet module we have a `puppet_version` setting. I think that's more flexible than a boolean. Because they're separate modules you will need to duplicate that setting.

modules/puppetca/puppetca_puppet_cert.rb
```diff
@@ -79,6 +125,23 @@ def find_puppetca
       end
     end
 
+    def find_puppetca_6
+      # Tell puppetca to use the ssl dir that Foreman has been told to use
+      @puppetca = "/opt/puppetlabs/bin/puppetserver ca "
+
+      if to_bool(::Proxy::PuppetCa::Plugin.settings.puppetca_use_sudo, true)
```
ekohl, line 132: I'd be nice to move this to a helper function instead of duplicating

modules/puppetca/puppetca_puppet_cert.rb
```diff
@@ -130,21 +203,32 @@ def revoked_serials(crl_cert_contents)
     end
 
     def puppetca mode, certname
+
       raise "Invalid mode #{mode}" unless mode =~ /^(clean|sign)$/
-      find_puppetca
       certname.downcase!
-      command = "#{@sudo} #{@puppetca} --#{mode} #{certname}"
+
+      if to_bool(::Proxy::PuppetCa::Plugin.settings.puppet_6, false)
```
ekohl, line 210: I think it would be cleaner if there also was a `find_puppetca_legacy` and `find_puppetca` would abstract it away.

modules/puppetca/puppetca_puppet_cert.rb
```diff
@@ -39,6 +47,44 @@ def list
...
+          next
+        elsif line.include? 'Requested Certificates:'
+          hash_name = 'pending'
+          next
+        elsif line.include? 'Revoked Certificates:'
+          hash_name = 'revoked'
+          next
+        end
+
+        (eval "#{hash_name}").merge! certificate_puppet_6(line, hash_name) rescue logger.warn("Failed to parse line: #{line}")
```
ekohl, line 78: Isn't there a way to avoid using `eval`? I don't see why you need to store `valid`, `revoked` and `pending` in separate hashes if they're going to be merged anyway. Otherwise I'd suggest something like:
```ruby
results = {
  'valid' => {},
  'revoked' => {},
  'pending' => {},
  'unknown' => {},
}
```

timogoebel:
@quadriq: Thanks for your efforts here.

I think we can call the [new ca api](https://puppet.com/docs/puppet/6.0/http_api/http_certificate.html) directly via http and do not need to parse the command outputs. I believe this is faster, more flexible and more robust. And we can implement proper test more easily.

In addition, we should implement this as a provider.

modules/puppetca/puppetca_puppet_cert.rb
```diff
@@ -19,6 +19,14 @@ def clean certname
 
     # list of all certificates and their state/fingerprint
     def list
+      if to_bool(::Proxy::PuppetCa::Plugin.settings.puppet_6, false)
```
timogoebel, line 22: I think it's better to inject an instance of a puppet 6 specific provider class here: 
https://github.com/theforeman/smart-proxy/blob/9d4605334ad7fec1b2e344cba15f15567db9c00d/modules/puppetca/plugin_configuration.rb#L10

ekohl:
> I think we can call the [new ca api](https://puppet.com/docs/puppet/6.0/http_api/http_certificate.html) directly via http and do not need to parse the command outputs. I believe this is faster, more flexible and more robust. And we can implement proper test more easily.
> 
> In addition, we should implement this as a provider.

Looks like I didn't actually comment this but this was also my thought. Looks like that CA API should cover our needs and was actually already available in earlier versions. For bonus points I think it means we can separate the proxy and puppet server onto separate physical servers because they no longer share a filesystem.

timogoebel:
@ekohl: Looks like we have found a topic for the hack day on Wednesday. ;-)

quadriq:
I definitely agree to use Puppet API instead of parsing files. However I would be able to look at the code only next week.

modules/puppetca/puppetca_puppet_cert.rb
```diff
@@ -39,6 +47,44 @@ def list
       hash.merge(ca_inventory) {|key, h1, h2| h1[:state] == "pending" ?  h1 : h1.merge(h2)}
     end
 
+    def list_puppet_6
```
timogoebel, line 50: Unfortunately, this does not list `serial`, `not_before` and `not_after` anymore.

timogoebel:
@quadriq: I just pushed a WIP implementation that used the http api to request the certificates. #615

mmoll:
GH-615 will be the way to go

PR #612, opened 2018-10-05
remove has_rdoc from gemspec

mmoll:
```
NOTE: Gem::Specification#has_rdoc= is deprecated with no replacement. It will be removed on or after 2018-12-01.
Gem::Specification#has_rdoc= called from /usr/local/share/foreman-proxy/smart_proxy.gemspec:14.
```


lzap (approved):
Sure.

lzap:
Thank you very much for your contribution, this is now merged.

PR #613, opened 2018-10-14
Fixes #25293 - Puma support for smart proxy

ik5:
Work in progress (still did not test it or wrote any tests for it), loading puma (or) webrick as an HTTP Server.

Started earlier today to work on it, most of it done, but not all :)

### DIFF
--- a/bundler.d/puma.rb
+++ b/bundler.d/puma.rb
@@ -1,0 +1,11 @@
+group :puma do
+  if RUBY_VERSION < '2.3'
+    # Important!
+    # The last actual version that supports v2.0.0 up to v2.2.0 is 3.10.0
+    # Puma version 3.11.0 changed the usage of socket to a feature that is not
+    # supported by Ruby 2.2.0 and lower, and it causes a crash on TLS!
+    gem 'puma', '3.10.0', :require => 'puma'
+  else
+    gem 'puma', '~>3.12', :require => 'puma'
+  end
+end
--- a/config/settings.yml.example
+++ b/config/settings.yml.example
@@ -58,6 +58,11 @@
 # default values for https_port is 8443
 #:https_port: 8443
 
+# http server type configuration
+# A string that indicates the server type (possible values: 'webrick', 'puma').
+# Default value is 'webrick'
+#:http_server_type: 'puma'
+
 # Log configuration
 # Uncomment and modify if you want to change the location of the log file or use STDOUT or SYSLOG values
 #:log_file: /var/log/foreman-proxy/proxy.log
--- a/lib/launcher.rb
+++ b/lib/launcher.rb
@@ -1,8 +1,20 @@
+require 'bundler'
 require 'proxy/log'
 require 'proxy/sd_notify'
 require 'proxy/settings'
 require 'proxy/signal_handler'
 require 'thread'
+require 'rack'
+require 'webrick'
+begin
+  require 'puma'
+  require 'rack/handler/puma'
+  require 'puma_patch'
+  $HAS_PUMA = true
+rescue LoadError
+  $stderr.puts 'Puma was requested but not installed'
+  $HAS_PUMA = false
+end
 
 module Proxy
   class Launcher
@@ -12,18 +24,24 @@ module Proxy
 
     def initialize(settings = SETTINGS)
       @settings = settings
+      @settings.http_server_type = Proxy::SETTINGS.http_server_type.to_sym
+      if @settings.http_server_type == :puma && !$HAS_PUMA
+        logger.warn 'Puma was requested but not installed, falling back to webrick'
+        @settings.http_server_type = :webrick
+      end
+      @servers = []
     end
 
     def pid_path
-      settings.daemon_pid
+      @settings.daemon_pid
     end
 
     def http_enabled?
-      !settings.http_port.nil?
+      !@settings.http_port.nil?
     end
 
     def https_enabled?
-      settings.ssl_private_key && settings.ssl_certificate && settings.ssl_ca_file
+      @settings.ssl_private_key && @settings.ssl_certificate && @settings.ssl_ca_file
     end
 
     def plugins
@@ -46,7 +64,7 @@ module Proxy
 
       {
         :app => app,
-        :server => :webrick,
+        :server => @settings.http_server_type,
         :DoNotListen => true,
         :Port => http_port, # only being used to correctly log http port being used
         :Logger => ::Proxy::LogBuffer::Decorator.instance,
@@ -72,8 +90,8 @@ module Proxy
       ssl_options |= OpenSSL::SSL::OP_NO_SSLv3 if defined?(OpenSSL::SSL::OP_NO_SSLv3)
       ssl_options |= OpenSSL::SSL::OP_NO_TLSv1 if defined?(OpenSSL::SSL::OP_NO_TLSv1)
 
-      if Proxy::SETTINGS.tls_disabled_versions
-        Proxy::SETTINGS.tls_disabled_versions.each do |version|
+      if @settings.tls_disabled_versions
+        @settings.tls_disabled_versions.each do |version|
           constant = OpenSSL::SSL.const_get("OP_NO_TLSv#{version.to_s.gsub(/\./, '_')}") rescue nil
 
           if constant
@@ -85,21 +103,31 @@ module Proxy
         end
       end
 
-      {
+      app_details = {
         :app => app,
-        :server => :webrick,
+        :server => @settings.http_server_type,
         :DoNotListen => true,
         :Port => https_port, # only being used to correctly log https port being used
         :Logger => ::Proxy::LogBuffer::Decorator.instance,
         :ServerSoftware => "foreman-proxy/#{Proxy::VERSION}",
         :SSLEnable => true,
         :SSLVerifyClient => OpenSSL::SSL::VERIFY_PEER,
-        :SSLPrivateKey => load_ssl_private_key(settings.ssl_private_key),
-        :SSLCertificate => load_ssl_certificate(settings.ssl_certificate),
-        :SSLCACertificateFile => settings.ssl_ca_file,
+        :SSLCACertificateFile => @settings.ssl_ca_file,
         :SSLOptions => ssl_options,
         :daemonize => false
       }
+      case @settings.http_server_type
+      when :webrick
+        app_details[:SSLPrivateKey] = load_ssl_private_key(@settings.ssl_private_key)
+        app_details[:SSLCertificate] = load_ssl_certificate(@settings.ssl_certificate)
+      when :puma
+        app_details[:SSLArgs] = {
+          :ca => @settings.ssl_ca_file,
+          :key => @settings.ssl_private_key,
+          :cert => @settings.ssl_certificate
+        }
+      end
+      app_details
     end
 
     def load_ssl_private_key(path)
@@ -147,17 +175,55 @@ module Proxy
       retry
     end
 
-    def webrick_server(app, addresses, port)
+    def add_puma_server(app, addresses, port, conn_type)
+      # IMPORTANT:
+      # The following code takes only a single host.
+      # The current reason for it, is that "run" is blocking, and in order to
+      # add support for more hosts, additional threads requires to be created
+      address = addresses.first
+      address = '0.0.0.0' if address == '*'
+      if conn_type == :ssl
+        host = "ssl://#{address}/"
+        require 'cgi'
+        query_list = []
+        app[:SSLArgs].each_pair do |name, value|
+          query_list << "#{CGI::escape(name.to_s)}=#{CGI::escape(value)}"
+        end
+        host = "#{host}?#{query_list.join('&')}"
+      else
+        host = address
+      end
+      Rack::Handler::Puma.run(app[:app],
+                              Verbose: true,
+                              Port: port,
+                              Host: host
+                             )
+    end
+
+    def add_webrick_server(app, addresses, port)
       server = ::WEBrick::HTTPServer.new(app)
-      addresses.each {|a| server.listen(a, port)}
-      server.mount "/", Rack::Handler::WEBrick, app[:app]
+      addresses.each { |a| server.listen(a, port) }
+      server.mount '/', Rack::Handler::WEBrick, app[:app]
       server
+    end
+
+    def add_threaded_server(server_name, conn_type, app, addresses, port)
+      case server_name
+      when :webrick
+        Thread.new do
+          @servers << add_webrick_server(app, addresses, port).start
+        end
+      when :puma
+        Thread.new do
+          add_puma_server(app, addresses, port, conn_type)
+        end
+      end
     end
 
     def launch
       raise Exception.new("Both http and https are disabled, unable to start.") unless http_enabled? || https_enabled?
 
-      if settings.daemon
+      if @settings.daemon
         check_pid
         Process.daemon
         write_pid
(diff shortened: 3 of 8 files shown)


lib/launcher.rb
```diff
@@ -10,6 +10,15 @@ class Launcher
 
     attr_reader :settings
 
+    RACK_SERVERS = {
+      :webrick => Rack::Handler::WEBrick,
+      :puma => Rack::Handler::Puma
```
ohadlevy: won't this will require both gems to load? e.g. if you dont have puma gem, this line will break?

ik5:
It does not work yet, and that's one of the reasons, also tests will be
broken at this time, due to refactoring of the code.

Also there are places still hard coded for webrick.

That's why I placed it as wip 😁

lib/launcher.rb
lzap: Does it hurt to load them both? If it don't let's ignore it? If it does, then let's definitely take care of this.

lib/launcher.rb
```diff
@@ -87,7 +96,7 @@ def https_app(https_port, plugins = https_plugins)
 
       {
         :app => app,
-        :server => :webrick,
+        :server => Proxy::SETTINGS.http_server_type || :webrick,
```
lzap: This does not correspondend with the line above `:server => Proxy::SETTINGS.http_server_type.to_sym,` so either one or another. The settings stack handles the defaults, so I think the version from 58 is better.

bundler.d/puma.rb
```diff
@@ -0,0 +1 @@
+gem 'puma', '3.11.4'
```
lzap: This should be in `group :puma` so users can turn it on or off via bundler. In RPM deployments, we need to do packaging change and do subpackage so installing `foreman-proxy-puma` will actually grab the dependency. Also I'd probably add '~> 3.11' requirement, not sure about the gemspec syntax tho.

lib/launcher.rb
lzap: So it looks like webrick is always present, dependency of sinatra or something. But we should be able to install foreman-proxy without puma bundler group, so this needs to be taken care of.

lzap:
Nicely done, if it boots up and can be properly disabled without the puma dependency, let's ship it to hear from our users after 1.20 is branched. We can test it many times, but only our community will really tell us.

lib/launcher.rb
ik5: so instead of the original conversation, you want it more plugable?
that is, if it present, then load all of it requirements?

ik5:
there are basics puma, but it does not handle the actual execution for the route properly, and still on it, and when that works, I will have to do two things:

A. Fix tests
B. Add tests for Puma
C. Group Puma, and only if it exists load things that belongs to it.

config/settings.yml.example
```diff
@@ -58,6 +58,11 @@
 # default values for https_port is 8443
 #:https_port: 8443
 
+# http server type configuration
+# A string that indicates the server type (possible values: 'webrick', 'puma').
+# Default value is 'webrick'
+#:http_server_type: 'puma'
```
iNecas, line 64: you're suggesting 'puma' but expecting `:puma` later in the code (when I tried 'puma', the server didn't even start

```
D, [2018-10-16T13:59:47.376959 ] DEBUG -- : ["/home/inecas/Projects/ws/foreman-rex/smart-proxy/lib/launcher.rb:172:in `block in add_server'", "/home/inecas/Projects/ws/foreman-rex/smart-proxy/lib/launcher.rb:172:in `each'", "/home/inecas/Projects/ws/foreman-rex/smart-proxy/lib/launcher.rb:172:in `add_server'", "/home/inecas/Projects/ws/foreman-rex/smart-proxy/lib/launcher.rb:192:in `block in launch'"]
```

iNecas:
Also, ssl doesn't work at all yet

config/settings.yml.example
ik5: I know, already fixed that, didn't push it yet :)

ik5:
> Also, ssl doesn't work at all yet

I haven't tested it yet, because the actual routing crashes, and now working on that part.

bundler.d/puma.rb
```diff
@@ -0,0 +1,7 @@
+group :puma do
+  if RUBY_VERSION < '2.2'
+    gem 'puma', '3.11.4', :require => 'puma'
+  else
+    gem 'puma', :require => 'puma'
+  end
```
lzap, line 10: Holy Moly, we are lucky that 3.11 was the last version supporting Ruby 2.0. Latest is 3.12 which only supports Ruby 2.2+. Unfortunately, SCLing proxy is imminent.

Can you avoid open-ended version requirement on line 5? Packagers will require us to have a specific version, something like '~> 3.12' or something like that.

bundler.d/development.rb
```diff
@@ -4,6 +4,9 @@
   gem "ruby-debug19", :platforms => :ruby_19
   gem 'single_test'
   gem 'pry'
+  gem 'pry-rescue'
+  gem 'pry-stack_explorer'
+  gem 'pry-byebug'
```
lzap: These belong to `bundler.d/Gemfile.local.rb`

lib/puma_patch.rb
```diff
@@ -0,0 +1,18 @@
...
+    end
+  end
+
+  def start
+    self.run.join
+  end
+end
+
+require 'puma'
+Puma::Server.class_eval { include PumaPatch }
```
lzap: I don't like monkey patching just because of one or two simple calls. In the method where you call `mount` you already have `case server_name` anyway, I suggest to do the mapping there, or in a launcher helper.

bundler.d/development.rb
ik5: ack, I thought it can help debug on development, but np :)

lzap: It can, we tend to be little more strict in core, I'd probably follow this rule. Everyone use her own tools.

lib/puma_patch.rb
ik5: ack

lib/launcher.rb
```diff
@@ -7,7 +7,7 @@
 require 'rack'
 require 'webrick'
 begin
-  require 'puma'
+  Bundler.require(:puma)
```
lzap: Please test this on RPM installation, I think this will not work. We actually turn bundler off for RPM installations and use special stub called `bundler_ext` so it works seamlessly with Bundler for development.

lib/launcher.rb
```diff
@@ -167,10 +191,33 @@ def launch
...
       http_app = http_app(settings.http_port)
       https_app = https_app(settings.https_port)
-      install_webrick_callback!(http_app, https_app)
+      install_http_server_callback!(http_app, https_app)
+
+      http_server_name = @settings.http_server_type
+      https_server_name = @settings.https_server_type
+      if https_app
+        t1 = case https_server_name
+             when :webrick
```
timogoebel: I would extract this to a method `add_server()` then then calls `add_webrick_server` or `add_puma_server` to reduce the duplication.

ik5: executing the server is blocking, so at the moment actually playing around with my threading design.

the need for it, is to support multiple hosts, that the current way of working as a server (puma that is) is blocking.

lib/launcher.rb
ik5: Will check it out at the end of my work (soon, very soon :) )

ik5:
Okay, almost there, now working on making SSL/TLS to work.

Important thing, after investing a lot of time testing it out, I can't bind to many hosts per port, only for a single one.
The reason is that I will require to manage multiple threads - one per host and port/connection type. The reason for it, is that each server running entry is block instead of non-block.
In order to manage multiple hosts/connection type, there is a need to manage a threadpool and other resources management system.

I think that as another task, such management can be written,

ik5:
Several notes:

A. Last working version of Puma with Ruby v2.0.0 is 3.10.0 (https://github.com/puma/puma/issues/1662).
B. Puma's version of 3.10.0 does not support of filtering Ciphers in any way (so no code for that).
C. Currently fixing some tests, Ruby 2.0.0 have bug with Kerberos that makes it crash with the tests, so handpicking the tests and testing them.

Once the tests are ready I'll remove the WIP part. 
@lzap now it's more or less ready for PR review regarding the sources of the logic without much of rewrites :)

lib/puma_patch.rb
```diff
@@ -0,0 +1,11 @@
+
+class PumaMapperMiddleware
+  def initialize(app)
```
iNecas: Is this middleware needed for anything right now?

lib/launcher.rb
```diff
@@ -12,18 +25,25 @@ class Launcher
 
     def initialize(settings = SETTINGS)
       @settings = settings
+      @settings.http_server_type = Proxy::SETTINGS.http_server_type.to_sym
+      @settings.https_server_type = Proxy::SETTINGS.https_server_type.to_sym
+      if @settings.http_server_type == :puma && !$HAS_PUMA
```
iNecas, line 28: I don't think we need an extra `$GLOBAL_VARIABLE`, we can just check on `defined? Puma`

lib/launcher.rb
```diff
@@ -1,8 +1,21 @@
+require 'bundler'
 require 'proxy/log'
 require 'proxy/sd_notify'
 require 'proxy/settings'
 require 'proxy/signal_handler'
 require 'thread'
+require 'rack'
+require 'webrick'
+require 'pry'
```
iNecas: still a leftover from requiring `pry` for development

lib/launcher.rb
```diff
@@ -1,8 +1,21 @@
+require 'bundler'
 require 'proxy/log'
 require 'proxy/sd_notify'
 require 'proxy/settings'
 require 'proxy/signal_handler'
 require 'thread'
+require 'rack'
+require 'webrick'
```
iNecas, line 8: Do we need to require `webrick` in case we will not use it?

config/settings.yml.example
```diff
@@ -58,6 +58,15 @@
...
+# https server type configuration
+# A string that indicates the server type (possible values: 'webrick', 'puma').
+# Default value is 'webrick'
+#:https_server_type: 'puma'
```
iNecas: I don't think we really need separate server_type setting for http and https and I think it's just misleading. One option should rule them all

iNecas:
When running via `bundle exec bin/smart-proxy`, I can't quit the process with `CTRL+C`, when running without `bundle exec` it fails with `LoadError: cannot load such file -- puma`. I think at least one of the variants should work

lib/puma_patch.rb
ik5: Yes, it is the only way for Puma to enter the route and deal with it.

lib/launcher.rb
ik5: my issue with `define puma` is that if something else regarding puma was not loaded we do not know of it, and then looking for every point that is needed is a bit of an issue in my opinion.

You will end up either writing a function that does this, or have duplicate places that check that code.

So knowing that everything was loaded properly, even on other locations at the project when needed is a bit easier in my opinion and easier for the long run.

lib/launcher.rb
ik5: that's my bad, I added it when debug puma crashes, removing it

lib/launcher.rb
ik5: if the loading of puma failed it is the fallback, and at the moment the `SETTINGS constant` is not available for me to know if to load it or not.

Do you have a better idea?

lib/puma_patch.rb
ik5: there is another way of doing it, but then we will ignore `rack`, and that by using `add_xxx_listener` ([for example](https://github.com/puma/puma/blob/v3.10.0/lib/puma/binder.rb#L261)) of Puma, but we can then support multiple Ethernet devices, and that's the balance I was playing with.

config/settings.yml.example
ik5: okay unifying them

ik5:
@iNecas the pressing of `^C` works for me, the only time it didn't was when i was inside a breakpoint.

screenshot at 2018-10-24 09-26-06

ik5:
fixed all known test issues (ruby 2.0.0 crashes on a bug of kerberos), so using the following test mechanism to test my code :)

ik5:
[test smart-proxy]

ik5:
[test]

lib/puma_patch.rb
iNecas: Sorry, I don't get it. The only place this middleware is being used is in here https://github.com/theforeman/smart-proxy/pull/613/files#diff-b8fbcbffabe3ae5c9935bb21e3e94852R178, and I don't see the `server` variable being used anywhere.

iNecas:
Interesting regarding the `^C` behavior: can you check with ruby 2.4 and 2.5 (2.4.1 and 2.5.1 were misbehaving with `^C` for me)

lzap:
That sounds more like a race condition than a problem with a particular Ruby version.

Anyway, what is the status of the PR? Is this still WIP? I want to do formal review and testing after it's finished.

lib/launcher.rb
```diff
@@ -147,30 +174,86 @@ def write_pid
...
     end
 
+    def add_threaded_server(server_name, conn_type, app, addresses, port)
+      case server_name
+      when :webrick
+        Thread.new { add_webrick_server(app, addresses, port).start }
+      when :puma
+        Thread.new do
+          add_puma_server(app, addresses, port, conn_type)
+        end
```
lzap, line 219: Any particular reason why the two blocks don't have the same syntax (`do` vs `{`)?

lzap:
I think the way we handle UNIX signals is incorrect in the current version, we kinda do this (`signal_handler.rb`):

    trap(:INT) { exit(0) }
    trap(:TERM) { exit(0) }

We should change that and do graceful shutdown. Puma server has methods for graceful stop for both Single and Cluster classes called `stop_blocked`. We should also correct this for webrick which offers method called `shutdown`. There is no need of checking a timelimit, systemd will kill the process if it takes too long (45 secs by default) to gracefully terminate.

lzap:
To be able to do that, you'd probably need to use `::Puma::Launcher` instead of `::Rack::Handler::Puma` wrapper.

lib/launcher.rb
iNecas: What about loading `webrick` or `puma` just after we know what server we're going to use? We don't have to load everything at the very beginning, if we have a reason for that. That would also allow us not to use the `$HAS_PUMA` global variable

lib/launcher.rb
ik5: originally there was a longer line and wanted to make it readable, use moved to `do` instead of `proc`.

ik5:
@lzap Puma has it's own handler for the signals: https://github.com/puma/puma/blob/v3.10.0/lib/rack/handler/puma.rb#L70

And as you can see at that file, it is using `launcher`.

lzap:
Interrupt Ruby exception is only for Ctrl-C interrupt for interactive mode. In production you run this as a daemon but it still needs to be able to respond to SIGTERM or SIGHUP and SIGUSR1. We use these to rotate logs and initiate graceful restart. So this is different thing, that's why I suggest to change how things are booted up.

Anyway, what is the status? This is entitled WIP, so wondering what are the plans and timeframe. I like your work, would totally love to see this happening.

ik5:
@lzap, the status is that  I waited for your comments.
I'm not sure that it is WIP anymore, but I will add what you have commented here.

lzap:
Sorry I haven't noticed you already removed the WIP tag, I would like to proceed with testing but it appears you closed the PR and deleted the branch?

lzap:
I opened new PR at https://github.com/theforeman/smart-proxy/pull/623 with your commit, I will push it through. @iNecas want to take a look?

ik5:
I can explain it not over this thread.

PR #614, opened 2018-10-16
Refs #25100 - do not use Process::CLOCK_MONOTONIC yet

lzap:
This is Ruby 2.2+ only. No big deal if measured time is negative, it only goes into logs.


tbrisker (approved):
Thanks @lzap !

lzap:
There is not test for these couple of lines, but I will be AFK for 2 hours now.

tbrisker:
[test core]

PR #617, opened 2018-10-19
Fixes #25244 - structured logging of exceptions

lzap:
Proxy now supports configuring against system journal (via `:logfile: JOURNALD`). This patch refactors `error` calls, previously it was accepting bactrace as a simple string putting it into log on single line. This was unreadable.

With this patch, exceptions are passed into the `error` call as objects. Backtrace is correctly parsed via `exception.backtrace` and put as an extra warning log line with structured fields.

To test, configure proxy with journald and misconfigure a module so it fails to initialize it, exception will be thrown into logs as text and also if you do `journalctl -e -o verbose` three fields named `EXCEPTION` should be logged.

### DIFF
--- a/lib/launcher.rb
+++ b/lib/launcher.rb
@@ -107,14 +107,14 @@ module Proxy
     def load_ssl_private_key(path)
       OpenSSL::PKey::RSA.new(File.read(path))
     rescue Exception => e
-      logger.error "Unable to load private SSL key. Are the values correct in settings.yml and do permissions allow reading?: #{e}"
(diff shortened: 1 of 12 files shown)


lzap:
@ares a nice to have for 1.22

ares:
LGTM, will give it a try before merging which should happen probably after 1.21 branching

lzap:
Feel free.

ares:
Thanks, I haven't tested each logger change here but proxy seems to generally work with the patch. Merging!

PR #618, opened 2018-10-19
Fixes #25249 - update Rubocop to 0.50.x

mmoll:
this is the last version supporting Ruby 2.0


mmoll:
The new .rubocop.yml merges the previous disablings with the ones we have in core.

bundler.d/test.rb
```diff
@@ -6,7 +6,6 @@
   gem 'public_suffix', '< 3' if RUBY_VERSION < '2.1'
   gem 'benchmark-ips'
   gem 'ruby-prof'
-  gem 'rubocop', '0.38.0'
```
ekohl, line 9: Won't this be an issue for jenkins?

mmoll: I didn't think so, but now that I checked it, yes... In core it's implemented differently, but anyway I'll revert this.

lzap: What kind of issue? I don't understand.

ekohl: Now that it's back in test.rb you probably can't see the full context but it was moved to development.rb. IIRC Jenkins does `bundle install --without development` and not have rubocop installed.

lzap (approved):
Let's merge and see thanks.

PR #619, opened 2018-10-22
Fixes #25267 - Provide Ruby logger API method formatter

lzap:
This puts back method `formatter` which disappeared since the logging stack is no longer Ruby logger but logging gem which does not implement it. Dynflow requires this method to be present, blocker for 1.20 release @tbrisker.

### DIFF
--- a/lib/proxy/log_buffer/decorator.rb
+++ b/lib/proxy/log_buffer/decorator.rb
@@ -7,6 +7,8 @@ module Proxy::LogBuffer
       @@instance ||= new(::Proxy::LoggerFactory.logger, ::Proxy::LoggerFactory.log_file)
     end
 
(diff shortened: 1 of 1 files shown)


iNecas (approved):
I can confirm this solves the issue for me

tbrisker:
I am not familiar with this are so i'll take @iNecas's word on it, i'm fine pulling this into 1.20.

tbrisker:
:cherries: to 1.20 as ad1f7fb

lzap:
So @dLobatog made a good point that this is only an issue in dev setup, but it is worth cherry picking for sure, thanks all!

iNecas:
Not completely true: it's also influencing debian quite a lot (as we use the dynflow-in-proxy version as well)

PR #621, opened 2018-11-02
Fixes #25390 - pass post via template API

lzap:
This is needed for https://github.com/theforeman/community-templates/pull/461 to work via proxy

### DIFF
--- a/lib/proxy/request.rb
+++ b/lib/proxy/request.rb
@@ -2,6 +2,7 @@ require 'net/http'
 require 'net/https'
 require 'uri'
 require 'cgi'
+require 'hash-patch'
 
 module Proxy::HttpRequest
   class ForemanRequestFactory
@@ -10,7 +11,7 @@ module Proxy::HttpRequest
     end
 
     def query_string(input={})
-      input.map{|k,v| "#{CGI.escape(k.to_s)}=#{CGI.escape(v)}"}.join("&")
+      input.compact.map{|k,v| "#{CGI.escape(k.to_s)}=#{CGI.escape(v)}"}.join("&")
     end
 
     def create_get(path, query={}, headers={})
@@ -26,15 +27,17 @@ module Proxy::HttpRequest
 
     def add_headers(req, headers={})
       req.add_field('Accept', 'application/json,version=2')
-      req.content_type = 'application/json'
+      req.content_type = headers["Content-Type"] || 'application/json'
       headers.each do |k, v|
         req.add_field(k, v)
       end
       req
     end
 
-    def create_post(path, body, headers={})
-      req = Net::HTTP::Post.new(uri(path).path)
+    def create_post(path, body, headers={}, query={})
+      uri = uri(path)
+      uri.query = query_string(query)
+      req = Net::HTTP::Post.new(uri)
       req = add_headers(req, headers)
       req.body = body
       req
--- a/modules/templates/template_proxy_request.rb
+++ b/modules/templates/template_proxy_request.rb
@@ -7,7 +7,7 @@ module Proxy::Templates
   class TemplateProxyRequest < ::Proxy::HttpRequest::ForemanRequest
     include Proxy::Log
 
-    def get_template(kind, env, params)
+    def call_template(method, kind, env, params, body = '')
       template_url = Proxy::Templates::Plugin.settings.template_url
       proxy_ip = URI.parse(template_url).host
       opts = params.clone.merge(:url => template_url)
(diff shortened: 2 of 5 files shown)


timogoebel:
This needs a separate redmine issue.

lib/proxy/request.rb
```diff
@@ -26,15 +26,16 @@ def uri(path)
...
         req.add_field(k, v)
       end
       req
     end
 
-    def create_post(path, body, headers={})
-      req = Net::HTTP::Post.new(uri(path).path)
+    def create_post(path, body, headers={}, query={})
+      uri = uri(path)
+      req = Net::HTTP::Post.new("#{uri.path || '/'}?#{query_string(query)}")
```
timogoebel: If there is a query string, can we set this in the uri? Manually building query string just causes trouble in my experience.

```
uri.query = query_string(query)
```

modules/templates/template_proxy_request.rb
```diff
@@ -19,8 +19,16 @@ def get_template(kind, env, params)
       # in hostgroup provisioning there are spaces
       kind = kind.split('/').map{|x| CGI.escape(x)}.join('/')
       logger.debug "Template: request for #{kind} using #{opts.inspect} at #{uri.host}"
-      proxy_headers = extract_request_headers(env).merge("X-Forwarded-For" => "#{env['REMOTE_ADDR']}, #{proxy_ip}")
-      proxy_req = request_factory.create_get("/unattended/#{kind}", opts, proxy_headers)
+      proxy_headers = extract_request_headers(env)
+      proxy_headers.merge!("X-Forwarded-For" => "#{env['REMOTE_ADDR']}, #{proxy_ip}")
```
timogoebel: You can avoid merge here,

```suggestion
      proxy_headers['X-Forwarded-For'] = "#{env['REMOTE_ADDR']}, #{proxy_ip}"
```

timogoebel (changes requested):

lib/proxy/request.rb
lzap: Sure.

modules/templates/template_proxy_request.rb
lzap: Copy pasted it, yeah :-)

lzap:
Amended the changes + monkey patch for Hash.compact (this is Ruby 2.4+ only) to fix tests on 2.0-2.3. Created RM issue for this.

modules/templates/templates_api.rb
```diff
@@ -25,4 +25,11 @@ class Proxy::TemplatesApi < Sinatra::Base
       Proxy::Templates::TemplateProxyRequest.new.get_template([kind].join('/'), request.env, params)
     end
   end
+
+  post "/:kind" do |kind|
+    log_halt(500, "Failed to retrieve #{kind} template for #{params.inspect}: ") do
+      params.merge!("Content-Type" => "text/plain")
```
timogoebel: I still see a `merge!` here. ;-)

lzap: Missed it somehow, fixed.

lzap:
All checks are finally passing on all Ruby versions. This patch needs to be tested with new template change. The way I test this is just a fake host, template preview and then I launch wget/curl from my laptop to exit the build.

timogoebel (approved):
Code looks good to me. I don't have time to test this, sorry.

lzap:
All relevant PRs merged, rebased, @iNecas interested in trying this out?

iNecas:
I'm interested, I just need to get my devel provisioning setup working again. I wont get to it this week, adding to my TODO list for next one

modules/templates/templates_api.rb
```diff
@@ -25,4 +25,11 @@ class Proxy::TemplatesApi < Sinatra::Base
       Proxy::Templates::TemplateProxyRequest.new.get_template([kind].join('/'), request.env, params)
     end
   end
+
+  post "/:kind" do |kind|
+    log_halt(500, "Failed to retrieve #{kind} template for #{params.inspect}: ") do
```
iNecas: Are we actually trying to retrieve template here? Perhaps more generic message would be a bit better

modules/templates/template_proxy_request.rb
```diff
@@ -29,6 +37,14 @@ def get_template(kind, env, params)
       res.body
     end
 
+    def get_template(kind, env, params)
```
iNecas: Since it's not about just tempaltes anymore, what about calling the methods just `get` and `post`?

lzap:
All done

lzap:
Rebased not sure why tests are not passing, locally fine.

iNecas (approved):
Tested and works well. Thanks @lzap

lzap:
Thanks a buch to @iNecas for review and test!

PR #622, opened 2018-11-12
Fixes #25432 - Support Huawei VRP

UXabre:
Allow support for provisioning Huawei switches running VRP via ZTP mechanism
This code also allows to add additional switch manufactures whom can utilize ZTP

### DIFF
--- a/modules/tftp/server.rb
+++ b/modules/tftp/server.rb
@@ -110,7 +110,7 @@ module Proxy::TFTP
       [pxeconfig_dir]
     end
     def pxeconfig_file mac
-      ["#{pxeconfig_dir}/"+mac.delete(':').upcase]
+      ["#{pxeconfig_dir}/"+mac.delete(':').upcase, "#{pxeconfig_dir}/"+mac.delete(':').upcase+".cfg"]
     end
   end
 
--- a/test/dhcp/isc_omapi_provider_test.rb
+++ b/test/dhcp/isc_omapi_provider_test.rb
@@ -118,6 +118,13 @@ class IscOmapiProviderTest < Test::Unit::TestCase
     assert_equal [], @dhcp.ztp_options_statements(:filename => 'foo.cfg')
     assert_equal ['option option-150 = c0:a8:7a:01;', 'option FM_ZTP.config-file-name = \\"ztp.cfg\\";'],
                  @dhcp.ztp_options_statements(:filename => 'ztp.cfg', :nextServer => '192.168.122.1')
+    assert_equal ['option option-150 = c0:a8:7a:01;', 'option FM_ZTP.config-file-name = \\"ztp.cfg\\";'],
+                 @dhcp.ztp_options_statements(:filename => 'ztp.cfg', :nextServer => '192.168.122.1', :ztp_vendor => 'junos')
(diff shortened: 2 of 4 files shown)


test/dhcp/isc_omapi_provider_test.rb
```diff
@@ -118,8 +118,15 @@ def test_ztp_quirks
...
+    assert_equal ['option option-150 = c0:a8:7a:01;', 'option FM_ZTP.config-file-name = \\"ztp.cfg\\";'],
+                 @dhcp.ztp_options_statements(:filename => 'ztp.cfg', :nextServer => '192.168.122.1', :ztp_vendor => 'junos')
+    assert_equal ['option option-150 = c0:a8:7a:01;', 'option FM_ZTP.config-file-name = \\"ztp.cfg\\";',
+                  'option option-143 = \\"vrpfile=firmware.cc;webfile=web.7z;\\";'],
+                 @dhcp.ztp_options_statements(:filename => 'ztp.cfg', :nextServer => '192.168.122.1',
+                                              :ztp_vendor => 'huawei', :ztp_firmware => { :core => 'firmware.cc',
+                                                                                          :web => 'web.7z'})
   end
-
+
```
ekohl: Nit: can you remove the trailing whitespace here?

UXabre: Sure, must've overlooked this one

modules/tftp/server.rb
```diff
@@ -110,7 +110,7 @@ def pxe_default
       [pxeconfig_dir]
     end
     def pxeconfig_file mac
-      ["#{pxeconfig_dir}/"+mac.gsub(/:/,"").upcase]
+      ["#{pxeconfig_dir}/"+mac.gsub(/:/,"").upcase+".cfg"]
```
lzap: Can you actually return the original and then the new (copy) with added extension for backward compatibility? Proxy will render two files with the same content actually.

lzap (approved):
Works for me, this needs a test once core PR is ready.

modules/tftp/server.rb
UXabre: Sure! Will change it tonight 😊 thanks for the review!

UXabre:
Made the proposed changes regarding writing both ".cfg"-suffixed TFTP files as well as the non-suffixed variant for backwards compatibility.

lzap (approved):

lzap (approved):
Thank you.

lzap:
Oh please fix the Rubocop:

Offenses:

modules/dhcp_common/isc/omapi_provider.rb:174:15: C: Avoid more than 3 levels of block nesting.
              if options.has_key?(:ztp_firmware) ...
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

370 files inspected, 1 offense detected

UXabre:
Oh, for sure! Seems I was to quick to call this one :-) Will notify when it is done!

UXabre:
Hey @lzap , I've resolved the issue by refactoring out the switch case into a seperate function. Please review to validate if this fits the qualitative constraints :-)

lzap:
That's fine, or you could increase Rubocop setting that's what we do. It's too strict.

PR #624, opened 2018-11-29
allow for PUPPET_VERSION to be set

komidore64:
this will help when testing smart-proxy against different puppet
versions

### DIFF
--- a/bundler.d/puppet.rb
+++ b/bundler.d/puppet.rb
@@ -1,4 +1,4 @@
 group :puppet_proxy_legacy do
-  gem 'puppet', '< 6.0.0'
+  gem 'puppet', ENV.key?('PUPPET_VERSION') ? "~> #{ENV['PUPPET_VERSION']}" : '< 6.0.0'
   gem 'ruby-augeas', :require => 'augeas'
 end


ekohl (commented):
I'm surprised this was never there and how we installed multiple puppet versions before. Do note that this is the legacy implementation and the current one talks to the REST API so then the gem isn't used AFAIK.

Given Puppet 3.x doesn't run on current rubies, it'd also be fine to ensure that part just isn't tested and we use `bundle install --without puppet_proxy_legacy`.

ehelms:
If I were to summarize, @ekohl are you suggesting to just reduce the proxy test matrix to Puppet 4+ and Ruby combinations only? Or are you saying that the huge matrix isn't needed at all because the gem is no longer used and only different Rubies matter?

ekohl:
We load the REST API provider when the puppet version is >= 4.0: https://github.com/theforeman/smart-proxy/blob/986c95cd8e9585df8ebc23d01836fbb958144423/modules/puppet_proxy/configuration_loader.rb#L8

That means we only need the Puppet gem with version 3.x. In reality we'll only see that on EL7 with EPEL (Ruby 2.0) and Ubuntu Xenial (Ruby 2.3). In all other Ruby versions you can skip the whole legacy puppet setup.

It's also possible to go a step further since 3.x support is [unsupported since 1.19](https://theforeman.org/manuals/1.19/index.html#Releasenotesfor1.19). That means you could even omit it altogether.

ekohl:
To be clear, I'm surprised this code wasn't here already. I assumed it was but assumptions ...

ekohl (approved):
I assumed this was already in place so yay for making reality match my expectations :)

PR #625, opened 2018-12-03
Fixes #25614 - v2 Features api with Capabilities and exposed settings

jlsherrill:


jlsherrill:
This is a take over of https://github.com/theforeman/smart-proxy/pull/585 (with his blessing).  It maintains most all of his original pr, but:

1) Adds authentication as a requirement to the v2 features api
2) Allows capabilities to be calculated dynamically by simply defining them with a proc:

```
capability(Proc.new { 
   calculate_something_here
})
```

modules/root/protected_root_api.rb
```diff
@@ -0,0 +1,42 @@
+class Proxy::ProtectedRootApi < Sinatra::Base
+  helpers ::Proxy::Helpers
+
+  # authorize_with_trusted_hosts
```
ekohl: Was this part of debugging?

modules/root/protected_root_api.rb
```diff
@@ -0,0 +1,42 @@
...
+      plugins.to_json
+    rescue => e
+      log_halt 400, e
+    end
+  end
+
+  def process_capabilities(capabilities)
+    capabilities = capabilities.map do |capability|
+      if capability.is_a?(Proc)
+        capability.call
```
ekohl: I like the ability of using a proc. I didn't think of that. Should we take some safe guards here with regards to exceptions or timeouts (if they call to external services)?

A possible test case in the current code would be to use puppet. If it has a provider set, then it should have the capability puppetrun. This can then be used by the foreman to detect it at runtime. Currently it's a setting in foreman's settings.yaml to enable the button globally but it's really a capability on the proxy for the puppet feature.

modules/root/protected_root_api.rb
```diff
@@ -0,0 +1,42 @@
...
+      plugins = enabled_plugins.inject({}) do |hash, plugin|
+        result = plugin.slice(*attributes)
+        result[:capabilities] = process_capabilities(plugin[:capabilities])
+        hash[plugin[:name]] = result
+        hash
+      end
+
+      plugins.to_json
+    rescue => e
+      log_halt 400, e
```
ekohl: Should this be an internal server error?

modules/root/protected_root_api.rb
```diff
@@ -0,0 +1,42 @@
...
+  helpers ::Proxy::Helpers
+
+  # authorize_with_trusted_hosts
+  # authorize_with_ssl_client
+
+  get "/v2/features" do
+    begin
+      enabled_plugins = ::Proxy::Plugins.instance.select do |p|
+        p[:name] != :foreman_proxy && \
+          p[:state] == :running && \
```
ekohl: Later I realized that sending the state can be useful. It would mean the server could detect misconfigurations (which is now done in the installer). That does mean the server must always filter on running plugins so it'd be good to get it in on the first go.

ekohl (commented):
Thanks for taking over the effort. It looks like I didn't add tests yet either.

modules/root/protected_root_api.rb
jlsherrill: Whoops yes, will uncomment :)

modules/root/protected_root_api.rb
jlsherrill: Yeah, after discussing with eric, i realized that the installer may not even know what content units are installed (for pulp 3), so dynamic capabilities became more important.  I went back and forth between explicitly dynamic properties, but this seemed very simple and worked well. 

I'm not sure that we should enforce a timeout outside of the proc call, RestClient will typically timeout after ~2 minutes, and i think it's simpler to leave it up to the plugin to ensure that a timeout will occur.  Otherwise the only other way i can think would be to spin up the proc in a thread and kill it after some timeout occurs.  I'm not sure thats worth it at this time, but open to other opinions.

ekohl: I'd be OK with mentioning in docs that you are expected to handle all exceptions and timeouts.

modules/root/protected_root_api.rb
jlsherrill: This was code you wrote :) but yes i think you are correct!

ekohl: I thought I could pin it on you now ;) But yes, my was very much WIP

modules/root/protected_root_api.rb
jlsherrill: That makes sense.  I think we'd want to ignore dynamic capabilities if its not running?

ekohl: Good thought

jturel:
After I click 'Refresh features' in the UI I see that the version API no longer works and the UI reports 
```
ERF50-5345 [Foreman::WrappedException]: Unable to connect ([ProxyAPI::ProxyException]: ERF12-1764 [ProxyAPI::ProxyException]: Unable to detect version ([RestClient::NotFound]: 404 Not Found) for proxy https://centos7-proxy-devel.jturel.example.com:9090/version)
```
With a 404 in the proxy logs:

```
2019-01-07T19:02:55 492926e0 [I] Started GET /version 
2019-01-07T19:02:55 492926e0 [I] Finished GET /version with 404 (0.77 ms)
```

However, it version check seems to work *before* I refresh features even running the new code

jlsherrill:
@jturel that's strange, my code shouldn't have affected that at all.  Any errors in the smart proxy log?

jturel:
@jlsherrill it could very well be a PICNIC. I didn't see any docs on how to test this so what I did was:

- spin up centos7-proxy-devel
- cd to /usr/share/foreman-proxy
- git init to preserve state
- cherry pick the commits on this PR
- restart foreman proxy

Is there a better way?

edit: no errors in the proxy log though - just the 404

jlsherrill:
@jturel yeah, that should be fine, although i actually tested running straight from git (which was a little bit more work).

To clarify, fetching the version worked until you clicked refresh?  and now it doesn't work?  Did you try curling it directly?

jturel:
When I switch to the branch with your changes upon first load of the UI for my external proxy I see that it actually doesn't fetch the version. I guess it's cached (hence refresh button). So .. ignore what I said about clicking refresh!

When I use the foreman API I see the error as well:

https://centos7-devel.jturel.example.com/api/v2/smart_proxies/2/version
```
{
"error": {
"message": "ERF50-5345 [Foreman::WrappedException]: Unable to connect ([ProxyAPI::ProxyException]: ERF12-1764 [ProxyAPI::ProxyException]: Unable to detect version ([RestClient::NotFound]: 404 Not Found) for proxy https://centos7-proxy-devel.jturel.example.com:9090/version)"
}
}
```
And the same 404 in the proxy logs:
```
2019-01-07T19:42:19 492926e0 [I] Started GET /version
2019-01-07T19:42:19 492926e0 [I] Finished GET /version with 404 (0.39 ms)
```

I guess you aren't seeing this problem?

jlsherrill:
@jturel good catch, updated with a new commit

jlsherrill:
[test core]

jturel:
@jlsherrill works now! could you enumerate some of the overall goals of these two PRs so I can be a little more targeted in my testing? Existing functionality seems to be met just fine

jlsherrill:
@jturel  See: https://community.theforeman.org/t/exposing-capabilities-in-the-smart-proxy/9860/1

Its kinda hard to test fully as not much utilizes it, the only thing actually utilized in this pr is the tftpserver name setting.  

What you can do is edit an enabled smart proxy plugin, such as puppet on here:
https://github.com/theforeman/smart-proxy/blob/develop/modules/puppet_proxy/puppet_plugin.rb#L7

on that line add:
```
expose_setting :puppet_version:

capability('static_capability')
capability(Proc.new {
  ['some', 'dynamic', 'capabilities']
})

```

This will expose a setting, a static capability, and a dynamic capability. 

and then do a refresh and check the SmartProxyFeatures model, specifically SmartProxyFeature#capabilities and SmartProxyFeature#settings.  You might try changing around the capabilities and settings, do a refresh, and verify that things get updated in the db.

jlsherrill:
I forgot to mention, you need to restart the smart proxy after any change.

test/plugins/plugin_test.rb
```diff
@@ -129,4 +129,11 @@ class TestPluginWithoutCustomValidators < ::Proxy::Plugin; end
   def test_custom_validators_returns_empty_hash_if_mappings_were_omitted
     assert TestPluginWithoutCustomValidators.custom_validators.empty?
   end
+
+  class TestCapabilityPlugin < ::Proxy::Plugin
```
jturel, line 133: Can you add a cabability test which uses a lambda as well?

I'm getting an error once I restart my proxy when defining a capability like this:
```
capability(lambda{ ['a', 'b'] })
```

```
Error during startup, terminating. comparison of String with Proc failed
```

I think your tests in root_v2_api_tests didn't reveal it because of the way you're stubbing the `loaded` method but I'm just guessing

jturel:
Thanks! This is meeting the design laid out by the RFC and I see that `:tftp_servername` made it into my DB correctly. I left one other comment regarding testing.

jlsherrill:
@jturel fixed, we were previously sorting capabilities at plugin registration time, this was moved to feature request time to enure that dynamic capabilities were sorted properly as well.  Also test was added to catch that

jlsherrill:
@lzap the failing test seems to also occur on master?

lib/proxy/plugin_initializer.rb
```diff
@@ -29,6 +29,15 @@ def https_enabled?
     @https_enabled
   end
 
+  def capabilities
+    [members.map { |m| m.capabilities }.compact].flatten
```
ekohl: I probably wrote this initially, but could this be `members.map(&:capabilities).compact.flatten`?

modules/root/root_v2_api.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  authorize_with_trusted_hosts
+  authorize_with_ssl_client
+
+  get "/features" do
+    begin
+
+      enabled_plugins = ::Proxy::Plugins.instance.select do |plugin|
+        plugin[:name] != :foreman_proxy && \
+          plugin[:class].ancestors.include?(::Proxy::Plugin)
```
jturel, line 11: Interesting. Is there a case where there'd be a plugin that would not inherit ::Proxy::Plugin, or is this more of a defensive check?

ekohl: I copied it from the v1 api:
https://github.com/theforeman/smart-proxy/blob/ce744b8bee2506fe8779f561ded3977dc15feabb/modules/root/root_api.rb#L6

jlsherrill: This is actually copied from https://github.com/theforeman/smart-proxy/blob/develop/modules/root/root_api.rb#L6   in @ekohl's original PR.  I don't actually know the reason for it :)

modules/root/root_v2_api.rb
```diff
@@ -0,0 +1,39 @@
...
+      end
+
+      plugins.to_json
+    rescue => e
+      log_halt 500, e
+    end
+  end
+
+  def process_capabilities(state, capabilities)
+    capabilities = capabilities.select{|cap| !cap.is_a?(Proc) || state == :running }
```
jturel, line 31: If I had written this I'd have checked for `respond_to? :call`

Not asking for a change but wanted to mention it in case you agree since I see checking for Proc in a few places

jturel (approved):
I'm acking because it works to the best of my knowledge, but probably better for someone with more smart-proxy background to ACK as well. I just wanted to get more familiar with this stuff :)

modules/root/root_v2_api.rb
jturel: Looks like it was https://projects.theforeman.org/issues/15025 I guess it's fair enough to keep

jturel:
Oh - I think https://projects.theforeman.org/projects/smart-proxy/wiki/API will need some updates as well :)

jlsherrill:
@jturel yeah, i will update that once merged

lzap:
Yes I can confirm that this is a failed test in develop branch, regression. Will look into it later.

lzap:
The fix is here @jlsherrill https://github.com/theforeman/smart-proxy/pull/628

jlsherrill:
Thanks @lzap 
[test core]

test/plugins/plugin_initializer_test.rb
```diff
@@ -6,14 +6,21 @@ class TestPlugin2 < Proxy::Plugin; plugin :plugin_2, "1.0"; default_settings :en
   class TestPlugin3 < Proxy::Plugin; plugin :plugin_3, "1.0"; uses_provider; default_settings :enabled => true, :use_provider => :plugin_4; end
   class TestPlugin4 < Proxy::Provider; plugin :plugin_4, "1.0"; default_settings :enabled => true; end
   class TestPlugin5 < Proxy::Plugin; plugin :plugin_5, "1.0"; default_settings :enabled => false; end
+  class TestPlugin6 < Proxy::Plugin
+    plugin :plugin_6, "1.0"
+    capability(lambda{})
+    capability(proc{})
+    capability('foo')
+  end
```
lzap, line 19: Can you enable this plugin? Seems disabled.

test/plugins/plugin_test.rb
lzap: Won't hurt.

lib/proxy/pluggable.rb
```diff
@@ -31,6 +35,10 @@ def default_settings(a_hash = {})
     @plugin_default_settings.merge!(a_hash)
   end
 
+  def expose_setting(setting)
+    exposed_settings << setting
+  end
```
lzap, line 40: Method not covered in tests.

lib/proxy/pluggable.rb
```diff
@@ -167,10 +175,18 @@ def loading_failed(message)
...
+    @capabilities ||= []
+  end
+
   def dependencies
     @dependencies ||= []
   end
 
+  def exposed_settings
+    @exposed_settings ||= []
+  end
```
lzap, line 188: Ditto.

lzap (changes requested):
I am okay with this, the major concern are tests now. Once rebased I am gonna try this.

test/plugins/plugin_test.rb
jlsherrill: this was added

jlsherrill:
[test core]

jlsherrill:
[test rubocop]

jlsherrill:
[test core]

lzap (approved):
WFM, @ekohl any comments?

The output is a bit lengthy, I wonder if we could drop empty arrays/hashes but not a big deal:

```json
{
    "discovery": {
        "http_enabled": true,
        "https_enabled": true,
        "settings": {

        },
        "state": "running",
        "capabilities": [

        ]
    },
    "discovery_image": {
        "http_enabled": true,
        "https_enabled": true,
        "settings": {

        },
        "state": "running",
        "capabilities": [

        ]
    },
    "salt": {
        "http_enabled": true,
        "https_enabled": true,
        "settings": {

        },
        "state": "running",
        "capabilities": [

        ]
    },
    "abrt": {
        "http_enabled": true,
        "https_enabled": true,
        "settings": {

        },
        "state": "running",
        "capabilities": [

        ]
    },
    "pulp": {
        "http_enabled": true,
        "https_enabled": true,
        "settings": {

        },
        "state": "running",
        "capabilities": [

        ]
    },
    "pulpnode": {
        "http_enabled": false,
        "https_enabled": false,
        "settings": {

        },
        "state": "disabled",
        "capabilities": [

        ]
    },
    "tail": {
        "http_enabled": false,
        "https_enabled": false,
        "settings": {

        },
        "state": "disabled",
        "capabilities": [

        ]
    },
    "dynflow": {
        "http_enabled": false,
        "https_enabled": false,
        "settings": {

        },
        "state": "disabled",
        "capabilities": [

        ]
    },
    "facts": {
        "http_enabled": false,
        "https_enabled": false,
        "settings": {

        },
        "state": "disabled",
        "capabilities": [

        ]
    },
    "dns": {
        "http_enabled": true,
        "https_enabled": true,
        "settings": {

        },
        "state": "running",
        "capabilities": [

        ]
    },
    "templates": {
        "http_enabled": true,
        "https_enabled": true,
        "settings": {

        },
        "state": "running",
        "capabilities": [

        ]
    },
    "tftp": {
        "http_enabled": true,
        "https_enabled": true,
        "settings": {
            "tftp_servername": "192.168.100.1"
        },
        "state": "running",
        "capabilities": [

        ]
    },
    "dhcp": {
        "http_enabled": true,
        "https_enabled": true,
        "settings": {

        },
        "state": "running",
        "capabilities": [

        ]
    },
    "puppetca": {
        "http_enabled": true,
        "https_enabled": true,
        "settings": {

        },
        "state": "running",
        "capabilities": [

        ]
    },
    "puppet": {
        "http_enabled": true,
        "https_enabled": true,
        "settings": {

        },
        "state": "running",
        "capabilities": [

        ]
    },
    "bmc": {
        "http_enabled": true,
        "https_enabled": true,
        "settings": {

        },
        "state": "running",
        "capabilities": [

        ]
    },
    "realm": {
        "http_enabled": false,
        "https_enabled": false,
        "settings": {

        },
        "state": "disabled",
        "capabilities": [

        ]
    },
    "logs": {
        "http_enabled": true,
        "https_enabled": true,
        "settings": {

        },
        "state": "running",
        "capabilities": [

        ]
    },
    "httpboot": {
        "http_enabled": true,
        "https_enabled": true,
        "settings": {

        },
        "state": "running",
        "capabilities": [

        ]
    }
}
```

ekohl (approved):
> WFM, @ekohl any comments?

Yay! Big thanks to @jlsherrill for finish what I started. 

> The output is a bit lengthy, I wonder if we could drop empty arrays/hashes but not a big deal:

In my experience an API is easier if you can assume data structures. Otherwise you need to check if a key is present or a value is nil. IMHO it's worth the added overhead.

lzap:
It's in! Thanks @jlsherrill and @ekohl and others. Let's start building on top of that.

lzap:
This is demo-worthy I think even if Foreman part is not yet finished.

PR #626, opened 2018-12-19
Fixes #25825 - fixed BMC SSH config names

lzap:
SSIA + typos in SSH BMC module example config

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,10 @@ Currently Supported modules:
  * Puppet CA - Manage certificate signing, cleaning and autosign on a Puppet CA server
  * Realm - Manage host registration to a realm (e.g. FreeIPA)
  * TFTP - any UNIX based tftp server
+ * Facts - module to gather facts from facter (used only on discovered nodes)
+ * HTTPBoot - endpoint exposing a (TFTP) directory via HTTP(s) for UEFI HTTP booting
+ * Logs - log buffer of proxy logs for easier troubleshooting
+ * Templates - unattended Foreman endpoint proxy
 
 # Installation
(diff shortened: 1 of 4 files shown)


lzap:
Someone? @ekohl perhaps? This is a trivial one thanks!

ekohl (approved):
From reading the code this does appear to be correct, but it also means https://github.com/theforeman/puppet-foreman_proxy/commit/668eb6525d0f246e4fbdddfecf159ad4de23665e#diff-4cec27937c8aef24f2e4df145925452cR8 is wrong and needs an update. Since it's unreleased there's still time but it'd be good to get it in before 1.21.

ekohl:
https://theforeman.org/manuals/1.20/index.html#4.3.3BMC will also need an update if this is incorrect.

lzap:
Done thanks.

lzap:
WAIT

Looks like puppet codebase also has this prefixed version, I am gonna change the code rather than this.

https://github.com/theforeman/puppet-foreman_proxy/commit/668eb6525d0f246e4fbdddfecf159ad4de23665e#diff-4cec27937c8aef24f2e4df145925452cR8

lzap:
I rebased the PR, now the config keys are actually correct.

ekohl:
The installer side hasn't been released yet. I can be ok with this change because the parameters only affect the SSH provider, but it is a breaking change.

lzap:
I can do migration if needed but I think it is not worth the effort.

lzap:
FYI I fixed the failing test in another PR.

tbrisker (approved):
Thanks @lzap !

tbrisker:
1.21- b708da3
do we need to cherry pick this to 1.20 as well?

lzap:
Thanks, I don't think so. Again, not in wide use. I am probably the only user when testing BMC API :-)

PR #628, opened 2019-01-10
Fixes #25824 - BMC provider nil exception test fixed

lzap:
I am not aware it's possible to send `nil` via request parameters, therefore this test is not correct. This fails with recent Sinatra versions due to this change:

https://github.com/sinatra/sinatra/pull/1479#issuecomment-453034886


mmoll:
[test rubocop]

mmoll:
[test core]

mmoll:
merged díky @lzap!

ekohl:
It also means https://github.com/theforeman/puppet-foreman_proxy/commit/668eb6525d0f246e4fbdddfecf159ad4de23665e#diff-4cec27937c8aef24f2e4df145925452cR8 is wrong and needs an update.

mmoll:
@ekohl is that comment correct here or does it belong to GH-626?

ekohl:
@mmoll it is indeed intended for that issue.

PR #633, opened 2019-01-24
Refs #15095 - improve cache debug logging

lzap:
A user reports memory leak, this is an attempt to add more debugging to see if the cache is not culprit of the problem.

### DIFF
--- a/modules/puppet_proxy_puppet_api/v3_environment_classes_api_classes_retriever.rb
+++ b/modules/puppet_proxy_puppet_api/v3_environment_classes_api_classes_retriever.rb
@@ -80,19 +80,24 @@ class Proxy::PuppetApi::V3EnvironmentClassesApiClassesRetriever
       begin
         response, etag = @puppet_api.new(puppet_url, ssl_ca, ssl_cert, ssl_key).list_classes(environment, etag, MAX_PUPPETAPI_TIMEOUT)
       rescue Exception => e
-        @m.synchronize { @futures_cache[environment] = nil }
+        @m.synchronize { @futures_cache.delete(environment) }
         logger.error "Error while retrieving puppet classes for '%s' environment" % [environment], e
         raise e
       end
 
       if response == Proxy::PuppetApi::EnvironmentClassesApiv3::NOT_MODIFIED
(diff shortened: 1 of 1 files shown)


ekohl:
All tests now fail with `undefined method `[]' for nil:NilClass (NoMethodError)` and I think it's related

lzap:
Yeah you are right, fixed.

modules/puppet_proxy_puppet_api/v3_environment_classes_api_classes_retriever.rb
```diff
@@ -96,7 +96,10 @@ def async_get_classes(environment)
         end
       end
     end
-    @m.synchronize { @futures_cache[environment] = future }
+    @m.synchronize do
+      @futures_cache[environment] = future
+      logger.debug { "Class cache: #{@classes_cache.size}, ETag cache: #{@etag_cache.size}, Futures cache: #{@futures_cache.size}" }
```
iNecas: It doesn't seem the logger has an implicit prefix, so it would be good to explicitly mention where this lines are coming from: something like `"Puppet cache counts: classes %d, ETag %d,  futures %d"`

iNecas: Also, I wonder, why we're only logging it at this specific place? (what about other ones where we touch any of the cache variables?)

lzap: Not sure what you mean by the prefix, but rebased as you suggested.

lzap:
Moved the block to the future.

Also I think that the idea behind setting some values to `nil` was to remove entries from cache. But in Ruby hash implementation, setting an entry to `nil` does *not* remove the entry - it's kept with reference to `nil` value. Therefore I changed this to `delete` which does affect `count` actually. This could improve memory possibly.

iNecas (approved):
Tested and works well. Also, using delete makes sense here IMO.

iNecas:
@ekohl any comment before merging?

ekohl (approved):

iNecas:
Thanks @lzap

PR #636, opened 2019-01-31
Refs #25614 - improve and add DNS capabilities

lzap:
When I started working on DNS capabilities, I realized that the feature does not work as expected. It publishes capabilities from plugins but completely ignores caps from providers. Therefore I am proposing slight change in V2 features API to also publish it from providers.

This changes the API result a bit, now it includes bunch of providers which are registered as "uninitialized". My first idea was to simply hide them, but I think it might be quite useful showing them flagged with state set to "uninitialized" because if we decide to add this later, this will be again a breaking change since the client code must now parse actually "state" value and not rely on simple element presence.

For this reason, I am leaving the "uninitialized" providers in the response. But now the output is getting out of control, it's very long with tons of "null", empty arrays and hashes. Therefore I am proposing to compact the result to more human-readable form (deleting all empty/nil elements). While I agree with Ewoud that generally it is good to avoid checking for non-existing elements, I think for JSON API this is not the case as this applies more for coding.

The patch also exposes "use_provider" setting value where appropriate. This way the consumer can eventually create a "tree" of initialized dependencies easily.

Second part of the patch is actually adding capabilities to DNS providers in core. We have three DNS providers from which ISC (nsupdate) is the most capable and libvirt is the most limited (supports only A type of records). I propose for `handles_X` pattern because we might add another set of capabilities so the prefix clearly differentiate each group.

### DIFF
--- a/lib/proxy/helpers.rb
+++ b/lib/proxy/helpers.rb
@@ -20,6 +20,7 @@ module Proxy::Helpers
     end
     content_type :json if request.accept?("application/json")
     logger.error message, exception
+    logger.exception(message, exception) if exception.is_a?(Exception)
     halt code, message
   end
 
--- a/modules/dhcp/dhcp_plugin.rb
+++ b/modules/dhcp/dhcp_plugin.rb
@@ -4,6 +4,7 @@ class Proxy::DhcpPlugin < ::Proxy::Plugin
 
   uses_provider
   default_settings :use_provider => 'dhcp_isc', :server => '127.0.0.1', :subnets => []
+  expose_setting :use_provider
   plugin :dhcp, ::Proxy::VERSION
 
   load_classes ::Proxy::DHCP::ConfigurationLoader
--- a/modules/dns_dnscmd/dns_dnscmd_plugin.rb
+++ b/modules/dns_dnscmd/dns_dnscmd_plugin.rb
@@ -6,6 +6,11 @@ module ::Proxy::Dns::Dnscmd
 
     requires :dns, ::Proxy::VERSION
 
+    capability "record_type_a"
+    capability "record_type_aaaa"
+    capability "record_type_cname"
+    capability "record_type_ptr"
+
     load_classes ::Proxy::Dns::Dnscmd::PluginConfiguration
     load_dependency_injection_wirings ::Proxy::Dns::Dnscmd::PluginConfiguration
   end
--- a/modules/root/root_v2_api.rb
+++ b/modules/root/root_v2_api.rb
@@ -15,8 +15,9 @@ class Proxy::RootV2Api < Sinatra::Base
       attributes = [:http_enabled, :https_enabled, :settings, :state]
 
       plugins = enabled_plugins.inject({}) do |hash, plugin|
-        result = Hash[attributes.map { |attribute| [attribute, plugin[attribute]] }]
-        result[:capabilities] = process_capabilities(plugin[:state], plugin[:capabilities])
+        result = Hash[attributes.map { |attribute| [attribute, plugin[attribute]] }.delete_if { |k, v| v.nil? || (v.instance_of?(Hash) && v.empty?) }]
+        capabilities = process_capabilities(plugin[:state], plugin[:capabilities])
+        result[:capabilities] = capabilities unless capabilities.empty?
         hash[plugin[:name]] = result
         hash
       end
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -113,3 +113,6 @@ Style/LineEndConcatenation:
 
 Style/ParallelAssignment:
   Enabled: false
+
+Metrics/LineLength:
+  Max: 250
--- a/lib/proxy/log.rb
+++ b/lib/proxy/log.rb
@@ -79,7 +79,7 @@ module Proxy
     def call(env)
       status = 500
       env['rack.logger'] = logger
-      logger.info { "Started #{env['REQUEST_METHOD']} #{env['PATH_INFO']} #{env['QUERY_STRING']}" }
+      logger.info { "Started #{env['REQUEST_METHOD']} #{env['REQUEST_PATH']} #{env['QUERY_STRING']}" }
       logger.trace { 'Headers: ' + env.select {|k,v| k.start_with? 'HTTP_'}.inspect }
       logger.trace { (body = env['rack.input'].read).empty? ? '' : 'Body: ' + body }
       before = Time.now.to_f
--- a/modules/dns/dns_plugin.rb
+++ b/modules/dns/dns_plugin.rb
@@ -5,6 +5,7 @@ module Proxy::Dns
 
     uses_provider
     default_settings :use_provider => 'dns_nsupdate', :dns_ttl => 86_400
+    expose_setting :use_provider
     plugin :dns, ::Proxy::VERSION
 
     load_classes ::Proxy::Dns::ConfigurationLoader
--- a/modules/dns_libvirt/dns_libvirt_plugin.rb
+++ b/modules/dns_libvirt/dns_libvirt_plugin.rb
@@ -6,6 +6,8 @@ module ::Proxy::Dns::Libvirt
 
     default_settings :url => "qemu:///system", :network => 'default'
 
+    capability "record_type_a"
+
     load_classes ::Proxy::Dns::Libvirt::PluginConfiguration
     load_dependency_injection_wirings ::Proxy::Dns::Libvirt::PluginConfiguration
   end
--- a/modules/dns_nsupdate/dns_nsupdate_plugin.rb
+++ b/modules/dns_nsupdate/dns_nsupdate_plugin.rb
@@ -8,6 +8,12 @@ module ::Proxy::Dns::Nsupdate
 
     validate_readable :dns_key
 
+    capability "record_type_a"
+    capability "record_type_aaaa"
+    capability "record_type_cname"
+    capability "record_type_ptr"
+    capability "record_type_srv"
+
     load_classes ::Proxy::Dns::Nsupdate::PluginConfiguration
     load_dependency_injection_wirings ::Proxy::Dns::Nsupdate::PluginConfiguration
   end
--- a/modules/puppetca/puppetca_plugin.rb
+++ b/modules/puppetca/puppetca_plugin.rb
@@ -5,6 +5,7 @@ module Proxy::PuppetCa
 
     uses_provider
     default_settings :use_provider => 'puppetca_hostname_whitelisting'
+    expose_setting :use_provider
 
     load_classes ::Proxy::PuppetCa::PluginConfiguration
     load_programmable_settings ::Proxy::PuppetCa::PluginConfiguration
--- a/modules/realm/realm_plugin.rb
+++ b/modules/realm/realm_plugin.rb
@@ -4,6 +4,7 @@ module Proxy::Realm
     https_rackup_path File.expand_path("http_config.ru", File.expand_path("../", __FILE__))
 
     default_settings :use_provider => 'realm_freeipa'
+    expose_setting :use_provider
 
     uses_provider
     load_classes ::Proxy::Realm::ConfigurationLoader
--- a/test/puppetca_http_api/ca_v1_api_request_test.rb
+++ b/test/puppetca_http_api/ca_v1_api_request_test.rb
@@ -52,7 +52,6 @@ class CaApiv1RequestTest < Test::Unit::TestCase
     stub_request(:get, 'https://puppet:8140/puppet-ca/v1/certificate_statuses/foreman').
       to_return(:status => 200, :body => fixture('ca_search.json'))
 
-    # rubocop:disable Metrics/LineLength
     expected = [
       {
         'dns_alt_names' => ['DNS:puppet', 'DNS:puppet.example.com'],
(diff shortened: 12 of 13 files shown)


lzap:
@ekohl and @jlsherrill here is some changes needed for using this with DNS. I wonder if it makes sense to introduce "type" field with values either "plugin" or "provider" so consumers are able to differentiate those. Example output after the patch:

```json
{
    "discovery": {
        "http_enabled": true,
        "https_enabled": true,
        "state": "running"
    },
    "discovery_image": {
        "http_enabled": true,
        "https_enabled": true,
        "state": "running"
    },
    "salt": {
        "http_enabled": true,
        "https_enabled": true,
        "state": "running"
    },
    "abrt": {
        "http_enabled": true,
        "https_enabled": true,
        "state": "running"
    },
    "pulp": {
        "http_enabled": true,
        "https_enabled": true,
        "state": "running"
    },
    "pulpnode": {
        "http_enabled": false,
        "https_enabled": false,
        "state": "disabled"
    },
    "tail": {
        "http_enabled": false,
        "https_enabled": false,
        "state": "disabled"
    },
    "dynflow": {
        "http_enabled": false,
        "https_enabled": false,
        "state": "disabled"
    },
    "dhcp_infoblox": {
        "state": "uninitialized"
    },
    "dns_infoblox": {
        "state": "uninitialized"
    },
    "facts": {
        "http_enabled": false,
        "https_enabled": false,
        "state": "disabled"
    },
    "dns": {
        "http_enabled": true,
        "https_enabled": true,
        "settings": {
            "use_provider": "dns_libvirt"
        },
        "state": "running",
        "capabilities": [
            "handles_a"
        ]
    },
    "dns_nsupdate": {
        "state": "uninitialized"
    },
    "dns_nsupdate_gss": {
        "state": "uninitialized"
    },
    "dns_dnscmd": {
        "state": "uninitialized"
    },
    "dns_libvirt": {
        "state": "running"
    },
    "templates": {
        "http_enabled": true,
        "https_enabled": true,
        "state": "running"
    },
    "tftp": {
        "http_enabled": true,
        "https_enabled": true,
        "settings": {
            "tftp_servername": "192.168.100.1"
        },
        "state": "running"
    },
    "dhcp": {
        "http_enabled": true,
        "https_enabled": true,
        "settings": {
            "use_provider": "dhcp_isc"
        },
        "state": "running"
    },
    "dhcp_isc": {
        "state": "running"
    },
    "dhcp_native_ms": {
        "state": "uninitialized"
    },
    "dhcp_libvirt": {
        "state": "uninitialized"
    },
    "puppetca": {
        "http_enabled": true,
        "https_enabled": true,
        "settings": {
            "use_provider": "puppetca_hostname_whitelisting"
        },
        "state": "running"
    },
    "puppetca_hostname_whitelisting": {
        "state": "running"
    },
    "puppetca_token_whitelisting": {
        "state": "uninitialized"
    },
    "puppet": {
        "http_enabled": true,
        "https_enabled": true,
        "state": "running"
    },
    "puppet_proxy_customrun": {
        "state": "uninitialized"
    },
    "puppet_proxy_legacy": {
        "state": "uninitialized"
    },
    "puppet_proxy_mcollective": {
        "state": "uninitialized"
    },
    "puppet_proxy_puppet_api": {
        "state": "running"
    },
    "puppet_proxy_puppetrun": {
        "state": "uninitialized"
    },
    "puppet_proxy_salt": {
        "state": "uninitialized"
    },
    "puppet_proxy_ssh": {
        "state": "running"
    },
    "bmc": {
        "http_enabled": true,
        "https_enabled": true,
        "state": "running"
    },
    "realm": {
        "http_enabled": false,
        "https_enabled": false,
        "settings": {
            "use_provider": null
        },
        "state": "disabled"
    },
    "realm_freeipa": {
        "state": "uninitialized"
    },
    "logs": {
        "http_enabled": true,
        "https_enabled": true,
        "state": "running"
    },
    "httpboot": {
        "http_enabled": true,
        "https_enabled": true,
        "state": "running"
    }
}
```

lib/proxy/helpers.rb
```diff
@@ -20,6 +20,7 @@ def log_halt code=nil, exception=nil
     end
     content_type :json if request.accept?("application/json")
     logger.error message, exception
+    logger.exception message, exception
```
lzap: We were missing the stacktrace from logs here.

ekohl (changes requested):
In my initial patch this worked without exposing all the additional "modules". This proposal isn't good because the foreman doesn't know how to handle the providers - it only knows about features. I'd really like to just expose modules in this API call.

modules/dns_dnscmd/dns_dnscmd_plugin.rb
```diff
@@ -6,6 +6,11 @@ class Plugin < ::Proxy::Provider
 
     requires :dns, ::Proxy::VERSION
 
+    capability "handles_a"
```
ekohl: Since these are record types I'd say `record_type_a` is more explicit.

lzap:
I figured it out, I was testing it incorrectly. It works as expected now, the patch is still adding DNS capabilities, exposes provider setting and compacts the output:

```json
{
    "discovery": {
        "http_enabled": true,
        "https_enabled": true,
        "state": "running"
    },
    "discovery_image": {
        "http_enabled": true,
        "https_enabled": true,
        "state": "running"
    },
    "salt": {
        "http_enabled": true,
        "https_enabled": true,
        "state": "running"
    },
    "abrt": {
        "http_enabled": true,
        "https_enabled": true,
        "state": "running"
    },
    "pulp": {
        "http_enabled": true,
        "https_enabled": true,
        "state": "running"
    },
    "pulpnode": {
        "http_enabled": false,
        "https_enabled": false,
        "state": "disabled"
    },
    "tail": {
        "http_enabled": false,
        "https_enabled": false,
        "state": "disabled"
    },
    "dynflow": {
        "http_enabled": false,
        "https_enabled": false,
        "state": "disabled"
    },
    "facts": {
        "http_enabled": false,
        "https_enabled": false,
        "state": "disabled"
    },
    "dns": {
        "http_enabled": true,
        "https_enabled": true,
        "settings": {
            "use_provider": "dns_libvirt"
        },
        "state": "running",
        "capabilities": [
            "record_type_a"
        ]
    },
    "templates": {
        "http_enabled": true,
        "https_enabled": true,
        "state": "running"
    },
    "tftp": {
        "http_enabled": true,
        "https_enabled": true,
        "settings": {
            "tftp_servername": "192.168.100.1"
        },
        "state": "running"
    },
    "dhcp": {
        "http_enabled": true,
        "https_enabled": true,
        "settings": {
            "use_provider": "dhcp_isc"
        },
        "state": "running"
    },
    "puppetca": {
        "http_enabled": true,
        "https_enabled": true,
        "settings": {
            "use_provider": "puppetca_hostname_whitelisting"
        },
        "state": "running"
    },
    "puppet": {
        "http_enabled": true,
        "https_enabled": true,
        "state": "running"
    },
    "bmc": {
        "http_enabled": true,
        "https_enabled": true,
        "state": "running"
    },
    "realm": {
        "http_enabled": false,
        "https_enabled": false,
        "settings": {
            "use_provider": null
        },
        "state": "disabled"
    },
    "logs": {
        "http_enabled": true,
        "https_enabled": true,
        "state": "running"
    },
    "httpboot": {
        "http_enabled": true,
        "https_enabled": true,
        "state": "running"
    }
}
```

lzap:
Pushed again, not sure what happened with CI.

modules/root/root_v2_api.rb
```diff
@@ -15,8 +15,9 @@ class Proxy::RootV2Api < Sinatra::Base
       attributes = [:http_enabled, :https_enabled, :settings, :state]
 
       plugins = enabled_plugins.inject({}) do |hash, plugin|
-        result = Hash[attributes.map { |attribute| [attribute, plugin[attribute]] }]
-        result[:capabilities] = process_capabilities(plugin[:state], plugin[:capabilities])
+        result = Hash[attributes.map { |attribute| [attribute, plugin[attribute]] }.delete_if { |k, v| v.nil? || (v.instance_of?(Hash) && v.empty?) }]
+        capabilities = process_capabilities(plugin[:state], plugin[:capabilities])
+        result[:capabilities] = capabilities unless capabilities.empty?
```
ekohl, line 20: IMHO we should always have a list, even if it's empty. Application code is much easier that way.

modules/dhcp/dhcp_plugin.rb
```diff
@@ -4,6 +4,7 @@ class Proxy::DhcpPlugin < ::Proxy::Plugin
 
   uses_provider
   default_settings :use_provider => 'dhcp_isc', :server => '127.0.0.1', :subnets => []
+  expose_setting :use_provider
```
ekohl, line 7: If you want to expose this manually, I think it'd be easier to do this in the API. You can use `Plugin.uses_provider?` to determine this.

modules/root/root_v2_api.rb
lzap: You can't assume that a string coming from (bad) internet has some empty elements. When parsing, be defensive. While generally in programming it's good practice to avoid null values, I do think in this case it's considered harmful.

modules/dhcp/dhcp_plugin.rb
lzap: Can you elaborate how exactly you want me to do this?

lzap:
I am going to rebase to get tests passed. I need a feedback here, because I want to test the core patch ideally on top of this version. I think that compacted JSON output is much better.

ekohl:
I'm looking into writing a test for this so we can verify capabilities on modules automatically. Will get back to you on this.

lzap:
Thanks please ping me I don't want to delay @jlsherrill any longer.

lzap:
Fixed Rubocop and amended one tiny change - smart proxy logger incorrectly reports query path.

```diff
diff --git a/lib/proxy/log.rb b/lib/proxy/log.rb
index e4142fb..cabacdd 100644
--- a/lib/proxy/log.rb
+++ b/lib/proxy/log.rb
@@ -79,7 +79,7 @@ module Proxy
     def call(env)
       status = 500
       env['rack.logger'] = logger
-      logger.info { "Started #{env['REQUEST_METHOD']} #{env['PATH_INFO']} #{env['QUERY_STRING']}" }
+      logger.info { "Started #{env['REQUEST_METHOD']} #{env['REQUEST_PATH']} #{env['QUERY_STRING']}" }
       logger.trace { 'Headers: ' + env.select {|k,v| k.start_with? 'HTTP_'}.inspect }
       logger.trace { (body = env['rack.input'].read).empty? ? '' : 'Body: ' + body }
       before = Time.now.to_f
```

It was reported as

```
2019-02-27T16:52:33 2e8814ad [I] Started GET /features 
2019-02-27T16:52:33 2e8814ad [I] Finished GET /features with 200 (0.64 ms)
```

and now it includes the full path correctly:

```
2019-02-27T16:52:33 2e8814ad [I] Started GET /v2/features 
2019-02-27T16:52:33 2e8814ad [I] Finished GET /v2/features with 200 (0.64 ms)
```

modules/dhcp/dhcp_plugin.rb
lzap: If this is your concern I will remove this from this PR. It was just an idea, I have no immediate plan to present this info in Foreman core in any way.

ekohl: I do think it's a valuable debug feature at least, but there's an easier way in `lib/proxy/plugin_initializer.rb` by extending the settings method with:
```ruby
exposed_settings << 'use_provider' if @plugin.uses_provider?
```

Now any module that uses `uses_provider` in the plugin DSL, exposes the `use_provider` setting.

I've found a way to test setting and capabilities for modules and their plugins and I'm preparing a PR with tests and the above fix.

ekohl:
I spent some time to understand all of this. In [my branch](https://github.com/ekohl/smart-proxy/tree/25614-more-caps-improvements) I've split your commit into separate commits to indicate the various changes. With that I've added two commits: 2f9521bc7d5977073cbdbcfad3f0435709e922fd adds basic integration tests and 2f8a59ec7e335e2dc1d0d285004700592647cfb8 implements the simplified provider exposing.

The tests are limited to two classes. Before I implement it for all modules and providers I'd like some feedback on the code. I'm strongly considering a base class to reduce duplication but not sure where to place it.

I'm also not sure about pruning empty settings and capabilities. While I can appreciate that you can't assume anything on network input, it can be useful when writing simple scripts.

I'm also going to add some tests around the DNS capabilities. Ideally we'd add the record_type_a to the dns module so every provider doesn't have to repeat it. Perhaps we should hold off on adding capabilities till we have some proper design around them. Especially with DNS I have some thoughts about this.

lzap:
Well I think that trying to reduce duplication is for the worse in this case. If I diff the two tests, almost every single line is different except:

```
get '/features'
response = JSON.parse(last_response.body)
```

plus the `app` helper and include statement which I consider to be pretty normal everywhere. I encourage you to copy and paste this.

With that said, you've improved it a lot and I am closing this PR, plese reopen with the individual commits which is much cleaner.

Lastly, I think that having `/v2/features` with human-readable output is more important than machine-shell-friendly output given amount of calls I give it during development and troubleshooting via `curl`. Make your own decision however.

PR #637, opened 2019-02-15
Fixes #26591 - Remove puppet_proxy_legacy (Puppet 3)

ekohl:
This drops legacy Puppet support and relies on the Puppet API to retrieve classes. This means at least Puppet 4.4 is required. Since Puppet 4 also removed puppet kick, that provider is removed as well.


ekohl:
Note this needs some discussion and probably some cleaner config migrations. https://community.theforeman.org/t/dropping-puppet-3-in-the-proxy/12901 is open to discuss this.

mmoll:
@ekohl this needs a rebase

mmoll:
@ekohl @ares also, what's the status for 1.22 for this PR?

ekohl:
I didn't have time to look at the needed migrations and it's rather late in the development cycle. Let's do this in 1.23.

ekohl:
Looks like there's an infra issue:
```
+ sed -i '/^\s*gem.*puppet/ s/$/, '\''~> 5.0'\''/' /home/jenkins/workspace/test_proxy_develop_pr_core/puppet/5.0/ruby/2.5/bundler.d/puppet.rb
sed: can't read /home/jenkins/workspace/test_proxy_develop_pr_core/puppet/5.0/ruby/2.5/bundler.d/puppet.rb: No such file or directory
```

https://github.com/theforeman/foreman-infra/pull/1058

mmoll:
[test core]

ekohl:
Not sure why the freeipmi tests now fail. Locally it works. This change shouldn't affect IPMI either. Rebasing again.

lzap:
They usually fail if ipmitool command line app is not installed.

mmoll:
That's installed via puppet on the slaves, I'm a bit puzzled what's the problem...

ekohl:
[test core]

ekohl:
It is reproducable since it failed again, but they should indeed be installed. However, `ipmipower` is in `/sbin` so it's relying on `PATH`:

```
===============================================================================whwich: no ipmipower in (/usr/local/rvm/gems/ruby-2.5.1@test_proxy_develop_pr_core-2/bin:/usr/local/rvm/gems/ruby-2.5.1@global/bin:/usr/local/rvm/rubies/ruby-2.5.1/bin:/usr/local/rvm/bin:/usr/local/bin:/usr/bin)
```

This looks like a badly mocked test since it's testing the API. Will look into this.

ekohl:
I suspect this is changing the path and IPMI was relying on that: https://github.com/puppetlabs/puppet/blob/84f2ae3a27e7bfb915203fbf4d4113aaab813adf/lib/puppet/defaults.rb#L320-L337

ekohl:
I've added a few commits that clean up BMC tests. I've tested these without freeipmi and ipmitool installed on my system. Not entirely sure about their implementation so I'd prefer a good review.

ekohl:
:green_heart: now

ekohl:
While the code is passing tests, I haven't looked at migrations.

config/settings.d/puppet_proxy_puppet_api.yml.example
```diff
@@ -1,9 +1,3 @@
-#
-# puppet_proxy_pupppet_api module is used for puppet versions 4.0 and higher
-#
-# puppet_proxy_pupppet_api is configured automatcially based on
-# :puppet_version setting in smart-proxy's puppet.yml configuration file.
-#
 ---
 # URL of the puppet master itself for API requests.
 #:puppet_url: https://puppet.example.com:8140
```
lzap, line 3: So to understand this, we are completely dropping Ruby Puppet and only support external (Java) Puppet, is that correct?

extra/migrations/20160413000000_migrate_puppet_settings.rb
```diff
@@ -80,8 +70,6 @@ def migrate_puppet_configuration(to_migrate)
       end
     end
 
-    migrated[:puppet][:puppet_version] = puppet_version
-
     migrated
   end
```
lzap, line 75: While I understand that you need to get rid of "require puppet" in this file, generally you should not change passed migrations. I'd expect a new migration that would remove/rename all "legacy" options.

modules/bmc/ipmi.rb
```diff
@@ -18,6 +18,7 @@ def self.logger=(log)
 
       # returns boolean true if the specified provider is installed
       def self.installed?(provider)
+        return false if provider.empty?
```
lzap, line 21: This does not belong here, I would not mind usually but this looks like a big refactoring/change and it can be delayed or whatever.

test/bmc/bmc_test.rb
```diff
@@ -171,6 +172,7 @@ def test_frulist_2xquirk_ipmitool
   end
 
   def test_frulist_2xquirk_fail
+    Rubyipmi.stubs(:is_provider_installed?).returns(true)
```
lzap, line 175: Ditto, if you don't mind pulling this painful bug out of this PR.

lzap (commented):
Nice, however I am not sure if I can test this - only if you guide me step by step setting puppet up.

For migrations, I sholud be able to test this, please provide starting point (YAML) and what's expected after the migration.

config/settings.d/puppet_proxy_puppet_api.yml.example
ekohl: Correct

extra/migrations/20160413000000_migrate_puppet_settings.rb
ekohl: The reason I'm changing this, is that it requires the Puppet gem but we know it's not going to be present. I could mock something up but thought it was easier to modify the migration.

modules/bmc/ipmi.rb
ekohl: This was needed to make the tests pass. Once you remove the Puppet gem, BMC tests start to fail. https://github.com/theforeman/smart-proxy/pull/637#issuecomment-499573545

I don't mind submitting it as a separate PR so we can review/merge it first.

ekohl:
Related installer PR: https://github.com/theforeman/puppet-foreman_proxy/pull/514

lzap (approved):
I am good, let me know how to test this.

ekohl:
To test this, you do a normal installation. Out of the box we enable Puppet. You should be able to provision a host. https://github.com/theforeman/forklift/blob/master/bats/fb-test-puppet.bats covers the Puppet ENC part, but since it's on the same host, it doesn't test the PuppetCA aspect and kickstart/preseed integration.

PR #639, opened 2019-03-05
Fixes #21605 - more authorization options

iNecas:
Before this change, when using authorization helpers, one got all or
nothing, without any chance to use the authorization just of a subset of
the requests.

This patch introduces `Sinatra::Authorization::Helpers` that provide
`do_authorize*` methods that are not wrapped in the before block. so that
they their usage is more flexible.

There are at least 3 places I know of that would benefit from this change:

- https://github.com/theforeman/smart_proxy_dynflow/pull/28
- https://github.com/theforeman/smart_proxy_dynflow/pull/54
- https://github.com/theforeman/smart_proxy_remote_execution_ssh/pull/40

### DIFF
--- a/lib/sinatra/authorization.rb
+++ b/lib/sinatra/authorization.rb
@@ -1,9 +1,30 @@
 module Sinatra
   module Authorization
-    def authorize_with_trusted_hosts
-      helpers ::Proxy::Helpers
+    # Authorization helper for more granular authorization (such as per-route authorization)
+    #
+    # Usage:
+    #
+    #  class MyPlugin < ::Sinatra::Base
+    #    include Sinatra::Authorization::Helpers
+    #
+    #    get '/my_public_api' do
+    #      # ...
+    #    end
+    #
+    #    get 'my_secret_api' do
+    #      # Authorize using the common set of authorization methods
+    #      do_authorize_any
+    #      # ...
+    #    end
+    #  end
+    #
+    module Helpers
+      def self.included(base)
+        base.helpers ::Proxy::Helpers
+        base.helpers ::Proxy::Log
+      end
 
-      before do
+      def do_authorize_with_trusted_hosts
         # When :trusted_hosts is given, we check the client against the list
         # HTTPS: test the certificate CN
         # HTTP: test the reverse DNS entry of the remote IP
@@ -21,16 +42,10 @@ module Sinatra
           unless Proxy::SETTINGS.trusted_hosts.include?(fqdn)
             log_halt 403, "Untrusted client #{fqdn} attempted to access #{request.path_info}. Check :trusted_hosts: in settings.yml"
           end
-
         end
       end
-    end
 
-    def authorize_with_ssl_client
-      helpers ::Proxy::Helpers
-      helpers ::Proxy::Log
-
-      before do
+      def do_authorize_with_ssl_client
         if ['yes', 'on', '1'].include? request.env['HTTPS'].to_s
           if request.env['SSL_CLIENT_CERT'].to_s.empty?
             log_halt 403, "No client SSL certificate supplied"
@@ -40,6 +55,27 @@ module Sinatra
         end
       end
 
+      # Common set of authorization methods used in foreman-proxy
+      def do_authorize_any
+        do_authorize_with_trusted_hosts
+        do_authorize_with_ssl_client
+      end
+    end
+
+    def authorize_with_trusted_hosts
+      include Helpers
+
+      before do
+        do_authorize_with_trusted_hosts
+      end
+    end
+
+    def authorize_with_ssl_client
+      include Helpers
+
+      before do
+        do_authorize_with_ssl_client
+      end
     end
   end
 end
--- a/test/sinatra/authorization_helpers_test.rb
+++ b/test/sinatra/authorization_helpers_test.rb
@@ -1,0 +1,37 @@
+require 'test_helper'
+require 'sinatra/base'
+
+class AuthorizationHelpersTest < Test::Unit::TestCase
+  include Rack::Test::Methods
+
+  def app
+    TestApp.new
+  end
+
+  def test_http
+    get '/public'
+    assert last_response.ok?
+    get '/private'
+    assert last_response.ok?
+  end
+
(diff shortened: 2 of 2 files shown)


test/sinatra/authorization_helpers_test.rb
```diff
@@ -0,0 +1,40 @@
+require 'test_helper'
+require 'json'
+require 'sinatra/base'
+
+ENV['RACK_ENV'] = 'test'
```
ekohl: I've seen this in multiple places, but the test_helper also sets this. Isn't it really redundant?

test/sinatra/authorization_helpers_test.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  def app
+    TestApp.new
+  end
+
+  def test_http
+    get '/public'
+    assert last_response.ok?
+    get '/private'
+    assert last_response.ok?
```
ekohl, line 15: This is interesting and not something I thought about. We added the v2 features endpoint in root, but I suspect that's exposed over https without authentication. Perhaps we need some authentication method that enforces some authentication.

lib/sinatra/authorization.rb
```diff
@@ -40,6 +55,27 @@ def authorize_with_ssl_client
         end
       end
 
+      # Common set of authorization methods used in foreman-proxy
+      def do_authorize
```
ares: perform_authorization perhaps?

ares (commented):
works well, makes sense, just one suggestion for name but I don't insist, let me know what you think

lib/sinatra/authorization.rb
iNecas: You've seen some old code of mine, so you know I don't mind long names, but I still, prefer a bit shorter variant:

```
["do_authorize_with_ssl_client", "perform_authorization_with_ssl_client"].map(&:size)
=> [28, 37]
```

Let's see if there is anyone else that would choose what version is better?

lzap: Hmmm I kinda don't like this very generic `do_authorize` because it does not tell anything about how the authorization is performed. I suggest `do_authorize_any` because basically this is the actual semantics.

lzap:
Works for me. Rubocop failed.

lib/sinatra/authorization.rb
iNecas: Changed to do_authorize_any

iNecas:
Rubocop failures unrelated

lib/sinatra/authorization.rb
```diff
@@ -1,9 +1,30 @@
...
+    #  end
+    #
+    module Helpers
+      def self.included(base)
+        base.helpers ::Proxy::Helpers
+        base.helpers ::Proxy::Log
+      end
 
-      before do
+      def do_authorize_with_trusted_hosts
```
ekohl, line 27: Why do you add the `do_` prefix to it?

test/sinatra/authorization_helpers_test.rb
ekohl: Having tested this: this relies on `localhost` being in the `trusted_hosts`. On a HTTP call it uses reverse DNS + forward DNS. This isn't very secure but since the default deployment only enabled modules on HTTP when they are designed to assume no security it's not that big of a deal.

iNecas: Yes, this is how it have been done for a while: using authenticated http is not very secure and people should count on it. Anyway, not trying to address it here

test/sinatra/authorization_helpers_test.rb
iNecas: Removed and seemed it still worked. Also removed the `require 'json'` as we didn't need it actually

test/sinatra/authorization_helpers_test.rb
ekohl: I do think it would be good to capture this in a test. Perhaps set the trusted hosts to empty or something different and assert that the response it not ok? Could be separate tests.

lib/sinatra/authorization.rb
iNecas: to make it different than already existing methods: I don't like to have methods with the same name doing a bit different thing

test/sinatra/authorization_helpers_test.rb
iNecas: The subject of this test is ensuring the authorization helpers are available. I'm not touching the authorization algorigth itself, and the tests for it are somewhere else: I would prefer keeping this PR focused on thing only: which is making the methods available outside of the before filters, that right now we are constraint with.

lzap (approved):
I am good, works fine. Thanks.

test/sinatra/authorization_helpers_test.rb
lzap: Agreed, more tests is good but I want this to be merged ASAP, let's improve somewhere else.

lib/sinatra/authorization.rb
ekohl: Ah, I missed that it's the helpers module vs methods on the higher level. This part of Ruby is still rather new to me.

test/sinatra/authorization_helpers_test.rb
ekohl: My argument was that if the response is ok, you still don't really know if the authorization was properly called.

lib/sinatra/authorization.rb
```diff
@@ -40,6 +55,27 @@ def authorize_with_ssl_client
         end
       end
 
+      # Common set of authorization methods used in foreman-proxy
+      def do_authorize_any
```
ekohl, line 59: IMHO any is a bit ambiguous. Any could also mean authorized any client without authentication. Perhaps `do_authorize_any_method` would be clearer but naming is hard.

lzap: Yeah I was thinking maybe `do_authorize_all` because we actually do not authorize but rather refuse unauthorized requests (the opposite). It's complicated, I was thinking like `any` reads a bit cleaner in a sense "any method will do it".

PR #638, opened 2019-02-26
Fixes #26322 - Expose the puppet URL settings

ekohl:
This allows the Foreman to serve the correct hostname to clients. This is important when the Puppet URL is different from the Proxy URL.

PR #642, opened 2019-03-11
Refs #25614 - Always expose the provider setting

ekohl:
This also adds integration tests for some modules.


mmoll (dismissed):
Only checked visually, but looks correct.

tbrisker:
Pushed another commit with tests for other modules. logs and templates seem to work without the setting file expectation, seems like for some reason this method isn't called for them.

test/realm_freeipa/integration_test.rb
```diff
@@ -0,0 +1,30 @@
+require 'test_helper'
+require 'json'
+require 'root/root_v2_api'
+# require 'realm_freeipa/configuration_loader'
```
ekohl: Could you remove the commented code here?

ekohl (commented):
I also considered testing whether it's running, but DHCP at least fails so I opted not to. If we can find an easy way to add why a plugin failed to the assertion that'd be a good test as well if we can fix it.

test/realm_freeipa/integration_test.rb
tbrisker: oops, fixed

tbrisker:
[test core]

ekohl:
Could you fix the rubocop issues?

tbrisker:
:robot: :policeman:  fixed

modules/templates/templates_plugin.rb
```diff
@@ -6,9 +6,5 @@ class Plugin < ::Proxy::Plugin
     plugin :templates, ::Proxy::VERSION
 
     validate_presence :template_url
-
-    after_activation do
-      loading_failed "missing :foreman_url: from configuration." unless Proxy::SETTINGS.foreman_url
```
tbrisker: This is actually required. TemplatesAPI creates a [TemplateProxyRequest](https://github.com/theforeman/smart-proxy/blob/develop/modules/templates/templates_api.rb#L19), which calls [`request_factory`](https://github.com/theforeman/smart-proxy/blob/develop/modules/templates/template_proxy_request.rb#L26) which [requires this setting](https://github.com/theforeman/smart-proxy/blob/develop/lib/proxy/request.rb#L57).

test/templates/integration_test.rb
```diff
@@ -0,0 +1,29 @@
...
+class TemplatesApiFeaturesTest < Test::Unit::TestCase
+  include Rack::Test::Methods
+
+  def app
+    Proxy::PluginInitializer.new(Proxy::Plugins.instance).initialize_plugins
+    Proxy::RootV2Api.new
+  end
+
+  def test_features
+    # not sure why, but this module loads without loading the config file
```
tbrisker: can this comment be removed then?

modules/templates/templates_plugin.rb
ekohl: Ah, that does explain it.

mmoll:
@ekohl which should go in first, GH-638 or this one?

ekohl:
Rebased and squashed. @tbrisker I've added you as `Co-Authored-By` on the integration tests. I think this is a cleaner history. Should be ready to be merged.

tbrisker (approved):
LGTM, but since i wrote part of the code i'll leave it to @mmoll to push the button

mmoll:
merged, bedankt @ekohl and thanks @tbrisker!

PR #643, opened 2019-03-13
Fixes #26329 - TFTP not needed active for HTTPBoot

lzap:
I don'ŧ remember why I added this, I think I tought that TFTP is a separate plugin. Not needed.


mmoll:
merged, díky @lzap!

PR #645, opened 2019-04-05
Refs #26537 - adjust xmlrpc dependency

mmoll:
this should be built into:

* [x] Nightly
* [x] 1.21
* [x] 1.20


mmoll:
While 0.2 required Ruby 2.4, 0.3 is also OK with 2.3. I'd need this to get the proxy going with Ubuntu/xenial and Debian/stretch, which bring Ruby 2.3.

ekohl (approved):
I do wonder why the gem is only required on newer versions. Do older versions of ruby bundle xmlrpc?

mmoll:
@ekohl It's indeed a 2.4+ thing... I'll solve this only in packaging then :+1:

PR #640, opened 2019-03-05
[WIP] Refs #25614 - Various changes to the capabilities

ekohl:
See the individual commits. Going to review the commits (will probably leave DNS capabilities out) and add more tests, ideally all modules and their providers.

PR #647, opened 2019-04-18
Fixes #26642 - logging post data fix

lzap:
My work and Ivan's review in smart proxy dynflow core revealed nasty bug in our logging stack in smart proxy. This luckily only hits when FOREMAN_PROXY_TRACE env variable is set to true which is only for debugging purposes. Rack input (StringIO) is read however it is never rewinded back, therefore later on the data is empty therefore POST/PUT services do not work correctly. Additionally, time tracking code can lead to nil error during exceptions.

### DIFF
--- a/lib/proxy/log.rb
+++ b/lib/proxy/log.rb
@@ -77,12 +77,19 @@ module Proxy
     end
 
     def call(env)
+      before = Time.now.to_f
       status = 500
       env['rack.logger'] = logger
-      logger.info { "Started #{env['REQUEST_METHOD']} #{env['PATH_INFO']} #{env['QUERY_STRING']}" }
+      logger.info { "Started #{env['REQUEST_METHOD']} #{env['REQUEST_PATH']} #{env['QUERY_STRING']}" }
       logger.trace { 'Headers: ' + env.select {|k,v| k.start_with? 'HTTP_'}.inspect }
-      logger.trace { (body = env['rack.input'].read).empty? ? '' : 'Body: ' + body }
-      before = Time.now.to_f
+      logger.trace do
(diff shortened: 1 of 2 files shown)


lib/proxy/log.rb
```diff
@@ -77,12 +77,15 @@ def initialize(app)
...
 
     def call(env)
+      before = Time.now.to_f
       status = 500
       env['rack.logger'] = logger
       logger.info { "Started #{env['REQUEST_METHOD']} #{env['PATH_INFO']} #{env['QUERY_STRING']}" }
       logger.trace { 'Headers: ' + env.select {|k,v| k.start_with? 'HTTP_'}.inspect }
-      logger.trace { (body = env['rack.input'].read).empty? ? '' : 'Body: ' + body }
-      before = Time.now.to_f
+      if logger.trace? && env['rack.input']
```
iNecas: What about using the block variant instead of the ifs:

```ruby
    logger.trace do
        if env['rack.input'] && !(body = env['rack.input'].read).empty?
          env['rack.input'].rewind
          "Body: #{body}"
        else
          ''
        end
      end
```

lzap:
Rebased, works fine.

lzap:
Two small logging improvements also added which were previously added in https://github.com/theforeman/smart-proxy/pull/640/files but @ekohl closed the PR:

* exceptions via halt helper are logged via structured helper logger
* path is logged as full instead of relative Sinatra path (it is now `/dhcp/MAC` instead of just confusing `/MAC`)

mmoll:
[test core]

mmoll:
merged, díky @lzap!

PR #649, opened 2019-05-04
Fixes #26742 - use Ruby 2.0 compatible gem versions

mmoll:


mmoll:
This should get cherry-picked down to all supported versions in order to have the tests running reliably also on Ruby 2.0.

bundler.d/dhcp_isc.rb
```diff
@@ -1,6 +1,10 @@
 group :dhcp_isc_inotify do
   gem 'rsec'
-  gem 'rb-inotify'
+  if RUBY_VERSION < '2.2'
+    gem 'rb-inotify', '< 0.9'
+  else
+    gem 'rb-inotify'
+  end
```
lzap, line 7: Version 0.9.x still works with 2.0 ruby according to rubygems.org so this would do it:

```ruby
  if RUBY_VERSION < '2.1'
    gem 'rb-inotify', '~> 0.9.10'
  else
    gem 'rb-inotify'
  end
```

Gemfile
```diff
@@ -5,6 +5,7 @@ gemspec
 if RUBY_VERSION < '2.2'
   gem 'sinatra', '< 2'
   gem 'rack', '>= 1.1', '< 2.0.0'
+  gem 'parallel', '< 1.14'
```
lzap: It's probably better to pin the dependency:

```ruby
  if RUBY_VERSION < '2.1'
    gem 'rubocop-checkstyle_formatter', '~> 0.3.0'
  else
    gem 'rubocop-checkstyle_formatter'
  end
```

lzap (dismissed):
Few issues, I've actually pinned those in my PR today if you prefer merging that one:

https://github.com/theforeman/smart-proxy/pull/650/files

Gemfile
mmoll: This might or might not work implicitly, but `rubocop-checkstyle_formatter` itself is not the problem, but `rubocop` with it's pin of `parallel` not being strict enough.

bundler.d/dhcp_isc.rb
mmoll: In theory, this allows me to use `rb-inotify` 0.8 on Ruby 2.2+, but not on Ruby 2.1-, updated the pin from `'< 0.9'` to `'< 0.10'`

mmoll:
@lzap I'd prefer this PR to get into a mergeable shape first, as this is a separate issue.

bundler.d/dhcp_isc.rb
mmoll: hm, my version does this, too... :man_shrugging:

ekohl (approved):
Tests are green so I don't mind merging this now.

lzap:
Sorry for letting you to wait, just rebase and I am good.

Gemfile
mmoll: I took this ouf of the Gemfile now, no idea why it didn't work only in test.rb at the first try :man_shrugging:

mmoll:
should be ready now :bowtie:

lzap (approved):
OK then. I will rebase mine.

PR #650, opened 2019-05-13
Fixes #26785 - improved exception handling

lzap:
Do not merge, please review and then I am gonna refactor all `log_halt` calls. The idea is to have this backward compatible because plugins use this method. It's a messy but what can I do.

### DIFF
--- a/lib/proxy/helpers.rb
+++ b/lib/proxy/helpers.rb
@@ -7,8 +7,12 @@ module Proxy::Helpers
   # Accepts a html error code and a message, which is then returned to the caller after adding to the proxy log
   # OR  a block which is executed and its errors handled in a similar way.
   # If no code is supplied when the block is declared then the html error used is 400.
-  def log_halt code=nil, exception=nil
-    message = exception.is_a?(String) ? exception : exception.to_s
+  def log_halt code=nil, exception_or_msg=nil, custom_msg=nil
+    message = exception_or_msg.to_s
+    message = "#{custom_msg}: #{message}" if custom_msg
+    exception = exception_or_msg.is_a?(Exception) ? exception_or_msg : Exception.new(exception_or_msg)
+    # just in case exception is passed in the 3rd parameter let's not loose the valuable info
+    exception = custom_msg.is_a?(Exception) ? custom_msg : exception
     begin
       if block_given?
         return yield
@@ -16,7 +20,7 @@ module Proxy::Helpers
     rescue => e
       exception = e
       message += e.message
-      code     = code || 400
+      code = code || 400
     end
     content_type :json if request.accept?("application/json")
     logger.error message, exception
--- a/modules/puppetca/puppetca_api.rb
+++ b/modules/puppetca/puppetca_api.rb
@@ -13,7 +13,7 @@ module Proxy::PuppetCa
       begin
         puppetca_impl.list.to_json
       rescue => e
-        log_halt 406, "Failed to list certificates: #{e}"
+        log_halt 406, e, "Failed to list certificates"
       end
     end
 
@@ -22,7 +22,7 @@ module Proxy::PuppetCa
       begin
         autosigner.autosign_list.to_json
       rescue => e
-        log_halt 406, "Failed to list autosign entries: #{e}"
+        log_halt 406, e, "Failed to list autosign entries"
       end
     end
 
@@ -33,7 +33,7 @@ module Proxy::PuppetCa
       begin
         autosigner.autosign(certname, token_ttl)
       rescue => e
-        log_halt 406, "Failed to enable autosign for #{certname}: #{e}"
+        log_halt 406, e, "Failed to enable autosign for #{certname}"
       end
     end
 
@@ -46,7 +46,7 @@ module Proxy::PuppetCa
         request.body.rewind
         autosigner.validate_csr(request.body.read) ? 200 : 404
       rescue => e
-        log_halt 406, "Failed to validate CSR: #{e}"
+        log_halt 406, e, "Failed to validate CSR"
       end
     end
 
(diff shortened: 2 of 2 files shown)


lib/proxy/helpers.rb
```diff
@@ -7,16 +7,18 @@ module Proxy::Helpers
   # Accepts a html error code and a message, which is then returned to the caller after adding to the proxy log
   # OR  a block which is executed and its errors handled in a similar way.
   # If no code is supplied when the block is declared then the html error used is 400.
-  def log_halt code=nil, exception=nil
-    message = exception.is_a?(String) ? exception : exception.to_s
+  def log_halt code=nil, exception_or_msg=nil, custom_msg=nil
```
ekohl, line 10: I never knew the second argument is an exception. I *always* thought it was a message since that's pretty much used everywhere. To me a signature this signature feels more natural but not sure if this is compatible:
```suggestion
  def log_halt(code=nil, message=nil, exception=nil)
```
I also find it much easier to read with braces, but that may be a personal preference.

Lastly, should code default to 400 instead of inside the body?

lzap: Yeah it's a top secret. I wish I could change this, but almost every plugin use this helper...

lzap:
Merged your suggestion, make sure to squash when merging!

tbrisker:
@lzap 360 failed tests seem related

lzap:
@ekohl or @tbrisker hey this is new:

    Gem::RuntimeRequirementNotMetError: parallel requires Ruby version >= 2.2.
    Gem::RuntimeRequirementNotMetError: rb-inotify requires Ruby version >= 2.2.

Strange, rb-inotify was updated in January this year, why jenkins asks now? Shall I pin those two?

lzap:
Strange Rubocop requires parallel 0.10 however something (not sure what it's in the root tree in Gemfile.lock) requires the latest parallel.

lzap:
I think this broke it: https://rubygems.org/gems/rubocop-checkstyle_formatter but not yet sure about rb-inotify (no change there).

lzap:
All green, pinned two gems.

lzap:
Dropped version pinning, rebased.

lib/proxy/helpers.rb
```diff
@@ -7,16 +7,18 @@ module Proxy::Helpers
...
+    exception = exception_or_msg.is_a?(Exception) ? exception_or_msg.message : exception_or_msg
     begin
       if block_given?
         return yield
       end
     rescue => e
       exception = e
       message += e.message
-      code     = code || 400
+      code = code || 400
```
ekohl, line 23: Sort of related is that I'd deprecate the calling of `log_halt` without a code. Probably better to do it in a separate PR, but it looks like everything in the proxy uses an explicit code anyway (as you should).

lib/proxy/helpers.rb
```diff
@@ -7,16 +7,18 @@ module Proxy::Helpers
   # Accepts a html error code and a message, which is then returned to the caller after adding to the proxy log
   # OR  a block which is executed and its errors handled in a similar way.
   # If no code is supplied when the block is declared then the html error used is 400.
-  def log_halt code=nil, exception=nil
-    message = exception.is_a?(String) ? exception : exception.to_s
+  def log_halt code=nil, exception_or_msg=nil, custom_msg=nil
+    message = exception_or_msg.is_a?(String) ? exception_or_msg : exception_or_msg.to_s
+    message = "#{custom_msg}: #{message}" if custom_msg
+    exception = exception_or_msg.is_a?(Exception) ? exception_or_msg.message : exception_or_msg
```
ekohl: This here breaks line 25 since it means `exception.is?(Exception)` will always be false meaning you no longer get backtraces.

ekohl (changes requested):

lib/proxy/helpers.rb
lzap: I am fine doing it in this PR however this will break all plugins, almost all of them use this and can possibly use this. Deprecating could do it but I feel like it is not probably worth it.

lzap:
Good find, here is result from my testing:

```
log_halt 500, "test"

2019-06-26T11:19:43 b93408ce [E] test
2019-06-26T11:19:43 b93408ce [W] test
Exception: test
```

```
raise("raised exception") rescue log_halt 400, e #pseudo code

2019-06-26T11:21:08 471de61c [E] raised exception
2019-06-26T11:21:08 471de61c [W] raised exception
RuntimeError: raised exception
/home/lzap/work/smart-proxy/modules/root/root_api.rb:6:in `block in <class:RootApi>'
... stacktrace ...

```

```
raise("raised exception") rescue log_halt 400, e, "Blah blah error" #pseudo code

2019-06-26T11:23:05 478060f0 [E] Blah blah error: raised exception
2019-06-26T11:23:05 478060f0 [W] Blah blah error: raised exception
RuntimeError: raised exception
/home/lzap/work/smart-proxy/modules/root/root_api.rb:6:in `block in <class:RootApi>'
/home/lzap/.rbenv/versions/2.6.3/lib/ruby/gems/2.6.0/gems/sinatra-2.0.5/lib/sinatra/base.rb:1635:in `call'
```

I made sure that even if the caller swaps 2nd and 3rd parameter by accident (it's hard to test error states) it prints everything nicely. The output is the same as the last example above.

lzap:
Rebased!

lzap:
Tests are green, ready for rereview.

lib/proxy/helpers.rb
ekohl: I also would like to see that deprecation. A quick grep though the plugins I have checked out all use log_halt with a code so in 1.25 we could make it a required parameter.

ekohl (approved):
It's not great, but given the previous API and its usage I think this is the only way.

On a related note, I was looking through the plugins that use this and see a log of forcing it to a string in various places. I'm wondering how we can communicate this to plugin authors.

@xprazak2 looks like the API allows for a much better debugging by keeping the traces and smart_proxy_openscap doesn't do this. Perhaps you have thoughts on how this could be communicated as well?

lzap:
Rebased, let's merge this one?

xprazak2:
@ekohl, I do not have a better idea TBH.

lzap:
Why I already see "This branch is out-of-date with the base branch" when I rebased yesterday?!? I don't get it, rebase was clean.

PR #653, opened 2019-06-05
Fixes #26962 - put session id into mdc too

lzap:
Proxy currently only sends request id to logs, session id can be useful too. One more improvement - logging pattern should only be the message, no logging level or request id because these are reported using structured fields.

Core required PR: https://github.com/theforeman/smart-proxy/pull/653


ehelms (approved):

tbrisker (approved):
Thanks @lzap !

PR #655, opened 2019-06-07
Fixes #26986 - reordered authorization checkers

lzap:
In trusted hosts authorization we do DNS reverse lookup. We do trusted hosts first and then SSL cert verify, we should do the other way around as SSL is more common on production deployments. Also, we do not set any DNS resolution timeout, therefore it defaults to 60/90 seconds on most systems which slows every single request.

### DIFF
--- a/modules/dhcp_common/isc/omapi_provider.rb
+++ b/modules/dhcp_common/isc/omapi_provider.rb
@@ -200,7 +200,7 @@ module Proxy::DHCP::CommonISC
         ns = ip2hex validate_ip(server)
       rescue
         begin
-          ns = ip2hex Resolv.new.getaddress(server)
+          ns = ip2hex resolver.getaddress(Resolv.new)
         rescue
           logger.warn "Failed to resolve IP address for #{server}"
           ns = "\\\"#{server}\\\""
--- a/lib/proxy/helpers.rb
+++ b/lib/proxy/helpers.rb
@@ -72,8 +72,10 @@ module Proxy::Helpers
     begin
       dns = Resolv.new
       fqdn = dns.getname(ip)
+    rescue Resolv::ResolvTimeout => e
+      log_halt 403, "timeout while resolving hostname for ip address #{ip}: #{e.message}"
     rescue Resolv::ResolvError => e
-      log_halt 403, "unable to resolve hostname for ip address #{ip}\n\n#{e.message}"
+      log_halt 403, "unable to resolve hostname for ip address #{ip}: #{e.message}"
     end
 
     unless forward_verify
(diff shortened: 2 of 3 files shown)


modules/dhcp_common/isc/omapi_provider.rb
```diff
@@ -200,7 +200,9 @@ def bootServer server
         ns = ip2hex validate_ip(server)
       rescue
         begin
-          ns = ip2hex Resolv.new.getaddress(server)
+          resolver = Resolv.new
```
ekohl: Should we introduce a resolv helper with the right settings?

modules/dhcp_common/isc/omapi_provider.rb
```diff
@@ -200,7 +200,9 @@ def bootServer server
...
+          resolver.timeouts = 5
```
ekohl: I don't think this exists so I doubt it has an effect.

```
[5] pry(main)> resolver = Resolv.new
=> #<Resolv:0x0000555ba96bd5f0
 @resolvers=
  [#<Resolv::Hosts:0x0000555ba96bd578
    @filename="/etc/hosts",
    @initialized=nil,
    @mutex=#<Thread::Mutex:0x0000555ba96bd528>>,
   #<Resolv::DNS:0x0000555ba96bd4b0
    @config=
     #<Resolv::DNS::Config:0x0000555ba96bd2d0
      @config_info=nil,
      @initialized=nil,
      @mutex=#<Thread::Mutex:0x0000555ba96bd258>,
      @timeouts=nil>,
    @initialized=nil,
    @mutex=#<Thread::Mutex:0x0000555ba96bd488>>]>
[6] pry(main)> resolver.timeouts
NoMethodError: undefined method `timeouts' for #<Resolv:0x0000555ba96bd5f0>
from (pry):6:in `__pry__'
```

ekohl (changes requested):
Could you split this into two PRs? The authorization needs a cherry pick to 1.22 and might even be CVE worthy.

cc @tbrisker

modules/dhcp_common/isc/omapi_provider.rb
lzap: I have the other PR, maybe there? https://github.com/theforeman/smart-proxy/pull/657

modules/dhcp_common/isc/omapi_provider.rb
lzap: Yeah because this is not Resolv::DNS but just Resolv generic class.

lzap:
Rebased, I removed all timeout related stuff because you are right that it does not belong here. Maybe except catching timeout error in a nicer way.

However I don't understand why this would needed to be cherry picked or related in any way to security - this is just better ordering of the checkers. But now when I am thinking about this change, this might actually be a change in behavior: when user have an entry in trusted hosts and calling proxy with an invalid SSL client cert, previously it was authorized, however after the patch this would be failure.

I might just close the PR actually, let's confirm.

ekohl:
I think that was the old way: first SSL cert checks and then DNS checks as a fall back.

lzap:
I am closing this PR now, I would really like to refactor https://github.com/theforeman/smart-proxy/pull/657/files patch and provide a generic resolver for all our code parts dealing with resolving DNS that would properly log information to the logger, specifically when something takes too much time to resolve.

PR #656, opened 2019-06-11
[WIP] fixes #27022 - settings via env variables

timogoebel:
This is still a very early version of the patch.

@ekohl & @ohadlevy: I'm using the definition of the default settings to guess the type of a setting. Is it safe to assume that every setting has a default setting defined? Or would we need to extend the plugin dsl to define settings and their type?

### DIFF
--- a/lib/proxy/plugin_validators.rb
+++ b/lib/proxy/plugin_validators.rb
@@ -8,7 +8,7 @@ module ::Proxy::PluginValidators
     end
 
     def required_setting?
-      @plugin.plugin_default_settings.has_key?(@setting_name)
+      @plugin.plugin_default_settings.has_key?(@setting_name) && !@plugin.plugin_default_settings[@setting_name].nil?
     end
 
     def evaluate_predicate(settings)
--- a/lib/proxy/settings.rb
+++ b/lib/proxy/settings.rb
@@ -10,15 +10,16 @@ module Proxy::Settings
   def self.initialize_global_settings(settings_path = nil, argv = ARGV)
     global = ::Proxy::Settings::Global.new(YAML.load(File.read(settings_path || SETTINGS_PATH)))
     global.apply_argv(argv)
+    global.apply_env(ENV)
     global
   end
 
   def self.load_plugin_settings(defaults, settings_file, settings_directory = nil)
     settings = {}
     begin
-      settings = YAML.load(File.read(File.join(settings_directory || ::Proxy::SETTINGS.settings_directory, settings_file))) || {}
+      settings = read_settings_file(settings_file, settings_directory)
     rescue Errno::ENOENT
-      logger.warn("Couldn't find settings file #{settings_directory || ::Proxy::SETTINGS.settings_directory}/#{settings_file}. Using default settings.")
+      logger.warn("Couldn't find settings file #{File.join(settings_directory || ::Proxy::SETTINGS.settings_directory, settings_file)}. Using default settings.")
     end
     ::Proxy::Settings::Plugin.new(defaults, settings)
   end
--- a/lib/proxy/settings/global.rb
+++ b/lib/proxy/settings/global.rb
@@ -1,3 +1,6 @@
+require 'proxy/settings_from_env'
+require 'ostruct'
+
 module ::Proxy::Settings
   class Global < ::OpenStruct
     DEFAULT_SETTINGS = {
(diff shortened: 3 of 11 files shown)


lib/proxy/settings.rb
```diff
@@ -10,6 +10,7 @@ module Proxy::Settings
   def self.initialize_global_settings(settings_path = nil, argv = ARGV)
     global = ::Proxy::Settings::Global.new(YAML.load(File.read(settings_path || SETTINGS_PATH)))
     global.apply_argv(argv)
+    global.apply_env(ENV)
```
ekohl, line 13: Would it make sense to make this a parameter as well, just like argv so it's easier to test?

lib/proxy/settings/global.rb
```diff
@@ -53,5 +53,51 @@ def apply_argv(args)
       self.daemon = true if args.include?('--daemonize')
       self.daemon = false if args.include?('--no-daemonize')
     end
+
+    def apply_env(env)
+      DEFAULT_SETTINGS.each do |setting_key, setting_value|
+        env_key = "PROXY_#{setting_key.to_s.upcase}"
```
ekohl: It may be longer, but maybe `FOREMAN_PROXY_` or `SMART_PROXY_`? Proxy can be ambiguous and I wonder if it can interfere with something.

ekohl (commented):
Plugins have their own settings. Any plan to incorporate that? I was thinking about `FOREMAN_PROXY_<plugin>_<setting>`. I like the idea of a DSL for it, but I can tell you that there are settings without defaults. IMHO a `setting` DSL would be good. Perhaps something like:
```ruby
setting :puppet_url, cast: :url, required: true, expose: true
setting :puppet_key, cast: :path, required: true
setting :puppet_cert, cast: :path, required: true
setting :puppet_ca, cast: :path, required: true, default: '/etc/puppetlabs/puppet/ssl/certs/ca.pem'
```

lib/proxy/settings.rb
timogoebel: It already is a parameter, for exactly that reason. ;-)

timogoebel:
> Plugins have their own settings. Any plan to incorporate that?

Yes, definitely.

> IMHO a `setting` DSL would be good

This would require changes across all plugins/modules. I'd love to have a solution that works without any changes to the plugin DSL. Ideally, all settings have a default. Or can have one. I'll work on this more and we'll see.

timogoebel:
Added settings for plugins.

TODOs:
- [x] add tests
- [ ] add missing default settings for all modules

timogoebel:
Basically, we have two options to choose from:

1. Require all settings to have `default_settings` set. If the default value is nil, the setting is treated as optional. If settings are defined in a config file but aren't specified as default settings, log a warning. These settings cannot be overridden by an env variable.

2. Add a new settings dsl and add a compatibility layer that adds settings definition based on default settings etc. Eventually, deprecate the current settings api.

Basically, 1 is already implemented and the least intrusive option. 2 would probably be the cleanest solution, but is a lot more effort for little gain.

@ekohl: What do you think?

lib/proxy/plugin_validators.rb
```diff
@@ -8,7 +8,7 @@ def initialize(plugin, setting_name, params, predicate)
     end
 
     def required_setting?
-      @plugin.plugin_default_settings.has_key?(@setting_name)
+      @plugin.plugin_default_settings.has_key?(@setting_name) && !@plugin.plugin_default_settings[@setting_name].nil?
```
iNecas, line 11: Why do we need this change?

timogoebel: Prior to this commit, not specifying a setting at all means that the setting is optional. For the setting to show up as an env variable, we need to explicitly define all settings. So setting the default to `nil` now means the setting is optional. Makes sense?

lib/proxy/settings_from_env.rb
```diff
@@ -0,0 +1,39 @@
...
+      :list
+    when TrueClass, FalseClass
+      :boolean
+    when Float
+      :float
+    when Integer
+      :integer
+    when Symbol
+      :symbol
+    else
```
iNecas, line 14: Isn't `Hash/:dict` missing here?

iNecas:
Waiting for the all the changes before testing this fully. One small comment code-wise

ares:
@timogoebel we started to also triage proxy PRs and wanted to check, if you see yourself working on this in foreseeable future. It's still labeled as WIP and there seems to be no update since June. Thanks for letting us know.

timogoebel:
I'll close this for the time being.

PR #657, opened 2019-06-12
Fixes #27032 - improve logging and timeout

lzap:
Foreman proxy DNS module performs record resolve operating during create request, but there is not a single log line about it. It is worth adding also timing information because this can be cruical for finding why some DNS requests are slow.

### DIFF
--- a/lib/proxy/helpers.rb
+++ b/lib/proxy/helpers.rb
@@ -1,5 +1,5 @@
 require 'openssl'
-require 'resolv'
+require 'proxy/logging_resolv'
 
 module Proxy::Helpers
   include Proxy::Log
@@ -68,13 +68,23 @@ module Proxy::Helpers
     json_data
   end
 
+  def dns_resolv(*args)
+      resolv = Resolv::DNS.new(*args)
+      resolv.timeouts = Proxy::SETTINGS.dns_resolv_timeouts
+      ::Proxy::LoggingResolv.new(resolv)
+  end
+
+  def resolv(*args)
+      ::Proxy::LoggingResolv.new(Resolv.new(*args))
+  end
+
   # reverse lookup an IP address while verifying it via forward resolv
   def remote_fqdn(forward_verify=true)
     ip = request.env['REMOTE_ADDR']
     log_halt 403, 'could not get remote address from environment' if ip.empty?
 
     begin
-      dns = Resolv.new
+      dns = resolv
       fqdn = dns.getname(ip)
     rescue Resolv::ResolvError => e
       log_halt 403, "unable to resolve hostname for ip address #{ip}\n\n#{e.message}"
--- a/lib/proxy/logging_resolv.rb
+++ b/lib/proxy/logging_resolv.rb
@@ -1,0 +1,65 @@
+require 'resolv'
+require 'proxy/time_utils'
+require 'forwardable'
+
+# Decorator for Resolv and DNS::Resolv Ruby classes that performs
+# logging into the smart proxy logger.
+module Proxy
+  class LoggingResolv
+    include ::Proxy::Log
+    include ::Proxy::TimeUtils
+    extend Forwardable
+
+    def_delegators :@resolv, :each_address, :each_name
+
+    SLOW_DNS_QUERY_MS = 9000
+
+    def initialize(resolv)
+      @resolv = resolv
+    end
+
+    def resolver(override_nameserver = @server)
+      dns_resolv(:nameserver => override_nameserver)
+    end
+
+    def getresource(name, typeclass)
+      call_with_timing(:getresource, name, typeclass)
+    end
+
+    def getresources(name, typeclass)
+      call_with_timing(:getresources, name, typeclass)
+    end
+
+    def getname(name)
+      call_with_timing(:getname, name)
+    end
+
+    def getnames(name)
+      call_with_timing(:getnames, name)
+    end
+
+    def getaddress(name)
+      call_with_timing(:getaddress, name)
+    end
+
+    def getaddresses(name)
+      call_with_timing(:getaddresses, name)
+    end
+
+    private
+
+    def call_with_timing(method, *args)
+      result = nil
+      duration = time_spent_in_ms do
+        result = @resolv.send(method, *args)
+      end
+      if duration < SLOW_DNS_QUERY_MS
+        logger.debug "Finished DNS query #{method} for '#{args.first}' in #{duration.round(2)} ms"
+      else
+        logger.warn "Slow DNS query #{method} for #{args.inspect} took #{duration.round(2)} ms"
+        logger.debug "Resolver used: #{@resolv.inspect}"
+      end
+      result
+    end
+  end
+end
--- a/modules/dns_common/dns_common.rb
+++ b/modules/dns_common/dns_common.rb
@@ -1,5 +1,6 @@
 require 'resolv'
 require 'ipaddr'
+require 'proxy/logging_resolv'
 
 module Proxy::Dns
   class Error < RuntimeError; end
@@ -8,16 +9,18 @@ module Proxy::Dns
 
   class Record
     include ::Proxy::Log
+    include ::Proxy::TimeUtils
+    include ::Proxy::Helpers
 
     attr_reader :server, :ttl
 
     def initialize(server = nil, ttl = nil)
       @server = server || "localhost"
-      @ttl    = ttl || "86400"
+      @ttl = ttl || "86400"
     end
 
-    def resolver
-      Resolv::DNS.new(:nameserver => @server)
+    def resolver(override_nameserver = @server)
+      dns_resolv(:nameserver => override_nameserver)
     end
 
     def create_srv_record(service, value)
--- a/config/settings.yml.example
+++ b/config/settings.yml.example
@@ -79,3 +79,10 @@
 # which is about 500 kB request.
 #:log_buffer: 2000
 #:log_buffer_errors: 1000
+
+# DNS resolver timeout(s). This may be a single positive number
+# or an array of positive numbers representing timeouts in seconds.
+# If an array is specified, a DNS request will retry and wait for
+# each successive interval in the array until a successful response
+# is received. See Ruby Resolv#timeouts documentation for more info.
+#:dns_resolv_timeouts: [5, 8, 13]
--- a/lib/proxy/log.rb
+++ b/lib/proxy/log.rb
@@ -1,5 +1,6 @@
 require 'logging'
 require 'proxy/log_buffer/decorator'
+require 'proxy/time_utils'
 
 module Proxy
   module Log
@@ -74,13 +75,14 @@ module Proxy
 
   class LoggerMiddleware
     include Log
+    include ::Proxy::TimeUtils
 
     def initialize(app)
       @app = app
     end
 
     def call(env)
-      before = Time.now.to_f
+      before = time_monotonic
       status = 500
       env['rack.logger'] = logger
       logger.info { "Started #{env['REQUEST_METHOD']} #{env['REQUEST_PATH']} #{env['QUERY_STRING']}" }
@@ -99,7 +101,7 @@ module Proxy
       raise e
     ensure
       logger.info do
-        after = Time.now.to_f
+        after = time_monotonic
         duration = (after - before) * 1000
         "Finished #{env["REQUEST_METHOD"]} #{env["REQUEST_PATH"]} with #{status} (#{duration.round(2)} ms)"
       end
--- a/lib/proxy/settings/global.rb
+++ b/lib/proxy/settings/global.rb
@@ -17,7 +17,8 @@ module ::Proxy::Settings
       :log_buffer => 2000,
       :log_buffer_errors => 1000,
       :ssl_disabled_ciphers => [],
-      :tls_disabled_versions => []
+      :tls_disabled_versions => [],
+      :dns_resolv_timeouts => [5, 8, 13] # Ruby default is [5, 20, 40] which is a bit too much for us
     }
 
     HOW_TO_NORMALIZE = {
--- a/lib/proxy/time_utils.rb
+++ b/lib/proxy/time_utils.rb
@@ -1,0 +1,23 @@
+module Proxy::TimeUtils
+  # monotonic timers are only on Ruby 2.1+
+  if defined?(Process::CLOCK_MONOTONIC)
+    def time_monotonic
+      Process.clock_gettime(Process::CLOCK_MONOTONIC)
+    end
+  else
+    def time_monotonic
+      Time.now.to_f
+    end
+  end
+
+  def time_spent_in_ms
+    before = time_monotonic
+    begin
+      yield
+    ensure
+      after = time_monotonic
+      duration = (after - before) * 1000
+    end
+    duration
+  end
+end
--- a/modules/dhcp_common/server.rb
+++ b/modules/dhcp_common/server.rb
@@ -4,6 +4,7 @@ require "dhcp_common/record/lease"
 require "dhcp_common/record/reservation"
 require 'dhcp_common/record/deleted_reservation'
 require 'dhcp_common/pingable'
+require 'proxy/logging_resolv'
 
 module Proxy::DHCP
   class Server
@@ -14,6 +15,8 @@ module Proxy::DHCP
     include Proxy::DHCP::Pingable
     include Proxy::Log
     include Proxy::Validations
+    include ::Proxy::TimeUtils
+    include ::Proxy::Helpers
 
     def initialize(name, managed_subnets, subnet_service, free_ips_service = nil)
       @name = name


modules/dns_common/dns_common.rb
```diff
@@ -17,9 +19,28 @@ def initialize(server = nil, ttl = nil)
     end
 
     def resolver
+      logger.warn("Proxy::Dns::Record#resolver has been deprecated and will be removed, use get_dns_resource(s)")
```
ekohl: :-1: to removing this. It's a good place to abstract it out. smart_proxy_dns_route53 also uses this to ignore the DNS servers since Route53 doesn't really provide stable IPs to query. That's why it uses the system resolver. You'd break this functionality.

https://github.com/theforeman/smart_proxy_dns_route53/blob/86e6393d75192e991f2f2128c57ad94471d8180d/lib/smart_proxy_dns_route53/dns_route53_main.rb#L52-L54

modules/dns_common/dns_common.rb
```diff
@@ -17,9 +19,28 @@ def initialize(server = nil, ttl = nil)
...
       Resolv::DNS.new(:nameserver => @server)
     end
 
+    def get_dns_resource(name, typeclass)
+      result = nil
+      duration = time_spent_in_ms do
+        result = Resolv::DNS.new(:nameserver => @server).getresource(name, typeclass)
```
ekohl:
```suggestion
        result = resolver.getresource(name, typeclass)
```

modules/dns_common/dns_common.rb
```diff
@@ -17,9 +19,28 @@ def initialize(server = nil, ttl = nil)
...
+      end
+      logger.info "Resolved #{name} via system resolver in #{duration.round(2)} ms"
```
ekohl: This is debug info
```suggestion
      logger.debug "Resolved #{name} via system resolver in #{duration.round(2)} ms"
```

modules/dns_common/dns_common.rb
```diff
@@ -17,9 +19,28 @@ def initialize(server = nil, ttl = nil)
...
+      result
+    end
+
+    def get_dns_resources(name, typeclass)
+      result = nil
+      duration = time_spent_in_ms do
+        result = Resolv::DNS.new(:nameserver => @server).getresources(name, typeclass)
```
ekohl:
```suggestion
        result = resolver.getresources(name, typeclass)
```

modules/dns_common/dns_common.rb
```diff
@@ -17,9 +19,28 @@ def initialize(server = nil, ttl = nil)
...
+      end
+      logger.info "Resolved #{name} via system resolver in #{duration.round(2)} ms"
```
ekohl:
```suggestion
      logger.debug "Resolved #{name} via system resolver in #{duration.round(2)} ms"
```

ekohl (changes requested):

modules/dns_common/dns_common.rb
lzap: This method is deprecated and files a warning into the logger.

modules/dns_common/dns_common.rb
lzap: This is intentional, I actually want to be very verbose here.

modules/dns_common/dns_common.rb
lzap: Ditto.

modules/dns_common/dns_common.rb
lzap: Ditto.

modules/dns_common/dns_common.rb
lzap: No, I am not breaking the compatibility, I am deprecating the functionality with a plan to remove (break) it. That's a difference.

One possible solution to avoid modification of plugins is to provide a decorator object that would keep the API. If you insist, I can do that, however I think that it's hacky solution. No other plugin uses that and I am willing to patch route53 until the method is deprecated.

lzap:
Thanks for review, however what I am doing here is all intentional. I've spent tremeneous time debugging this issue with two customers affected by this, it's not yet clear if this was the real cause for DNS requests to be served with 90 second delays, but this will make much easier.

I also intentionally want to inform users via an INFO message that there is a resolution being done via *system* resolver. I think this is worth an INFO message, every request already does several INFO messages and this is only for the DNS creation ones.

modules/dns_common/dns_common.rb
ekohl: IMHO this method still makes sense. I'd be ok with deprecating it and making it private if there's a good reason solution to route53, but duplicating the creation of a resolver object is something I'd like to avoid.

modules/dns_common/dns_common.rb
ekohl: It's actually also incorrect. It uses specific nameservers from the configuration so it's not using the system resolver.

I'm against such verbose timing logging. Didn't we introduce a timing measuring API to store this kind of data in an easily parsable way?

ekohl:
> there is a resolution being done via _system_ resolver

But it's not the system resolver. The idea is that you can specify the IP of your authoritative nameserver and it would query that. If you're running your host on 192.168.100.2 and an infoblox server on 192.168.100.3 then you should specify that nameserver IP. If you're using the system resolver, it's a misconfiguration.

This still isn't perfect and for API-based systems it should really query those (PowerDNS, Route53) to detect whether there are conflicts and the whole system needs an overhaul, but that's out of scope now.

modules/dns_common/dns_common.rb
lzap: But what do you recommend then? I believe that defining a better contract (two methods) is so much better than just a method which creates new instance.

ekohl: Use all 3 methods. The new methods should be recommended. I'd also copy the caching of the instance that route53 does so it becomes a method that lazy loads the resolver instance.

modules/dns_common/dns_common.rb
lzap: You are right that this does not use system resolver, there is a fallback to localhost tho. I will correct the string, also I am willing to change this to debug if you think it's really not something that should appear every DNS creation.

I am not aware of any kind of timing API, if you mean telemetry API that has been added only to Foreman core. Of course, once we add it here (and it's on my long-term todo) this will be included.

lzap:
Yeah, I figured out that this is not system resolver, gonna correct that and this also explains it. Also the customer confirmed that this is the root cause of their issue.

Looks like route 53 does not even use the helper, so I am actually changing my mind and I will provide a decorator around Resolv Ruby object with measuring and a helper that plugins can optionally use.

https://github.com/theforeman/smart_proxy_dns_route53/blob/master/lib/smart_proxy_dns_route53/dns_route53_main.rb

lzap:
I am looking in our installer codebase and I believe that the "server" DNS module setting is not parametrized, so effectively this is localhost only. Can you confirm?

lzap:
How about this, completely rewritten from scratch. There is a Resolv decorator available now, nothing is changing for plugins and it only creates one debug line, however if the duration is longer it will create a warning with more details for easier troubleshooting. Unfortuateny there is no way to pull nameservers out of the Resolv object other than inspect or reading private attribute.

This is how log would look like when there was a slow query:

```
2019-06-13T10:50:00 49db72b5 [W] Slow DNS query (getresources) in 12030.66 ms: ["test228.ibx.lan", Resolv::DNS::Resource::IN::A]
2019-06-13T10:50:00 49db72b5 [D] Resolver used: #<Resolv::DNS:0x000055555759fc20 @mutex=#<Thread::Mutex:0x000055555759fbd0>, @config=#<Resolv::DNS::Config:0x000055555759fb30 @mutex=#<Thread::Mutex:0x000055555759fae0>, @config_info={:nameserver=>"192.168.201.2"}, @initialized=true, @timeouts=[10], @nameserver_port=[["192.168.201.2", 53]], @search=[[#<Resolv::DNS::Label::Str home>, #<Resolv::DNS::Label::Str lan>]], @ndots=1>, @initialized=true>
```

modules/dns_common/dns_common.rb
```diff
@@ -8,16 +9,20 @@ class Collision < RuntimeError; end
...
 
     attr_reader :server, :ttl
 
-    def initialize(server = nil, ttl = nil)
+    def initialize(server = nil, ttl = nil, resolv_timeout = nil)
       @server = server || "localhost"
-      @ttl    = ttl || "86400"
+      @ttl = ttl || "86400"
+      @resolv_timeout = resolv_timeout || 10
     end
```
lzap, line 20: Optionally timeout can be now specified, Ruby default is quite tricky to understand.

modules/dns_common/dns_common.rb
```diff
@@ -8,16 +9,20 @@ class Collision < RuntimeError; end
...
 
-    def resolver
-      Resolv::DNS.new(:nameserver => @server)
+    def resolver(override_nameserver = @server)
+      resolv = Resolv::DNS.new(:nameserver => override_nameserver)
+      resolv.timeouts = @resolv_timeout
+      ResolvLogDecorator.new(resolv)
     end
```
lzap, line 24: Also nameserver can be overriden by plugins now, I could add this to the initializer above but I think here it's also fine since this is actually "constructor" of the resolver object.

lzap:
Fixed rubocop and tests, this is now ready for re-review.

ekohl (commented):
I still think this is just making it hard to follow. The core problem is that this class is overloaded and has too many concerns. That was already the case before this change.

I do like the wrapper that extracts the resolving, but maybe it should go further. I would actually like to introduce a `ConflictChecker` interface. Using plain DNS is one implementation, but networked services like Route53 and Infoblox should be able to introduce their own that may not rely on DNS queries at all.

lzap:
Honestly, I can't commit for refactoring of this class with breaking changes, however I like the idea of plugins being easily able to provide own checking interface. How about changing the code in more simple way introducing method `do_get` alongside `do_create` and `do_remove` with the default implementation using Ruby DNS resolver. This would give ability to override this and also would not break existing plugins at all.

The class is not that bad, it's quite long because of all the record types, but if you strip it to the bare minimum code, it's basically CREATE and DELETE with lots of deprecated stuff. Maybe if we remove deprecated stuff that would help to clean it a bit too.

ekohl:
So I started with https://github.com/theforeman/smart-proxy/pull/659.

And you're right: my goal was to just have a very simple interface which passes just the right DNS data and let implementations handle the specifics. https://github.com/theforeman/smart-proxy/pull/507 as an early attempt, but I soon realized we need a V2 API that doesn't have the oddness that comes from the REST API. The creation of A and PTR records is very confusing because it deals with IPs rather than just DNS names and their values. I haven't picked this up because I can't really commit to it either (though the whole capabilities framework was the first step in optionally providing it).

I'm going to need a bit to think about how to do this cleanly but I think there's a cleaner way. Will try to get back to you tomorrow.

lzap:
Here is a rough idea (untested):

```diff
diff --git a/modules/dns_common/dns_common.rb b/modules/dns_common/dns_common.rb
index 31c1c69..e9318e5 100644
--- a/modules/dns_common/dns_common.rb
+++ b/modules/dns_common/dns_common.rb
@@ -19,12 +19,37 @@ module Proxy::Dns
       @resolv_timeout = resolv_timeout || 10
     end
 
+    # Methods which MUST be implemented by providers
+    def do_create(name, value, type)
+      raise(Proxy::Dns::Error, "Creation of #{type} not implemented")
+    end
+
+    def do_remove(name, type)
+      raise(Proxy::Dns::Error, "Deletion of #{type} not implemented")
+    end
+
+    # Methods which CAN be implemented by providers
+    def do_get_resources(name, type)
+      # must return an array
+      do_get_via_ruby_resolver(name, type)
+    end
+
+    # The rest is common code for all providers
     def resolver(override_nameserver = @server)
       resolv = Resolv::DNS.new(:nameserver => override_nameserver)
       resolv.timeouts = @resolv_timeout
       ResolvLogDecorator.new(resolv)
     end
 
+    def do_get_via_ruby_resolver(name, type)
+      resolver.getresources(name, type)
+    end
+
+    def do_get_resource(name, type)
+      do_get_resources.each(name, type) {|resource| return resource}
+      raise Resolv::ResolvError.new("DNS result has no information for #{name}")
+    end
+
     def create_srv_record(service, value)
       do_create(service, value, 'SRV')
     end
@@ -73,10 +98,6 @@ module Proxy::Dns
       end
     end
 
-    def do_create(name, value, type)
-      raise(Proxy::Dns::Error, "Creation of #{type} not implemented")
-    end
-
     def remove_a_record(fqdn)
       do_remove(fqdn, "A")
     end
@@ -97,10 +118,6 @@ module Proxy::Dns
       do_remove(name, "PTR")
     end
 
-    def do_remove(name, type)
-      raise(Proxy::Dns::Error, "Deletion of #{type} not implemented")
-    end
-
     def dns_find(key)
       logger.warn(%q{Proxy::Dns::Record#dns_find has been deprecated and will be removed in future versions of Smart-Proxy.
                       Please use ::Proxy::Dns::Record#get_name or ::Proxy::Dns::Record#get_address instead.})
@@ -138,13 +155,13 @@ module Proxy::Dns
     end
 
     def get_resource_as_string(value, resource_type, attr)
-      resolver.getresource(value, resource_type).send(attr).to_s
+      do_get_resource(value, resource_type).send(attr).to_s
     rescue Resolv::ResolvError
       false
     end
 
     def get_resource_as_string!(value, resource_type, attr)
-      resolver.getresource(value, resource_type).send(attr).to_s
+      do_get_resource(value, resource_type).send(attr).to_s
     rescue Resolv::ResolvError
       raise Proxy::Dns::NotFound.new("Cannot find DNS entry for #{value}")
     end
@@ -195,14 +212,14 @@ module Proxy::Dns
         raise Proxy::Dns::Error.new("Not an IP Address: '#{ip}'")
       end
 
-      resources = resolver.getresources(fqdn, type)
+      resources = do_get_resources(fqdn, type)
       return -1 if resources.empty?
       return 0 if resources.any? {|r| IPAddr.new(r.address.to_s) == ip_addr }
       1
     end
 
     def record_conflicts_name(fqdn, type, content)
-      resources = resolver.getresources(fqdn, type)
+      resources = do_get_resources(fqdn, type)
       return -1 if resources.empty?
       return 0 if resources.any? {|r| r.name.to_s.casecmp(content) == 0 }
       1
```

ekohl:
Have a look at my deprecation PR. Most of these functions are unused.

Perhaps we can have a more generic `record_conflicts(name, type, content)`. Then the infoblox plugin can just query the HTTP API to see if there's a conflict.

Alternatively, if the creation can be done in a single REST call it should override the `create_*_record` methods to perform it as part of that.

lzap:
I am completely fine with a separate commit deprecating stuff, but I will likely need to backport this change if we agree on an interface/method for Infoblox to provide conflict resulution on its own so I'd appreciate if this commit could go first.

lzap:
> Perhaps we can have a more generic record_conflicts(name, type, content). Then the infoblox plugin can just query the HTTP API to see if there's a conflict.

Isn't better to keep conflict handling code in the common module and define the contract as minimal possible? Just small operations, CRUD basically.

lzap:
I see that dropping of some unused code has started already, so now I don't mind testing and merging your other deprecation patch.

lzap:
@ekohl I have a suggestion, how about I turn this PR into generic resolver API for all core and plugins. On every place where we do `Resolv.new` we'd use this new decorator which provides additional logging. Then we can also work on improving the DNS API towards better support for DNS conflict detection.

ekohl:
That was roughly what I was thinking about yes.

lzap:
Rebased and pushed a WIP version, one question let me find the code.

lib/proxy/helpers.rb
```diff
@@ -64,13 +64,24 @@ def parse_json_body
...
   # reverse lookup an IP address while verifying it via forward resolv
   def remote_fqdn(forward_verify=true)
     ip = request.env['REMOTE_ADDR']
     log_halt 403, 'could not get remote address from environment' if ip.empty?
 
     begin
-      dns = Resolv.new
+      # TODO - shall we actually make this dns_resolv?
+      dns = resolv
       fqdn = dns.getname(ip)
```
lzap, line 88: Here in this remote_fqdn method I realized that we don't actually do DNS resolve but generic resolve, this will use any OS provider (kerberos, NSS, DNS). Do we actually want to keep this behavior or shell I change this to strictly DNS only?

modules/dns_common/dns_common.rb
lzap: I still want to keep this, I might need to do a hotfix/quickfix for affected customer because this whole DNS refactoring and legacy code removal might block me from doing full Infoblox conflict query rewrite backport.

lib/proxy/helpers.rb
lzap: Ping @ekohl

ekohl: Sorry, I missed this. I'm leaning to the current resolv. I think the value of using NSS is actually pretty big and something we should be preferring unless we really need DNS.

lzap:
Then I amended and removed the note in the code.

lzap:
I have just found that the reverse lookup in bootServer method does not work properly, because we throw away exception it was actually not working. Gonna fix this and add tests.

modules/dhcp_common/isc/omapi_provider.rb
```diff
@@ -195,22 +195,25 @@ def poap_options_statements(options)
...
           ns = "\\\"#{server}\\\""
         end
       end
       "next-server = #{ns};"
     end
 
-    def ip2hex ip
-      ip.split(".").map{|i| "%02x" % i }.join(":")
+    def ip2hex(ip)
+      ip.to_s.split(".").map{|i| "%02x" % i }.join(":")
```
lzap: This was actually not working, because we swallowed the exception we haven't noticed. Resolver returns IPv4 object which has no `split` method. Probably the reason why we added reverse search into Foreman core?! Not sure. It now has tests and this PR also adds nice logging around this so it's very obvious what is going on.

lzap:
Rebased, fixed issue with reverse DNS search and added couple of tests around it.

modules/dhcp_common/isc/omapi_provider.rb
ekohl: Good find. Almost feels like it deserves a separate Redmine issue.

lib/proxy/logging_resolv.rb
```diff
@@ -0,0 +1,63 @@
...
+    end
+
+    private
+
+    def call_with_timing(method, *args)
+      result = nil
+      duration = time_spent_in_ms do
+        result = @resolv.send(method, *args)
+      end
+      if duration < 9000
```
ekohl: Small nit: I like constants for this. Perhaps a `SLOW_DNS_QUERY_MS` definition? Note I'm not asking for a config value, just a constant in the class.

modules/dhcp_common/isc/omapi_provider.rb
```diff
@@ -195,22 +195,25 @@ def poap_options_statements(options)
...
-        ns = ip2hex validate_ip(server)
+        ns = ip2hex(validate_ip(server))
       rescue
         begin
-          ns = ip2hex Resolv.new.getaddress(server)
-        rescue
-          logger.warn "Failed to resolve IP address for #{server}"
+          logger.info "Next-server option not IPv4, trying to resolve PTR record for #{server}"
+          ns = ip2hex(dns_resolv.getaddress(server))
+        rescue Exception => e
```
ekohl: Is there a `ResolvException` or similar we can catch just as a generic resolution failed? That could use `logger.warn` and this could log at the exception level.

ekohl (commented):
Overall it's looking good. I do feel like the bootserver fixes could use a separate Redmine issue for visibility in the changelog.

lib/proxy/logging_resolv.rb
lzap: Sure, it's probably an overkill but with all the issues, cases and confused users around DNS queries I want to be as clear as possible when this happens.

modules/dhcp_common/isc/omapi_provider.rb
lzap: Yes it's `Resolv::ResolvError`.

lzap:
Rebased.

modules/dhcp_common/isc/omapi_provider.rb
```diff
@@ -195,22 +195,25 @@ def poap_options_statements(options)
...
+    def bootServer(server)
       begin
-        ns = ip2hex validate_ip(server)
+        ns = ip2hex(validate_ip(server))
       rescue
         begin
-          ns = ip2hex Resolv.new.getaddress(server)
-        rescue
-          logger.warn "Failed to resolve IP address for #{server}"
+          logger.info "Next-server option not IPv4, trying to resolve PTR record for #{server}"
```
ekohl: PTR is reverse DNS but here it's forward DNS.

lzap:
Force-pushed, fixed the logging message.

ekohl (changes requested):
Please extract the omapi fixes to a separate PR. It's incomplete (you missed `ztp_options_statements` which will fail if `nextServer` is a DNS name) and holding the logging PR.

lzap:
Can you give it another review? I removed the omapi bits and moved them to: https://github.com/theforeman/smart-proxy/pull/682

ekohl (approved):
Thanks!

PR #658, opened 2019-06-12
Fixes #27036 - improve logging of error backtraces

lzap:
Looks like it's not visible at all.

### DIFF
--- a/lib/proxy/log_buffer/decorator.rb
+++ b/lib/proxy/log_buffer/decorator.rb
@@ -17,7 +17,7 @@ module Proxy::LogBuffer
       self.roll_log = false
     end
 
-    def add(severity, message = nil, progname = nil, backtrace = nil)
+    def add(severity, message = nil, progname = nil, exception_or_backtrace = nil)
       severity ||= UNKNOWN
       if message.nil?
         if block_given?
@@ -40,16 +40,21 @@ module Proxy::LogBuffer
         @logger.add(severity, message)
         # add add to the buffer
         if severity >= @logger.level
-          # we accept backtrace, exception and simple string
-          backtrace = backtrace.is_a?(Exception) ? backtrace.backtrace : backtrace
-          backtrace = backtrace.respond_to?(:join) ? backtrace.join("\n") : backtrace
+          # accepts backtrace, exception and simple string for historical reasons
+          backtrace = if exception_or_backtrace.is_a?(Exception) && !exception_or_backtrace.backtrace.nil?
+                        exception_or_backtrace.message + ': ' + exception_or_backtrace.backtrace.join("\n")
+                      elsif backtrace.respond_to?(:join)
+                        exception_or_backtrace.backtrace.join("\n")
+                      else
+                        exception_or_backtrace
+                      end
           rec = Proxy::LogBuffer::LogRecord.new(nil, severity, message.to_s.encode('UTF-8', invalid: :replace, undef: :replace, replace: '?'), backtrace, request_id)
           @buffer.push(rec)
         end
       end
       info("Logging file reopened via USR1 signal") if reopened
       # exceptions are also sent to structured log if available
-      self.exception("Error details", backtrace) if backtrace&.is_a?(Exception)
+      self.exception("Error details for #{message}", exception_or_backtrace) if exception_or_backtrace&.is_a?(Exception)
     end
 
     def trace?
@@ -111,7 +116,7 @@ module Proxy::LogBuffer
       extra_fields[:foreman_code] = exception.code if exception.respond_to?(:code)
       with_fields(extra_fields) do
         public_send(level) do
-          ([context_message, "#{exception.class}: #{exception.message}"] + backtrace).join("\n")
+          (["#{context_message}: <#{exception.class}>: #{exception.message}"] + backtrace).join("\n")
         end
       end
     end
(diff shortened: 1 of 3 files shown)


modules/dhcp_libvirt/subnet_service_initializer.rb
```diff
@@ -37,7 +37,7 @@ def parse_config_for_subnets
       raise Proxy::DHCP::Error("Only one subnet is supported") if ret_val.size > 1
       ret_val
     rescue Exception => e
-      msg = "Unable to parse subnets XML"
+      msg = "Unable to parse subnets XML: #{e}"
       logger.error msg, e
```
tbrisker: Isn't the issue than actually that the logger.error ignores the second parameter and other places are also affected?

lzap: Yeah I was thinking the same this morning.

lzap:
Ok improved it, actually regression made in: https://github.com/theforeman/smart-proxy/commit/8652a7e5c63009cae026f6350178583dcadc5f34

```
2019-06-14T08:56:28 9302404b [E] TEST
2019-06-14T08:56:28 9302404b [W] Error details for TEST: <StandardError>: Test
```

It's slightly more readable now.

After IRC discussion with @ekohl we agreed that adding support for `try` method from Rails will not hurt, it clearly makes code easier to read. After we upgrade to 2.3+ we can drop this.

lzap:
Rebased, in case you can't boot smart-proxy with "daemon" dependency error - rebase. It was an issue we've fixed this morning.

lib/try-patch.rb
```diff
@@ -0,0 +1,50 @@
+#
```
ekohl: Given there's only one usage of this, I'd prefer to hold off on this. If you do have a larger block of code that uses it, then I'd be happy to include it.

lzap:
Removed, rebased.

lib/proxy/log_buffer/decorator.rb
```diff
@@ -38,6 +38,14 @@ def add(severity, message = nil, progname = nil, backtrace = nil)
           message = progname
         end
       end
+      # accepts backtrace, exception and simple string for historical reasons
+      backtrace = if exception_or_backtrace.is_a?(Exception) && !exception_or_backtrace.backtrace.nil?
```
ekohl: I don't see why you're moving this out here. It's only used on line 57 so nesting it inside the if there should be good enough?

lzap: Right, this was done for no particular reason. Rebased.

ekohl (commented):
Could you add a test for `add`? This is complicated (for historical reasons) and I'd feel better if it had tests.

lzap:
Rebased, added tests.

lzap:
Fixed rubocop issue, ready for rereview.

modules/dhcp_libvirt/subnet_service_initializer.rb
```diff
@@ -37,7 +37,7 @@ def parse_config_for_subnets
       raise Proxy::DHCP::Error("Only one subnet is supported") if ret_val.size > 1
       ret_val
     rescue Exception => e
-      msg = "Unable to parse subnets XML"
+      msg = "Unable to parse subnets XML: #{e}"
```
ekohl, line 40: Doesn't the `, e` to `logger.error` already add this information?

lzap: Yes but apparently I want the error message to be returned to the client. Our error reporting is sometimes very weak.

lib/proxy/log_buffer/decorator.rb
```diff
@@ -46,15 +46,20 @@ def add(severity, message = nil, progname = nil, backtrace = nil)
         @logger.add(severity, message)
         # add add to the buffer
         if severity >= @logger.level
-          # we accept backtrace, exception and simple string
-          backtrace = backtrace.is_a?(Exception) ? backtrace.backtrace : backtrace
-          backtrace = backtrace.respond_to?(:join) ? backtrace.join("\n") : backtrace
+          # accepts backtrace, exception and simple string for historical reasons
+          backtrace = if exception_or_backtrace.is_a?(Exception) && !exception_or_backtrace.backtrace.nil?
+                        exception_or_backtrace.message + ': ' + exception_or_backtrace.backtrace.join("\n").to_s
```
tbrisker: not crucial but the `.to_s` here is redundent

tbrisker (approved):
Any further comments @ekohl? lgtm except for a super minor comment

lzap:
Rebased.

modules/dhcp_libvirt/subnet_service_initializer.rb
ekohl: Would `logger.exception` be useful here (and later in the file)?

ekohl (approved):
Would appreciate to know how the answer to my question but overall :+1: and can be merged regardless of the answer.

lzap:
I can't comment in the file, so let me put it here. I can add that, this code is probably pre-exception helper. Rebased.

modules/dhcp_libvirt/subnet_service_initializer.rb
ekohl: Is the specific exception message still needed when using `logger.exception`? I think that already formats the message to include the specific exception message.

tbrisker:
please rebase @lzap

lzap:
Rebased.

tbrisker:
Thanks @lzap !

PR #660, opened 2019-06-14
Refs #26979 - Remove left over require

ekohl:
bf476d0ad900e36f25865592538942a4b91610ad removed lib/daemon.rb since it only contained Ruby 1.x compatibility but left the require causing it to fail to start up.


ekohl:
Added 2 more commits that add more cleanups. This time I ran the proxy manually to ensure it still booted.

PR #661, opened 2019-06-20
Fixes #27088 - Handle capabilities for uninitialized plugins

lzap:
For some reason, this test fails locally for me but not on Jenkins:

```
Error: test_features(BmcApiFeaturesTest): JSON::ParserError: 809: unexpected token at 'private method `select' called for nil:NilClass'
/home/lzap/.rbenv/versions/2.7.2/lib/ruby/gems/2.7.0/gems/json-2.5.1/lib/json/common.rb:216:in `parse'
/home/lzap/.rbenv/versions/2.7.2/lib/ruby/gems/2.7.0/gems/json-2.5.1/lib/json/common.rb:216:in `parse'
/home/lzap/work/smart-proxy/test/bmc/integration_test.rb:21:in `test_features'
     18:
     19:     get '/features'
     20:
  => 21:     response = JSON.parse(last_response.body)
     22:
     23:     mod = response['bmc']
     24:     refute_nil(mod)
```

### DIFF
--- a/modules/root/root_v2_api.rb
+++ b/modules/root/root_v2_api.rb
@@ -25,6 +25,7 @@ class Proxy::RootV2Api < Sinatra::Base
   end
 
   def process_capabilities(state, capabilities)
+    return [] if state == :uninitialized
     capabilities = capabilities.select { |cap| !cap.is_a?(Proc) || state == :running }
     capabilities = capabilities.map do |capability|
       capability.is_a?(Proc) ? capability.call : capability
--- a/test/bmc/integration_test.rb
+++ b/test/bmc/integration_test.rb
@@ -1,7 +1,7 @@
 require 'test_helper'
 require 'json'
 require 'root/root_v2_api'
-require 'bmc/bmc'
+require 'bmc/bmc_plugin'
 require 'bmc/ipmi'
 
 class BmcApiFeaturesTest < Test::Unit::TestCase


lzap:
Rebased this is still a thing for me locally, unsure what is different on my environment but the patch is harmless I guess. @ekohl

ekohl:
My issue with this is that it should never happen so we should understand why this code path is being triggered. The intended API is that it's always an array and code should be able to rely on that.

https://github.com/theforeman/smart-proxy/blob/47f859148ee63e09bcaadd97568e5eb677025c20/lib/proxy/pluggable.rb#L177-L179 should ensure it's always an array for anything pluggable. That's both `Plugin` and `Provider`. However, https://github.com/theforeman/smart-proxy/blob/47f859148ee63e09bcaadd97568e5eb677025c20/test/plugins/plugin_initializer_test.rb#L52-L53 explicitly expected it with nil. Perhaps this is something we should be doing in the constructor instead?

lzap:
Sorry I fail to understand both the issue and also the mentioned test. Why the heck `test_4` plugin is expected to have capabilities set to `nil`? They are all the same. Can you probably suggest the change?

lzap:
Ping @ekohl any ideas?

lzap:
@ekohl since you implemented this code, can you please investigate this one? It is driving me crazy, locally this fails on my PC and I am out of ideas why.

ekohl:
Apologies for the late reply. I can't reproduce a failure. If you run just on that single file, does it still fail for you?
```
ekohl@wisse ~/dev/smart-proxy (develop) $ be rake test TEST=test/plugins/plugin_group_test.rb
/usr/bin/ruby -w -I"lib:.:lib:test" -W1 -I"/home/ekohl/.gem/ruby/gems/rake-13.0.1/lib" "/home/ekohl/.gem/ruby/gems/rake-13.0.1/lib/rake/rake_test_loader.rb" "test/plugins/plugin_group_test.rb" 
Loaded suite /home/ekohl/.gem/ruby/gems/rake-13.0.1/lib/rake/rake_test_loader
Started
..................
Finished in 0.01022557 seconds.
----------------------------------------------------------------------------------------------------------------------------------------
18 tests, 44 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
----------------------------------------------------------------------------------------------------------------------------------------
1760.29 tests/s, 4302.94 assertions/s
```

lzap:
Yes, that is passing actually.

ekohl:
Can you still reproduce this? It can be some loading order dependent issue. IIRC the rake task generates some seed for the test order. Can you share that from a failed run?

lzap:
Yep, every time. This is how I start tests: `bundle exec rake test`

mmoll:
I start the tests the same way and it works on my machine... :face_with_head_bandage:

lzap:
Do you guys have `freeipmi` and `ipmitool` installed? Our integration tests rely on that, otherwise they are skipped I think. Here is the stack:

```

 "/home/lzap/work/smart-proxy/modules/root/root_v2_api.rb:28:in `process_capabilities'",
 "/home/lzap/work/smart-proxy/modules/root/root_v2_api.rb:18:in `block (2 levels) in <class:RootV2Api>'",
 "/home/lzap/work/smart-proxy/modules/root/root_v2_api.rb:16:in `each'",
 "/home/lzap/work/smart-proxy/modules/root/root_v2_api.rb:16:in `each_with_object'",
 "/home/lzap/work/smart-proxy/modules/root/root_v2_api.rb:16:in `block in <class:RootV2Api>'",
 "/home/lzap/work/smart-proxy/test/bmc/integration_test.rb:17:in `test_features'",
```

mmoll:
> Do you guys have freeipmi and ipmitool installed?

Yes.

lzap:
Ok found it, there are plugins which are created by other test:

```
 {:name=>:plugin_6,
  :version=>"1.0",
  :class=>PluginInitializerTest::TestPlugin6,
  :state=>:failed,
  :di_container=>#<Proxy::DependencyInjection::Container:0x0000555557e527a0>,
  :http_enabled=>false,
  :https_enabled=>false,
  :capabilities=>
   [#<Proc:0x00005555576fcc08@/home/lzap/work/smart-proxy/test/plugins/plugin_initializer_test.rb:31 (lambda)>,
    #<Proc:0x00005555576fcd48@/home/lzap/work/smart-proxy/test/plugins/plugin_initializer_test.rb:30>,
    "foo"],
  :settings=>{:foo=>nil}},
```

In plugin_initializer_test.rb:

```
       {:name => :plugin_4, :version => '1.0', :class => TestPlugin4, :state => :running, :http_enabled => nil, :https_enabled => nil,
          :settings => nil, :capabilities => nil},
```

So we can either merge my solution or remove this test if you feel like testing it with `nil` is not good. LIkely tests are executed in a different order than in your case, or I don't know why only I see this.

lzap:
What's weird is when I print `enabled_plugins` in IRB I don't actually see offending plugin_4. But I I do this, then `nil` values are there. I hate our initialization code, dependency injection and all the lambda magic. This all could be achievable by regular object composition.

```ruby
[5] pry(#<Proxy::RootV2Api>)> enabled_plugins.each_with_object({}) do |plugin, hash| puts plugin[:capabilities].inspect end
[]
[]
[]
[]
[]
[]
[]
[]
[]
[]
[]
[]
[]
[]
[]
[]
[#<Proc:0x00005555576fcc08@/home/lzap/work/smart-proxy/test/plugins/plugin_initializer_test.rb:31 (lambda)>, #<Proc:0x00005555576fcd48@/home/lzap/work/smart-proxy/test/plugins/plugin_initializer_test.rb:30>, "foo"]
[]
[]
[]
[]
[]
[]
[]
[]
[]
[]
[]
nil
nil
nil
nil
nil
nil
=> {}
```

lzap:
I think I got that. During testing, we create bunch of plugins which are in `:uninitialized` state or other non-running state. These plugins can have capabilities set to `nil`. These only appear in tests.

Now, I have no idea WHY I do see them and you don't. Ruby, hello!

```
> ::Proxy::Plugins.instance.loaded.each_with_object({}) do |plugin, hash| puts "#{plugin[:name]} #{plugin[:state]} #{plugin[:capabilities].inspect}" end
bmc running []
dhcp running ["dhcp_filename_ipv4", "dhcp_filename_hostname"]
dhcp_isc running nil
dhcp_libvirt running nil
dhcp_native_ms running nil
dns running []
dns_dnscmd running nil
dns_libvirt running nil
dns_nsupdate running nil
dns_nsupdate_gss uninitialized nil
facts running []
httpboot running []
logs running []
test4 uninitialized nil
test5 disabled []
test6 disabled []
test7 disabled []
test8 disabled []
test9 disabled []
test10 disabled []
test11 running []
test12 running nil
test13 running []
test14 disabled []
plugin_4 uninitialized nil
plugin_5 disabled []
plugin_6 running [#<Proc:0x0000555556276798@/home/lzap/work/smart-proxy/test/plugins/plugin_initializer_test.rb:31 (lambda)>, #<Proc:0x00005555562767e8@/home/lzap/work/smart-proxy/test/plugins/plugin_initializer_test.rb:30>, "foo"]
test_bundler_group disabled []
test_default_bundler_group disabled []
plugin_1 running []
plugin_2 failed []
plugin_3 running []
puppet_proxy_customrun uninitialized nil
puppet disabled []
puppet_proxy_puppet_api uninitialized nil
puppet_proxy_mcollective uninitialized nil
puppet_proxy_salt uninitialized nil
puppet_proxy_ssh uninitialized nil
puppetca failed []
puppetca_hostname_whitelisting uninitialized nil
puppetca_http_api uninitialized nil
puppetca_puppet_cert uninitialized nil
puppetca_token_whitelisting uninitialized nil
realm disabled []
realm_freeipa uninitialized nil
templates failed []
foreman_proxy disabled []
test2 disabled []
test3 disabled []
tftp running []
discovery running []
discovery_image running []
dhcp_infoblox uninitialized nil
dns_infoblox uninitialized nil
dynflow failed []
ssh failed []
ansible failed []
externalipam running []
```

lzap:
Looks like plugin *providers* do not have capabilities, thus they are set to `nil`. Problem is, current implementation does not differentiate between them - they are being kept in a single array. Information about ancestor is thrown away after they are initialized.

Do you have any providers enabled in your development smart-proxy? As you can see above, I have many (e.g. infoblox, GSS).

lzap:
I have found the offender, enable "smart-proxy-discovery" plugin. Looks like these are not the tests, jesus christ this spagetti code is killing me.

lzap:
For some reason, during this test which is failing some plugins (including my smart-proxy-discovery) are uninitialized. I see other tests do this: `Proxy::PluginInitializer.new(Proxy::Plugins.instance).initialize_plugins` however when I do this it breaks bunch of other unrelated tests. I am running out of ideas, @jlsherrill actually contributed this code maybe he has more insights into this.

modules/root/root_v2_api.rb
```diff
@@ -25,6 +25,7 @@ class Proxy::RootV2Api < Sinatra::Base
   end
 
   def process_capabilities(state, capabilities)
+    return [] if capabilities.nil?
```
ekohl: Thanks to @adamruzicka and smart_proxy_salt we figured out how this can happen: if the plugin isn't properly initialized, `capabilities` can be nil. That's why you can see it in tests.

Now I wonder if this should be looking at `state` as well. If state is `:running` and `capabilities` is nil then it is a bug and we shouldn't accept that.

ekohl (changes requested):
I'd like this to take the plugin state in consideration.

lzap:
How about this, @ekohl ?

lzap:
I have found out today an interesting thing:

When I enable https://github.com/theforeman/smart_proxy_shellhooks I get the error:

```
$ grep shellhooks bundler.d/Gemfile.local.rb
gem 'smart_proxy_shellhooks', :path => "../smart_proxy_shellhooks/"
```

Can you try this to confirm?

lzap:
Correction: Any enabled plugin triggers that on my system.

lzap:
I can confirm, if you enable any plugin this error will showup. Shellhooks is one to test with it is super small.

ekohl:
I still think that it's loading a plugin and somehow fails to initialize. I wonder what the plugin state is when it triggers. Could you find out?

lzap:
It is in my previous comment, they are in state running, some are uninitialized. The reason why its BMC what is failing is IMHO it comes first in the alphabet.

ekohl (commented):
I'm OK with this change, but I think the description should be different. Instead, I think it should be something like "Handle capabilities for uninitialized plugins". Do you agree?

lzap:
Rebased.

ekohl (approved):

PR #665, opened 2019-07-02
Fixes #27218: revamps regex for finding realm name

phess:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1726380 -- When parsing /etc/ipa/default.conf to find the realm name, instead of matching any line containing 'realm' match only lines starting with 'realm' and containing an = sign followed by optional spaces and mandatory 1+ non-spaces.

### DIFF
--- a/test/realm/realm.conf
+++ b/test/realm/realm.conf
@@ -2,7 +2,7 @@
 basedn = dc=demo1,dc=freeipa,dc=org
 realm = DEMO1.FREEIPA.ORG
 domain = demo1.freeipa.org
-server = ipa.demo1.freeipa.org
+server = ipa-realm.demo1.freeipa.org
 host = lucid-nonsense
 xmlrpc_uri = https://ipa.demo1.freeipa.org/ipa/xml
 enable_ra = True
--- a/modules/realm_freeipa/ipa_config_parser.rb
+++ b/modules/realm_freeipa/ipa_config_parser.rb
@@ -40,11 +40,11 @@ module Proxy::FreeIPARealm
       parsed_uri, realm_name = nil
 
       io.readlines.each do |line|
-        if line =~ /xmlrpc_uri/
+        if line =~ /^\s*xmlrpc_uri\s*=\s*\S+/
           uri = line.split("=")[1].strip
           parsed_uri = URI.parse(uri)
           logger.debug "freeipa: uri is #{uri}"
-        elsif line =~ /realm/
+        elsif line =~ /^\s*realm\s*=\s*\S+/
           realm_name = line.split("=")[1].strip
           logger.debug "freeipa: realm #{realm_name}"
         end


ekohl:
ok to test

We do require Redmine issues. You can create a new one and add the BZ number there. There is some automation that will link the two.

adamruzicka:
Cloned the BZ to redmine[1], please change the commit message to begin with `Fixes #27218`

[1] - http://projects.theforeman.org/issues/27218

phess:
"please change the commit message to begin with Fixes #27218"
Done. Thanks.

ekohl:
Looks like you added another commit and a merge commit. You need to make sure it's just a single commit based on develop with the right message.

ekohl (approved):
It's too bad there's no built in ini parser in Ruby.

Could you have a look at some tests?

phess:
> It's too bad there's no built in ini parser in Ruby.

Yes it is. We definitely need one.

> 
> Could you have a look at some tests?

Me? I would gladly. The tests for ipa_config_parser.rb under this project's tests dir look fine to me in terms of ensuring a realm and an API URI are obtained from /etc/ipa/default.conf.

phess:
> > It's too bad there's no built in ini parser in Ruby.
> 
> Yes it is. We definitely need one.

Hmmm. https://github.com/TwP/inifile looks pretty good but the latest commit was in 2014.

ekohl:
> Me? I would gladly. The tests for ipa_config_parser.rb under this project's tests dir look fine to me in terms of ensuring a realm and an API URI are obtained from /etc/ipa/default.conf.

I was also thinking about the negative test. I assume this regex was tightened so it wouldn't match lines which just had some of these words. A test that would fail before the patch was applied.

phess:
> I was also thinking about the negative test. I assume this regex was tightened so it wouldn't match lines which just had some of these words. A test that would fail before the patch was applied.

Right on your assumption.
Ack on the test that would fail before the patch was applied. I'll be working on it and adding to this PR if this is what you're asking for.

test/realm/realm.conf
```diff
@@ -2,7 +2,7 @@
 basedn = dc=demo1,dc=freeipa,dc=org
 realm = DEMO1.FREEIPA.ORG
 domain = demo1.freeipa.org
-server = ipa.demo1.freeipa.org
+server = ipa-realm.demo1.freeipa.org
```
lzap, line 5: What's this change for?

Anyway I am good, please rebase and we can merge.

ekohl: Because previously this triggered a failure in parsing. It would assume this was the option `realm = ...` instead of `server = ...`.

ares:
there's update branch button that I clicked, if tests are green, I think this can be squash and merged.

ekohl:
It created a merge commit and it makes the PR processor very unhappy so I've done the rebase and force pushed which should allow merged. However, I do wonder why we require every commit to be exactly up to date in this project. We do it nowhere else.

ekohl (approved):

ekohl:
Thanks @phess!

PR #667, opened 2019-07-19
Fixes #27351 - don't use rolling appender by default

lzap:
It looks like TwP/logging gem's Rolling File implementation is not thread safe and it can be causing issues during reopens. Smart-proxy configures Rolling File appender and this can't be turned off. Default behavior is weekly rolling or by 100 MiB size, this patch changes it to use simple File appender and rely on logrotate which we configure by default anyway.

The implementation is based on dynflow where we do the very same approach when `file_rolling_size` is set to zero. So for consistency I'd prefer to take the same approach here. Also, installer change will be probably needed.

### DIFF
--- a/config/settings.yml.example
+++ b/config/settings.yml.example
@@ -64,8 +64,8 @@
 # Uncomment and modify if you want to change the log level
 # WARN, DEBUG, ERROR, FATAL, INFO, UNKNOWN
 #:log_level: INFO
-# The maximum size of a log file before it's rolled (in MiB)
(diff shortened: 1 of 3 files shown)


ekohl (approved):
:+1: for relying on logrotate and having just one rolling mechanism.

mmoll:
merged, díky @lzap!

lzap:
Doublechecked the puppet codebase and this is not yet parametrized, so no change needed.

PR #668, opened 2019-08-05
Fixes #27509 - logging reopen USR1 handler fixed

lzap:
After refactoring to "logging" gem the file reopening was not implemented properly.

### DIFF
--- a/lib/proxy/log_buffer/decorator.rb
+++ b/lib/proxy/log_buffer/decorator.rb
@@ -7,26 +7,15 @@ module Proxy::LogBuffer
       @@instance ||= new(::Proxy::LoggerFactory.logger, ::Proxy::LoggerFactory.log_file)
     end
 
-    attr_accessor :formatter
+    attr_accessor :formatter, :roll_log
+    alias_method :roll_log?, :roll_log
 
     def initialize(logger, log_file, buffer = Proxy::LogBuffer::Buffer.instance)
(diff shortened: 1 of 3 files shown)


lzap:
I have deleted the three tests, rolling is now fully managed by "logging" gem. It will only reopen relevant appenders (e.g. file appender) and not stdout/stderr.

lib/proxy/log_buffer/decorator.rb
```diff
@@ -53,6 +52,7 @@ def add(severity, message = nil, progname = nil, backtrace = nil)
           @buffer.push(rec)
         end
       end
+      debug("Reopened logging files") if reopened
```
ekohl: I would like this to show up as the very first line after reopening. Otherwise you'll end up with a log file that says `<message>` and then `Reopened logging files`. However, I'm aware this can cause (endless) recursing. Would it be better to use the raw `@buffer.push for this right after reopening?

Note I actually do like this log line and wonder if it should actually be an `info` message. As an admin it can be very useful to know when a log file was created.

lzap: This does work as expected actually, it appears as the very first log line and it does not end up in endless loop. Let me change that to INFO level then.

lzap:
Rebased, fixed the missed test.

ekohl (approved):

PR #670, opened 2019-08-08
Fixes #27550 - grub2 native config files

lzap:
See the RM issue for the reasoning. We already do this for PXELinux, this will vastly simplify how we load custom config files.


ekohl (approved):
Slightly surprising to me that `pxeconfig_file` returns an array of files, but that's not really something this PR should touch.

lzap:
Hehe yeah we use this already for PXELinux and iPXE too I think. It was refactored, `files` would be definitely better. Thanks for quick review!

PR #671, opened 2019-08-09
Fixes #27557 - support EUI-48, 64 and InfiniBand MACs

lzap:
Our DHCP API does not seem to verify MAC address form which is good, so it is up to the DHCP providers to error out in case they don't support them. Looks like ISC DHCP does work fine with all the three, so I added some extra tests and fixed our parser so it recognize them properly.

### DIFF
--- a/test/dhcp/conf_parser_test.rb
+++ b/test/dhcp/conf_parser_test.rb
@@ -356,6 +356,40 @@ EOMULTILINE_HOST
     ]]], Proxy::DHCP::CommonISC::ConfigurationParser.new.conf.parse!(MULTILINE_HOST)
   end
 
+  HOST_WITH_MAC64 =<<EOHOST_WITH_MAC64
+    host eui64 {
+      hardware ethernet 00:25:96:FF:FE:12:34:56;
+      fixed-address 192.168.1.1;
+      option routers 204.254.239.1;
+      filename "pxelinux.0";
+    }
+EOHOST_WITH_MAC64
+  def test_multiline_host_parser_with_mac64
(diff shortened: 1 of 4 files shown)


ekohl:
Failure looks relevant: `<Proxy::DHCP::CommonISC::ConfigurationParserTest#test_multiline_host_parser> was redefined`

lzap:
The reporter told me in the RM it works for them. Will fix the tests.

lzap:
Aaaand rebased.

test/dhcp/conf_parser_test.rb
```diff
@@ -356,6 +356,40 @@ def test_multiline_host_parser
     ]]], Proxy::DHCP::CommonISC::ConfigurationParser.new.conf.parse!(MULTILINE_HOST)
   end
 
+  HOST_WITH_MAC64 =<<EOHOST_WITH_MAC64
```
ekohl, line 359: If you use this version, can't you indent the end tag?
```suggestion
  HOST_WITH_MAC64 = <<-EOHOST_WITH_MAC64
```

ekohl (approved):
The inline comment is mostly a question to increase my understanding. I'm fine either way.

test/dhcp/conf_parser_test.rb
lzap: I can indent it regardless, whitespace must not break the parser. I don't care, honestly I'd rather avoid big refactoring because I would probably need to reindent the whole test.

ekohl: Good point that it's following the current convention in the file. :+1: for consistency within a file.

PR #676, opened 2019-08-19
Fixes #27654 - expose http(s) port setting

lzap:
This is needed for HTTP Boot functionality as Proxy is not running on the standard 80.

PR #677, opened 2019-08-23
Refs #27654 - port is nil for disabled endpoints

lzap:
This fixes the previous commit - when endpoint is disabled, port was not set to `nil` which was confusing design. This fixes it.

https://github.com/theforeman/smart-proxy/pull/676

### DIFF
--- a/modules/httpboot/httpboot_plugin.rb
+++ b/modules/httpboot/httpboot_plugin.rb
@@ -4,11 +4,16 @@ module Proxy::Httpboot
     https_rackup_path File.expand_path("http_config.ru", File.expand_path("../", __FILE__))
 
     plugin :httpboot, ::Proxy::VERSION
-    load_programmable_settings ::Proxy::Httpboot::PluginConfiguration
 
     default_settings :root_dir => '/var/lib/tftpboot'
 
     expose_setting :http_port
     expose_setting :https_port
+
+    after_activation do
+      plugin = ::Proxy::Plugins.instance.find{|p| p[:name] == :httpboot}
+      settings[:http_port] = plugin[:http_enabled] ? Proxy::SETTINGS.http_port : nil
+      settings[:https_port] = plugin[:https_enabled] ? Proxy::SETTINGS.https_port : nil
+    end
   end
 end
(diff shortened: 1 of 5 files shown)


test/httpboot/integration_test.rb
```diff
@@ -13,7 +12,7 @@ def app
   end
 
   def test_features
-    Proxy::DefaultModuleLoader.any_instance.expects(:load_configuration_file).with('httpboot.yml').returns(enabled: true, root_dir: '/var/lib/tftpboot')
+    Proxy::LegacyModuleLoader.any_instance.expects(:load_configuration_file).with('httpboot.yml').returns(enabled: true, root_dir: '/var/lib/tftpboot')
```
lzap, line 15: This is a change which is weird, the only way of achieving what I want was to use `after_activation` block, however when that's present the class loading falls back to legacy module for some reason:

```ruby
def module_loader_class
    @module_loader_class ||= after_activation_blk.nil? ? ::Proxy::DefaultModuleLoader : ::Proxy::LegacyModuleLoader
  end
```

I don't feel like I want to refactor this in this PR.

modules/httpboot/httpboot_plugin.rb
```diff
@@ -4,11 +4,16 @@ class Plugin < ::Proxy::Plugin
...
     default_settings :root_dir => '/var/lib/tftpboot'
 
     expose_setting :http_port
     expose_setting :https_port
+
+    after_activation do
+      plugin = ::Proxy::Plugins.instance.find{|p| p[:name] == :httpboot}
+      settings[:http_port] = plugin[:http_enabled] ? Proxy::SETTINGS.http_port : nil
+      settings[:https_port] = plugin[:https_enabled] ? Proxy::SETTINGS.https_port : nil
+    end
```
lzap, line 17: Whether http(s) is enabled or not is not visible in the plugin itself, only in the plugins "wrapper". This must be queried after activation was done, otherwise it's not set.

test/httpboot/integration_test.rb
ekohl: In https://github.com/theforeman/smart-proxy/pull/679 I implemented it using programmable settings.

lzap:
Nice that's much better.

PR #678, opened 2019-08-23
Fixes #27692 - ping optional for unused_ip call

lzap:
This should be pretty straightforward. Will do installer change once this is merged.


ekohl (approved):

lzap:
In regard to discussion at https://projects.theforeman.org/issues/23524 I think this is probably sub-optimal solution to the problem. Let me suggest this:

* Make the IP availablility method to be configurable in core and extendable by plugins
* Provide implementations for core:
  * ping ICMP
  * ping TCP
  * DNS reverse lookup
* Provide implementation for Infoblox DHCP plugin (HTTP REST API call)

Users could pick which options would be in use (array setting of options). Plugins could add to the array of methods. When performing IP availability check, if any method would return IP as not free next available IP would be taken.

lzap:
Created: https://projects.theforeman.org/issues/27841

PR #679, opened 2019-08-23
Refs #27654 - Account for module enabled over http or not

ekohl:
The module can be enabled for http-only while the proxy is https-only (or other combinations). This patch takes the enabled setting into account and guarantees that the http_port is only set when the module is actually available over http. The same goes for https_port.

This did need a minor refactor to make this easier to use.


lzap:
Works great. Thanks.

PR #680, opened 2019-08-23
Fixes #27693 - Port facts to the DefaultModuleLoader

ekohl:


lzap:
Test failures do seem to be related.

ares:
@ekohl thanks for this PR. It seems there's no update since August, when tests were failing.

```
Error Message
Disabling all modules in the group ['facts'] due to a failure in one of them: cannot load such file -- facter....
Stacktrace
Disabling all modules in the group ['facts'] due to a failure in one of them: cannot load such file -- facter.
<"running"> expected but was
<"failed">. (Test::Unit::AssertionFailedError)
/home/jenkins/workspace/test_proxy_develop_pr_core/ruby/2.3/test/facts/integration_test.rb:23:in `test_features'    
```

I'm marking this as waiting on contributor and would be interested if you plan to fix them in foreseeable future. We'd prefer to close this if not as it can be easily reopened later. We'll revisit this in next 14 days, please let us know meanwhile.

ekohl:
Added a `facts` bundler group and then it passes. I'm not sure what the best path forward is. Should we exclude this in packaging and deliver it broken? That might be ok since it's only used in discovery.

cc @evgeni

ares:
Sorry I didn't get that from your comment, what will be broken. Does that mean discovery needs to adapt as well? There's not much info in here or the redmine ticket so I have no idea what this try to change. E.g. if that's fixing something or refactors something.

ekohl:
So the situation is that the facts module needs facter (as a gem). Previously this code was untested. By adding a test, I triggered the error.

There are 2 solutions:
1. Not adding tests
2. Adding facter as a dependency.

At the time of writing, this PR implements option 2. However, this means in packaging we will need to deal with this. Again, 2 options:
1. Add it as a hard dependency even though in most cases it's unused
2. Exclude the dependency as we do now

Maybe there's a way to mock the require as well.

Right now I'm leaning to excluding the dependency since @lzap is working on a different implementation and this will go away. In the mean time, this will help moving away from the legacy module loader.

evgeni:
I am not totally sure I understand the situation here, so let me summarize.

* `facts` is a smart-proxy module/plugin
* it is only used during discovery, not when we deploy a regular smart-proxy
* it was using an old module interface, and this PR moves it to the modern interface
* this PR also adds tests, which fail if we don't add the `facter` dependency
* this dependency, as the module itself, is only needed in FDI, so the idea is to exclude it from the normal smart proxy packaging

If all of the above is correct, I think the current approach is OK. An alternative would be to add `facter` to `bundler.d/test.rb` because we only need it during tests and FDI handles dependencies differently?

ekohl:
That is indeed correct. This part now takes care of the facter dependency:

https://github.com/theforeman/foreman-discovery-image/blob/8f2c2a2d55605f35cd809355999c2f4310ce1611/20-packages.ks#L23-L29

Adding it to `bundler.d/test.rb` sounds like the easiest solution in the short term. Hadn't thought of that but implemented that now.

lzap:
On that note, facts on FDI will be removed next year after SSH/REX refactoring.

lzap (approved):
Thanks, I am ok with having this until we get rid of this on FDI. On my TODO.

lzap:
Works fine, thanks.

PR #681, opened 2019-09-05
Fixes #27789 - skip debug logging of large bodies

lzap:
While I find logging of whole body when TRACE environmental variable is enabled
very useful, OpenSCAP plugin uploads quite large artifact (bzi2 of XML report)
and the log line is not very useful. I am going to create a patch that will
filter out body which is longer than 2000 characters, that is exactly one 80x25
terminal screen.

### DIFF
--- a/lib/proxy/log.rb
+++ b/lib/proxy/log.rb
@@ -77,6 +77,7 @@ module Proxy
 
     def initialize(app)
       @app = app
+      @max_body_size = ENV['FOREMAN_LOG_MAX_BODY_SIZE'] || 2000
     end
 
     def call(env)
@@ -88,7 +89,11 @@ module Proxy
       logger.trace do
         if env['rack.input'] && !(body = env['rack.input'].read).empty?
           env['rack.input'].rewind
-          "Body: #{body}"
+          if env['CONTENT_TYPE'] = 'application/json' && body.size < @max_body_size
+            "Body: #{body}"
+          else
+            "Body: [unknown content type or body too large - filtered out]"
+          end
         else
           ''
         end


lib/proxy/log.rb
```diff
@@ -88,7 +88,11 @@ def call(env)
       logger.trace do
         if env['rack.input'] && !(body = env['rack.input'].read).empty?
           env['rack.input'].rewind
-          "Body: #{body}"
+          if body.size < 2000
```
ekohl: I'm wondering if we should also look at the `env['HTTP_CONTENT_TYPE']` to see if it's gzip/bzip2. Even small binary data is probably not very readable.

I'm also a fan of using constants for these values. `TRACE_MAX_BODY_SIZE` perhaps?

ares:
Will that be limited to file logger? Audits can contain more data but it's still useful for syslog consumers

lib/proxy/log.rb
lzap: Good ideas, implemented both. Actually more strict content type filtering - only JSON for now.

lzap:
> Will that be limited to file logger? Audits...

It's generic yes. There is no audit in smart proxy, I don't think that's a valid concern.

ekohl (approved):

ekohl:
LGTM but leaving the final merge to @ares since he commented.

lzap:
@ares this body logging only happens when A) debug mode is turned on, B) special ENV variable FOREMAN_TRACE is set. so it will not have any sideeffect on regular installs, this is for dev environment only.

ares:
Ah yes, totally makes sense. Sorry for the delay, merging!

ekohl:
This caused a regression:
https://ci.theforeman.org/job/smart-proxy-develop-source-release/34/
```
undefined method `to_sym' for ["starts", 2014-07-12 10:08:29 UTC, ";"]:Array
```

Edit: Apologies, looking closer it's actually the rsec release that https://github.com/theforeman/smart-proxy/pull/689 should fix.

PR #682, opened 2019-09-20
Fixes #27896 - DHCP providers filename option as hostname

lzap:
And expose the behavior preferably via capability.

### DIFF
--- a/modules/dhcp_common/isc/omapi_provider.rb
+++ b/modules/dhcp_common/isc/omapi_provider.rb
@@ -42,7 +42,7 @@ module Proxy::DHCP::CommonISC
       # TODO: Extract this block into a generic dhcp options helper
       statements = []
       statements << "filename = \\\"#{options[:filename]}\\\";" if options[:filename]
-      statements << bootServer(options[:nextServer])            if options[:nextServer]
+      statements << "next-server = #{bootServer(options[:nextServer])};" if options[:nextServer]
       statements << "option host-name = \\\"#{options[:hostname] || record.name}\\\";"
 
       statements += solaris_options_statements(options)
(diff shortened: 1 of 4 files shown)


ekohl (changes requested):
In `ztp_options_statements` there's:
```ruby
        opt150 = ip2hex validate_ip(options[:nextServer])`
```
This doesn't have a rescue for when this fails and needs some care. Probably best to introduce a helper that returns a quoted `bootServer` value. Might also be good to cache the result so there's just a single DNS query rather than multiple.

Looking at `poap_options_statements` which also uses `nextServer` I wonder if you can also quote IPs rather than putting them in hex. Maybe that would simplify some code as well as making lease files easier to read when debugging but feel free to ignore that.

lzap:
Rebased, your concerns were attacked except caching - that's a bad idea to do for a method with parameters.

lzap:
Fixed tests, all green.

modules/dhcp_common/isc/omapi_provider.rb
```diff
@@ -195,22 +194,22 @@ def poap_options_statements(options)
...
-        begin
-          ns = ip2hex Resolv.new.getaddress(server)
-        rescue
-          logger.warn "Failed to resolve IP address for #{server}"
-          ns = "\\\"#{server}\\\""
-        end
+        logger.info "Next-server option not IPv4, trying to resolve '#{server}'"
+        ip2hex(dns_resolv.getaddress(server))
+      rescue Resolv::ResolvError => e
+        logger.warn "Unable to resolve record, using '#{server}' as next-server"
```
ekohl: Nit: this is not entirely true anymore: it can be option-150. On the other hand, we use the `nextServer` parameter so I'm fine either way.

ekohl (approved):
Thank you. This is indeed what I had in mind.

I do wonder about completely dropping the `dns_resolv.getaddress` since we're allowed to use DNS. It would end up more predictable in what we generate so the proxy behavior becomes easier to understand. Just don't know about the operational side of it because it might mean you need to debug DNS issues on clients. I'd like to leave the choice up to you and merge what you think is best. Asking the community on discourse would also be fine by me.

lzap:
Well I am pretty sure we'd break existing deployments, many users don't have correct DNS setup yet they do PXE just fine. Also I am not sure if there are clients in the wild which don't support DNS at all - given amount of client-related workarounds in the TFTP server codebase I would not be surprised at all. I am all for keeping the current behavior.

modules/dhcp_common/isc/omapi_provider.rb
lzap: Improved the warning message a bit.

lzap:
Tests are green.

ares:
Makes sense to me, given an ack here, I take the liberty to merge. Thanks @lzap and @ekohl, merging!

lzap:
Great, thanks.

PR #684, opened 2019-10-03
Fixes #27993 - remove CBC ciphers

lzap:
They are no longer used by OpenSSL anyway.


ekohl (approved):
Might be good to add newer ciphers as well (GCM-SHA384 isn't listed), but this is well scoped.

PR #685, opened 2019-11-14
DHCP ISC: Try to remove lease also

jeremy-albinet:
Hi,

I guess we can try to "remove" lease when it's a lease.
We can't remove a lease but we can set the end of the lease to 0.

Let me know if I forgot something.

### DIFF
--- a/modules/dhcp_common/isc/omapi_provider.rb
+++ b/modules/dhcp_common/isc/omapi_provider.rb
@@ -16,13 +16,23 @@ module Proxy::DHCP::CommonISC
 
     def del_record(record)
       validate_record record
-      raise InvalidRecord, "#{record} is static - unable to delete" unless record.deleteable?
 
-      om_connect
-      omcmd "set hardware-address = #{record.mac}"
-      omcmd "open"
-      omcmd "remove"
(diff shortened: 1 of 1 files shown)


ares:
ok to test

lzap:
Hello, sorry for the delay. I am not willing to accept this for several reasons.

First, this is not a lease deletion or edit, you are creating a new lease which is sneaky. ISC DHCP handles that well and updates its database in a way that existing lease end time is reset, but this does not mean this is correct. Let me explain.

ISC DHCP does not support updating existing lease and informing client, there is a RFC 3203 (FORCERENEW) in the wild but it hasn't been implemented into ISC DHCP. I've checked using `tcpdump` that ISC DHCP does *not* inform the client about the change.

When Foreman user would perform this delete/update operation on a lease, there would be a client thinking it still has a valid lease and use it through the whole lifetime until the next update. And ISC DHCP would handle this IP address as free which could result in another lease being created for a different clientt. This is terrible thing to do and I am not going to debug and help people with this.

Realize that the moment lease is handed over to a client you can't do anything about it. At least with ISC DHCP, some DHCP servers do support this. I strongly suggest you to change your workflow and set a shorter lease time if you need leases to expire sooner. Consider using larger subnets if you are running out of IP addresses.

PR #701, opened 2020-02-19
Fixes #29085 - Fix Style/SafeNavigation cop

mmoll:


modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -50,7 +50,7 @@ def execute cmd, msg=nil, error_only=false
       rescue Timeout::Error
         raise Proxy::Dns::Error.new("dnscmd did not respond within #{tsecs} seconds")
       ensure
-        std_in.close  unless std_in.nil?
+        std_in&.close
         std_out.close unless std_in.nil?
         std_err.close unless std_in.nil?
```
mmoll, line 55: While I have no idea about this code, I guess these two should get adjusted.

lib/proxy/log_buffer/decorator.rb
```diff
@@ -50,7 +50,7 @@ def add(severity, message = nil, progname = nil, backtrace = nil)
       end
       info("Logging file reopened via USR1 signal") if reopened
       # exceptions are also sent to structured log if available
-      self.exception("Error details", backtrace) if backtrace && backtrace.is_a?(Exception)
+      self.exception("Error details", backtrace) if backtrace&.is_a?(Exception)
```
ekohl, line 53: Is this actually needed? Doesn't `is_a?` always work, even on nil?

modules/dns_dnscmd/dns_dnscmd_main.rb
ekohl: This looks like a bug indeed

ekohl (approved):

ekohl (approved):

lib/proxy/log_buffer/decorator.rb
lzap: Correct.

modules/dns_dnscmd/dns_dnscmd_main.rb
lzap: Yeah definitely, just adjust everything to `&.` please.

lzap:
This still needs a fix, let me know if you are planning a PR:

```
        std_in&.close
        std_out.close unless std_in.nil?
```

modules/dns_dnscmd/dns_dnscmd_main.rb
ekohl: Note that it's always checking `std_in` here so the cop can't detect it's the same.

mmoll:
https://github.com/theforeman/smart-proxy/pull/703

PR #702, opened 2020-02-19
Fixes #29086 - Configure and fix Style/TrailingCommaIn*Literal

mmoll:


lzap:
Is it possible to make this a non-fatal check? Like a warning. While I understand it's good, I also hate to hear from Rubocop when I miss it.

modules/dhcp_common/dhcp_common.rb
```diff
@@ -19,7 +19,7 @@ module Proxy::DHCP
     :install_server_name   => {:code => 11, :kind => "String"    }, # mediahost
     :install_path          => {:code => 12, :kind => "String"    }, # /vol/solgi_5.10/sol10_hw0910
     :sysid_server_path     => {:code => 13, :kind => "String"    }, # 192.168.216.241:/vol/jumpstart/sysidcfg/sysidcfg_primary
-    :jumpstart_server_path => {:code => 14, :kind => "String"    }  # 192.168.216.241:/vol/jumpstart
+    :jumpstart_server_path => {:code => 14, :kind => "String"    },  # 192.168.216.241:/vol/jumpstart
```
ekohl: Perhaps we should keep the comments aligned
```suggestion
    :jumpstart_server_path => {:code => 14, :kind => "String"    }, # 192.168.216.241:/vol/jumpstart
```

ekohl (approved):

mmoll:
> Is it possible to make this a non-fatal check?

AFIACT no.

modules/dhcp_common/dhcp_common.rb
mmoll: Fixed.

PR #713, opened 2020-02-21
Fixes #29117 - Fix Style/SpecialGlobalVars cop

mmoll:


ekohl (commented):
I think things are failing unless you `require 'English'`

mmoll:
I thought newer Rubies don't need the require, but that's not the case...

ekohl (approved):

PR #714, opened 2020-02-21
Fixes #29118 - Fix Style/SingleLineMethods cop

mmoll:


test/dependency_injection_test.rb
```diff
@@ -2,11 +2,15 @@
 
 class TestDependencyOne
   attr_reader :arg
-  def initialize(arg = nil); @arg = arg; end
+  def initialize(arg = nil);
```
ekohl: are the semi colons still needed?

mmoll: Nope, changed for these ones now. There's another cop for it, that finds more. ;)

ekohl (approved):

PR #717, opened 2020-02-21
Fixes #29121 - Fix Style/RedundantBegin cop

mmoll:


lib/proxy/util.rb
```diff
@@ -18,7 +18,6 @@ def start(&ensured_block)
       # run the task in its own thread
       logger.debug "Starting task: #{@command}"
       @task = Thread.new(@command) do |cmd|
-        begin
           status = nil
```
ekohl: The indenting probably also needs to be modified, right?

ekohl (commented):
Looks like a merge conflict has shown up in the todo file as well.

lib/proxy/util.rb
mmoll: Should get auto-corrected automatically, after GH-723 and GH-724 got in.

ekohl (approved):

PR #719, opened 2020-02-21
Sort Rubocop rules

mmoll:


ekohl (approved):
The goal is to copy the one from Foreman?

mmoll:
> The goal is to copy the one from Foreman?

At the end to have as much as possible the same configs for core, proxy and installer.

mmoll:
Only sorting and minor rewordings.

PR #721, opened 2020-02-21
Fixes #29127 - Fix Style/SymbolProc cop

mmoll:


mmoll:
We do have this one enabled in core and it's an easy auto-correct.

ekohl (approved):

PR #722, opened 2020-02-21
Fixes #29128 - Fix Style/Lambda cop

mmoll:


mmoll:
We do have this one enabled in core and it's an easy auto-correct.

ekohl:
I'm not sure about this one. Is this purely a style thing? Are there differences?

mmoll:
AFAICT, there should be no difference: http://rubyforadmins.com/lambda-functions

The cop enforces the arrow notation specifially only for one-liners.

lzap:
Honestly I see zero value in this one. Actually I like `lambda` syntax more easy to read, that's probably just me.

mmoll:
> Honestly I see zero value in this one.

If we wouldn't have this in core since many years, I'd just leave it disabled. The alternative would be to disable it in core.

test/root/root_v2_api_test.rb
```diff
@@ -25,8 +25,8 @@ def app
   end
 
   def test_features
-    proc2 = lambda{ ['a', 'b']}
-    proc3 = lambda{raise "Should not be called"}
+    proc2 = -> { ['a', 'b'] }
+    proc3 = -> {raise "Should not be called" }
```
ekohl: This whitespacing looks inconsistent.

mmoll: aw, I fixed up the `proc2` line manually. Now also `proc3`.

mmoll: I guess, there would be another (currently disabled) cop also for that.

ekohl (approved):

PR #723, opened 2020-02-21
Fixes #29129 - Fix Layout/IndentationWidth cop

mmoll:


tasks/proxy_tasks.rake
```diff
@@ -1,8 +1,8 @@
 desc "Send patch information to the foreman-dev list"
```
ekohl, line 1: This whole task should be dropped

ekohl (approved):

tasks/proxy_tasks.rake
ekohl: https://github.com/theforeman/smart-proxy/pull/725

PR #734, opened 2020-02-24
Fixes #29156 - Fix Style/Dir cop

mmoll:


modules/tftp/server.rb
```diff
@@ -41,7 +41,7 @@ def create_default config
     # returns the absolute path
     def path(p = nil)
       p ||= Proxy::TFTP::Plugin.settings.tftproot
-      (p =~ /^\//) ? p : Pathname.new(File.expand_path(File.dirname(__FILE__))).join(p).to_s
+      (p =~ /^\//) ? p : Pathname.new(__dir__).join(p).to_s
```
ekohl, line 44: Is `__dir__` guaranteed to be an absolute path?

mmoll: I'd say yes: https://ruby-doc.org/core-2.7.0/Kernel.html#method-i-__dir__

ekohl: Ok, my test case was in an interactive shell which is probably not representative.
```console
$ pry
/usr/share/ruby/irb/input-method.rb:16: warning: already initialized constant IRB::STDIN_FILE_NAME
/usr/share/gems/gems/irb-1.0.0/lib/irb/input-method.rb:16: warning: previous definition of STDIN_FILE_NAME was here
[1] pry(main)> Pathname.new(__dir__).join('bla').to_s
=> "bla"
[2] pry(main)> __dir__
=> "."
[3] pry(main)> Pathname.new(File.expand_path(__dir__)).join('bla').to_s
=> "/home/ekohl/bla"
```

lzap (approved):

PR #735, opened 2020-02-24
Fixes #29157 - Fix Style/NegatedIf cop

mmoll:


extra/changelog
```diff
@@ -13,7 +13,7 @@ cmd.each_line do |l|
   date, _, _ = date.split(" ")
 
   id = "#{date}\t#{author}"
-  if not list[id]
+  if !(list[id])
```
ekohl: Can this be `unless list[id]` as well?

modules/dhcp_native_ms/dhcp_native_ms_main.rb
```diff
@@ -95,11 +95,11 @@ def unused_ip(subnet_address, mac_address, from_address, to_address)
         foreman_end_addr_ip = IPAddr.new(to_address)
         dhcp_start_addr_ip = IPAddr.new(dhcp_start_addr)
         dhcp_end_addr_ip = IPAddr.new(dhcp_end_addr)
-        if not (dhcp_start_addr_ip..dhcp_end_addr_ip).cover? foreman_start_addr_ip
+        if !(dhcp_start_addr_ip..dhcp_end_addr_ip).cover? foreman_start_addr_ip
```
ekohl: Logically this is equivalent to `unless` as well, right? Same be low

extra/changelog
mmoll: Style/NegatedIf does fix this

modules/dhcp_native_ms/dhcp_native_ms_main.rb
mmoll: Style/NegatedIf does fix this

extra/changelog
```diff
@@ -13,7 +13,7 @@ cmd.each_line do |l|
   date, _, _ = date.split(" ")
 
   id = "#{date}\t#{author}"
-  if not list[id]
+  unless (list[id])
```
ekohl: Maybe remove the braces as well?

ekohl (approved):

extra/changelog
mmoll: removed.

PR #745, opened 2020-02-25
Fixes #25369 - Fix Lint/StringConversionInInterpolation cop

mmoll:


ekohl (approved):
`option.to_s.start_with?(mod.to_s)` is probably easier to understand, but let's keep this minimal.

PR #750, opened 2020-02-28
Fixes #29185 - expose TFTP via /EFI/BOOT

lzap:
Javier Martinez Canillas of Red Hat explained to me that grub2 from grub2-efi-x64-cdboot package expects files to be on /EFI/BOOT path and this is where we should expose them both via TFTP and HTTP. Let's change this to comply with how grub2 expects things, currently we expose the TFTP directory via /httpboot and /EFI root paths.

This patch retains backward compatibility, so a file `/var/lib/tftpboot/TEST` is available under three paths:

* http://localhost:8448/EFI/BOOT/TEST
* http://localhost:8448/EFI/TEST
* http://localhost:8448/httpboot/TEST

An attempt to list BOOT subdirectory fails, that's expected:

```
$ curl http://localhost:8448/EFI/BOOT
Invalid or empty path
```

### DIFF
--- a/config/settings.d/httpboot.yml.example
+++ b/config/settings.d/httpboot.yml.example
@@ -1,7 +1,7 @@
 ---
-# Enable publishing of a given directory under /EFI and /httpboot paths.
-# Directory listing is not possible, symlinks are followed but not outside
-# of the root directory specified in this file.
+# Enable publishing of a given directory under /EFI/BOOT, /EFI and /httpboot
+# paths. Directory listing is not possible, symlinks are followed but not
+# outside of the root directory specified in this file.
 
 # Enables the module, make sure to enable TFTP module as well to allow
 # configuration files deployment.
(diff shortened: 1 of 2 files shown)


ekohl:
It would be nice to have the same text as in this PR for the git commit. The history is very useful. Would it also make sense to write some integration test for this to make sure we don't break this in the future?

lzap:
They do have the same text, or what do you mean?

We do have integration tests in `httpboot_api_test.rb` thing is they are written on the Sinatra App level, therefore they are all executed from the app root namespace. The patch simply adds new mapping, I don't see huge benefit in adding that.

lzap:
Ping @ekohl

lzap:
Rebased, I think this can be useful for non-RH/Debian Foreman users with some different (untested) grub2 builds.

lzap:
This is still valid and should be merged.

PR #752, opened 2020-03-03
Fixes #29252 - Stop accepting TLS 1.1 connections

ekohl:
TLS 1.1 is not needed since all tools connecting to the Smart Proxy should support TLS 1.2. A general rule in security is to disable what's not needed.


ehelms (approved):

lzap (approved):
This only affects Webrick, Puma initializes OpenSSL on its own and it has configurable flags for it. I am fine with this short-lived change tho.

tbrisker:
should we update the 2.1 upgrade warnings section about this?

ekohl:
Probably good and perhaps also a headline feature.

tbrisker:
I don't think a change in accepted tls version is big enough for a headline feature, just upgrade warning is enough. @lzap can you please open a PR to the manual?

ekohl:
It could be useful for PCI compliance which disallows TLS 1.1, but smart_proxy_dynflow should also do this. On the katello side candlepin should also do this. Foreman itself could also be done, but can be trickier on Katello due older RHEL versions with Pulp.

ekohl:
https://github.com/theforeman/smart_proxy_dynflow/pull/69 does the same for smart_proxy_dynflow_core.

ekohl:
And https://github.com/theforeman/puppet-candlepin/pull/152 for Candlepin.

PR #757, opened 2020-04-03
Fixes #29468 - Add default URL validator for plugins

ekohl:
URL settings are common and this moves it from a custom validator in the Puppet module to a default. Also adds tests and ensures the dhcp and dns libvirt providers use it too.


ezr-ondrej:
Would it make sense to do what rails validation do and allow `Hash` as a value of the validation?
We could allow e.g. `:url => { :allow_nil => true }`

ekohl:
At this point we don't have optional URLs: we either need one or there is none so I'm not sure how much value there is in that.

mmoll:
merged, bedankt @ekohl!

PR #758, opened 2020-04-27
Fixes #30906 - Add support for Redfish to BMC smart proxy.

xray-sky:
Initial commit includes support for variations in Redfish
implementation among an arbitrary range of Supermicro, Dell,
and HPE servers. Modular implementation should make it easier
to accommodate additional variations, as they are discovered.

There is a corresponding pull request in the Foreman main project 
to allow selection of Redfish provider from front end.

### DIFF
--- a/bundler.d/bmc.rb
+++ b/bundler.d/bmc.rb
@@ -1,3 +1,4 @@
 group :bmc do
   gem 'rubyipmi', '>= 0.10.0'
+  gem 'redfish_client', '>= 0.5.1'
 end
--- a/modules/bmc/bmc.rb
+++ b/modules/bmc/bmc.rb
@@ -4,7 +4,7 @@ module Proxy
   module BMC
     # Just a bunch of stubs
     def installed_providers
-      Proxy::BMC::IPMI.providers_installed + ['shell']
+      Proxy::BMC::IPMI.providers_installed + ['redfish', 'shell']
     end
 
     def installed_ipmi_providers
@@ -12,15 +12,11 @@ module Proxy
     end
 
     def providers
-      Proxy::BMC::IPMI.providers + ['shell']
+      Proxy::BMC::IPMI.providers + ['redfish', 'shell']
     end
 
     def installed?(provider)
-      if provider == 'shell'
-        true
-      else
-        Proxy::BMC::IPMI.installed?(provider)
-      end
+      %w(redfish shell).include?(provider) || Proxy::BMC::IPMI.installed?(provider)
     end
   end
 end
--- a/modules/bmc/bmc_api.rb
+++ b/modules/bmc/bmc_api.rb
@@ -1,4 +1,5 @@
 require 'bmc/ipmi'
+require 'bmc/redfish'
 
 module Proxy::BMC
   class Api < ::Sinatra::Base
@@ -360,7 +361,7 @@ module Proxy::BMC
     end
 
     def non_ipmi_providers
-      ['ssh', 'shell']
+      ['redfish', 'ssh', 'shell']
     end
 
     # returns a provider type by validating the given type.  If for some reason the type is invalid
@@ -417,6 +418,18 @@ module Proxy::BMC
             :bmc_provider => provider_type,
           }
           @bmc = Proxy::BMC::IPMI.new(args)
+        when 'redfish'
+          log_halt 401, "unauthorized" unless auth.provided?
+          log_halt 401, "bad_authentication_request, credentials are not in auth.basic format" unless auth.basic?
+          username, password = auth.credentials
+
+          args = {
+            :host     => params[:host],
+            :username => username,
+            :options  => body_parameters['options'],
+            :password => password,
+          }
+          @bmc = Proxy::BMC::Redfish.new(args)
         when "shell"
           require 'bmc/shell'
           @bmc = Proxy::BMC::Shell.new
--- a/modules/bmc/redfish.rb
+++ b/modules/bmc/redfish.rb
@@ -1,0 +1,238 @@
+require 'redfish_client'
+require 'bmc/base'
+require 'bmc/redfish/dell'
+require 'bmc/redfish/hpe'
+
+module Proxy
+  module BMC
+    class Redfish < Base
+      include Proxy::Log
+      include Proxy::Util
+
+      def initialize(args)
+        @redfish_verify_ssl = Proxy::BMC::Plugin.settings.redfish_verify_ssl
+        super
+        load_vendor_overrides
+      end
+
+      def connect(args = { })
+        connection = RedfishClient.new("https://#{args[:host]}/", verify: @redfish_verify_ssl)
+        connection.login(args[:username], args[:password])
+        connection
+      end
+
+      def load_vendor_overrides
+        mod = case manufacturer
+              when 'Dell Inc.' then :RedfishVendorOverridesDellInc
+              when 'HPE'       then :RedfishVendorOverridesHPE
+              else
+                logger.debug "No #{manufacturer} specific overrides available - using generic Redfish calls"
+                return
+              end
+
+        # Get the actual Module corresponding to the name stored in mod (based on the
+        # manufacturer data returned from this Redfish interface), then extend this
+        # instance of Proxy::BMC::Redfish with the methods in that module (if there is
+        # one). Methods from the override module can use super, etc. as if it had been
+        # a static subclass, but it is a dynamic/runtime effect.
+        #
+        # The manufacturer method must work without overrides -- so all vendor-specific
+        # code may be delegated into override modules _except_ for in manufacturer()
+
+        logger.debug "Extending Redfish with vendor overrides for #{manufacturer}"
+        extend Kernel.const_get(mod)
+      end
+
+      def manufacturer
+        system.Manufacturer
+      end
+
+      # returns boolean if the test is successful
+      def test
+        host.Managers.Members.any?
+      rescue NoMethodError
+        false
+      end
+
+      def identifystatus
+        system.IndicatorLED&.downcase
+      end
+
+      def identifyon
+        system.patch(payload: { 'IndicatorLED' => 'On' })
+      end
+
+      def identifyoff
+        system.patch(payload: { 'IndicatorLED' => 'Off' })
+      end
+
+      def poweroff(soft = false)
+        poweraction(soft ? 'GracefulShutdown' : 'ForceOff')
+      end
+
+      def powercycle
+        poweraction('ForceRestart')
+      end
+
+      def poweron
+        poweraction('On')
+      end
+
+      def powerstatus
+        system.PowerState&.downcase
+      end
+
+      def poweron?
+        powerstatus == 'on'
+      end
+
+      def poweroff?
+        powerstatus == 'off'
+      end
+
+      def bootdevice
+        system.Boot['BootSourceOverrideTarget']
+      end
+
+      def bootdevices
+        # RedFish will tell you which devices can be put in the boot source override
+        # but Foreman seems structurally arranged at present for these values to be
+        # hardcoded. I can think of a few reasons to continue with that pattern. But
+        # if in the future it becomes advantageous to do something more dynamic --
+        # this is how to find out:
+        #
+        #    system.Boot['BootSourceOverrideTarget@Redfish.AllowableValues']
+        #
+        # There's an override for older HPs which have the list of boot devices at a
+        # different spot. (see HPE vendor overrides)
+        #
+        # Dell gives a list like [ None, Floppy, Cd, Hdd, SDCard, Utilities,
+        #                          BiosSetup, Pxe ]
+        # HPE -        [ None, Floppy, Cd, Hdd, Usb, Utilities, BiosSetup, Pxe ]
+        # SuperMicro - [ None, Pxe, Hdd, Diags, Cd, BiosSetup, FloppyRemovableMedia,
+        #                UsbKey, UsbHdd, UsbFloppy, UsbCd, UefiCd, UefiHdd, ... ]
+        #
+        # For the four devices Foreman hardcodes, lucky for us all three use the same set
+        # of values.
+        ['pxe', 'disk', 'bios', 'cdrom']
+      end
+
+      def bootdevice=(args = { :device => nil, :reboot => false, :persistent => false })
+        devmap = { 'bios'  => 'BiosSetup',
+                   'cdrom' => 'Cd',
+                   'disk'  => 'Hdd',
+                   'pxe'   => 'Pxe' }
+
+        system.patch(
+          payload: {
+            'Boot' => {
+              'BootSourceOverrideTarget' => devmap[args[:device]],
+              'BootSourceOverrideEnabled' => args[:persistent] ? 'Enabled' : 'Once',
+            },
+          })
+        powercycle if args[:reboot]
+      end
+
+      def bootpxe(reboot = false, persistent = false)
+        self.bootdevice = { :device => 'pxe', :reboot => reboot, :persistent => persistent }
+      end
+
+      def bootdisk(reboot = false, persistent = false)
+        self.bootdevice = { :device => 'disk', :reboot => reboot, :persistent => persistent }
+      end
+
+      def bootbios(reboot = false, persistent = false)
+        self.bootdevice = { :device => 'bios', :reboot => reboot, :persistent => persistent }
+      end
+
+      def bootcdrom(reboot = false, persistent = false)
+        self.bootdevice = { :device => 'cdrom', :reboot => reboot, :persistent => persistent }
+      end
+
+      def ip
+        bmc_nic.IPv4Addresses&.first&.[]('Address')
+      end
+
+      def mac
+        bmc_nic.MACAddress
+      end
+
+      def gateway
+        bmc_nic.IPv4Addresses&.first&.[]('Gateway')
+      end
+
+      def netmask
+        bmc_nic.IPv4Addresses&.first&.[]('SubnetMask')
+      end
+
+      def vlanid
+        vlaninfo = bmc_nic.VLAN
+        return unless vlaninfo # some older Proliants don't have a VLAN setting at all.
+        return unless vlaninfo.VLANEnable
+        vlaninfo.VLANId
+      end
+
+      def ipsrc
+        bmc_nic.IPv4Addresses&.first&.[]('AddressOrigin')
+      end
+
+      def guid
+        manager.UUID
+      end
+
+      def version
+        manager.FirmwareVersion
+      end
+
+      def reset(type = nil)
+        logger.debug("BMC reset arg #{type.inspect} unused for Redfish - standard reset only") if type
+        host.post(path: manager.Actions&.[]('#Manager.Reset')&.[]('target'), payload: { 'ResetType' => 'Reset' })
+      end
+
+      def model
+        system.Model
+      end
+
+      def serial
+        system.SerialNumber
+      end
+
+      def asset_tag
+        system.AssetTag
+      end
+
+      protected
+
+      attr_reader :host
+
+      private
+
+      def poweraction(action)
+        host.post(path: system.Actions&.[]('#ComputerSystem.Reset')&.[]('target'), payload: { 'ResetType' => action })
+      end
+
+      # I haven't yet encountered a system (apart from a blade chassis, which I think
+      # wouldn't be modeled in Foreman as a host?) which has multiple BMCs, managed systems,
+      # or BMC NICs. But it's part of the Redfish design, so it's at least possible that
+      # someone could. Warn if we encounter such a system, but try to carry on regardless.
+
+      def manager
+        managers = host.Managers&.Members
+        logger.warn("Chassis has multiple BMCs? - using first") if managers.length > 1
+        managers.first
+      end
+
+      def system
+        systems = host.Systems&.Members
+        logger.warn("BMC has multiple managed systems? - using first") if systems.length > 1
+        systems.first
+      end
+
+      def bmc_nic
+        nics = manager.EthernetInterfaces&.Members
+        logger.warn("BMC has multiple NICs? - using first") if nics.length > 1
+        nics.first
+      end
+    end
+  end
+end
--- a/modules/bmc/redfish/dell.rb
+++ b/modules/bmc/redfish/dell.rb
@@ -1,0 +1,22 @@
+module RedfishVendorOverridesDellInc
+  def powercycle
+    # ForceRestart was added in Lifecycle Controller 3.30.30.30
+    # prior to that, it required ForceOff followed by On.
+    if system.Actions&.[]('#ComputerSystem.Reset')&.[]('ResetType@RedfishAllowableValues')&.include? 'ForceRestart'
+      poweraction('ForceRestart')
+    else
+      poweraction('ForceOff')
+      # it only takes a couple seconds to force off, but if you send the 'On' action too
+      # quickly, you'll get an error that the server is already on, because it hasn't
+      # actually shut off yet. fifteen seconds is chosen arbitrarily; hopefully it covers
+      # all scenarios.
+      sleep 15
+      poweraction('On')
+    end
+  end
+
+  def reset(type = nil)
+    logger.debug("BMC reset arg #{type.inspect} unused for Dell Redfish - GracefulRestart only") if type
+    host.post(path: manager.Actions&.[]('#Manager.Reset')&.[]('target'), payload: { 'ResetType' => 'GracefulRestart' })
+  end
+end
--- a/modules/bmc/redfish/hpe.rb
+++ b/modules/bmc/redfish/hpe.rb
@@ -1,0 +1,14 @@
+module RedfishVendorOverridesHPE
+  def poweroff(soft = false)
+    poweraction(soft ? 'PushPowerButton' : 'ForceOff')
+  end
+
+  # -- See note about dynamic bootdevices in redfish.rb --
+  # this override will let Foreman get the list of supported boot devices
+  # from older HPs, if and when potential boot devices become dynamically discovered
+  #
+  # def bootdevices
+  #   # older HPs have the list of supported values at a different location
+  #   system.Boot.BootSourceOverrideSupported || super
+  # end
+end
--- a/config/settings.d/bmc.yml.example
+++ b/config/settings.d/bmc.yml.example
@@ -4,9 +4,15 @@
 
 # Available providers:
 # - freeipmi / ipmitool - requires the appropriate package installed, and the rubyipmi gem
+# - redfish - requires the redfish_client gem
 # - shell - for local reboot control (requires sudo access to /sbin/shutdown for the proxy user)
 # - ssh - limited remote control (status, reboot, turn off)
 #:bmc_default_provider: freeipmi
+
+# Redfish provider HTTPS certificate verification:
+# - If your BMCs just have the vendor-supplied self-signed certificates, you can set
+#   bmc_redfish_verify_ssl to false. The default is to perform certificate verification.
+#:bmc_redfish_verify_ssl: true
 
 # default user and ssh key for ssh provider
 #:bmc_ssh_user: root
--- a/lib/proxy/default_plugin_validators.rb
+++ b/lib/proxy/default_plugin_validators.rb
@@ -4,6 +4,7 @@ class Proxy::DefaultPluginValidators
       file_readable: ::Proxy::PluginValidators::FileReadable,
       presence: ::Proxy::PluginValidators::Presence,
       url: ::Proxy::PluginValidators::Url,
+      boolean: ::Proxy::PluginValidators::Boolean,
     }
   end
 end
--- a/lib/proxy/plugin_validators.rb
+++ b/lib/proxy/plugin_validators.rb
@@ -56,4 +56,12 @@ module ::Proxy::PluginValidators
       raise ::Proxy::Error::ConfigurationError.new("Setting '#{@setting_name}' contains an invalid url")
     end
   end
+
+  class Boolean < Base
+    def validate!(settings)
+      setting_value = settings[@setting_name]
+      raise ::Proxy::Error::ConfigurationError, "Setting '#{@setting_name}' is expected to be true/false" unless setting_value.is_a?(TrueClass) || setting_value.is_a?(FalseClass)
+      true
+    end
+  end
 end
--- a/modules/bmc/bmc_plugin.rb
+++ b/modules/bmc/bmc_plugin.rb
@@ -3,6 +3,8 @@ module Proxy::BMC
     http_rackup_path File.expand_path("http_config.ru", File.expand_path(__dir__))
     https_rackup_path File.expand_path("http_config.ru", File.expand_path(__dir__))
 
+    default_settings :redfish_verify_ssl => true
+    validate :redfish_verify_ssl, :boolean => true
     plugin :bmc, ::Proxy::VERSION
   end
 end
--- a/test/bmc/bmc_api_test.rb
+++ b/test/bmc/bmc_api_test.rb
@@ -218,7 +218,7 @@ class BmcApiTest < Test::Unit::TestCase
     get "/providers", args
     assert last_response.ok?, "Last response was not ok: #{last_response.body}"
     data = JSON.parse(last_response.body)
-    expected = ['freeipmi', 'ipmitool', 'ssh', 'shell']
+    expected = ['freeipmi', 'ipmitool', 'redfish', 'ssh', 'shell']
     assert_equal(expected, data["providers"])
   end
 
@@ -227,7 +227,7 @@ class BmcApiTest < Test::Unit::TestCase
     get "/providers/installed", args
     assert last_response.ok?, "Last response was not ok: #{last_response.body}"
     data = JSON.parse(last_response.body)
-    expected = ['freeipmi', 'ipmitool', 'ssh', 'shell']
+    expected = ['freeipmi', 'ipmitool', 'redfish', 'ssh', 'shell']
     assert_equal(expected, data["installed_providers"])
   end
 
--- a/test/plugins/validator_test.rb
+++ b/test/plugins/validator_test.rb
@@ -133,3 +133,37 @@ class UrlValidatorTest < Test::Unit::TestCase
     assert_match(/expected to contain a url/, error.message)
   end
 end
+
+class BooleanValidatorTest < Test::Unit::TestCase
+  class BooleanValidatorTestPlugin < ::Proxy::Plugin
+    default_settings :a_settting => true
+  end
+
+  def test_required_parameter_with_a_value_passes_validation
+    assert ::Proxy::PluginValidators::Boolean.new(BooleanValidatorTestPlugin, 'a_setting', nil, nil).validate!(:a_setting => true)
+  end
+
+  def test_empty_string_treated_as_missing_value
+    error = assert_raises ::Proxy::Error::ConfigurationError do
+      ::Proxy::PluginValidators::Boolean.new(BooleanValidatorTestPlugin, 'a_setting', nil, nil).validate!(:a_setting => '')
+    end
+
+    assert_match(%r{expected to be true/false}, error.message)
+  end
+
+  def test_required_parameter_without_a_value_fails_validation
+    error = assert_raises ::Proxy::Error::ConfigurationError do
+      ::Proxy::PluginValidators::Boolean.new(BooleanValidatorTestPlugin, 'a_setting', nil, nil).validate!(:a_setting => nil)
+    end
+
+    assert_match(%r{expected to be true/false}, error.message)
+  end
+
+  def test_optional_parameter_without_a_value_fails_validation
+    error = assert_raises ::Proxy::Error::ConfigurationError do
+      ::Proxy::PluginValidators::Boolean.new(BooleanValidatorTestPlugin, 'optional_setting', nil, nil).validate!(:optional_setting => nil)
+    end
+
+    assert_match(%r{expected to be true/false}, error.message)
+  end
+end


modules/bmc/redfish/dell.rb
```diff
@@ -0,0 +1,18 @@
+module RedfishVendorOverridesDellInc
+
+  def powercycle
+    # ForceRestart was added in Lifecycle Controller 3.30.30.30
+    # prior to that, it required ForceOff followed by On.
+    if host.System.Members[0].Actions['#ComputerSystem.Reset']['ResetType@RedfishAllowableValues'].include? 'ForceRestart'
+      poweraction('ForceRestart')
+    else
+      poweraction('ForceOff')
+      # REVIEW - I wonder if any kind of delay is needed for this to be reliable.
```
ohadlevy: Anyway to test this? probably in the grand scheme of things (e.g. how long it actually take to reboot), adding some delay here won't hurt.

ohadlevy:
[ok to test]

ohadlevy (commented):
LGTM on a high level, I dont have the ability to actually test this so YMMV :) tests are missing of course :)

ohadlevy:
ah, and note the test failures due to syntax error..

```
test_proxy_develop_pr_core/ruby/2.7/modules/bmc/redfish/dell.rb:18: syntax error, unexpected end-of-input, expecting `end' 
```

modules/bmc/redfish/dell.rb
xray-sky: I wanted to resolve this question, but the way my access to real BMCs and dev/test instance of Foreman shook out meant I couldn't do it. Theoretically the iDRAC will queue the events, but _theoretically_ the iDRAC will do all _kinds_ of things it doesn't actually do. A sleep may not be sufficient, in case it doesn't -- it may require using the API to check status of "long running" Redfish actions. Perhaps it's better just to not support the older iDRAC without ForceRestart, in this initial commit. I'm open to direction.

xray-sky:
whoops.

ps. I don't understand how to correct the Redmine check failure.

modules/bmc/redfish/dell.rb
ohadlevy: are you worried about a race condition? or simply ignoring the older command (so its on and stays on?)
I think its acceptable to power off, busy wait and turn on again, I'm sure folks with older hardware would still prefer an option vs no option at all? (assuming its not too much of an effort..?)

modules/bmc/bmc.rb
```diff
@@ -4,19 +4,19 @@ module Proxy
...
     end
 
     def providers
-      Proxy::BMC::IPMI.providers + ['shell']
+      Proxy::BMC::IPMI.providers + ['redfish', 'shell']
     end
 
     def installed?(provider)
-      if provider == 'shell'
+      if provider == 'redfish' or provider == 'shell'
```
ekohl: Ruby typically uses `||` rather than `or`
```suggestion
      if provider == 'redfish' || provider == 'shell'
```

modules/bmc/redfish.rb
```diff
@@ -0,0 +1,206 @@
...
+
+      def initialize(args)
+        super
+        load_vendor_overrides
+        @bmc
+      end
+
+      def connect(args = { })
+        # TODO probably verify should be an option.
+        connection = RedfishClient.new("https://#{args[:host]}/", verify: false)
```
ekohl:
```suggestion
        connection = RedfishClient.new("https://#{args[:host]}/", verify: args.fetch(:verify, false))
```

ekohl (commented):
Overall this looks like a good implementation but like @ohadlevy I don't have the hardware to test this. Since it doesn't interfere with other providers, that doesn't have to be a problem.

Some small inline comments.

modules/bmc/redfish.rb
lzap: This really must be a setting, by default it must be turned on and be only opt-out flag for test/dev environments. In the README/docs we should tell users to put X509 cert into system cert store. You can do this as a global flag for the whole bmc module that would affect all providers (since you can only select just one).

modules/bmc/redfish.rb
```diff
@@ -0,0 +1,206 @@
...
+        mfr.tr!('^A-Za-z', '')
+        begin
+          mod = Kernel.const_get("RedfishVendorOverrides#{mfr}".to_sym)
+        rescue NameError
+          # no extensions for this vendor
+          return
+        end
+        logger.debug "Extending Redfish with vendor overrides for #{mfr}"
+        self.extend mod
+      end
```
lzap, line 44: Uh. Okay...

modules/bmc/redfish/dell.rb
lzap: Looks like the underlaying library does support async calls. We could use that, that's a good start: https://github.com/xlab-si/redfish-client-ruby#handling-asynchronous-operations

Our BMC API is really simplistic and request-response. If you expect more from this, then we should probably talk about something completely new.

lzap (commented):
Nice, this is on our TODO for a long time.

Until we reserve some hardware for testing for this feature, do you think it's possible to test this against mocks? For example against https://redfish.dmtf.org/redfish/mockups/v1/922 or similar ones there.

ekohl:
Could you also update the example config to include redfish:
https://github.com/theforeman/smart-proxy/blob/ba31df8e859a5a0cb5f1fb5057e3533e7ea26a4d/config/settings.d/bmc.yml.example#L5-L8

lzap:
The biggest question is - our BMC API is ultra simple and not the cleanest thing in the world. Redfish, on the other hand, is well designed, generic and complex API. Building on top of our BMC API narrows the scope of Redfish to bare minimum. The moment we want more from it can be problematic, we have experienced this with DHCP (Foreman has simplified IP management). Now we have implemented External IPAM which is another service providing similar thing.

I think it's more complex question to ask, Foreman has been what I call "traditional old school datacenter" approach but things are changing now. Everything needs to be automated. It's not just about management of servers but inventory, discovery, management of storage, networks and relationships.

Having said that, the implementation is currently so small that it totally makes sense to take this as the first step towards bigger goals. Since it solves your problem, it makes a lot of sense.

modules/bmc/bmc.rb
xray-sky: it uses both, and the precedence is slightly different, but I don't mind changing it. I'll include it in the next update.

modules/bmc/redfish.rb
xray-sky: I've never worked anywhere that deployed valid (non-self-signed) certificates to BMC interfaces, but I don't have a problem making it a config file option.

xray-sky:
Only some Redfish calls use the asynchronous interface. I'm not sure if power state changes do.  But as luck would have it, I just found out that we received a brand new Dell into our testing environment yesterday, and it hasn't been handed off to anybody yet. I will see if I can get access to get at least the beginning of an answer to the power transition question (odds are good it has a new enough lifecycle controller to not need that workaround).

Testing is a tough question for me to answer, because it's not clear to me specifically what the goals of your automated testing are in this area. Implementing some mocks heedlessly could result in just testing that the redfish_client gem is installed, which doesn't seem very useful. I am not opposed to implementing tests, but this is an area I am weak in and I would be glad for guidance. 

Give me a day or two to put together some changes that address the other comments, and get them through our internal code review process.

modules/bmc/redfish.rb
timogoebel: Well, if you have self-signed certificates you can also disable the encryption. That's easier. :-)

Sorry for being a smart-ass here. I know most folks don't use signed certificates for BMC interfaces. But that's so important for security.

xray-sky: Not at all, you guys are absolutely right that this shouldn't be default disable, even if I suspect in most deployments it will get disabled.

ekohl:
Note to self: https://github.com/theforeman/puppet-foreman_proxy/blob/d2f07f4a7e460231015f91469f298eb6831c7479/manifests/init.pp#L429

lzap:
I am looking into some options of finding some available hardware internally so we can test your patches. We don't like merging untested stuff :-)

Ideally we would merge this after 2.1 is branched off (two weeks), this should give us more time to prepare for what's coming. I expect many customers will be asking about this feature.

lzap:
@ekohl FYI I found one available ProLiant DL60 Gen9 I can play with today. Going to try this now.

lzap:
@ohadlevy do you remember why we have this "heuristic" in our code: https://community.theforeman.org/t/associating-bmc-proxies/18599

modules/bmc/redfish/dell.rb
xray-sky: Power state changes aren't exposed as asynchronous redfish calls (at least not on Dell), so we can't use that. It also turns out the Dells are more restrictive about how they accept power state change actions and they weren't working at all the way I had it coded, so it's a good thing I went back to explicitly test.

modules/bmc/bmc.rb
xray-sky: changed.

modules/bmc/redfish.rb
xray-sky: added as config item, default enabled.

xray-sky:
I don't know "why" from your perspective, but I do know that we are waiting eagerly for it, as latency on these BMC calls has a big impact on web UI performance for hosts pages, and we have many very geographically distant subnets, plus ones that aren't directly routable from our production Foreman instance. Subnet associations take care of this for our other proxies (dns, dhcp, tftp). My next project after redfish was to be to volunteer in whatever way would help get bmc proxy subnet associations implemented.

lzap:
It looks like I have already worked on making BMC proxy to be configurable by users: https://github.com/theforeman/foreman/pull/7208 so I will try to get that into develop.

ekohl:
We have capabilities in the Smart Proxy now. It would be nice if the available IPMI providers would be exposed as such. That way the Foreman can correctly list them without calling out to the proxy.

domitea:
FYI, there is emulator (https://github.com/DMTF/Redfish-Interface-Emulator) by DMTF of Redfish. Maybe it can help with testing.

xray-sky:
I see the 2.1-stable branch has been created. May I please have a summary of where the maintainers think we're at with this PR? I'm having a hard time following the various threads and being sure of what is still an actual outstanding request to me for further change, vs. thinking out loud.

lzap:
Apologies, I've been super busy with 2.1 release. This is definitely on my radar, I've secured a credentials for some testing servers with Redfish and all I need is to find time to test and review your contributions. I will do my best to get to that as soon as 2.1 comes out.

xray-sky:
No problem, I just wanted to make sure you guys weren't waiting on something from me.

lzap:
Comment removed - irrelevant.

lzap:
Oh, ignore my last comment it was supposed to go somewhere else.

I am currently looking into tests for BMC patch https://github.com/theforeman/foreman/pull/7208 which I need to be able to easily configure Redfish. Then I will start looking into this.

modules/bmc/bmc.rb
```diff
@@ -4,19 +4,19 @@ module Proxy
...
     end
 
     def providers
-      Proxy::BMC::IPMI.providers + ['shell']
+      Proxy::BMC::IPMI.providers + ['redfish', 'shell']
     end
 
     def installed?(provider)
-      if provider == 'shell'
+      if provider == 'redfish' || provider == 'shell'
```
adamruzicka: This whole method could be reduced to
```ruby
%w(redfish shell).include?(provider) || Proxy::BMC::IPMI.installed?(provider)
```

modules/bmc/redfish.rb
```diff
@@ -0,0 +1,210 @@
...
+        # enforce SSL verification as a default, if the value isn't set in the config
+        # file, or is set to a "garbage" value (anything but boolean false)
+        @redfish_verify_ssl = true unless @redfish_verify_ssl == false
+        super
+        load_vendor_overrides
+        @bmc
+      end
+
+      def connect(args = { })
+        # TODO probably verify should be an option.
```
adamruzicka: I think this was already done

modules/bmc/redfish.rb
```diff
@@ -0,0 +1,210 @@
...
+  module BMC
+    class Redfish < Base
+      include Proxy::Log
+      include Proxy::Util
+
+      def initialize(args)
+        @redfish_verify_ssl = Proxy::BMC::Plugin.settings.bmc_redfish_verify_ssl
+        # enforce SSL verification as a default, if the value isn't set in the config
+        # file, or is set to a "garbage" value (anything but boolean false)
+        @redfish_verify_ssl = true unless @redfish_verify_ssl == false
```
adamruzicka: This feels odd, if you really want true/false here, wouldn't it be better to crash hard rather than silently ignore a "wrong" value and use a different one instead?

modules/bmc/redfish.rb
```diff
@@ -0,0 +1,210 @@
...
+        super
+        load_vendor_overrides
+        @bmc
```
adamruzicka: Why?

modules/bmc/redfish.rb
```diff
@@ -0,0 +1,210 @@
...
+      end
+
+      def connect(args = { })
+        # TODO probably verify should be an option.
+        connection = RedfishClient.new("https://#{args[:host]}/", verify: @redfish_verify_ssl)
+        connection.login(args[:username], args[:password])
+        connection
+      end
+
+      def load_vendor_overrides
```
adamruzicka, line 24: Wouldn't it be cleaner to somewhere have a hash mapping manufacturer names to modules?

modules/bmc/redfish.rb
```diff
@@ -0,0 +1,210 @@
...
+        # pretty low-effort try, here. sufficient, though?
+        begin
+          host.Managers.Members.any?
+        rescue NoMethodError
+          false
+        end
+      end
+
+      def identifystatus
+        host.Systems.Members[0].IndicatorLED.downcase
```
adamruzicka: `host.Systems.Members[0]` is repeated here a lot, maybe it would be worth extracting to a separate method?

adamruzicka (commented):
I don't have any hardware to test this, so I just left a couple of comments about the code itself

lzap:
I can test, the other PR unit tests are finally passing locally had some issues with taxnomy: https://github.com/theforeman/foreman/pull/7208

bundler.d/bmc.rb
```diff
@@ -1,3 +1,4 @@
 group :bmc do
   gem 'rubyipmi', '>= 0.10.0'
+  gem 'redfish_client', '>= 0.5.1'
```
lzap, line 3: Note: We need a RPM/DEB packages for the dependency.

modules/bmc/bmc_api.rb
```diff
@@ -412,6 +413,14 @@ def bmc_setup
           args = { :host     => params[:host], :username     => username, :options => body_parameters['options'],
                    :password => password, :bmc_provider => provider_type }
           @bmc = Proxy::BMC::IPMI.new(args)
+        when 'redfish'
+          log_halt 401, "unauthorized" unless auth.provided?
+          log_halt 401, "bad_authentication_request, credentials are not in auth.basic format" unless auth.basic?
+          username, password = auth.credentials
+
+          args = { :host     => params[:host], :username     => username, :options => body_parameters['options'],
+                   :password => password }
```
lzap: The formatting is ugly, feel free to reformat this and even above.

modules/bmc/redfish.rb
lzap: Definitely, this is very hacky :-)

modules/bmc/redfish.rb
```diff
@@ -0,0 +1,210 @@
...
+      end
+
+      def load_vendor_overrides
+        mfr = manufacturer
+        return unless mfr
+        mfr.tr!('^A-Za-z', '')
+        begin
+          mod = Kernel.const_get("RedfishVendorOverrides#{mfr}".to_sym)
+        rescue NameError
+          # no extensions for this vendor
```
lzap: A debug line indicating that this vendor does not have any overrides would be appropriate, only visible in debug mode.

modules/bmc/redfish.rb
```diff
@@ -0,0 +1,210 @@
...
+        self.extend mod
+      end
+
+      def manufacturer
+        host.Systems.Members[0].Manufacturer
+      end
+
+      # returns boolean if the test is successful
+      def test
+        # pretty low-effort try, here. sufficient, though?
```
lzap: No idea, I will test this so remove the comment please :-)

modules/bmc/redfish.rb
lzap: Also check for the boundaries in that method and error out nicely.

modules/bmc/redfish.rb
```diff
@@ -0,0 +1,210 @@
...
+        #
+        # Dell gives a list like [ None, Floppy, Cd, Hdd, SDCard, Utilities,
+        #                          BiosSetup, Pxe ]
+        # HPE -        [ None, Floppy, Cd, Hdd, Usb, Utilities, BiosSetup, Pxe ]
+        # SuperMicro - [ None, Pxe, Hdd, Diags, Cd, BiosSetup, FloppyRemovableMedia,
+        #                UsbKey, UsbHdd, UsbFloppy, UsbCd, UefiCd, UefiHdd, ... ]
+        #
+        # For the four devices Foreman hardcodes, lucky for us all three use the same set
+        # of values.
+        [ 'pxe', 'disk', 'bios', 'cdrom' ]
```
lzap: I will only be testing Dell, so we trust you for this one. :-)

modules/bmc/redfish.rb
```diff
@@ -0,0 +1,210 @@
...
+        bootdevice = { :device => 'disk', :reboot => reboot, :persistent => persistent }
+      end
+
+      def bootbios(reboot=false, persistent=false)
+        bootdevice = { :device => 'bios', :reboot => reboot, :persistent => persistent }
+      end
+
+      def bootcdrom(reboot=false, persistent=false)
+        bootdevice = { :device => 'cdrom', :reboot => reboot, :persistent => persistent }
+      end
```
lzap, line 150: Are these good candidates to be moved to the Base class?

I don't like method names, these are "side effect" methods which should be named with bang e.g. `bootpxe!` but I see almost zero value in these, does not make the code much shorter.

modules/bmc/redfish.rb
```diff
@@ -0,0 +1,210 @@
...
+
+      def ip
+        host.Managers.Members[0].EthernetInterfaces.Members[0].IPv4Addresses.first['Address']
```
lzap: This chain (and the others below) needs more testing for nil values. You can use `try` Ruby method I think since we upgraded to Ruby 2.6 already.

modules/bmc/redfish/hpe.rb
```diff
@@ -0,0 +1,13 @@
...
+
+  def poweroff(soft=false)
+    poweraction( soft ? 'PushPowerButton' : 'ForceOff' )
+  end
+
+  # -- See note about dynamic bootdevices in redfish.rb --
+  #def bootdevices
+  #  # older HPs have the list of supported values at a different location
+  #  host.Systems.Members[0].Boot.BootSourceOverrideSupported || super
+  #end
```
lzap: Do you want to delete this?

lzap (changes requested):
Please incorporate changes and I can start testing this.

modules/bmc/redfish.rb
xray-sky: I was thinking about the fact that there are many implementations of redfish, with many variations, and I could only test a small number of them. It seemed like it might be valuable to give folks a way to provide their own overrides, either to maintain locally, or as a first step toward contributing them, without necessarily needing to hack up the "official" part of the code. If you don't think it's valuable, or prefer something else, I'll change it.

xray-sky: Oops, I ran out of time to edit - the overrides may not look like much now, but that has a lot to do with luck ultimately deriving from the limited scope of what's implemented today in the proxy (I only implemented what the IPMI provider currently offers, though Redfish could realistically provide much more). If the feature list starts to grow beyond what's already here, even a little, there starts to be much more divergence between even just these four vendor implementations.

bundler.d/bmc.rb
xray-sky: Is this a note for me as the submitter? Or a note for you as the maintainer?

modules/bmc/redfish.rb
xray-sky: Are you asking me that? My hastily formed opinion is, maybe not for the base class, since the ssh and shell methods don't (can't) provide them. I have tried to avoid doing much more than implementing Redfish as if it were effectively another IPMI provider, and trying not to make too many philosophical decisions about how I think existing code "should" work instead. I've deliberately preferred to work as much as possible with what's already present, thinking (perhaps naively) that it would help keep the focus on getting Redfish support added in, and not drift into philosophical debates about coding convention or quality -- my opinion on that has been, you guys are the maintainers, it's your call and not mine. I should be clear, too, I'm not complaining about how anything has gone so far!

Regarding the method names and the value of them, the API code calls these methods. I don't have anything to say if you want to change the way that all works. But would involve a more invasive change.

modules/bmc/redfish/hpe.rb
xray-sky: It's not used currently, but I was leaving it as tested code (or at least I tested it) for the day when boot devices might become more dynamic than the current four hardcoded selections. The other vendors (and recent enough HPs) didn't need anything special to retrieve the list of possible boot devices. I'll delete it if you would prefer that to leaving the comment. Or I can add some more comments describing what I just said.

xray-sky:
I'm working on an updated patchset to address the comments. I also noticed that the API exposes a fair bit more functionality of the IPMI provider than Foreman makes use of. I'll look into extending the Redfish implementation to cover those methods, as well, since the documentation provides examples of interacting directly with the BMC proxy API to get at those extra IPMI methods.

modules/bmc/redfish.rb
xray-sky: GOOD. QUESTION. I think the best I can say is "I was confused," or possibly "I was thinking of an unrelated thing."

modules/bmc/redfish.rb
xray-sky: I guess since I have to require the vendor extensions anyway, "without necessarily needing to hack up the official part of the code" isn't a coherent idea. I'll think a little harder about it. If you want to think out loud about it here, too, I wouldn't mind the collaboration.

xray-sky:
didn't notice I'd pushed a merge conflict in modules/bmc/bmc_api.rb -- fixed

bundler.d/bmc.rb
lzap: You are free to do it if you want to, it should be just a single command, see the README: https://github.com/theforeman/foreman-packaging

Most if not all dependencies we already do carry: https://rubygems.org/gems/redfish_client

modules/bmc/redfish.rb
```diff
@@ -0,0 +1,235 @@
...
+    class Redfish < Base
+      include Proxy::Log
+      include Proxy::Util
+
+      def initialize(args)
+        @redfish_verify_ssl = Proxy::BMC::Plugin.settings.bmc_redfish_verify_ssl
+        # enforce SSL verification as a default, if the value isn't set in the config
+        # and bail if it's set to something nonsensical
+        @redfish_verify_ssl = true if @redfish_verify_ssl.nil?
+        raise ::Proxy::Error::ConfigurationError.new("bmc_redfish_verify_ssl must be boolean (true/false)") unless [true, false].include?(@redfish_verify_ssl)
```
lzap: We actually have a mechanism now to provide default values to settings via `default_setting` method, see `modules/dhcp/dhcp_plugin.rb` as an example. But there is no validation API.

modules/bmc/redfish.rb
```diff
@@ -0,0 +1,235 @@
...
+        super
+        load_vendor_overrides
+      end
+
+      def connect(args = { })
+        connection = RedfishClient.new("https://#{args[:host]}/", verify: @redfish_verify_ssl)
+        connection.login(args[:username], args[:password])
+        connection
```
lzap, line 21: You could memorize the connection, it's a Ruby pattern that is quite common.

modules/bmc/redfish.rb
lzap: The point is, instead of playing with strings and replacing/tring etc you could have just a simple map which is safer. Granted if the implementation adds a new vendor the map would probably need an update too but in regard to readability and sustainability it's probably better.

modules/bmc/redfish.rb
lzap: I was just thinking out loud. True, these are not implemented SSH fake implementation.

The other comment was just a nitpicking - it's a common practice in Ruby to end methods with `bang!` characters if they only have a side effect and does not return a relevant value. Feel free to ignore this.

modules/bmc/redfish/hpe.rb
lzap: Feel free to leave it, I just thought it's some leftover.

lzap:
Ok this looks good, no major comments.

lzap:
Can you make Rubocop happy? https://ci.theforeman.org/job/test_proxy_develop_pr_rubocop/290/ruby=2.5/console

modules/bmc/redfish.rb
xray-sky: I had originally thought I would do just that, but as it happens (assuming I've understood the rest of the code correctly, which maybe I haven't) it wouldn't serve any purpose -- every call to the proxy instantiates a new BMC object, for which connect is only called once. It looks to me like an extremely invasive change.

The redfish gem itself does its own memoizing, and also provides some caching. It keeps its auth token with the object, and will not perform a subsequent remote call to the BMC to reference data it's already seen. I ran into some problems with the cache needing flushed manually after calls that would update BMC data (e.g. changing the one-time boot device), but given the way the Proxy instantiates BMC objects it wasn't necessary to account for cache flushing in the plugin (a new BMC object with a new connection and an empty cache will be created for the next BMC Proxy request).

I think the main incentive for doing a bigger change to reuse the BMC objects across multiple Proxy calls would be to reduce the impact of network latency on querying this data, as currently there are quite a few round trips needed to query all the data needed to, for example, render a Host page in the Foreman UI. Any non-trivial latency is immediately obvious as it can take 30-40 seconds for all that data to return for some of our most distant subnets. But the work that's being done to put subnet associations on the BMC Proxy ought to take care of maybe 90% of that, and my sense is that's probably good enough given the effort that would need to go into redoing the app and the proxy to support longer lived BMC connections.

On the other hand I could easily be wrong about any or all of this. I don't mind finding out.

xray-sky:
> Can you make Rubocop happy? https://ci.theforeman.org/job/test_proxy_develop_pr_rubocop/290/ruby=2.5/console

Gladly. Done. Unless something new occurs.

modules/bmc/redfish.rb
```diff
@@ -0,0 +1,232 @@
...
+      def load_vendor_overrides
+        mod = case manufacturer
+              when 'Dell Inc.' then 'RedfishVendorOverridesDellInc'
+              when 'HPE'       then 'RedfishVendorOverridesHPE'
+              else
+                logger.debug "No #{manufacturer} specific overrides available - using generic Redfish calls"
+                return
+              end
+        logger.debug "Extending Redfish with vendor overrides for #{manufacturer}"
+        extend Kernel.const_get(mod.to_sym)
```
lzap: Okay this is called on every request, we want this only to be done once during initialization.

modules/bmc/redfish.rb
lzap: Fair enough with the memorizing. Yeah I have a oversea HP server I am testing against and it is slow as hell.

lzap (changes requested):
It works but please explain or change the extending. We should not be extending on every request.

bundler.d/bmc.rb
lzap: https://github.com/theforeman/foreman-packaging/pull/5720

modules/bmc/redfish.rb
xray-sky: If it's done once during initialization, the provider can only ever work with a single vendor's redfish implementation, which defeats the point. We have systems from four different vendors in production, and some BMCs behave differently even within a single vendor, owing to function changes between versions. It has to happen at runtime, based on the specific Redfish implementation that BMC proxy instance is communicating with.

Unless you're advocating for an arrangement where there are case statements in every method to support every possible vendor? I don't consider that an improvement in any way, but if that's the will of the maintainers, that's what I'll do. Or... something else I don't understand yet?

xray-sky: Actually.. is it the extend specifically? I just recently learned I don't quiiite understand extend the way I thought I did, so there may be side effects I don't understand. But my intent was to allow some methods to be overridden at runtime based on data coming back from the specific BMC, and (secondarily) to allow the use of super to simplify those overrides (since in many cases it'll just be static data that needs changing, either values passed to an endpoint, or the location of an endpoint; the interaction with an endpoint otherwise being the same). And extend does this (even if it doesn't do it quite the way I thought it did). If include or prepend can do this, great, but I might need a little help understanding how to make that change. If that's where you were going.

xray-sky: I did an experiment to check some of the assumptions I'd had about extend's outward behavior and... although I still don't quite understand the way it works, it does at least behave the way I'd assumed - extending gets the module's methods available in that single instance. I had read (and I am probably butchering this, since I don't yet understand it, etc.) that extending worked on the _class_'s singleton object -- that had me worried that perhaps the methods were becoming available to all instances, and the Class itself would be dragged back and forth between one vendor or another's overrides. Anyway, that doesn't seem to be the case. For whatever that's worth.

adamruzicka: It took me a bit to wrap my head around what's going on and I must say I'm impressed, this is genius. A bit on the crazy side, but genius nevertheless.

I would still prefer having it done the boring way with some kind of adapters or something, but this should work.

xray-sky: well, I just noticed I missed a trick changing it from the dynamic lookup - no need to assign a string to mod, then convert it to symbol. I could have just assigned the symbol directly to mod.

I wouldn't mind learning what you mean by "the boring way with some kind of adapters or something", if you have the inclination to explain it. Just for my own growth.

adamruzicka: Something along the lines of this ~~https://github.com/theforeman/smart-proxy/commit/655a8c711c005ecfd024f0b39b894420f09a78db~~ https://github.com/theforeman/smart-proxy/commit/9889bc5f640d273a20c1c30f74f788a1fc5ac9aa . It is a bit more explicit than what you have here. And also untested :)

xray-sky: I see. The cost of doing it that way is approximately doubling the number of round trips to the BMC in the case one wants overrides -- one must connect, authenticate, and query the manufacturer, before creating the new object which must connect, authenticate, etc. (Unless, I suppose, one passes a reference to the original session to the new object.) 

What are the benefits?

adamruzicka: The connection is indeed passed [here](https://github.com/theforeman/smart-proxy/commit/9889bc5f640d273a20c1c30f74f788a1fc5ac9aa#diff-3b223db7ff1577c5ae9efd3ac7a328c6R32) so the it gets opened only once.

I'd say the benefits are that the code is more explicit about what is going on and one doesn't need to know advanced ruby stuff like that extend when called within an instance extends the instance's eigenclass.

xray-sky: I missed that. Actually I will need to look closer at this because you've made redfish_adapter a class method, which then apparently has the somewhat perverse side effect of returning an instance of itself. Believe it or not this is more confusing to me! But I guess it could be thought of as a somewhat pathological angle on new(), which makes it seem less threatening. Is this what Java and Python people refer to as a "factory" pattern?

xray-sky: I guess what I'm driving at, is... I think there's an element in both of these directions that effectively boils down to, "this is what's familiar to me." I agree that there's value in doing things that are understood, and in that case your preference as a maintainer I think legitimately carries more weight. To counter that, all I have to say is something like... I think it's fair to say Foreman is pretty solidly committed to ruby. Why avoid taking advantage of legit features of the language in favor of something that feels to me more like a Java pattern? It's admittedly not much of an argument. If it's your collective will that I change it, I will; however reluctantly.

My motivations are entirely selfish -- I want this in the next release of Foreman. We desperately need it. (@;

adamruzicka: > Is this what Java and Python people refer to as a "factory" pattern?

Most likely yes, but I must admit I'm a bit rusty with design patterns :)

I'd be fine with your way if you just added a comment briefly explaining why it works the way it does so it is not so surprising. Not sure how others look at this though

xray-sky: It'd be simple for me to add a comment. So let me do that (and fix the assignment thing I mentioned) while you guys come to a consensus on keeping it or not. If I may ask -- could you describe what you find surprising about the use of extend? It would help me write a more useful comment.

modules/bmc/redfish.rb
xray-sky: now with 100% more default_setting

modules/bmc/redfish.rb
xray-sky: Took a stab at it; focused on concrete effects rather than anything philosophical. Let me know if it misses the mark?

modules/bmc/redfish.rb
```diff
@@ -0,0 +1,239 @@
...
+
+module Proxy
+  module BMC
+    class Redfish < Base
+      include Proxy::Log
+      include Proxy::Util
+
+      def initialize(args)
+        @redfish_verify_ssl = Proxy::BMC::Plugin.settings.redfish_verify_ssl
+        raise ::Proxy::Error::ConfigurationError.new("redfish_verify_ssl must be boolean (true/false)") unless [true, false].include?(@redfish_verify_ssl)
```
ekohl: It would be cleaner to do this in a validator. In 8b27dabfcbcd6a1e8873c7f75b90a253aeb27bec I added the URL validator. You can implement a generic bool test in the same way. It also had an example of a custom validator that you can use. The benefit is that it's validated on application start up rather than on the first API request.

I would recommend to write a test for it because it's too easy to mess up otherwise.

modules/bmc/redfish.rb
lzap: I am reading this right now, sorry for the delay. Let me review what I thought when I made that short comment :-)

lzap: Allright, my first quick impression is this code smells and there's the traditional way of workarounding it, for example the one that Adam described. Since you insist on the Ruby way, and it's totally fine here this indeed is a Ruby project, I had to drill down the codebase to confirm some things. One of my biggest concern was - what if we keep the Redfish instance in memory? Then we will be extending the singleton class on every request until the call chain is long AF practically slowing down every processing by a bit. That's not the case, we create new instance on every request, at least this is what I expect from Sinatra/Rack contract. Note if this is fine today, this might change in the future creating terrible bug that will cost a lot of money to track it down (customers complaining that the service slows down over time, memory consumption etc).

My second concern is performance, this was just an impression without raw numbers but quick google finds that creating an object and then extending [it is nearly 4 times as slow](https://stackoverflow.com/questions/8616275/how-costly-is-extend-in-ruby) as only creating the object. Probably not the big issue for Redfish where this request will spend 99% in waiting for I/O, but for me it's a flag raised.

And finally, I am ex-Java guy - Hello! :-) - and I still find the traditional design patterns easier to understand for an average software engineer. You need to understand who is gonna maintain the code - we are most commonly people with some traditional programming background (java, c/c++, python etc) than native-rubyists. It's not the language what defines how things should be written, it's the project and people around it.

Having said that, I was ultra slow responding to your PR, this is a feature we want and if I am not mistaken there are no known bugs other than the minor performance concern. If you can confirm @adamruzicka that this extension approach is correct, let's merge this.

lzap (approved):
This is ready to go, you can fix the boolean setting while I will be working with Ewoud on packaging side of this. It has a dependency so we need to wait until it's in nightlies until I can finally propose the gem redfish_client.

modules/bmc/redfish.rb
adamruzicka: :+1: from me

lzap:
Just a quick reminder - please fix the setting and we are good to go. Packaging PR (1 of 2): https://github.com/theforeman/foreman-packaging/pull/5720

xray-sky:
I believe I have addressed the setting validator comments.

Would you like me to rebase this down to a single commit, with a correct commit message? I have avoided doing it because I haven't been entirely sure of the expected workflow here.

lzap:
Please go ahead ans quash into one commit, I had a delay on the packaging work I pushed accidentally and closed the PR, but we are all good now. As soon as packages are ready we are merging this, thank you!

xray-sky:
Opened Smart Proxy feature #30906 for Redmine test.

lzap:
The dependency RPM package is finally in, here is the other one: https://github.com/theforeman/foreman-packaging/pull/5847

lzap:
@ekohl any other comments? I am good.

ekohl:
I haven't seen a PR for Debian packaging. Not sure if the gem is already in Debian and Ubuntu (and new enough).

lzap:
> I haven't seen a PR for Debian packaging. Not sure if the gem is already in Debian and Ubuntu (and new enough).

Aren't these just regular Rubygem dependencies which we do "bundle install"?

lzap:
I am merging this, this is just a regular gem dependency, just like `gssapi` in `bundler.d/krb5.rb`. Thanks.

ekohl:
> Aren't these just regular Rubygem dependencies which we do "bundle install"?

No, smart-proxy on Debian requires native packages. By merging this you probably broke the BMC module there. We don't see this in our pipelines because there the module is not enabled.

lzap:
Can you elaborate what needs to be done and I will do it. However I don't understand why I could not find `gssapi` in deb/packaging at all: this looks limilar.

ekohl:
Here it recommends ruby-rkerberos:
https://github.com/theforeman/foreman-packaging/blob/a1da82fdfaeebe72a1879eb52be65d9b07945db5/debian/buster/foreman-proxy/control#L16

This is then packaged here:
https://github.com/theforeman/foreman-packaging/tree/deb/develop/dependencies/buster/rkerberos

Links are for Debian Buster, but we do exactly the same for Ubuntu Bionic.

lzap:
Oh now I understand, the gssapi is actually in Debian itself! Damn, redfish is not. I am too used to think in RHEL context when we already ship 4k packages (while Debian is like 20k).

So I need to create a new Debian package and carry that in our own repos? Do I put this into `dependencies/` directory?

ekohl:
Exactly. I must admit that my Debian knowledge is limited. There is a tool called gem2deb that should at least provide a solid base. Perhaps @theforeman/packaging could assist.

evgeni:
Essentially, yes. You'll have to package redfish_client (and  server_sent_events) into `dependencies` and then make `foreman-proxy` depend on it.

And yeah, my Debian knowledge in this area is also limited.

lzap:
So, can you guide me @mmoll ? I created a debian10 VM, installed required software and executed the script. It successfully created a package and some artifacts, but I don't know what to do next. README is sparse on dependencies:

```
root@debian10:~/foreman-packaging/dependencies# git status
On branch deb/develop
Your branch is up to date with 'origin/deb/develop'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)

	ruby-server-sent-events-0.1.2/
	ruby-server-sent-events_0.1.2-1.debian.tar.xz
	ruby-server-sent-events_0.1.2-1.dsc
	ruby-server-sent-events_0.1.2-1_all.deb
	ruby-server-sent-events_0.1.2-1_amd64.buildinfo
	ruby-server-sent-events_0.1.2-1_amd64.changes
	ruby-server-sent-events_0.1.2.orig.tar.gz
	server-sent-events-0.1.2.tar.gz
	server_sent_events-0.1.2.gem
```

lzap:
https://github.com/theforeman/foreman-packaging/pull/5912

lzap:
Hello, I am a bit stuck since our debian packager maintainer left the project and the infra is a bit of a black-box for me. Will do my best to solve this.

PR #760, opened 2020-04-28
Fixes #29648 - Move Windows service registration script

mmoll:


mmoll:
Seems like this file is not referenced specifically in RPMS/DEBs, so we can just move it as the last one remaining. :)

ekohl (approved):
This is indeed only relevant on Windows, but maybe the commit message or issue should state that? Right now as a Windows user I wouldn't bother to look at this while it may actually be useful to know. Other than that :+1:

mmoll:
`Rename script for Windows service registration` or similar for the changelog? It's also mentioned in the docs repo, which needs then updating, too.

ekohl:
Yes, I think that'd be more useful.

PR #762, opened 2020-08-20
Fixes #30429 - Remove puppetrun endpoints

shiramax:


modules/puppet_proxy/puppet_api.rb
```diff
@@ -7,14 +7,6 @@ class Proxy::Puppet::Api < ::Sinatra::Base
 
   inject_attr :class_retriever_impl, :class_retriever
   inject_attr :environment_retriever_impl, :environment_retriever
-  inject_attr :puppet_runner_impl, :puppet_runner
-
-  post "/run" do
```
ekohl, line 12: Thinking out loud: is it a good idea to maintain the endpoint but return an explicit HTTP 501 Not Implemented?

config/settings.d/puppet.yml.example
```diff
@@ -1,10 +0,0 @@
----
```
ekohl, line 1: This file needs to stay since it's still used for the Puppet module. You just need to drop the provider part. In 3eaa9cc26ddcdb2c5dfd3f3b88447c661fc6da31 I moved the Puppet HTTP API provider into the module itself. This is a leftover from when there was a legacy provider. 7e7015aba068db541bd4f4cf1856bf69ebbe2953 removed that.

ekohl (changes requested):
You can also drop `modules/puppet_proxy_common/runner.rb` and its tests.

You also need drop the provider loading:

https://github.com/theforeman/smart-proxy/blob/a644146e800fbffeb7a8d899bf5122ba6d63bb9d/modules/puppet_proxy/configuration_loader.rb#L4-L5

modules/puppet_proxy/puppet_api.rb
ares: For how long do you think we should keep it? Major version bump?

ekohl: I don't know. Maybe it doesn't make sense - if we initially had the Not Implemented as potentially Not Configured that meant Foreman could show a proper error. However, keeping it and removing it on the next major might make sense. This is easier to distinguish than a 404, which may also happen if you're talking to a wrong server.

ekohl (changes requested):
Looks like you referenced the incorrect Redmine issue.

I also still see `modules/puppet_proxy_common/runner.rb` but that class is now unused. Also means you can drop `test/puppet/runner_test.rb`.

shiramax:
@ekohl done, thanks!

modules/puppet_proxy/configuration_loader.rb
```diff
@@ -1,8 +1,7 @@
 module ::Proxy::Puppet
   class ConfigurationLoader
     def load_programmable_settings(settings)
-      use_provider = settings[:use_provider]
-      use_provider = [use_provider].compact unless use_provider.is_a?(Array)
+      use_provider = []
       use_provider << :puppet_proxy_puppet_api
       settings[:use_provider] = use_provider
```
ekohl: Should we simplify this:
```suggestion
      settings[:use_provider] = [:puppet_proxy_puppet_api]
```

ekohl (commented):
This code now also looks unused:
https://github.com/theforeman/smart-proxy/blob/a644146e800fbffeb7a8d899bf5122ba6d63bb9d/test/puppet/puppet_api_test.rb#L60-L66

Other than that I think this is good

modules/puppet_proxy/configuration_loader.rb
shiramax: yes. make sense, thanks @ekohl

ekohl:
Could you also remove `ApiTestPuppetRunner`?

modules/puppet_proxy/puppet_api.rb
ares: makes sense, how do we make sure we drop this in next major release? do we have redmine issue for that? do we have deprecation messages similar to core?

ekohl: I think we need some proper maintenance on the Smart Proxy project. There are many things we deprecated, but never removed. ee1dcd16bdba277294cb9c744b667f911e1ea7c7 for example stated the methods would be removed in 1.24 but haven't to this day. 5427599db2fbf258d7513260e852ad44ee41ef2c (part of 1.23) removed methods deprecated in 1.14 / 1.16. Then we have the `LegacyModuleLoader` which is still used by 2 modules without a clear plan to remove them.

In this case it's an external API. Keeping it around makes it easier to debug. Its usefulness is gone when there's no real chance a Foreman with Puppetrun functionality has a Smart Proxy. While in many cases we say Foreman should be newer than the Proxy, in reality I still think Dmitri was right that just Foreman <-> Foreman Proxy it's much easier to say that Foreman Proxy should be upgraded first. It's just that with Pulp that may not be true.

In short: I'd keep this around for at least 2 releases but probably 4 just to be safe. Or wait for the next major version.

shiramax:
@ekohl done:) thanks for the review!

modules/puppet_proxy/puppet_api.rb
```diff
@@ -7,13 +7,9 @@ class Proxy::Puppet::Api < ::Sinatra::Base
...
   inject_attr :class_retriever_impl, :class_retriever
   inject_attr :environment_retriever_impl, :environment_retriever
-  inject_attr :puppet_runner_impl, :puppet_runner
 
   post "/run" do
-    log_halt 400, "Failed puppet run: No nodes defined" unless params[:nodes]
-    log_halt 500, "Failed puppet run: Check Log files" unless puppet_runner.run([params[:nodes]].flatten)
-  rescue => e
-    log_halt 500, "Failed puppet run: #{e}"
+    raise NotImplementedError.new
```
ekohl: Isn't this just a regular exception? Does it return a HTTP 500 or 501? Perhaps a test would be nice. Also a message that puppetrun support has been removed in version 2.3. is nice for the consumer for the API.

shiramax: @ekohl done.  thanks!

ekohl (approved):
From reviewing the code this should be good. This should also be reflected in https://github.com/theforeman/puppet-foreman_proxy but that doesn't have to block merging.

ekohl:
Thanks!

PR #763, opened 2020-08-24
Fixes #30677 - Global registration endpoint

stejskalleos:
Allows to render Global registration template.

How to test:
```
curl -X GET http://smart-proxy.devv/register --user admin:changeme
```

Notes:
* PR is part of **Simple & automatic host registration WF**, [RFC](https://community.theforeman.org/t/rfc-simple-automatic-host-registration-wf/19588) & [Redmine](https://projects.theforeman.org/issues/30440)

### DIFF
--- a/modules/templates/templates_register_api.rb
+++ b/modules/templates/templates_register_api.rb
@@ -1,0 +1,11 @@
+require 'templates/proxy_request'
+
+class Proxy::TemplatesRegisterApi < Sinatra::Base
+  helpers ::Proxy::Helpers
+
+  get "/" do
+    log_halt(500, "Failed to retrieve Global registration template.") do
+      Proxy::Templates::ProxyRequest.new.get(['register'], request.env, params)
+    end
+  end
+end
--- a/test/templates/templates_register_api_test.rb
+++ b/test/templates/templates_register_api_test.rb
@@ -1,0 +1,26 @@
+require 'test_helper'
+require 'templates/templates_register_api'
+require 'templates/templates'
+require 'webmock/test_unit'
+
+class TemplatesRegisterApiTest < Test::Unit::TestCase
+  include Rack::Test::Methods
+
+  def app
+    Proxy::TemplatesRegisterApi.new
+  end
+
+  def setup
+    @foreman_url = 'http://foreman.example.com'
+    Proxy::SETTINGS.stubs(:foreman_url).returns(@foreman_url)
+    @template_url = 'http://smart-proxy.example.com'
+    Proxy::Templates::Plugin.settings.stubs(:template_url).returns(@template_url)
+  end
+
+  def test_api_global_registration
+    stub_request(:get, "#{@foreman_url}/register").with(query: {"url" => @template_url}).to_return(:body => 'GRT content')
+    get "/", {}
+    assert last_response.ok?, "Last response was ok"
+    assert_match("GRT content", last_response.body)
+  end
+end
--- a/modules/templates/http_config.ru
+++ b/modules/templates/http_config.ru
@@ -1,5 +1,6 @@
 require 'templates/templates_unattended_api'
 require 'templates/templates_userdata_api'
+require 'templates/templates_register_api'
 
 map "/unattended" do
   run Proxy::TemplatesUnattendedApi
@@ -7,4 +8,8 @@ end
 
 map "/userdata" do
   run Proxy::TemplatesUserdataApi
+end
+
+map "/register" do
+  run Proxy::TemplatesRegisterApi
 end
--- a/modules/templates/templates_plugin.rb
+++ b/modules/templates/templates_plugin.rb
@@ -7,6 +7,7 @@ module Proxy::Templates
 
     validate :template_url, url: true
     expose_setting :template_url
+    capability 'global_registration'
 
     after_activation do
       loading_failed "missing :foreman_url: from configuration." unless Proxy::SETTINGS.foreman_url
--- a/test/templates/integration_test.rb
+++ b/test/templates/integration_test.rb
@@ -21,7 +21,7 @@ class TemplatesApiFeaturesTest < Test::Unit::TestCase
     mod = response['templates']
     refute_nil(mod)
     assert_equal('running', mod['state'], Proxy::LogBuffer::Buffer.instance.info[:failed_modules][:templates])
-    assert_equal([], mod['capabilities'])
+    assert_equal(['global_registration'], mod['capabilities'])
 
     assert_equal({'template_url' => 'http://smart-proxy.example.com:8000'}, mod['settings'])
   end


ekohl (changes requested):
Could you expose a capability for this? d93261cc26c3e9ea435a1d649c77cca9e62af91c has an example. The benefit is that on the Foreman side you know for any registered Smart Proxy whether it actually support this. Then when a user forgot to upgrade a Smart Proxy, Foreman can show an error ahead of time rather than failing at runtime.

ekohl (commented):
Looks like there is test coverage for all other endpoinds. Mind adding the same for register?

test/templates/templates_register_api_test.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  def setup
+    @foreman_url = 'https://foreman.example.com'
+    Proxy::SETTINGS.stubs(:foreman_url).returns(@foreman_url)
+    @template_url = 'http://proxy.lan:8443'
+    Proxy::Templates::Plugin.settings.stubs(:template_url).returns(@template_url)
+    @args = { :token => "test-token" }
+  end
+
+  def test_api_can_ask_for_a_cloud_init_template
```
ekohl: This looks like a direct copy

test/templates/templates_register_api_test.rb
```diff
@@ -0,0 +1,28 @@
...
+    stub_request(:get, "#{@foreman_url}/register").with(query: {"url" => @template_url}).to_return(:body => 'GRT content')
```
ekohl, line 21: I didn't see the query part in https://github.com/theforeman/foreman/commit/57d95c1685f28e0a52bd40fc26dd66d24adc0288 so I'm wondering if there's a follow up PR that I'm missing.

test/templates/templates_register_api_test.rb
```diff
@@ -0,0 +1,28 @@
...
+  include Rack::Test::Methods
+
+  def app
+    Proxy::TemplatesRegisterApi.new
+  end
+
+  def setup
+    @foreman_url = 'https://foreman.example.com'
+    Proxy::SETTINGS.stubs(:foreman_url).returns(@foreman_url)
+    @template_url = 'http://proxy.lan:8443'
```
ekohl: I'm always in favor of using the RFC2606 domains: https://community.theforeman.org/t/using-ips-and-domains-in-tests-documentation/16600

test/templates/templates_register_api_test.rb
stejskalleos: Yes,
the PR for the :url is here: https://github.com/theforeman/foreman/pull/7928

test/templates/templates_register_api_test.rb
stejskalleos: I didn't know about it before that,
definitively will use it later, thx.

stejskalleos:
> Looks like there is test coverage for all other endpoinds. Mind adding the same for register?

Tests added.

ehelms:
I'm a bit of a novice on this plugin. What is the templates plugin doing generally speaking and in particular for this use case? This on the surface appears to be a reverse proxy back to the Foreman server.

stejskalleos:
> appears to be a reverse proxy back to the Foreman server.

Yes, that's it. And you can specify `:template_url:` in `settings.d/templates.yml`, so all the URLs in the template are linked to the smart-proxy (see https://github.com/theforeman/foreman/pull/7928)

test/templates/templates_register_api_test.rb
```diff
@@ -0,0 +1,28 @@
...
+  include Rack::Test::Methods
+
+  def app
+    Proxy::TemplatesRegisterApi.new
+  end
+
+  def setup
+    @foreman_url = 'https://foreman.example.com'
+    Proxy::SETTINGS.stubs(:foreman_url).returns(@foreman_url)
+    @template_url = 'https://foreman2.example.com'
```
ekohl: Perhaps this example is more realistic. Doesn't really matter for the actual result, but in the future it'll be easier to understand what's going on.
```suggestion
    @template_url = 'http://smart-proxy.example.com:8000'
```

Edit: yes, this is the typical configuration and the default what the installer provides. That's because provisioning setups like kickstart and preseed typically don't deal well with HTTPS. Either they don't support it at all or have no way to verify the SSL chain. Be aware of this.

test/templates/templates_register_api_test.rb
```diff
@@ -0,0 +1,28 @@
...
+    Proxy::Templates::Plugin.settings.stubs(:template_url).returns(@template_url)
+    @args = { :token => "test-token" }
```
ekohl: Where is this used?

stejskalleos:
@ekohl `templates_register_api_test.rb` updated.

test/templates/templates_register_api_test.rb
```diff
@@ -0,0 +1,27 @@
...
+  include Rack::Test::Methods
+
+  def app
+    Proxy::TemplatesRegisterApi.new
+  end
+
+  def setup
+    @foreman_url = 'http://foreman.example.com'
+    Proxy::SETTINGS.stubs(:foreman_url).returns(@foreman_url)
+    @template_url = 'http://foreman2.example.com:8000'
```
ekohl: I still think `foreman2` is a weird hostname for a smart proxy.

test/templates/templates_register_api_test.rb
```diff
@@ -0,0 +1,27 @@
+require 'test_helper'
+require 'json'
```
ekohl: This file doesn't use json, right?

test/templates/templates_register_api_test.rb
```diff
@@ -0,0 +1,26 @@
...
+  include Rack::Test::Methods
+
+  def app
+    Proxy::TemplatesRegisterApi.new
+  end
+
+  def setup
+    @foreman_url = 'http://foreman.example.com'
+    Proxy::SETTINGS.stubs(:foreman_url).returns(@foreman_url)
+    @template_url = 'http://foreman.example.com:8000'
```
ekohl: I wonder why you didn't use my suggestion. The reason I suggested it is that in a typical deployment you have a machine that runs Foreman on `foreman.example.com`. The Smart Proxy can run on the same machine, but can also run on another machine. My suggestion to use `smart-proxy.example.com` as a hostname was to clearly indicate that there doesn't have to be any relation between `foreman_url` and `template_url`.

This is a good practice to keep. If there is any code that parses URLs and rewrites, it will fail. That is a good thing. There's a reason you shouldn't use `@template_url = "#{foreman_url}:8000"`.

stejskalleos: I missed that somehow, changed to smart-proxy.example.com

ekohl (approved):
Thanks, LGTM.

@ehelms any more comments?

lzap (approved):

lzap:
> curl -X GET http://smart-proxy.devv/register --user admin:changeme

This does not appear to be correct address tho, it should be  `/unattended/register` if I read the code correctly.

ekohl:
> This does not appear to be correct address tho, it should be `/unattended/register` if I read the code correctly.

No, it's mapping `/register` to `Proxy::TemplatesRegisterApi`. That has a `get '/'` which means it's just `/register`.

lzap:
Yeah I just tested it, you are right. I am wondering if we want to make it 2nd level.

Anyway, it works fine. I also confirmed that Foreman throws 401 error when login/password is not present however I have one security concern - when HTTP plain is not provided AND hostname is in trusted_hosts, the request *could* be forwaded without authorization. I haven't tested this as on my dev setup I only have HTTP (and not HTTPS). @ekohl by chance do you have a dev setup with HTTPS?

Anyway, the code allows requests without HTTP authorization. None of the above would be a thread if we enforced it and throw 401 if login/pass is missing. What do you think?

modules/templates/templates_register_api.rb
```diff
@@ -0,0 +1,11 @@
+require 'templates/proxy_request'
+
+class Proxy::TemplatesRegisterApi < Sinatra::Base
+  helpers ::Proxy::Helpers
+
+  get "/" do
+    log_halt(500, "Failed to retrieve Global registration template.") do
+      Proxy::Templates::ProxyRequest.new.get(['register'], request.env, params)
```
ekohl, line 8: Btw, I hope you are aware that this sends client certificates from the Smart Proxy. That is an authentication channel. Right now I think you don't have `SmartProxyAuth` in the register controller, but if you do then passing a username and password becomes irrelevant because those client certificates can also authorize. It's quite easy to create a security vulnerability here.

lzap: > I hope

The worst possible thing when it comes to security :-) I do have the same concern and I think we should enforce HTTP auth. See my comment above.

ekohl: Well, HTTP auth that's then ignored by Foreman isn't really adding a lot.

ekohl:
> @ekohl by chance do you have a dev setup with HTTPS?

No, I don't.

ehelms:
I think I am being dumb here. What is the case where I would want to set the registration URL to be different than the hostname in the request ?

ekohl:
> I think I am being dumb here. What is the case where I would want to set the registration URL to be different than the hostname in the request ?

Multi homing is common in many complex environments. You would have a private lan. Foreman talks to Foreman Proxy on the infra lan. Then clients talk on the client lan.

So you would have:
* foreman.infra.example.com on 192.0.2.10
* foreman-proxy.infra.example.com on 192.0.2.11

Then on the client lan you have:
* foreman-proxy.client.example.com on 198.51.100.1
* some-machine.client.example.com on 198.51.100.20

Note foreman-proxy.infra.example.com and foreman-proxy.client.example.com are the same machine, just with multiple NICs.

The way you would configure this is that you register foreman-proxy.infra.example.com, but configure the template URL to http://foreman-proxy.client.example.com:8000.

This means you don't require any routing and it's all local lan traffic. Similarly, you can configure the TFTP module to listen on 198.51.100.1. This provides proper isolation: nothing from 198.51.100.0/24 can ever reach 192.0.2.0/24 except if the the foreman-proxy machine allows it.

Katello doesn't understand this concept at all, but it really should IMHO. Right now it ties service names to the system hostname. I started https://github.com/ekohl/smart_proxy_rhsm which would allow Katello to know the proper client URLs, but the Katello side is completely missing for now. That's probably the reason why the Satellite docs say such a scenario is unsupported. However, I think it also makes sense in a containerized world where you have private networks for internal communication and external routes.

ehelms:
As a client machine, why would I register to ` foreman-proxy.infra.example.com` if I am going to talk to the smart proxy via `foreman-proxy.client.example.com` ?

I admittedly don't know exactly what the Templates smart proxy plugin is intended for and doing in this case compared to a reverse proxy that would route the request to Foreman.

ekohl:
> As a client machine, why would I register to ` foreman-proxy.infra.example.com` if I am going to talk to the smart proxy via `foreman-proxy.client.example.com` ?

You wouldn't. However, that template URL setting allows Foreman and Foreman Proxy to **know** that. How else would it figure that out?

> I admittedly don't know exactly what the Templates smart proxy plugin is intended for and doing in this case compared to a reverse proxy that would route the request to Foreman.

Technically you can, however the templates plugin provides only those workflows clients would need. You don't need to expose your entire Foreman installation. If there's a remote exploit in Foreman, there's no way clients could exploit it. It's about encapsulating services and only exposing the minimal surface it needs to function. There is also client certificate authentication you so actually know which Smart Proxy you're talking to if needed.

Katello doesn't use this at all. For example, in the following code it's guessed from untrusted headers:

https://github.com/Katello/katello/blob/9dc14ebc5e8d5351b15e1d9c8bb82855b7964bbd/app/controllers/katello/api/rhsm/candlepin_proxies_controller.rb#L270-L274

Any client can send these headers and pretend it has a parent host. I don't know if this can actually be exploited, but it's a bad practice.

It would be better to verify if this actually came from a trusted IP. For example, this is how the unattended controller verifies it:

https://github.com/theforeman/foreman/blob/e46a838b004cdbf9042bf1c0122d3fba409eb5c5/app/controllers/concerns/foreman/controller/ip_from_request_env.rb#L8-L15

This is a common pattern when it comes to dealing with reverse proxies. Ideally we would actually use [ActionDispatch::RemoteIp](https://api.rubyonrails.org/classes/ActionDispatch/RemoteIp.html) as middleware.

However, that's only the security aspect. By default Apache sets `X-Forwarded-Host` to the `Host` header. That in my example it would be `foreman-proxy.client.example.com`. Katello can't find a Smart Proxy with that name so it would likely break.

In short: secure reverse proxies are hard and Katello doesn't do any of the things to actually make it secure. It also doesn't consider multi homing at all.

ehelms:
> You wouldn't. However, that template URL setting allows Foreman and Foreman Proxy to **know** that. How else would it figure that out?

I thought the request would contain the host information and could glean from that the target of the registration.

ehelms:
I am worried a little bit about naming conventions here and the confusion that can cause when discussing this overall with users and other developers. The setting is currently within the context of Templates and is `template_url` which does not connect it back to the target concept: registration. I do get the underlying software is related to the templates, but given this is targeting a new workflow we want users to utilize, I see the value in trying to align nomenclature related to registration.

lzap:
Please resolve conflicts.

So what is the resolution here? Generally this is not the first time we need a "global" template. For example iPXE bootstrap and also user-data cloud init bootstrap templates are both examples when we introduced new "global" endpoint. I know it is little bit late but I am wondering if we should name the new endpoint simply "/global" and this registration would be "/global/register" so we can reuse the same controller and render stack for other global templates.

stejskalleos:
Rebased.

ares:
> I am worried a little bit about naming conventions here and the confusion that can cause when discussing this overall with users and other developers.

Users will only see this if they enable the template module. The template module already exists, this just adds another endpoint to it. I'm not sure whether it's better to create a new registration module while it extends existing templates module.

> I know it is little bit late but I am wondering if we should name the new endpoint simply "/global" and this registration would be "/global/register" so we can reuse the same controller and render stack for other global templates.

We may end up having an endpoint for serving global templates, however we want to have the register url as simple as possible so you can call `https://$foreman/register` (where $foreman can also be a smart proxy). I see the similarity here with `cloud-init` templates, but this URL is something, that's user facing and should be easy to remember.

I think this PR is now ready with 2 ACKs. My understanding of the discussion is - Katello takes a different approach that we don't want to use, because we don't want to expose the whole Foreman API to client machines. We also want to keep the advantage of proxies identificating themselves with certs. The Foreman part actually won't use smart proxy certs for authentication, in fact the plan right now is to limit the registration endpoint explicitly to user/pass or JWT with a specific purpose.

:+1: (haven't tested)

ekohl:
> The Foreman part actually won't use smart proxy certs for authentication, in fact the plan right now is to limit the registration endpoint explicitly to user/pass or JWT with a specific purpose.

I still see use cases where the proxy identifies itself. It'll need code on the foreman side to require both authentication methods, but the good thing is that this code sends it and Foreman can use it if it wants to.

ekohl:
> I thought the request would contain the host information and could glean from that the target of the registration.

You are right that it could do that, but that's always a passive approach. Foreman can never reach out and determine it. Clients need to reach out. With a configuration option, Foreman can learn it. That means you can have instructions on the Smart Proxy page telling the user how to register.

@stejskalleos IMHO that's actually a good idea and I don't know if it came up already. On the Smart Proxy page you can create a tab where you check if the capability is present and then display "To register a client, run `curl $template_url/register`".

I will agree that Templates is getting overloaded and becoming obscure to the user. They probably want to Provision and don't care as much that this is achieved through Templates. However, that point may have already been reached when userdata was added so IMHO it doesn't block this.

ehelms:
I am OK with the use of the internal templates module for achieving this functionality. What I still am less a fan of is the use of `template_url` as the setting and the overloading of that setting from a user understanding point of view. 

Could there be a case where you want to set `templates_url` to something different than a registration URL based upon the `template_url` current use case?

I think this change will need some updates to the documentation in various places, and when I consider that I start to get overwhelmed with the various caveats that must be included as opposed to a `global_registration_url` that would be more straight forward. Consider the language today in each (the latter of which should get an update in this PR to reflect the use case):

 * https://github.com/theforeman/puppet-foreman_proxy/blob/master/templates/templates.yml.erb#L9-L15
 * https://github.com/theforeman/puppet-foreman_proxy/blob/master/manifests/init.pp#L118
 * https://github.com/theforeman/smart-proxy/blob/develop/config/settings.d/templates.yml.example#L9-L15

ekohl:
Historically this was used for unattended installs like kickstart and preseed. Some of them even had no HTTPS support at all, others had no way to pass in the trusted chain.

That's why on the Foreman side we have `unattended_url` (typically http://foreman.example.com) and `foreman_url` (typically https://foreman.example.com).

With client registration it may actually make sense to do this over HTTPS since these systems support HTTPS and there are ways to get a trusted chain established. I think if you really want to start using this, you may need HTTP and HTTPS versions of the two. A single `templates_url` may be insufficient since you still want to support the classic kickstart/preseed.

ares:
@ehelms what is your suggestion? would you prefer to have another setting for `registration_url` so we don't reuse `template_url`?

@ekohl so you're saying we already have this HTTP vs HTTPS problem today? e.g. if new anaconda supports HTTP but debian installer only supports HTTP, user needs to use 2 different proxies right? I wouldn't try to solve that here, perhaps adding new `registration_url` would be the best option.

ekohl:
> @ekohl so you're saying we already have this HTTP vs HTTPS problem today? e.g. if new anaconda supports HTTP but debian installer only supports HTTP, user needs to use 2 different proxies right?

Sort of. Right now, we allow an all-or-nothing approach. Some companies have a no-HTTP, HTTP-only policy. Setting unattended_url to HTTPS allows them to achieve that today.

In https://community.theforeman.org/t/removing-unattended-url/16700 I proposed to get rid of the HTTP-only part in Foreman and make it HTTPS-only. This hasn't received a lot of traction now. This is also more in line with how Rails 6.1 sees it.

Less relevant now that we use Puma, but with Passenger it meant we effectively ran the application twice. Now we just have Puma and Apache tells Puma whether it was HTTP or HTTPS.

> I wouldn't try to solve that here, perhaps adding new `registration_url` would be the best option.

Perhaps we should use common names like `http_url` and `https_url`. Then Foreman can decide whether it should serve HTTP or HTTPS.

ares:
Actually thinking about this more, I think we should set the url to `request.env['REQUEST_URI']` and allow overriding through new setting `registration_url`. That means, we'd use the URL the client machine used to contact proxy, unless admin wants to override it for some reason.

> Perhaps we should use common names like http_url and https_url. Then Foreman can decide whether it should serve HTTP or HTTPS.

I don't want to overengineer this, but would those urls be used then also for other purposes? I'm wondering if there's a use case to have different Foreman for templates and different for registration. These features may need different scaling, but again, maybe that's overengineering. If we call it http_url, do you want to keep templates_url as well? Or do you want this PR to modify and unify this? (sorry I just want to be crystal clear about next steps)

ares (changes requested):
After additional conversation with both Eric and Ewoud, the summary is, it would be best to take this and https://github.com/theforeman/smart-proxy/pull/764 and combine it all in a new module called `Registration` instead of hosts. That will combine both endpoint for getting the templates and host API.

We agreed we only want to expose endpoints necessary for registration.

The logic that will determine what url to take should be based on `request.env['REQUEST_URI']` and does not need to be overridable by user. We should not use `templates_url` for this or any other preconfigured url.

PR #764, opened 2020-08-24
Fixes #30683 - Hosts module

stejskalleos:
Add new Hosts module to the Smart proxy, allowing to call :get & :post requests to the Foreman /hosts api (needed for the Global registration feature)

Notes:
* PR is part of **Simple & automatic host registration WF**, [RFC](https://community.theforeman.org/t/rfc-simple-automatic-host-registration-wf/19588) & [Redmine](https://projects.theforeman.org/issues/30440)

### DIFF
--- a/config/settings.d/hosts.yml.example
+++ b/config/settings.d/hosts.yml.example
@@ -1,0 +1,2 @@
+---
+:enabled: false
--- a/lib/smart_proxy_main.rb
+++ b/lib/smart_proxy_main.rb
@@ -77,6 +77,7 @@ module Proxy
   require 'realm_freeipa/realm_freeipa'
   require 'logs/logs'
   require 'httpboot/httpboot'
+  require 'hosts/hosts'
 
   def self.version
     {:version => VERSION}
--- a/modules/hosts/hosts.rb
+++ b/modules/hosts/hosts.rb
@@ -1,0 +1,1 @@
+require 'hosts/hosts_plugin'
--- a/modules/hosts/hosts_api.rb
+++ b/modules/hosts/hosts_api.rb
@@ -1,0 +1,13 @@
+require 'hosts/proxy_request'
+
+class Proxy::Hosts::Api < ::Sinatra::Base
+  helpers ::Proxy::Helpers
+
+  get '/*' do
+    Proxy::Hosts::ProxyRequest.new.get(request)
+  end
+
+  post '/*' do
+    Proxy::Hosts::ProxyRequest.new.post(request)
+  end
+end
--- a/modules/hosts/hosts_plugin.rb
+++ b/modules/hosts/hosts_plugin.rb
@@ -1,0 +1,8 @@
+module Proxy::Hosts
+  class Plugin < ::Proxy::Plugin
+    http_rackup_path File.expand_path("http_config.ru", File.expand_path(__dir__))
+    https_rackup_path File.expand_path("http_config.ru", File.expand_path(__dir__))
+
+    plugin :hosts, ::Proxy::VERSION
+  end
+end
--- a/modules/hosts/http_config.ru
+++ b/modules/hosts/http_config.ru
@@ -1,0 +1,5 @@
+require 'hosts/hosts_api'
+
+map "/hosts" do
+  run Proxy::Hosts::Api
+end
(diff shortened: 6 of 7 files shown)


ekohl (changes requested):
Could you elaborate a bit on why this is needed? When is the hosts API needed?

One concern I have is that there is a risk that the Smart Proxy authentication is used. In some cases this is allowed to create hosts. Looking at `ForemanRequestFactory` it does set those certificates meaning you open a huge security hole: without authentication you allow anyone to create hosts.

Another is that in the Katello scenario I think that we already have a full reverse proxy to Foreman using Apache. That makes me wonder if we need to expose specific endpoints or perhaps go the route of fully exposing the API. That also properly sets the reverse proxy headers you would expect, like X-Forwarded-For & friends.

I'd also like to see some tests. Various modules already have some integration tests to see if they at least load completely.

cc @ehelms

ehelms:
This does appear like it's intending to just provide reverse proxy capabilities for certain APIs that Katello solves with an Apache based reverse proxy. This new endpoint and method for registering a host within Foreman raises that question more broadly: should we move towards smart-proxy providing that reverse proxy for the API or should we just deploy Apache in that capacity more broadly (i.e. outside the Katello scenario)?

stejskalleos:
> Could you elaborate a bit on why this is needed? When is the hosts API needed?
This is for Global registration template, where we want to create the host and access the information.

UC is following:
* Admin wants to register host to the Foreman, but doesn't have access to the Foreman directly. What admin needs is:
  * Render the Global Registration template (done in: https://github.com/theforeman/smart-proxy/pull/763)
  * Create the host & get info about the host - This PR, that's why only `GET` & `POST` request are enabled.

> Another is that in the Katello scenario I think that we already have a full reverse proxy to Foreman using Apache.

Okay I'll look into it. But how about users using foreman without Katello?

lzap:
> it's intending to just provide reverse proxy capabilities for certain APIs

I will reply to that that we already have the same for templates endpoint. Although I agree that the ultimate solution would be to use a generic HTTP proxy, if this new global endpoint is generic enough so even other features can use it (cloud-init, iPXE) that it is probably still a better solution than introducing another component.

ares:
> One concern I have is that there is a risk that the Smart Proxy authentication is used. In some cases this is allowed to create hosts. Looking at ForemanRequestFactory it does set those certificates meaning you open a huge security hole: without authentication you allow anyone to create hosts.

I think the risk is minimal. Given we specify the endpoints and given there needs to be [explicit call](https://github.com/theforeman/foreman/pull/7928#discussion_r486314938) in Foreman controller to start accepting smart proxy certificates as authentication mechanism, I think we can safely ignore this.

> Another is that in the Katello scenario I think that we already have a full reverse proxy to Foreman using Apache. That makes me wonder if we need to expose specific endpoints or perhaps go the route of fully exposing the API. That also properly sets the reverse proxy headers you would expect, like X-Forwarded-For & friends.

As you [commented here](https://github.com/theforeman/smart-proxy/pull/763#issuecomment-687199600), we should allow just needed endpoints. I don't think we want to make Smart proxy a full reverse proxy. If we want to go that path, proxy could easily become a bottleneck.

> This does appear like it's intending to just provide reverse proxy capabilities for certain APIs that Katello solves with an Apache based reverse proxy. This new endpoint and method for registering a host within Foreman raises that question more broadly: should we move towards smart-proxy providing that reverse proxy for the API or should we just deploy Apache in that capacity more broadly (i.e. outside the Katello scenario)?

If we want to expose the whole API for some reason, I vote for using some existing reverse proxy software. But it needs to be secured as Ewoud says elsewhere. I'd like to know about more usecases for this first though. In last 10 years, registration seems to be the second. Clients usually consumes services running on proxy side. Clients shouldn't have many needs to talk to Foreman. Even templates could be cached on proxy side.

@stejskalleos this will also needs changes in the installer modules, so we can enable this module, could you please create a ticket for that so we remember?

ekohl:
The alternative is to implement a `POST` action in the templates module just to register. That's a much smaller change. We can make the Foreman understand this and produce the correct URLs.

ares:
Even if the register wouldn't actually be a template request? Feels like abusing of that module, but I'm not against.

ares (changes requested):
Please see https://github.com/theforeman/smart-proxy/pull/763 for more information. I don't mind if we combine both in single PR or use this as a base for the new module, whatever you prefer.

PR #765, opened 2020-08-27
Fixes #30719 - allow spawning without shell

lzap:
For https://github.com/theforeman/smart_proxy_shellhooks we need to be able to spawn a command through the utility helper without shell to prevent possible problems. Popen3 supports this via multiple arguments, but the utility helper did not allow passing an array.

Also logging was improved ("Started task" now contains also the thread PID) and logging of plain/text body was improved for easy to understand examples in the plugin README.

### DIFF
--- a/lib/proxy/util.rb
+++ b/lib/proxy/util.rb
@@ -11,16 +11,21 @@ module Proxy::Util
     # create new task and spawn new thread logging all the cmd
     # output to the proxy log. only the process' output is connected
     # stderr is redirected to proxy error log, stdout to proxy debug log
-    def initialize(acommand)
-      @command = acommand
+    # command can be either string or array (command + arguments)
+    # input is passed into STDIN and must be string
+    def initialize(command, input = nil)
+      @command = command
+      @input = input
     end
 
     def start(&ensured_block)
       # run the task in its own thread
-      logger.debug "Starting task: #{@command}"
-      @task = Thread.new(@command) do |cmd|
+      @task = Thread.new(@command, @input) do |cmd, input|
         status = nil
-        Open3.popen3(cmd) do |stdin, stdout, stderr, thr|
+        Open3.popen3(*cmd) do |stdin, stdout, stderr, thr|
+          logger.info "[#{thr.pid}] Started task #{cmd}"
+          stdin.write(input) if input
+          stdin.close
           stdout.each do |line|
             logger.debug "[#{thr.pid}] #{line}"
           end
--- a/lib/proxy/http_download.rb
+++ b/lib/proxy/http_download.rb
@@ -14,8 +14,13 @@ module Proxy
       dns_timeout ||= DEFAULT_CONNECT_TIMEOUT
       connect_timeout ||= DEFAULT_DNS_TIMEOUT
 
-      super("#{wget} --connect-timeout=#{connect_timeout} --dns-timeout=#{dns_timeout} --read-timeout=#{read_timeout}"\
-            " --tries=3 --no-check-certificate -nv -c \"#{escape_for_shell(src.to_s)}\" -O \"#{escape_for_shell(dst.to_s)}\"")
+      super([wget,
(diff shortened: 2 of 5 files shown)


lib/proxy/util.rb
```diff
@@ -11,16 +11,21 @@ class CommandTask
     # create new task and spawn new thread logging all the cmd
     # output to the proxy log. only the process' output is connected
     # stderr is redirected to proxy error log, stdout to proxy debug log
-    def initialize(acommand)
+    # command can be either string or array (command + arguments)
+    # input must be string
+    def initialize(acommand, aninput = nil)
```
ekohl: I thought `aninput` as hard to read. Is `input` a reserved value/method? If so, I think `input_value` might be easier to read.

lib/proxy/util.rb
```diff
@@ -11,16 +11,21 @@ class CommandTask
...
       @command = acommand
+      @input = aninput
     end
 
     def start(&ensured_block)
       # run the task in its own thread
-      logger.debug "Starting task: #{@command}"
```
ekohl, line 20: This still looks useful.

ekohl (commented):
Small details, overall this makes sense. I'm certainly a big fan of passing arrays as commands rather than trying to shell escape.

Does make me wonder: should we update HttpDownload as well to use this? After Puppetrun is removed, it's the only place that still uses `escape_for_shell`. I'm fine with doing that in a follow up PR.

lib/proxy/util.rb
ares: I wish we also renamed `acommand` to just `command` at the same time. I know it's not changed by this PR :-)

lzap: Sure it's rare naming convention and I just followed what was there but can refactor.

lib/proxy/util.rb
lzap: Absolutely, I just moved it few lines down so I could actually use thread ID in the message so it's obvious which thread ID messages belongs to which command.

lzap:
No problem, I can incorporate the wget change into this PR. Amended, tested. Looks way cleaner now, no escaping anymore.

lib/proxy/util.rb
ekohl: The thing is that you now move it to within the `Thread.new` and `popen3`. If either fails, you don't really know. For example, `popen3` can raise an exception. While the command may not say a huge amount, I'm not sure what the error output is if the command in the thread failed. Those were my considerations for keeping it in. Since it's debug, most users won't see it, but can be useful when things break for a weird reason.

Not going to block on it because it may be a bit of an edge case so I'll leave it up to you.

lib/proxy/util.rb
```diff
@@ -11,16 +11,21 @@ class CommandTask
...
-      @task = Thread.new(@command) do |cmd|
+      @task = Thread.new(@command, @input) do |cmd, input|
         status = nil
-        Open3.popen3(cmd) do |stdin, stdout, stderr, thr|
+        Open3.popen3(*cmd) do |stdin, stdout, stderr, thr|
```
ekohl, line 25: Coming from Python, it surprised me you can do this with strings. In Python `my-cmd` would end up as `['m', 'y', '-', 'c', 'm', 'd']`. I learned that here Python and Ruby differ.

ekohl (approved):

lib/proxy/util.rb
lzap: Yeah I tested this. There's a test for this as well.

ekohl: I did indeed test it in a shell. Just one of those Today I Learned moments.

ekohl:
I'm going to assume @ares has no objections other than the one he mentioned. Since that's addressed, I'm going to merge this. Thanks!

After merging this and the puppetrun removal I want to deprecate the shell escape method in a separate PR.

PR #767, opened 2020-09-03
Fixes #30773 - expose template_url setting

ekohl:
It can be useful if Foreman has the templates_url statically available.  Right now this is retrieved via the /templates/templateServer API call but if it's stored statically, Foreman can be modified to not need this API call anymore.


lzap (approved):
Agreed, this is already exposed URL in templates themselves.

PR #768, opened 2020-09-15
Fixes #30677 - Registration module

stejskalleos:
SmartProxy Registration module allows users to register hosts via GlobalRegistration feature through the Smart Proxy.

**Summary**
Two new endpoints:
* `GET /register` - Render Global registration template
* `POST /register` - Render Host registration template (and create Host)

Both endpoints pass all arguments & headers set by user to the Foreman, plus it adds `:url` parameter - URL of the Smart Proxy. This parameter will be used later in the templates to replace Foreman URLs with the Smart Proxy URL.

4d6455aeafe4ba320a3f3714c6bc33f204fcd4fa

**How to test**
* For Vanilla Foreman: https://github.com/theforeman/foreman/pull/8018
* With Katello: https://github.com/Katello/katello/pull/8965
```
curl --user admin:changeme -X GET http://smart-proxy.example.com/register" | bash
```

Note that this won't change URLs in the templates (not implemented yet), but there is a [RM#30714](https://projects.theforeman.org/issues/30714) waiting.

**Not covered in this PR:**
* Changing URL in Global & Host registration templates - [RM#30714](https://projects.theforeman.org/issues/30714)
* Secure connection

---
PR is part of **Simple & automatic host registration WF**, see:
* [RFC](https://community.theforeman.org/t/rfc-simple-automatic-host-registration-wf/19588)
* [Redmine (this PR)](https://projects.theforeman.org/issues/30677)
* [Redmine (Parent task)](https://projects.theforeman.org/issues/30440)

### DIFF
--- a/modules/registration/proxy_request.rb
+++ b/modules/registration/proxy_request.rb
@@ -1,0 +1,37 @@
+require 'proxy/request'
+
+module Proxy::Registration
+  class ProxyRequest < ::Proxy::HttpRequest::ForemanRequest
+    def global_register(request)
+      proxy_req = request_factory.create_get '/register',
+                                             request_params(request),
+                                             headers(request)
+
+      send_request(proxy_req)
+    end
+
+    def host_register(request)
+      proxy_req = request_factory.create_post '/register',
+                                              request.body.read,
+                                              headers(request),
+                                              request_params(request)
+
+      send_request(proxy_req)
+    end
+
+    private
+
+    def request_params(request)
+      params = request.params
+      params[:url] = request.env['REQUEST_URI']&.split('/register')&.first
+      params
+    end
+
+    def headers(request)
+      Hash[request.env.select { |k, v| k =~ /^HTTP_/ && k !~ /^HTTP_(VERSION|HOST)$/ }.map { |k, v| [k[5..-1], v] }]
+    rescue Exception => e
+      logger.warn "Unable to extract request headers: #{e}"
+      {}
+    end
+  end
+end
--- a/modules/registration/registration_api.rb
+++ b/modules/registration/registration_api.rb
@@ -1,0 +1,40 @@
+require 'registration/proxy_request'
+
+class Proxy::Registration::Api < ::Sinatra::Base
+  get '/' do
+    response = Proxy::Registration::ProxyRequest.new.global_register(request)
+    handle_response(response)
+  rescue StandardError => e
+    logger.exception "Error when rendering Global Registration Template", e
+    render_error(default_error_msg)
+  end
+
+  post '/' do
+    response = Proxy::Registration::ProxyRequest.new.host_register(request)
+    handle_response(response)
+  rescue StandardError => e
+    logger.exception "Error when rendering Host Registration Template", e
+    render_error(default_error_msg)
+  end
+
+  private
+
+  def handle_response(response)
+    if response.code.start_with? '2'
+      response.body
+    else
+      # Return error message only if it is not HTML.
+      message = response["content-type"].include?('text/plain') ? response.body : default_error_msg
+      render_error(message, code: response.code)
+    end
+  end
+
+  def render_error(message, code: 500)
+    status code
+    message
+  end
+
+  def default_error_msg
+    "echo \"Internal Server Error\"\nexit 1\n"
+  end
+end
--- a/modules/registration/registration_plugin.rb
+++ b/modules/registration/registration_plugin.rb
@@ -1,0 +1,8 @@
+module Proxy::Registration
+  class Plugin < ::Proxy::Plugin
+    http_rackup_path  File.expand_path("http_config.ru", File.expand_path(__dir__))
+    https_rackup_path File.expand_path("http_config.ru", File.expand_path(__dir__))
+
+    plugin :registration, ::Proxy::VERSION
+  end
+end
--- a/test/registration/registration_api_test.rb
+++ b/test/registration/registration_api_test.rb
@@ -1,0 +1,97 @@
+require 'test_helper'
+require 'registration/registration_api'
+
+class RegistrationRegisterApiTest < Test::Unit::TestCase
+  include Rack::Test::Methods
+
+  def app
+    Proxy::Registration::Api.new
+  end
+
+  def setup
+    @foreman_url = 'http://foreman.example.com'
+    Proxy::SETTINGS.stubs(:foreman_url).returns(@foreman_url)
+  end
+
+  def test_global_register_template
+    stub_request(:get, "#{@foreman_url}/register").to_return(body: 'template')
+
+    get "/"
+    assert last_response.ok?
+    assert_match('template', last_response.body)
+  end
+
+  def test_global_register_template_with_args
+    stub_request(:get, "#{@foreman_url}/register?param=test").to_return(body: 'template')
+
+    get '/', { param: 'test' }
+    assert last_response.ok?
+    assert_match('template', last_response.body)
+  end
+
+  def test_host_register_template
+    stub_request(:post, "#{@foreman_url}/register").to_return(body: 'template')
+
+    post '/'
+    assert last_response.ok?
+    assert_match('template', last_response.body)
+  end
+
+  def test_host_register_template_with_args
+    stub_request(:post, "#{@foreman_url}/register?host=test.example.com").to_return(body: 'template')
+
+    post '/', { host: 'test.example.com' }
+    assert last_response.ok?
+    assert_match('template', last_response.body)
+  end
+
+  def test_global_401
+    stub_request(:get, "#{@foreman_url}/register").to_return(body: '401', status: 401, headers: { "Content-Type" => 'text/plain; charset=UTF-8' })
+
+    get '/'
+    assert last_response.unauthorized?
+    assert_match('401', last_response.body)
+  end
+
+  def test_host_401
+    stub_request(:post, "#{@foreman_url}/register").to_return(body: '401', status: 401, headers: { "Content-Type" => 'text/plain; charset=UTF-8' })
+
+    post '/'
+    assert last_response.unauthorized?
+    assert_match('401', last_response.body)
+  end
+
+  def test_global_401_html_response
+    stub_request(:get, "#{@foreman_url}/register").to_return(body: '401', status: 401, headers: { "Content-Type" => 'text/html; charset=UTF-8' })
+
+    get '/'
+    assert last_response.unauthorized?
+    assert_match("echo \"Internal Server Error\"\nexit 1\n", last_response.body)
+  end
+
+  def test_host_401_html_response
+    stub_request(:post, "#{@foreman_url}/register").to_return(body: '401', status: 401, headers: { "Content-Type" => 'text/html; charset=UTF-8' })
+
+    post '/'
+    assert last_response.unauthorized?
+    assert_match("echo \"Internal Server Error\"\nexit 1\n", last_response.body)
+  end
+
+  def test_global_500
+    Rack::NullLogger.any_instance.stubs(:exception)
+    stub_request(:get, "#{@foreman_url}/register").to_timeout
+
+    get '/'
+    assert last_response.server_error?
+    assert_match("echo \"Internal Server Error\"\nexit 1\n", last_response.body)
+  end
+
+  def test_host_500
+    Rack::NullLogger.any_instance.stubs(:exception)
+    stub_request(:post, "#{@foreman_url}/register").to_timeout
+
+    post '/'
+    assert last_response.server_error?
+    assert_match("echo \"Internal Server Error\"\nexit 1\n", last_response.body)
+  end
+end
--- a/config/settings.d/registration.yml.example
+++ b/config/settings.d/registration.yml.example
@@ -1,0 +1,2 @@
+---
+:enabled: false
--- a/lib/smart_proxy_main.rb
+++ b/lib/smart_proxy_main.rb
@@ -73,6 +73,7 @@ module Proxy
   require 'realm_freeipa/realm_freeipa'
   require 'logs/logs'
   require 'httpboot/httpboot'
+  require 'registration/registration'
 
   def self.version
     {:version => VERSION}
--- a/modules/registration/http_config.ru
+++ b/modules/registration/http_config.ru
@@ -1,0 +1,5 @@
+require 'registration/registration_api'
+
+map '/register' do
+  run Proxy::Registration::Api
+end
--- a/modules/registration/registration.rb
+++ b/modules/registration/registration.rb
@@ -1,0 +1,1 @@
+require 'registration/registration_plugin.rb'
--- a/test/registration/integration_test.rb
+++ b/test/registration/integration_test.rb
@@ -1,0 +1,24 @@
+require 'test_helper'
+require 'json'
+require 'root/root_v2_api'
+require 'registration/registration'
+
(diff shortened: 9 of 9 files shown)


stejskalleos:
@ares, @ekohl, @ehelms

test/registration/hosts_api_test.rb
```diff
@@ -0,0 +1,29 @@
...
+    Proxy::Registration::HostsApi.new
+  end
+
+  def setup
+    @foreman_url = 'http://foreman.example.com'
+    Proxy::SETTINGS.stubs(:foreman_url).returns(@foreman_url)
+  end
+
+  def test_get_hosts_api
+    stub_request(:get, "#{@foreman_url}/api/hosts").to_return(:body => 'GET hosts')
```
ekohl: Can we verify the body is returned in `last_response`? A pretty crucial part of this.

modules/registration/registration_template_api.rb
```diff
@@ -0,0 +1,7 @@
+require 'registration/proxy_request'
+
+class Proxy::Registration::TemplateApi < ::Sinatra::Base
+  get '/*' do
```
ekohl: This only has `/register`. I'd keep it strict.

modules/registration/registration_plugin.rb
```diff
@@ -0,0 +1,9 @@
+module Proxy::Registration
+  class Plugin < ::Proxy::Plugin
+    http_rackup_path  File.expand_path("http_config.ru", File.expand_path(__dir__))
+    https_rackup_path File.expand_path("http_config.ru", File.expand_path(__dir__))
+
+    plugin     :registration, ::Proxy::VERSION
+    capability :global_registration
```
ekohl: Now that it's a plugin, the capability doesn't make a lot of sense anymore. It's now promoted to a real Feature. Note that Foreman needs a PR as well to recognize the Feature.

ekohl (changes requested):
This still creates a vulnerability. The hosts API allows client certificates to be used and `ForemanRequest` sends client certificates. That means you allow unauthenticated attackers to create new hosts by using `POST /api/hosts/facts`.

IMHO we should not use the hosts API at all and host registration should introduce its own endpoints. As suggested elsewhere, `POST /register` should return the HRT which avoids the need for the pure Foreman case.

For the Katello case, a specific endpoint should be introduced that returns the HRT. When talking directly to Foreman, it should allow client certificates to be used. When talking to a Smart Proxy, the Smart Proxy should do that authentication. The Smart Proxy should then authenticate itself to the Foreman HRT endpoint using client certificates and passing the actual client hostname via a GET parameter.

This would avoid all the security risk. Keep in mind that using `ForemanRequest` can be very dangerous.

ares:
@stejskalleos the vulnerability that @ekohl talks about is, that you expost all /api/hosts/* endpoints, one of them actually can rely on client certificate for authentication [definition](https://github.com/theforeman/foreman/blob/develop/app/controllers/api/v2/hosts_controller.rb#L18) and [the problematic action](https://github.com/theforeman/foreman/blob/develop/app/controllers/api/v2/hosts_controller.rb#L298-L310)

While I wasn't able to fully exploit that, since the Foreman responds with `"error": {"message":"Unable to authenticate user "}` but I agree we should only limit these to `GET /api/hosts/$host` and `POST /api/hosts/`, no other endpoint. Then there's no vulnerability. Given we don't need the cert auth though, a better solution may be to enhance the ForemanReuqest so that we can disable client certificate completely. That could be achieved by assing an optional param to [send_request](https://github.com/theforeman/smart-proxy/blob/develop/lib/proxy/request.rb#L47-L49), which would pass it to [http](https://github.com/theforeman/smart-proxy/blob/develop/lib/proxy/request.rb#L59) method, that would then skip [client cert part](https://github.com/theforeman/smart-proxy/blob/develop/lib/proxy/request.rb#L80-L83). We don't need this at all, we rely on user credentials or JWT for authentication there.

EDIT: to reproduce the vulnerability `curl -X POST https://foreman.example.tst:8443/api/hosts/facts -d '{"name": "bart.example.tst", "facts": {}}'`, note that I don't specify any username/password, but Foreman starts the action...

ares:
> IMHO we should not use the hosts API at all and host registration should introduce its own endpoints. As suggested elsewhere, POST /register should return the HRT which avoids the need for the pure Foreman case.

I replied in the core PR, I would prefer not to introduce new API endpoint if we don't need it.

> For the Katello case, a specific endpoint should be introduced that returns the HRT. When talking directly to Foreman, it should allow client certificates to be used. When talking to a Smart Proxy, the Smart Proxy should do that authentication. The Smart Proxy should then authenticate itself to the Foreman HRT endpoint using client certificates and passing the actual client hostname via a GET parameter.

Also replied in core PR, I disagree, that does not solve user identification and therefore user authorization.

> This would avoid all the security risk. Keep in mind that using ForemanRequest can be very dangerous.

we can simply disable client certs for these forwarded requests, we rely on JWT, no need to authenticate the proxy itself

modules/registration/registration_hosts_api.rb
```diff
@@ -0,0 +1,11 @@
+require 'registration/proxy_request'
+
+class Proxy::Registration::HostsApi < ::Sinatra::Base
+  get '/*' do
```
ares: get '/:host'

modules/registration/registration_hosts_api.rb
```diff
@@ -0,0 +1,11 @@
...
+    Proxy::Registration::ProxyRequest.new.get(request)
+  end
+
+  post '/*' do
```
ares: post '/' do

ares (dismissed):
see my suggestions for limiting the host endpoints

ekohl:
> Given we don't need the cert auth though, a better solution may be to enhance the ForemanReuqest so that we can disable client certificate completely.

We don't for vanilla Foreman registration, but if you want to authenticate and identify the Smart Proxy (as you should for Katello), then it's good to keep it.

I still want to avoid exposing the whole hosts API. Exposing only parts of it is even worse IMHO.

I'd like to get back to the RFC and have an updated status. I feel like I'm explaining the same thing in multiple PRs which makes it hard to follow. I think both @timogoebel and I misunderstood the subscription-manager aspect of it and that you want a second HTTP call. I can't speak for him, but I think we should centralize the discussion, take it back to the original design and see if we need to make adjustments.

I've created https://community.theforeman.org/t/rfc-simple-automatic-host-registration-wf/19588/33 where I outline how I think the basic flows (without Smart Proxy) should work. I've added some security concerns to think about that scope. If we agree on those, then we can think about adding a Smart Proxy.

timogoebel:
> I can't speak for him, but I think we should centralize the discussion, take it back to the original design and see if we need to make adjustments.

By all means: Yes. :-)

stejskalleos:
Code & description updated according to changes in https://github.com/theforeman/foreman/pull/8018

lzap (approved):
LGTM

ekohl (commented):
It looks like you're not passing HTTP status codes along and I don't see any error handling in case Foreman returns some bad response. Maybe a generic HTTP 503 is sufficient when there's a non-2xx code. Perhaps the content type is also important to pass along.

stejskalleos:
> It looks like you're not passing HTTP status codes along and I don't see any error handling in case Foreman returns some bad response

If there is an error, we will handle it properly, for example case with not-existing organization will return:
```
echo "ERROR: not_found";
echo 'Organization with id 23 not found';
exit 1
```

ekohl:
I'm talking about a HTTP 5xx errors. That means the content may be HTML rendered by Apache (in case of HTTP 502/503). If someone executes that, it may lead to very unpredictable behavior.

modules/registration/proxy_request.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    def headers(request)
+      Hash[request.env.select { |k, v| k =~ /^HTTP_/ && k !~ /^HTTP_(VERSION|HOST)$/ }.map { |k, v| [k[5..-1], v] }]
+    rescue Exception => e
+      logger.warn "Unable to extract request headers: #{e}"
+      {}
+    end
+
+    def handle_response(response)
+      error_codes = (500..511).to_a
```
timogoebel:
```suggestion
      error_codes = (500..511).to_a
```

I believe this can stay a range, the conversation to an array isn't necessary.
https://apidock.com/ruby/Range/include%3F

modules/registration/registration_api.rb
```diff
@@ -0,0 +1,11 @@
+require 'registration/proxy_request'
+
+class Proxy::Registration::Api < ::Sinatra::Base
+  get '/' do
+    Proxy::Registration::ProxyRequest.new.global_register(request)
```
ekohl: I would consider not having `global_register` and `host_register` as separate methods and inline them here. The reason is that currently you don't do anything with the response code. However, I think it is very useful for responses from the Smart Proxy to also have a correct response code. For example, on a timeout (which you currently don't capture) you should probably return a HTTP 503 code.

There may also be other HTTP errors, such as invalid certificates etc. Since the expected use of this API is to pipe it directly to bash, it's crucial you capture any exception and always return something valid but with the appropriate context.

It's also useful to have logging when a request fails.

test/registration/registration_api_test.rb
```diff
@@ -0,0 +1,79 @@
...
+  def test_host_register_template_with_args
+    stub_request(:post, "#{@foreman_url}/register?host=test.example.com").to_return(body: 'template')
+
+    post '/', { host: 'test.example.com' }
+    assert last_response.ok?
+    assert_match('template', last_response.body)
+  end
+
+  def test_global_401
+    stub_request(:get, "#{@foreman_url}/register?error=401").to_return(body: 'error404')
```
ekohl: This looks odd. It suggests there's a HTTP GET parameter error, but there really isn't.
```suggestion
    stub_request(:get, "#{@foreman_url}/register").to_return(body: 'error401', status: 401)
```

(Same below with host_401)

modules/registration/registration_api.rb
stejskalleos: > I would consider not having global_register and host_register as separate methods and inline them here. The reason is that currently you don't do anything with the response code. 

I do, here: https://github.com/theforeman/smart-proxy/pull/768/files#diff-45f33bea51f0f06fd02907550e854ae2R37-R42
Or should I do it in different way? If yes could you show me how do you want to do it?

> There may also be other HTTP errors, such as invalid certificates etc. Since the expected use of this API is to pipe it directly to bash, it's crucial you capture any exception ...

Can you show me exactly how you want me to do it, so we are both on same boat.

ekohl: That is only the body you deal with for incoming requests. Clients for the Smart Proxy don't see those, do they? The fact `assert last_response.ok?` in the tests pass indicate that it returns a HTTP 200 instead of 500 (or similar).

stejskalleos: >  in the tests pass indicate that it returns a HTTP 200 instead of 500 (or similar)

Yeah but it is used in templates, so if there is an error we render it like this:
```
echo '<message>'
exit 1
```
So the script will exit anyway, no matter what status code is.

ekohl: But if you pass the HTTP status code, you at least know that it failed. You can see this in proxy.log and it's just a good practice.

For example:

```
2020-10-01T16:49:29 bf9e90fb [I] Started GET /features 
2020-10-01T16:49:29 bf9e90fb [I] Finished GET /features with 200 (24.44 ms)
```

If you see there are HTTP 5xx status codes, you know there's a problem. This can be used by various logging tools as well. Out of experience I know that debugging such failures can be painful and that's why I strongly feel that this is important.

modules/registration/registration_api.rb
```diff
@@ -0,0 +1,34 @@
...
+  rescue StandardError => e
+    logger.error "Error when rendering Host Registration Template: #{e}"
+    render_error
+  end
+
+  private
+
+  def handle_response(response)
+    if response.code == '200'
+      response.body
```
ekohl: All 2xx codes are considered valid. On a HTTP POST you would expect a HTTP 201 if a resource was created.

ares (commented):
No more code related comments, I tested the basic functionality. I'd be happy to test the whole flow once there's core PR adding url parameter support to test with.

modules/registration/registration_api.rb
```diff
@@ -0,0 +1,34 @@
+require 'registration/proxy_request'
+
+class Proxy::Registration::Api < ::Sinatra::Base
+  get '/' do
+    response = Proxy::Registration::ProxyRequest.new.global_register(request)
+    handle_response(response)
+  rescue StandardError => e
+    logger.error "Error when rendering Global Registration Template: #{e}"
```
ekohl: We have `logger.exception` which sounds appropriate.

modules/registration/registration_api.rb
```diff
@@ -0,0 +1,34 @@
...
+    render_error
+  end
+
+  private
+
+  def handle_response(response)
+    if response.code.start_with? '2'
+      response.body
+    else
+      render_error(response.code, response.body)
```
ekohl: Would it be useful log something here?

Another thing that came to mind is look at the content type of the response. If it's `text/html`, you don't want to pass the body but if it's `text/plain` it's probably safe to pass along. This may help when Apache generates some HTTP 502 error (in HTML) vs something that the Foreman application did. It may require a change on the Foreman side as well.

modules/registration/registration_plugin.rb
```diff
@@ -0,0 +1,8 @@
+module Proxy::Registration
+  class Plugin < ::Proxy::Plugin
+    http_rackup_path  File.expand_path("http_config.ru", File.expand_path(__dir__))
+    https_rackup_path File.expand_path("http_config.ru", File.expand_path(__dir__))
```
ekohl, line 4: Since 88fbc8e67d665e2c3b19acb53b31ff30acf078b7 there's a new preferred way
```suggestion
    rackup_path File.expand_path('http_config.ru', __dir__)
```

ekohl (commented):
I am also thinking about response headers, in particular the `Content-Type` header.

ares:
@stejskalleos I think there's one more suggestion, hopefully it's ready afterwards

ekohl (approved):
The code looks good. Please squash into a single commit with a short description of what this module aims to do. A reference to some documentation is also fine. That should make the Redmine issues check (and me) happy :)

That does point me to some issue that I've felt with all smart proxy modules: it's hard to figure out exactly why it's there, what it's supposed to do and how the API works. https://projects.theforeman.org/projects/smart-proxy/wiki/API is the closest we have and it's out of date.

ares:
@stejskalleos I guess the short description may be something like:

"This introduces a new registration module described at https://community.theforeman.org/t/rfc-simple-automatic-host-registration-wf/19588/59. The module proxies host registration request to Foreman, which typically responds with the template that contains the registration instructions. There are two endpoints, one for obtaining the global registration template with generic instructions for any host, the second for host registration template, which is typically customized by parameters set for this specific host."

@ekohl if you're happy with the description perhaps you can squash the message on merge, @stejskalleos is out at least for today, merging would unblock other PRs in core.

ekohl:
@ares looks good. I do wonder if Github wraps commit messages if you squash merge. I think I always did it manually but never checked.

ares:
Thanks, it does not, so I formatted it manually. The link itself is longer than 80 chars, so I kept it with the first line. Thanks @stejskalleos and @ekohl, this one is in.

PR #770, opened 2020-09-16
Fixes #30856 - Drop  Puppet class cache initializer

ekohl:
Retrieving all classes for all environments only makes sense if caching is enabled on the Puppetserver. By default this is the case. Even if this is not the case, it only works for environments that don't change.  It's sufficient to lazy load this.

This saves load on start up of services.


ekohl:
I didn't do any benchmarking, but AFAIK by default Puppetserver doesn't return the etags needed for caching so in most setups this is pure overhead.

tbrisker (approved):
Thanks @ekohl !

PR #772, opened 2020-09-16
Refs #30856 - Fix class loading

ekohl:
025c1f3098f139ba5056cb2e97f3ceeefda68d78 removed this require, but that file also contains the V3EnvironmentClassesApiClassesRetriever class which is still used. This is a case of a bad rebase from my side.


tbrisker (approved):
Thanks @ekohl !

PR #773, opened 2020-10-01
Fixes #30951 - Provide a rackup_path plugin DSL

ekohl:
Currently there's http_rackup_path and https_rackup_path. In practice all built in modules and most (all?) plugins that I know define the same rackup config for both HTTP and HTTPS. Having a DSL method to do this simplifies code.

### DIFF
--- a/modules/bmc/bmc_plugin.rb
+++ b/modules/bmc/bmc_plugin.rb
@@ -1,7 +1,6 @@
 module Proxy::BMC
   class Plugin < Proxy::Plugin
-    http_rackup_path File.expand_path("http_config.ru", File.expand_path(__dir__))
-    https_rackup_path File.expand_path("http_config.ru", File.expand_path(__dir__))
+    rackup_path File.expand_path("http_config.ru", __dir__)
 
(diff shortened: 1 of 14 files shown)


ehelms (approved):
Seems like a reasonable cleanup

modules/bmc/bmc_plugin.rb
```diff
@@ -1,7 +1,6 @@
 module Proxy::BMC
   class Plugin < Proxy::Plugin
-    http_rackup_path File.expand_path("http_config.ru", File.expand_path(__dir__))
-    https_rackup_path File.expand_path("http_config.ru", File.expand_path(__dir__))
+    rackup_path File.expand_path("http_config.ru", File.expand_path(__dir__))
```
ekohl: Reading this, I now wonder if we should also simplify this to:
```suggestion
    rackup_path File.expand_path("http_config.ru", __dir__)
```

That's equivalent, right?

ehelms: Seems like from testing that it should resolve to the same value.

ehelms (approved):

PR #775, opened 2020-10-05
Add x-forwared-for to forward_verify

chr1s692:
add x-forwared-for to the verification of remote addr

### DIFF
--- a/lib/proxy/helpers.rb
+++ b/lib/proxy/helpers.rb
@@ -30,7 +30,12 @@ module Proxy::Helpers
 
   # read the HTTPS client certificate from the environment and extract its CN
   def https_cert_cn
-    certificate_raw = request.env['SSL_CLIENT_CERT'].to_s
+    if request.env['HTTP_SSL_CLIENT_CERT']
+      certificate_raw = request.env['HTTP_SSL_CLIENT_CERT'].to_s
+      certificate_raw = "-----BEGIN CERTIFICATE-----\n#{certificate_raw}\n-----END CERTIFICATE-----"
+    else
+      certificate_raw = request.env['SSL_CLIENT_CERT'].to_s
+    end
     log_halt 403, 'could not read client cert from environment' if certificate_raw.empty?
 
     begin
@@ -80,11 +85,7 @@ module Proxy::Helpers
 
   # reverse lookup an IP address while verifying it via forward resolv
   def remote_fqdn(forward_verify = true)
-    if request.env['HTTP_X_FORWARDED_FOR']
-      ip = request.env['HTTP_X_FORWARDED_FOR'].split(',')[0]
-    else
-      ip = request.env['REMOTE_ADDR']
-    end
+    ip = request.env['REMOTE_ADDR']
     log_halt 403, 'could not get remote address from environment' if ip.empty?
 
     begin
--- a/lib/proxy/helpers.rb
+++ b/lib/proxy/helpers.rb
@@ -80,7 +80,11 @@ module Proxy::Helpers
 
   # reverse lookup an IP address while verifying it via forward resolv
   def remote_fqdn(forward_verify = true)
-    ip = request.env['REMOTE_ADDR']
+    if request.env['HTTP_X_FORWARDED_FOR']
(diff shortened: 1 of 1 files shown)


lib/proxy/helpers.rb
```diff
@@ -80,7 +80,11 @@ def resolv(*args)
 
   # reverse lookup an IP address while verifying it via forward resolv
   def remote_fqdn(forward_verify = true)
-    ip = request.env['REMOTE_ADDR']
+    if request.env['HTTP_X_FORWARDED_FOR']
+      ip = request.env['HTTP_X_FORWARDED_FOR'].split(',')[0]
+    else
+      ip = request.env['REMOTE_ADDR']
+    end
```
ekohl: The major issue with blindly accepting `X-Forwarded-For` is that you're opening up a massive security issue since anyone can spoof these headers. That means you need proper middleware to secure the requests.

This weekend I was looking at Rack and I think [request.ip](https://rubydoc.info/github/rack/rack/Rack/Request/Helpers#ip-instance_method) should do the same. It appears to have an ip filter that you can set. I'd consider the default there insecure since it accepts RFC1918 IPs as well, which is common for a Smart Proxy deployment.

I'd say that an explicit list of allowed reverse proxies is more secure.

Also note that when using HTTPS client certificates (as you should) this is only used over HTTP. In a secure deployment, all authenticated modules should only be accessible over HTTPS.

I was actually thinking about proposing to drop this altogether.

lzap: Yeah, this is a massive hole. We can't merge this as-is. I agree that we need to simplify SP's security.

What exactly do you need this for? Maybe we can help finding a better alternative.

chr1s692: @lzap : We need to place the foreman smart proxy behind a reverse proxy. We would much rather check the forwarded certificate, like in foreman, but i don't know that much about your source code/ruby. This was a dirty try to get reverse proxy support and i appreciate, not the securest one. 
Would the verification of the forwarded certificate like in foreman be a secure alternative?

ekohl: >  Would the verification of the forwarded certificate like in foreman be a secure alternative?

Yes

chr1s692: @ekohl : ok, and what have to be done to implement this?

lzap: You can generate valid SSL client cert for the reverse proxy and add it to the "trusted hosts" configuration. All requests coming from this client will be authorized. If this is what you want.

chr1s692: Currently thats our "workaround". But all clients, which can access or talk to the reverse proxy, are verified to the backend. Ok, i can add an require to some IP, but this isn't a very nice solution.
So, if you say the verification of an forwareded certificate is better, let me have a secound try. i will update the code.

lzap: Then we could actually accept your patch but only with a new (required) setting - list of X509 common names that are allowed to resped x-forwarded-for. This way administrator must explicitly state which client cert is allowed to do this dangerous action. @ekohl what you think? Other way?

lib/proxy/helpers.rb
```diff
@@ -30,7 +30,12 @@ def log_halt(code = nil, exception_or_msg = nil, custom_msg = nil)
 
   # read the HTTPS client certificate from the environment and extract its CN
   def https_cert_cn
-    certificate_raw = request.env['SSL_CLIENT_CERT'].to_s
+    if request.env['HTTP_SSL_CLIENT_CERT']
```
ekohl, line 33: This can be spoofed by any user. Reverse proxies without the appropriate middleware are **dangerous**. I'd suggest to read up on this, for example https://telekom-security.github.io/2020/05/smuggling-http-headers-through-reverse-proxies.html.

What this patch now does is that for the default deployment the user can send any certificate using `curl -H 'SSL-Client-Cert: <certificate>' https://smart-proxy.example.com:8443`. It doesn't have to be signed by the correct CA and the CN on it will be used.

If you forget to clear the header in the reverse proxy, you open yourself to the same security issue.

In short: ff you're applying a reverse proxy for security, you must **really** know what you're doing because it's trivial to open gaping security problems.

PR #776, opened 2020-10-17
Fixes #31086 - Use __dir__ when possible

ekohl:
`__dir__` is equal to `File.dirname(__FILE__)` and easier to read.


lzap (approved):
Thanks

PR #777, opened 2020-10-21
Fixes #31118 - Remove daemonize support

ekohl:
Currently Smart Proxy is able to daemonize and manage a pid file.  However, in the common deployments under systemd and Windows this isn't actually used. There are also other solutions out there to manage it externally if it's needed.

Removing this simplifies the code and config file.

### DIFF
--- a/config/settings.yml.example
+++ b/config/settings.yml.example
@@ -42,12 +42,6 @@
 #:foreman_ssl_cert: ssl/certs/fqdn.pem
 #:foreman_ssl_key: ssl/private_keys/fqdn.pem
 
-# by default smart_proxy runs in the foreground. To enable running as a daemon, uncomment 'daemon' setting
-#:daemon: true
-# Only used when 'daemon' is set to true.
-# Uncomment and modify if you want to change the default pid file '/var/run/foreman-proxy/foreman-proxy.pid'
-#:daemon_pid: /var/run/foreman-proxy/foreman-proxy.pid
-
 # host and ports configuration
 # an array of interfaces to bind ports to (possible values: *, localhost, 0.0.0.0)
 #:bind_host: ['*']
--- a/extra/systemd/foreman-proxy.service
+++ b/extra/systemd/foreman-proxy.service
@@ -6,7 +6,7 @@ After=basic.target network.target
 [Service]
 Type=notify
 User=foreman-proxy
-ExecStart=/usr/share/foreman-proxy/bin/smart-proxy --no-daemonize
+ExecStart=/usr/share/foreman-proxy/bin/smart-proxy
 Environment=MALLOC_ARENA_MAX=2
 
 [Install]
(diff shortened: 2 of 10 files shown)


ekohl:
@adamruzicka any idea if this breaks SmartProxyDynflowCore? From a quick glance I don't think it reuses any code, but can't tell for sure.

adamruzicka:
This shouldn't affect dynflow core in any way. It doesn't really care about daemonization related stuff when it runs as a plugin inside the smart proxy and when it runs as a separate service it uses its own implementation.

adamruzicka (approved):
LGTM

ekohl:
I noticed we still ship a sysvinit file here: https://github.com/theforeman/foreman-packaging/blob/deb/develop/debian/buster/foreman-proxy/foreman-proxy.init

I'll need to check if that breaks.

lzap (approved):
I am good, what's the status with debian initscript? It looks like this change will break it:

https://github.com/theforeman/foreman-packaging/blob/deb/develop/debian/buster/foreman-proxy/foreman-proxy.init#L53-L55

We need to rewrite it a bit and launch the process on the background, we can even use nohup or similar command if needed - this should be replaced by systemd soon once we stop supporting older Debian versions?

adamruzicka:
> once we stop supporting older Debian versions?

Don't all Debian and derivatives versions we support already have systemd?

ekohl:
Technically Debian still supports both sysvinit and systemd. However, I think we can choose to only support systemd (like we did with main Foreman).

@evgeni thoughts?

evgeni:
My life is to short for multiple inits.

ekohl:
I opened https://github.com/theforeman/foreman-packaging/pull/6045

lzap:
Debian packaging looks fine to me, but I will let @evgeni or others to do the final blessing.

ekohl:
I reached out to the FreeBSD developers and they said it's no problem to drop this. I'll rebase it shortly and push for this in Foreman Proxy 3.5 while deprecating it in Foreman Proxy 3.4.

evgeni:
@ekohl the diff looks sane to me, but I do not see the relevant deprecations in `3.4-stable` (or a PR doing so).

Given technically today is GA day for 3.4, we might be a bit late?

evgeni (approved):
formal ack, pending deprecations actually land in 3.4

ekohl:
> formal ack, pending deprecations actually land in 3.4

I'm wondering about those deprecations. In our systemd unit file we have an explicit `--no-daemonize` so the setting doesn't matter. It is then up to the packagers outside of RPM/deb to do the right thing. Would a changelog entry be sufficient?

evgeni:
> Would a changelog entry be sufficient?

You mean a changelog entry in `smart-proxy` (and in the release notes or whatever), but no code that does `logging.warn()`? That sounds good to me.

PR #778, opened 2020-10-26
Refs #30677 - support also urlencoded data

ares:
The default registration template in Foreman started to send the data in
urlencoded form instead of a JSON, as of https://projects.theforeman.org/issues/31043

In order to support that, we need to remove host parameter from query,
since Sinatra parses it from body automatically. We only need to proxy
the body of the request. ForemanRequestProxy also defaults to
application/json if not set explicitly, in this case we need to enforce
the correct content-type through headers.

---

comment to this PR: I tried to add a test for this, turns out the Rack test methods use url-encoded by default. I didn't find a good way to correctly distinguish between the two requests, given the testing target is stubbed anyway.

### DIFF
--- a/lib/proxy/request.rb
+++ b/lib/proxy/request.rb
@@ -26,7 +26,7 @@ module Proxy::HttpRequest
 
     def add_headers(req, headers = {})
       req.add_field('Accept', 'application/json,version=2')
-      req.content_type = headers["Content-Type"] || 'application/json'
+      req.content_type = headers.delete("Content-Type") || 'application/json'
       headers.each do |k, v|
         req.add_field(k, v)
       end
--- a/modules/registration/proxy_request.rb
+++ b/modules/registration/proxy_request.rb
@@ -10,11 +10,25 @@ module Proxy::Registration
       send_request(proxy_req)
     end
 
+    # we support two way of sending data - either a JSON or url encoded data
     def host_register(request)
-      proxy_req = request_factory.create_post '/register',
-                                              request.body.read,
-                                              headers(request),
-                                              request_params(request)
+      if request.content_type == 'application/x-www-form-urlencoded'
+        # the request has a different content type, ForemanRequestFactory sets content type to json, unless
+        # specified explicitly
+        # also request.params contain the same data that is in request.body, just parsed to hash,
+        # in case they are nested (e.g. host hash) we need this causes problem during CGI escaping
+        # therefore we only add url, everything else should be in body in this type of request
+        proxy_req = request_factory.create_post '/register',
+                                                request.body.read,
+                                                headers(request).merge("Content-Type" => request.content_type),
+                                                { url: register_url(request) }
+      else
+        # the application/json request body contains the data - JSON as a string, query contains only the URL
+        proxy_req = request_factory.create_post '/register',
+                                                request.body.read,
+                                                headers(request),
+                                                request_params(request)
+      end
 
       send_request(proxy_req)
     end
@@ -23,8 +37,12 @@ module Proxy::Registration
 
     def request_params(request)
       params = request.params
-      params[:url] = request.env['REQUEST_URI']&.split('/register')&.first
+      params[:url] = register_url(request)
       params
+    end
+
+    def register_url(request)
+      request.env['REQUEST_URI']&.split('/register')&.first
     end
 
     def headers(request)
--- a/test/registration/registration_api_test.rb
+++ b/test/registration/registration_api_test.rb
@@ -38,9 +38,17 @@ class RegistrationRegisterApiTest < Test::Unit::TestCase
   end
 
   def test_host_register_template_with_args
-    stub_request(:post, "#{@foreman_url}/register?host=test.example.com").to_return(body: 'template')
+    stub_request(:post, "#{@foreman_url}/register").to_return(body: 'template')
 
-    post '/', { host: 'test.example.com' }
+    post '/', { host: { name: 'test.example.com', build: false } }
+    assert last_response.ok?
+    assert_match('template', last_response.body)
+  end
+
+  def test_host_register_template_with_args_using_json
+    stub_request(:post, "#{@foreman_url}/register").to_return(body: 'template')
+
+    post '/', { host: { name: 'test.example.com', build: false } }, { 'CONTENT_TYPE' => 'application/json' }
     assert last_response.ok?
     assert_match('template', last_response.body)
   end


ares:
seems this breaks tests, will need to take a look in that

lzap (approved):
It works fine.

```
curl -u admin:changeme --data "host[name]=test.dummy.lan" --data "host[managed]=false" --data "host[operatingsystem_id]=1" http://localhost:8448/register

2020-10-26T17:44:37 6484fdfe [I] Started POST /register 
2020-10-26T17:44:37 6484fdfe [D] Headers: {"HTTP_HOST"=>"localhost:8448", "HTTP_AUTHORIZATION"=>"Basic YWRtaW46Y2hhbmdlbWU=", "HTTP_USER_AGENT"=>"curl/7.69.1", "HTTP_ACCEPT"=>"*/*", "HTTP_VERSION"=>"HTTP/1.1"}
2020-10-26T17:44:37 6484fdfe [D] Body: [filtered out]
2020-10-26T17:44:38  [D] accept: 127.0.0.1:47734
2020-10-26T17:44:38  [D] Rack::Handler::WEBrick is invoked.
2020-10-26T17:44:38 9f4a1150 [I] Started DELETE /ssh/known_hosts/test.dummy.lan 
2020-10-26T17:44:38 9f4a1150 [D] Headers: {"HTTP_ACCEPT"=>"application/json", "HTTP_USER_AGENT"=>"rest-client/2.1.0 (linux-gnu x86_64) ruby/2.7.1p83", "HTTP_X_REQUEST_ID"=>"9f4a1150-6e0a-4510-a611-fd637ad1912c", "HTTP_X_SESSION_ID"=>"9f4a1150-6e0a-4510-a611-fd637ad1912c", "HTTP_ACCEPT_ENCODING"=>"gzip;q=1.0,deflate;q=0.6,identity;q=0.3", "HTTP_HOST"=>"localhost:8448", "HTTP_VERSION"=>"HTTP/1.1"}
2020-10-26T17:44:38 9f4a1150 [I] Finished DELETE /ssh/known_hosts/test.dummy.lan with 404 (2.02 ms)
2020-10-26T17:44:38  [D] close: 127.0.0.1:47734
2020-10-26T17:44:38 6484fdfe [I] Finished POST /register with 200 (364.88 ms)
2020-10-26T17:44:38  [D] close: 127.0.0.1:47730
```

lzap:
There are test failures tho.

modules/registration/proxy_request.rb
```diff
@@ -10,11 +10,24 @@ def global_register(request)
...
-                                              headers(request),
-                                              request_params(request)
+      if request.content_type == 'application/x-www-form-urlencoded'
+        # the request has a different content type, ForemanRequestProxy sets content type to json, unless
+        # specified explicitly
+        # also request.params contain the same data that is in request.body, just parsed to hash, so we need
+        # to remove them
+        proxy_req = request_factory.create_post '/register',
+                                                request.body.read,
+                                                headers(request).merge("Content-Type" => "application/x-www-form-urlencoded"),
```
ekohl: This looks odd - it implies we filter it out which shouldn't happen, right?

modules/registration/proxy_request.rb
```diff
@@ -10,11 +10,24 @@ def global_register(request)
...
     end
 
+    # we support two way of sending data - either a JSON or url encoded data
     def host_register(request)
-      proxy_req = request_factory.create_post '/register',
-                                              request.body.read,
-                                              headers(request),
-                                              request_params(request)
+      if request.content_type == 'application/x-www-form-urlencoded'
+        # the request has a different content type, ForemanRequestProxy sets content type to json, unless
```
ekohl:
```suggestion
        # the request has a different content type, ForemanRequestFactory sets content type to json, unless
```

modules/registration/proxy_request.rb
ares: I'm not sure what you mean, but I'll try to explain a bit better

1) headers don't contain "Content-Type", this is how we build the headers list https://github.com/theforeman/smart-proxy/blob/develop/modules/registration/proxy_request.rb#L30-L35, we could keep it there, however that would always override the default that we later set (see point 2)
2) the problem then is https://github.com/theforeman/smart-proxy/blob/develop/lib/proxy/request.rb#L29, which defaults to JSON if no content-type is specified in headers, normally you set the content-type directly on request object

Therefore we need to add this explicitly.

ekohl: > headers don't contain "Content-Type", this is how we build the headers list https://github.com/theforeman/smart-proxy/blob/develop/modules/registration/proxy_request.rb#L30-L35, we could keep it there, however that would always override the default that we later set (see point 2)

So this is the confusing part for me: I assumed that the header was present. The problem here is normalization.

The request comes in with the env var `HTTP_CONTENT_TYPE`, which remains. So `header_request` will have `CONTENT_TYPE` set, but in point 2 as you pointed out only looks at `Content-Type`. It's not case insensitive (as HTTP headers are) and only looks at `-` instead of `_`. At least, that's my theory but I didn't check it out.

ares: ah it seems this will not work reliably though, the `add_headers` method adds the content type twice then which causes tests to fail, I'll need to find a better way how to set the custom content_type, probably the change won't be just in the registration module :-(

ares: > The request comes in with the env var HTTP_CONTENT_TYPE, which remains.

it does not, this is the result form pry above this line
```ruby
{"USER_AGENT"=>"curl/7.69.1",
 "ACCEPT"=>"*/*",
 "AUTHORIZATION"=>
  "Bearer eyJhbGciOiJIUzI1NiJ9.eyJ1c2VyX2lkIjo0LCJpYXQiOjE2MDM3NDAxMDcsImp0aSI6ImUzYThjZjZiYTY0MzgxZGE4ODAwZTEzZWY0NGYwMGVlMmZiMGU5NDZhYzExZjRiYTFmYjFmMTc5ZDA5NjczNzgiLCJleHAiOjE2MDM3NTQ1MDd9.DgnqIoWrhSZYsHqeUxY-08jwLtB5iBD8rApUrUuVYuM"}
```

request.env only contains these that start with HTTP
```ruby
{"HTTP_HOST"=>"foreman.example.tst:8443",
 "HTTP_USER_AGENT"=>"curl/7.69.1",
 "HTTP_ACCEPT"=>"*/*",
 "HTTP_AUTHORIZATION"=>
  "Bearer eyJhbGciOiJIUzI1NiJ9.eyJ1c2VyX2lkIjo0LCJpYXQiOjE2MDM3NDAxMDcsImp0aSI6ImUzYThjZjZiYTY0MzgxZGE4ODAwZTEzZWY0NGYwMGVlMmZiMGU5NDZhYzExZjRiYTFmYjFmMTc5ZDA5NjczNzgiLCJleHAiOjE2MDM3NTQ1MDd9.DgnqIoWrhSZYsHqeUxY-08jwLtB5iBD8rApUrUuVYuM",
 "HTTP_VERSION"=>"HTTP/1.1"}
```

`CONTENT_TYPE` is not prefixed with HTTP. But you're right, even if we kept that header, it wouldn't match because of case sensitivity.

There's another problem with `add_headers` method I mentioned in previous post, we must avoid adding the content-type twice. New version of the patch contains the fix for it. I also updated the git commit message for this change.

modules/registration/proxy_request.rb
ares: fixed

lib/proxy/request.rb
```diff
@@ -26,7 +26,7 @@ def uri(path)
 
     def add_headers(req, headers = {})
       req.add_field('Accept', 'application/json,version=2')
-      req.content_type = headers["Content-Type"] || 'application/json'
+      req.content_type = headers.delete("Content-Type") || 'application/json'
```
ares, line 29: without this, `req.add_field` on line 31 would add it again because we iterate over all headers, that caused the test problems

test/registration/registration_api_test.rb
```diff
@@ -38,7 +38,7 @@ def test_host_register_template
   end
 
   def test_host_register_template_with_args
-    stub_request(:post, "#{@foreman_url}/register?host=test.example.com").to_return(body: 'template')
+    stub_request(:post, "#{@foreman_url}/register").to_return(body: 'template')
```
ares, line 41: this test was always wrong, as it turned out, rack test methods defaults to url encoded meaning the data is passed in the body and should not be passed as the query string, this is what this PR implements, therefore changing it the stub this test calls

adding also a test for JSON branch

ares:
Tests are green now. This should get into 2.3, so a quick merge would be great. Trust me, I've tested this thoroughly :-) I can imagine the `headers` method keepint the `CONTENT_TYPE` header, however the code wouldn't be much nicer, since we'd need to convert it to `Content-Type` anyway. I'd prefer to keep it passing just HTTP_ headers.

modules/registration/proxy_request.rb
```diff
@@ -10,11 +10,24 @@ def global_register(request)
...
-                                              request_params(request)
+      if request.content_type == 'application/x-www-form-urlencoded'
+        # the request has a different content type, ForemanRequestFactory sets content type to json, unless
+        # specified explicitly
+        # also request.params contain the same data that is in request.body, just parsed to hash, so we need
+        # to remove them
+        proxy_req = request_factory.create_post '/register',
+                                                request.body.read,
+                                                headers(request).merge("Content-Type" => "application/x-www-form-urlencoded"),
+                                                request_params(request).tap { |params| params.delete("host") }
```
ekohl: Is this the only param to remove? If we already pass the body, do we even need to pass these?

modules/registration/proxy_request.rb
```diff
@@ -10,11 +10,24 @@ def global_register(request)
...
-                                              headers(request),
-                                              request_params(request)
+      if request.content_type == 'application/x-www-form-urlencoded'
+        # the request has a different content type, ForemanRequestFactory sets content type to json, unless
+        # specified explicitly
+        # also request.params contain the same data that is in request.body, just parsed to hash, so we need
+        # to remove them
+        proxy_req = request_factory.create_post '/register',
+                                                request.body.read,
+                                                headers(request).merge("Content-Type" => "application/x-www-form-urlencoded"),
```
ekohl:
```suggestion
                                                headers(request).merge("Content-Type" => request.content_type),
```

ekohl (approved):
Makes sense. Small inline questions but no objection to merging.

modules/registration/proxy_request.rb
lzap: My understanding is that Sinatra creates this, cleaner would be probably disable Sinatra parsing but I don't know how to do this.

lzap:
LGTM commit Ewouds suggestion and rebase to pass RM test please.

modules/registration/proxy_request.rb
ares: Right now, I believe it's the only parameter, there will be more in the future (setup rex, setup insigts), but at this moment it's the only param. Also it's the only param causing problems, because it's nested hash, if it's single level, e.g. `setup_rex: false` it properly escaped and doesn't cause issues.

Considering all pros and cons, I think it's better to turn the logic and keep the url parameter, drop everything else (not just the host).

ekohl (approved):

ares:
Updated both, retested and works. I'm not sure why, but the redmine issue check is stuck since yesterday evening. That should not block the merge I hope :-) Tests are green.

ekohl:
Thanks!

ares:
Thanks both for the prompt review, registration now works end to end in 2.3 even through proxy.

PR #782, opened 2020-12-21
Fixes #31537 - Expose BMC providers as capabilities

ekohl:
Currently the BMC module exposes an API endpoint /providers/installed but this can also be provided as capabilities or settings. This allows Foreman to determine up front if a proxy is even a valid choice without making API requests over the network.

The load_classes part is needed because capabilities are exectuted before loading the main API (which already has the same require). The require is explicit in the tests to stub out the implementation, making the tests reliable, regardless of which providers are installed.

While writing up the issue, I realized that perhaps this is hard to consume on the Foreman side. There is no way to recognize a capability is actually a provider. Assuming every capability is a provider is also wrong.

For this I can think of two solutions:
* Prefix all provider capabilities with `provider_`
* Expose installed providers as a setting

Rather than coming up with essentially 3 PRs I decided to submit this one (since I already wrote the code) and have the discussion here.

### DIFF
--- a/modules/bmc/bmc_plugin.rb
+++ b/modules/bmc/bmc_plugin.rb
@@ -5,5 +5,16 @@ module Proxy::BMC
     default_settings :redfish_verify_ssl => true
     validate :redfish_verify_ssl, :boolean => true
     plugin :bmc, ::Proxy::VERSION
+
+    # Various installed providers are exposed as capabilties
+    capability 'redfish'
+    capability 'shell'
+    capability 'ssh'
+    capability -> { Proxy::BMC::IPMI.providers_installed }
+
+    # Load IPMI to ensure the capabilities can be determined
+    load_classes do
+      require 'bmc/ipmi'
+    end
   end
 end
--- a/test/bmc/integration_test.rb
+++ b/test/bmc/integration_test.rb
@@ -2,6 +2,7 @@ require 'test_helper'
 require 'json'
 require 'root/root_v2_api'
 require 'bmc/bmc'
+require 'bmc/ipmi'
 
 class BmcApiFeaturesTest < Test::Unit::TestCase
   include Rack::Test::Methods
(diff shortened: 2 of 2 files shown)


modules/bmc/bmc_plugin.rb
```diff
@@ -5,5 +5,16 @@ class Plugin < Proxy::Plugin
     default_settings :redfish_verify_ssl => true
     validate :redfish_verify_ssl, :boolean => true
     plugin :bmc, ::Proxy::VERSION
+
+    # Various installed providers are exposed as capabilties
+    capability 'redfish'
+    capability 'shell'
+    capability 'ssh'
+    capability -> { Proxy::BMC::IPMI.providers_installed }
```
lzap, line 13: Please extract non-ipmi providers into a constant and use that here and also in the `non_ipmi_providers` method (or elsewhere when appropriate) so there's only one place to edit.

test/bmc/integration_test.rb
```diff
@@ -21,7 +23,23 @@ def test_features
     mod = response['bmc']
     refute_nil(mod)
     assert_equal('running', mod['state'], Proxy::LogBuffer::Buffer.instance.info[:failed_modules][:bmc])
-    assert_equal([], mod['capabilities'])
+    assert_equal(['redfish', 'shell', 'ssh'], mod['capabilities'])
```
lzap, line 26: This could use the frozen constant too.

lzap (changes requested):
Sure, that works. A nitpick thos.

test/bmc/integration_test.rb
ekohl: are frozen constants really that useful in tests?

modules/bmc/bmc_plugin.rb
ekohl: What would be a good place to store it? `Proxy::BMC::IPMI.non_ipmi_providers` doesn't make sense.

lzap:
After a second thought, I think this is fine. No big deal.

PR #783, opened 2021-01-15
Fixes #32329 - Introduce instance id

adamruzicka:
Part of https://community.theforeman.org/t/infrastructure-roles/22001

### DIFF
--- a/config/settings.yml.example
+++ b/config/settings.yml.example
@@ -86,3 +86,5 @@
 # each successive interval in the array until a successful response
 # is received. See Ruby Resolv#timeouts documentation for more info.
 #:dns_resolv_timeouts: [5, 8, 13]
+
+#:instance_uuid:
--- a/modules/root/root_api.rb
+++ b/modules/root/root_api.rb
@@ -14,7 +14,7 @@ class Proxy::RootApi < Sinatra::Base
     content_type :json
     enabled_plugins = ::Proxy::Plugins.instance.select { |p| p[:state] == :running && p[:class].ancestors.include?(::Proxy::Plugin) }
     modules = Hash[enabled_plugins.map { |plugin| [plugin[:name].to_s, plugin[:version].to_s] }].reject { |key| key == 'foreman_proxy' }
-    {:version => Proxy::VERSION, :modules => modules}.to_json
+    {:version => Proxy::VERSION, :modules => modules, :uuid => ::Proxy::SETTINGS[:instance_uuid]}.to_json
   rescue => e
     log_halt 400, e
   end
--- a/lib/proxy/settings/global.rb
+++ b/lib/proxy/settings/global.rb
@@ -19,7 +19,7 @@ module ::Proxy::Settings
       :ssl_disabled_ciphers => [],
       :tls_disabled_versions => [],
       :dns_resolv_timeouts => [5, 8, 13], # Ruby default is [5, 20, 40] which is a bit too much for us
-      :instance_uuid => nil
+      :instance_uuid => nil,
     }
 
     HOW_TO_NORMALIZE = {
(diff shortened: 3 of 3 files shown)


config/settings.yml.example
```diff
@@ -86,3 +86,5 @@
 # each successive interval in the array until a successful response
 # is received. See Ruby Resolv#timeouts documentation for more info.
 #:dns_resolv_timeouts: [5, 8, 13]
+
+#:instance_uuid:
```
lzap: Missing endline, other editors will insist on it, can you add it?

lzap:
Generally, I have no problems with this patch. There is ongoing discussion in the discourse, let me know when get back to this when we get consensus.

lzap:
I was wondering if you can use machine id provided by `hostnamectl`, a systemd utility. There is a big chance this is/will be reported by puppet/ansible, if not we can probably push a patch upstream so no custom facts are needed in the future.

adamruzicka:
> I was wondering if you can use machine id

It was briefly touched on in the thread on discourse. This would be a way if we didn't support running the proxy on non-systemd platforms such as windows. https://community.theforeman.org/t/infrastructure-roles/22001/14

lzap:
We also have `modules/root/root_v2_api.rb` which needs authorization by default. I am wondering it that is a better place to put UUID in. In the long term, I think we want to deprecate the old endpoint too.

I am thinking how capabilities could be helpful here, something like `has_uuid` but I do not see any advantage from that.

lzap:
Oh I think there is a mechanism to expose settings via `expose_setting` but that can probably be only used for plugins. This could actually be (always on) plugin but that smells like overengineering. I am fine with this, maybe moving it into V2 root? @ekohl

adamruzicka:
> I am wondering it that is a better place to put UUID in. In the long term,

It would indeed be a better place, but as far as I know, Foreman doesn't hit this endpoint to get information about the proxy, which is a bit of a showstopper

modules/root/root_api.rb
```diff
@@ -14,7 +14,7 @@ class Proxy::RootApi < Sinatra::Base
     content_type :json
     enabled_plugins = ::Proxy::Plugins.instance.select { |p| p[:state] == :running && p[:class].ancestors.include?(::Proxy::Plugin) }
     modules = Hash[enabled_plugins.map { |plugin| [plugin[:name].to_s, plugin[:version].to_s] }].reject { |key| key == 'foreman_proxy' }
-    {:version => Proxy::VERSION, :modules => modules}.to_json
+    {:version => Proxy::VERSION, :modules => modules, :uuid => ::Proxy::SETTINGS[:instance_uuid]}.to_json
```
ekohl, line 17: The version endpoint is a public unauthenticated API. Don't we consider the instance UUID as something private?

ekohl (commented):
Like @lzap, I am worried about exposing the data. While the root is a "plugin" but it does not show up in the V2 features.

In hindsight, I am unhappy I didn't introduce another layer in the v2 features. Right now it's a hash plugin -> plugin_data. However, I think it should have been a generic proxy information block at the top level and then a plugins -> $CURRENT_OUTPUT property. That would have been the ideal place for something like this. Other things that we may want there is the HTTP and HTTPS ports so you know the full identity of the Smart Proxy. Perhaps this effort is a good excuse to introduce that new endpoint.

modules/root/root_api.rb
adamruzicka: I wouldn't say the instance uuid is something private. The signature is.

ares: Personally I don't see this as a private information. It's random, so it only allows correlation, not identification. The version is IMHO much more sensitive as it allows to find out known vulnerabilities quicker.

ares:
Is this still POC? Seems ready to me.

lzap (approved):
I am good but more important review is Ewoud's one who specialize in proxy, distributed setups and installer too.

lzap:
There is also redmine issue error, please rebase.

adamruzicka:
Not needed anymore with v2 https://github.com/theforeman/foreman/pull/8632 , closing.

PR #785, opened 2021-02-09
Fixes #2412 - don't continue wget downloads

lzap:
With CentOS 8 Stream this is a more painful issue, I was planning rewriting this code but there is no time - let's change -c to --timestamping --no-if-modified-since. This will keep the current behavior - if a file is present and the same size and its timestamp matches what server reports, do not redownload it. If it's different, redownload it from the beginning - no longer attempt to continue and append.

The second option `--no-if-modified-since` makes `wget` to request the timestamp via HEAD HTTP call instead of If-Modified-Since because not all mirrors support this option:

```
$ LC_ALL=C wget --timestamping --no-if-modified-since http://mirror.centos.org/centos/8-stream/BaseOS/x86_64/os/images/pxeboot/vmlinuz
--2021-02-09 15:02:49--  http://mirror.centos.org/centos/8-stream/BaseOS/x86_64/os/images/pxeboot/vmlinuz
Resolving mirror.centos.org (mirror.centos.org)... 95.216.19.148, 2a00:1dc0:3061::10
Connecting to mirror.centos.org (mirror.centos.org)|95.216.19.148|:80... connected.
HTTP request sent, awaiting response... 200 OK
Length: 9936008 (9.5M)
Server file no newer than local file 'vmlinuz' -- not retrieving.

$ LC_ALL=C wget --timestamping http://mirror.centos.org/centos/8-stream/BaseOS/x86_64/os/images/pxeboot/vmlinuz
--2021-02-09 15:02:55--  http://mirror.centos.org/centos/8-stream/BaseOS/x86_64/os/images/pxeboot/vmlinuz
Resolving mirror.centos.org (mirror.centos.org)... 95.216.19.148, 2a00:1dc0:3061::10
Connecting to mirror.centos.org (mirror.centos.org)|95.216.19.148|:80... connected.
HTTP request sent, awaiting response... 304 Not Modified
File 'vmlinuz' not modified on server. Omitting download.

$ LC_ALL=C wget --timestamping http://ftp.sh.cvut.cz/centos/8-stream/BaseOS/x86_64/os/images/pxeboot/initrd.img
--2021-02-09 15:03:20--  http://ftp.sh.cvut.cz/centos/8-stream/BaseOS/x86_64/os/images/pxeboot/initrd.img
Resolving ftp.sh.cvut.cz (ftp.sh.cvut.cz)... 147.32.127.196, 2001:718:2::196
Connecting to ftp.sh.cvut.cz (ftp.sh.cvut.cz)|147.32.127.196|:80... connected.
HTTP request sent, awaiting response... 200 OK
Server ignored If-Modified-Since header for file 'initrd.img'.
You might want to add --no-if-modified-since option.
```

### DIFF
--- a/lib/proxy/http_download.rb
+++ b/lib/proxy/http_download.rb
@@ -14,11 +14,15 @@ module Proxy
       dns_timeout ||= DEFAULT_CONNECT_TIMEOUT
       connect_timeout ||= DEFAULT_DNS_TIMEOUT
 
-      args = [wget, "--connect-timeout=#{connect_timeout}",
+      args = ["--connect-timeout=#{connect_timeout}",
               "--dns-timeout=#{dns_timeout}",
               "--read-timeout=#{read_timeout}",
-              "--tries=3", "-nv", "-c", src.to_s, "-O", dst.to_s]
-      args << "--no-check-certificate" unless verify_server_cert
+              "--tries=3",
+              "--timestamping", # turn on timestamping to prevent redownloads
+              "--no-if-modified-since", # but use HTTP HEAD instead If-Modified-Since
+              "-nv", src.to_s, "-O", dst.to_s]
+      args.unshift("--no-check-certificate") unless verify_server_cert
+      args.unshift(wget)
       super(args)
     end
 
--- a/lib/proxy/plugin_initializer.rb
+++ b/lib/proxy/plugin_initializer.rb
@@ -114,7 +114,7 @@ class ::Proxy::PluginGroup
   end
 
   def fail_group(an_exception)
-    fail_group_with_message("Disabling all modules in the group #{printable_module_names(member_names)} due to a failure in one of them: #{an_exception}", an_exception.backtrace)
+    fail_group_with_message("Disabling all modules in the group #{printable_module_names(member_names)} due to a failure in one of them: #{an_exception}", an_exception)
   end
 
   def fail_group_with_message(a_message, an_exception = nil)
--- a/test/http_download_test.rb
+++ b/test/http_download_test.rb
@@ -12,10 +12,14 @@ class HttpDownloadsTest < Test::Unit::TestCase
     default_dns = Proxy::HttpDownload::DEFAULT_DNS_TIMEOUT
 
     expected = ["/wget",
+                "--no-check-certificate",
                 "--connect-timeout=#{default_connect}",
                 "--dns-timeout=#{default_dns}",
                 "--read-timeout=#{default_read}",
-                "--tries=3", "-nv", "-c", "src", "-O", "dst", "--no-check-certificate"]
+                "--tries=3",
+                "--timestamping",
+                "--no-if-modified-since",
+                "-nv", "src", "-O", "dst"]
     Proxy::HttpDownload.any_instance.stubs(:which).returns('/wget')
     assert_equal expected, Proxy::HttpDownload.new('src', 'dst').command
   end
@@ -29,7 +33,10 @@ class HttpDownloadsTest < Test::Unit::TestCase
                 "--connect-timeout=#{default_connect}",
                 "--dns-timeout=#{default_dns}",
                 "--read-timeout=#{default_read}",
-                "--tries=3", "-nv", "-c", "src", "-O", "dst"]
+                "--tries=3",
+                "--timestamping",
+                "--no-if-modified-since",
+                "-nv", "src", "-O", "dst"]
     Proxy::HttpDownload.any_instance.stubs(:which).returns('/wget')
     assert_equal expected, Proxy::HttpDownload.new('src', 'dst', nil, nil, nil, true).command
   end
@@ -40,10 +47,14 @@ class HttpDownloadsTest < Test::Unit::TestCase
 
     read_timeout = 1000
     expected = ["/wget",
+                "--no-check-certificate",
                 "--connect-timeout=#{default_connect}",
                 "--dns-timeout=#{default_dns}",
                 "--read-timeout=#{read_timeout}",
-                "--tries=3", "-nv", "-c", "src", "-O", "dst", "--no-check-certificate"]
+                "--tries=3",
+                "--timestamping",
+                "--no-if-modified-since",
+                "-nv", "src", "-O", "dst"]
     Proxy::HttpDownload.any_instance.stubs(:which).returns('/wget')
     assert_equal expected, Proxy::HttpDownload.new('src', 'dst', read_timeout, nil, nil).command
   end
@@ -57,7 +68,10 @@ class HttpDownloadsTest < Test::Unit::TestCase
                 "--connect-timeout=#{default_connect}",
                 "--dns-timeout=#{default_dns}",
                 "--read-timeout=#{read_timeout}",
-                "--tries=3", "-nv", "-c", "src", "-O", "dst"]
+                "--tries=3",
+                "--timestamping",
+                "--no-if-modified-since",
+                "-nv", "src", "-O", "dst"]
     Proxy::HttpDownload.any_instance.stubs(:which).returns('/wget')
     assert_equal expected, Proxy::HttpDownload.new('src', 'dst', read_timeout, nil, nil, true).command
   end
@@ -67,10 +81,14 @@ class HttpDownloadsTest < Test::Unit::TestCase
     connect_timeout = 99
     dns_timeout = 27
     expected = ["/wget",
+                "--no-check-certificate",
                 "--connect-timeout=#{connect_timeout}",
                 "--dns-timeout=#{dns_timeout}",
                 "--read-timeout=#{read_timeout}",
-                "--tries=3", "-nv", "-c", "src", "-O", "dst", "--no-check-certificate"]
+                "--tries=3",
+                "--timestamping",
+                "--no-if-modified-since",
+                "-nv", "src", "-O", "dst"]
     Proxy::HttpDownload.any_instance.stubs(:which).returns('/wget')
     assert_equal expected, Proxy::HttpDownload.new('src', 'dst', read_timeout, connect_timeout, dns_timeout).command
   end
@@ -83,7 +101,10 @@ class HttpDownloadsTest < Test::Unit::TestCase
                 "--connect-timeout=#{connect_timeout}",
                 "--dns-timeout=#{dns_timeout}",
                 "--read-timeout=#{read_timeout}",
-                "--tries=3", "-nv", "-c", "src", "-O", "dst"]
+                "--tries=3",
+                "--timestamping",
+                "--no-if-modified-since",
+                "-nv", "src", "-O", "dst"]
     Proxy::HttpDownload.any_instance.stubs(:which).returns('/wget')
     assert_equal expected, Proxy::HttpDownload.new('src', 'dst', read_timeout, connect_timeout, dns_timeout, true).command
   end


ekohl:
Haven't looked at the code yet, but there was already a rewrite of this code: https://github.com/theforeman/smart-proxy/pull/472

lzap:
And if you read through it you will come to conclusion that I did not like the idea of replacing a simple `wget` call with our own Ruby code. We'd need to handle edge cases, timestamp, etag, redirects. And then the discussion went into changing our architecture from fire-and-forget to request-response.

Our TFTP orchestration has some flaws and I think we should tackle them, this patch provides a simple solution to the "continue and corrupt files" problem we will see more often with CentOS 8 Stream. This is a simple change, have a look.

lzap:
I think this one is important one to solve, anyone can take a look please? @ezr-ondrej @ares

lzap:
Any chance to get this in @ekohl ? I do not want to rewrite HTTP downloading class from scratch when it's not necessary.

ares:
@lzap could you please rebase? I'm happy to test then, please ping me directly when it's ready, I sometimes miss the github notifications.

lzap:
Rebased, resolved conflicts.

lib/proxy/http_download.rb
```diff
@@ -14,11 +14,15 @@ def initialize(src, dst, read_timeout = nil, connect_timeout = nil, dns_timeout
...
               "--dns-timeout=#{dns_timeout}",
               "--read-timeout=#{read_timeout}",
-              "--tries=3", "-nv", "-c", src.to_s, "-O", dst.to_s]
-      args << "--no-check-certificate" unless verify_server_cert
+              "--tries=3",
+              "--timestamping", # turn on timestamping to prevent redownloads
+              "--no-if-modified-since", # but use HTTP HEAD instead If-Modified-Since
+              "-nv", src.to_s, "-O", dst.to_s]
+      args.unshift("--no-check-certificate") unless verify_server_cert
+      args.unshift(wget)
```
lzap, line 25: Took the opportunity to change behavior in this new code - instead of appending cert flag, I am prepending it. Looks cleaner to me.

lib/proxy/plugin_initializer.rb
```diff
@@ -114,7 +114,7 @@ def configure
   end
 
   def fail_group(an_exception)
-    fail_group_with_message("Disabling all modules in the group #{printable_module_names(member_names)} due to a failure in one of them: #{an_exception}", an_exception.backtrace)
+    fail_group_with_message("Disabling all modules in the group #{printable_module_names(member_names)} due to a failure in one of them: #{an_exception}", an_exception)
```
lzap, line 117: This was not logging backtrace properly, I ran into this when debugging this.

lzap:
Bump, this is coming back over and over again. Can we fix this? I propose 3.0 backport because CentOS Stream made this bug much worse, symptoms are extremely difficult to figure out (Anaconda showing "Dead pane" errors, booting randomly stuck, kernel reports invalid module versions etc depending on how lucky you are).

lib/proxy/http_download.rb
ekohl: This makes it much more effort to review these patches. This is really the reason these things aren't merged quickly. If you would at least separate the commits then it'd be easier.

ekohl (approved):

lib/proxy/http_download.rb
gvde: > Took the opportunity to change behavior in this new code - instead of appending cert flag, I am prepending it. Looks cleaner to me.

In the context of cleaning up: wouldn't it be better to consistently only use long options, i.e. --no-verbose and --output-document instead of -nv and -O?

lzap:
Thanks, CP: https://github.com/theforeman/smart-proxy/pull/798

lib/proxy/http_download.rb
ekohl: Yes, I thought about the same thing but didn't want to drag on this PR further. IMHO it should have been 1 PR to change the behavior and one PR to change the coding style.

lib/proxy/http_download.rb
```diff
@@ -14,11 +14,15 @@ def initialize(src, dst, read_timeout = nil, connect_timeout = nil, dns_timeout
...
 
-      args = [wget, "--connect-timeout=#{connect_timeout}",
+      args = ["--connect-timeout=#{connect_timeout}",
               "--dns-timeout=#{dns_timeout}",
               "--read-timeout=#{read_timeout}",
-              "--tries=3", "-nv", "-c", src.to_s, "-O", dst.to_s]
-      args << "--no-check-certificate" unless verify_server_cert
+              "--tries=3",
+              "--timestamping", # turn on timestamping to prevent redownloads
+              "--no-if-modified-since", # but use HTTP HEAD instead If-Modified-Since
```
gvde, line 22: --no-if-modified-since doesn't work for EL7 wget. This will break it for proxies running on EL7.

lzap: Damn, I did not expect a recent argument in wget and tested on Fedora. Does it work reliably with Katello without this option?

The argument is useful, some mirrors do not implement the header. Maybe a setting to be able to enable this? But on RHEL7 this will redownload the file from mirrors which do not support this.

gvde: > The argument is useful, some mirrors do not implement the header. Maybe a setting to be able to enable this? But on RHEL7 this will redownload the file from mirrors which do not support this.

I am afraid timestamping won't be helpful at all. Manual page of wget says:
```
For this reason, -N (for timestamp-checking) is not supported in
combination with -O: since file is always newly created, it will
always have a very new timestamp. A warning will be issued if this
combination is used.
```

Trying it warns about it:
```
$ wget -N -O /dev/null
WARNING: timestamping does nothing in combination with -O. See the manual
for details.
```

lzap: We don't use `-O /dev/null`, where did you find that?

lzap: Damn, the patch is all wrong.

lzap:
So after reading a ton of manual pages I recommend to ditch `wget` and replace it with `curl`. The version in RHEL7 is more capable than `wget` and offers more flexibility. This is what I think are the best arguments (tested on Katello deployment on RHEL7):

Download fresh kernel file:

```
# LC_ALL=C curl --silent --show-error --connect-timeout 5 --retry 5 --retry-delay 10 --max-time 9000 -o /tmp/test --remote-time --time-cond /tmp/test --location http://zzzap.xxx.redhat.com/pulp/repos/Default_Organization/Library/content/dist/rhel8/8.0/s390x/baseos/kickstart/images/kernel.img --write-out "Download done, result: %{http_code}, size downloaded: %{size_download}b, speed: %{speed_download}b/s, time: %{time_total}ms\n"
Download done, result: 200, size downloaded: 4799272b, speed: 81357382.000b/s, time: 0.059ms
```

Re-download the same file again:

```
# LC_ALL=C curl --silent --show-error --connect-timeout 5 --retry 5 --retry-delay 10 --max-time 9000 -o /tmp/test --remote-time --time-cond /tmp/test --location http://zzzap.xxx.com/pulp/repos/Default_Organization/Library/content/dist/rhel8/8.0/s390x/baseos/kickstart/images/kernel.img --write-out "Download done, result: %{http_code}, size downloaded: %{size_download}b, speed: %{speed_download}b/s, time: %{time_total}ms\n"
Download done, result: 304, size downloaded: 0b, speed: 0.000b/s, time: 0.191ms
```

Download and redownload from CentOS official mirror:

```
# LC_ALL=C curl --silent --show-error --connect-timeout 5 --retry 5 --retry-delay 10 --max-time 9000 -o /tmp/test --remote-time --time-cond /tmp/test --location --write-out "Download done, result: %{http_code}, size downloaded: %{size_download}b, speed: %{speed_download}b/s, time: %{time_total}ms\n" http://mirror.centos.org/centos/8-stream/BaseOS/x86_64/os/images/pxeboot/vmlinuz
Download done, result: 200, size downloaded: 10218632b, speed: 12865244.000b/s, time: 0.794ms

# LC_ALL=C curl --silent --show-error --connect-timeout 5 --retry 5 --retry-delay 10 --max-time 9000 -o /tmp/test --remote-time --time-cond /tmp/test --location --write-out "Download done, result: %{http_code}, size downloaded: %{size_download}b, speed: %{speed_download}b/s, time: %{time_total}ms\n" http://mirror.centos.org/centos/8-stream/BaseOS/x86_64/os/images/pxeboot/vmlinuz
Download done, result: 304, size downloaded: 0b, speed: 0.000b/s, time: 0.136ms
```

The same but from one of the Czechia mirrors without support of the HTTP header mentioned above:

```
# LC_ALL=C curl --silent --show-error --connect-timeout 5 --retry 5 --retry-delay 10 --max-time 9000 -o /tmp/test --remote-time --time-cond /tmp/test --location --write-out "Download done, result: %{http_code}, size downloaded: %{size_download}b, speed: %{speed_download}b/s, time: %{time_total}ms\n" http://ftp.sh.cvut.cz/centos/8-stream/BaseOS/x86_64/os/images/pxeboot/vmlinuz
Download done, result: 200, size downloaded: 10218632b, speed: 53411485.000b/s, time: 0.191ms

# LC_ALL=C curl --silent --show-error --connect-timeout 5 --retry 5 --retry-delay 10 --max-time 9000 -o /tmp/test --remote-time --time-cond /tmp/test --location --write-out "Download done, result: %{http_code}, size downloaded: %{size_download}b, speed: %{speed_download}b/s, time: %{time_total}ms\n" http://ftp.sh.cvut.cz/centos/8-stream/BaseOS/x86_64/os/images/pxeboot/vmlinuz
Download done, result: 304, size downloaded: 0b, speed: 0.000b/s, time: 0.040ms
```

So overall it works on RHEL7 in the following scenarios:

* Katello / Pulp (with redirects)
* Official CentOS mirror running http server with HTTP header support
* Official CentOS mirror running http server without HTTP header support

I am going to do a new Refs patch that will refactor this completely. 

Advantages: minimum changes, reliable downloader, support for FTP and other protocols for the future
Disadvantages: does not support dns/read timeouts

gvde:
Are you sure you want to do the timestamp checking? There are a lot of mirrors out there and if you use a mirrorlist you can never be sure where you end up. Do you really want to rely on that all possible mirrors put the correct upstream timestamp on all files at all times?

Pulp for instance doesn't do that. I think it just uses the time when the file was synced. Other people probably use their own local mirror and you also cannot make too many assumptions on how they sync their local mirror to the upstream repo.

I am not sure that there won't be any scenarios where the upstream timestamp goes backward. I think a correct implementation would rely on the http cache information given, i.e. basically on the etag and would check for that.

lzap:
> Pulp for instance doesn't do that. I think it just uses the time when the file was synced. Other people probably use their own local mirror and you also cannot make too many assumptions on how they sync their local mirror to the upstream repo.

Yes, I can confirm that. And that is the correct behavior - when content is synced I want TFTP proxy to redownload the file.

> I think a correct implementation would rely on the http cache information given, i.e. basically on the etag and would check for that.

I believe ETag is implemented as a hash of modification time anyway, look the source and Czechia mirror have both the very same ETag:

```
$ curl -v -o /dev/null http://mirror.centos.org/centos/8-stream/BaseOS/x86_64/os/images/pxeboot/vmlinuz
< Server: Apache/2.4.6 (CentOS)
< Last-Modified: Fri, 27 Aug 2021 17:41:20 GMT
< ETag: "9bec88-5ca8dfdc07000"

$ curl -v -o /dev/null http://mirror.centos.org/centos/8-stream/BaseOS/x86_64/os/images/pxeboot/vmlinuz
< Server: Apache/2.4.6 (CentOS)
< Last-Modified: Fri, 27 Aug 2021 17:41:20 GMT
< ETag: "9bec88-5ca8dfdc07000"
```

Out of curiousity, Katello/Pulp supports ETag as well. But as I said above, I do not think it is getting us anywhere, also, it will not be possible to use `curl` or `wget`:

```
$ curl -v -o /dev/null http://zzzap.xxx.redhat.com/pulp/repos/Default_Organization/Library/content/dist/rhel8/8.0/s390x/baseos/kickstart/images/kernel.img
< Server: Apache
< ETag: "d41d8cd98f00b204e9800998ecf8427e"
< Last-Modified: Thu, 02 Sep 2021 07:15:19 GMT
```

Edit: Wikipedia confirms: "Common methods of ETag generation include using a collision-resistant hash function of the resource's content, a hash of the last modification timestamp, or even just a revision number."

Edit: I think you mean if it is possible not to compare newer/older but simply "if it differs redownload". Curl cannot do that unfortunately, but I think compared to what we have today (continue, corrupting files) it is much better to use `curl`.

lzap:
Here is my proposal: https://github.com/theforeman/smart-proxy/pull/799

I do not want to maintain Ruby code that would do ETag comparisons honestly. It should work fine with Katello and with regular mirrors which copy files keeping modification timestamp - that is a good practice. I agree that there could be a problem when a new boot files were released with older modification time, but that is a rare situation.

gvde:
The etag for http://ftp.sh.cvut.cz/centos/8-stream/BaseOS/x86_64/os/images/pxeboot/vmlinuz is
```
ETag: "612923c0-9bec88"
```

It's different. 

Downloading when the timestamp is different should always be the case. But even if the timestamp was identical, it could be an issue, if the download was interrupted and the file is incomplete, but still got the correct timestamp. I am not saying this is a very likely scenario or even realistic, but the timestamp is really just a timestamp and it can be modified quite easily and doesn't indicate what the content really is.

From the past, I have seen files with incorrectly timestamp offsets because a server was incorrectly configured to interpret the hardware clock as UTC but as local time, please ntpd was not configured to boost and jump quickly, so basically your server ran a couple of minutes with a local clock which was off a couple hours until the time synchronization adjusted the time again.

The purpose of etag for http is exactly what you technically need: it allows you to check if the resource you access is the same or not. I know that with curl or wget it requires you to store the etag separately. But for what you really want etag is the real deal. Timestamps are just a workaround with a lot of variables in my opinion which makes me wonder if you wouldn't simply download the file each time instead of hoping the all timestamps on all mirrors are really correct and in sync...

ekohl:
`man curl` on Fedora says there is `--etag-compare <file>` and `--etag-save <file>` since 7.68.0 but sadly EL7 has curl 7.29.0. Even EL8 only has 7.61.1. That would have been exactly what we were looking for but sadly we can't use it.

gvde:
> `man curl` on Fedora says there is `--etag-compare <file>` and `--etag-save <file>` since 7.68.0 but sadly EL7 has curl 7.29.0. Even EL8 only has 7.61.1. That would have been exactly what we were looking for but sadly we can't use it.

Well, for the time being, you could use --dump-header to get access to the etag header and then call curl with --header to add the necessary header for the etag check.

ekohl:
Sure, but relying on built in functionality with dedicated parameters is a much better guarantee for a correct implementation than building it yourself. I much rather rely on a well tested tool like curl than my own code ;)

lzap:
> It's different.

Yeah you are right, I accidently pasted wrong output and made an incorrect conclusion. It can be a different webserver or even there is some kind of seed that is generated on installation. Looks like Apache httpd calculates it from inode too:

http://httpd.apache.org/docs/2.2/mod/core.html#FileETag

Basically, any copy has a different ETag, even if its modified to have the same last modified attribute.

> I much rather rely on a well tested tool like curl than my own code

I agree that not writing a HTTP client is key, there are lot of quirks we would need to resolve. Having said that, parsing STDOUT for header is something that can also break apart. That is also quite bad code to write and maintain.

How about this, since we are very close to upgrading Foreman to EL8 by default for end users, I will introduce a new setting and feature to use etags from newer curl. On EL7 deployments it will default to last modified, on EL8 deployments it will use etags. Our puppet installer could even detect if curl supports it or not and set appropriate value automatically so even Debian users would benefit from this.

ekohl:
> I agree that not writing a HTTP client is key, there are lot of quirks we would need to resolve. Having said that, parsing STDOUT for header is something that can also break apart. That is also quite bad code to write and maintain.

I was talking about curl 7.68.0's CLI options for etag handling, not dumping headers.

> How about this, since we are very close to upgrading Foreman to EL8 by default for end users, I will introduce a new setting and feature to use etags from newer curl

As I mentioned, curl is too old on EL8 so you can't use those CLI options. Debian Buster and Ubuntu Bionic don't either. Debian Bullseye (not supported yet) and Debian Focal do have sufficiently new versions.

So while I wouldn't mind a setting for this, it wouldn't benefit the majority of our users today.

lzap:
> it wouldn't benefit the majority of our users today.

But as the time passes it would.

> Well, for the time being, you could use --dump-header to get access to the etag header and then call curl with --header to add the necessary header for the etag check.

This code needs to work for both HTTP and FTP and ETags are not available for FTP at all.

Guys, let's be realistic - what are we trying to solve here with more complex etags (the value must be stored in a separate file perhaps)? It's an artificial case when someone pushes out PXE files with last modified older than previous version. Major linux distributions do have strict release engineering policies, packages are being built in build systems which ensure consistency. At least in Fedora and Red Hat, you cannot build packages anywhere else than Koji.

TFTP downloads in Proxy are broken for a long time and CentOS Stream made it worse. On top of that, we just merged my patch which broke it completely, let's find a reasonable solution or revert the patch.

gvde:
> It's an artificial case when someone pushes out PXE files with last modified older than previous version. Major linux distributions do have strict release engineering policies, packages are being built in build systems which ensure consistency. At least in Fedora and Red Hat, you cannot build packages anywhere else than Koji.

In this case, it's not about packages but extra files and their timestamp in the file system of any mirror. Not everyone syncs/connects to the upstream, original servers. Just like if you connect to katello/pulp you'll see different timestamps than the original.

I think the tradeoff is between simply downloading each time it's needed and the risk that it doesn't download the latest version because the timestamp for whatever reason is older resulting in this really weird issue during provisioning which some may find hard to troubleshoot and figure out.

lzap:
> In this case, it's not about packages but extra files

You cannot ship a byte of any file in repositories that would not go through full building process in Fedora. Even those PXE files, they are properly built in Koji, then extracted, sha sums are created, index and listing files created etc. What I am trying to express is that the problem you describe is unlikely to happen at all.

ETags are a lot more complex than last modified time, there are strong and weak etags, the can be in different forms or shapes, server implementations can also have bugs and the code in smart proxy needs to be much more complex due to missing feature in `curl`. If something breaks, I honestly think it's the etags implementation.

I sense a dead end here, shall I simply revert the previous patch that broke nightly?

lib/proxy/http_download.rb
lzap: I agree, this is exactly what I did in: https://github.com/theforeman/smart-proxy/pull/799/files

Generally, I tend not to do such changes, but in this case I was planning to do backport in a more simple way (just replacing the arguments in question in a new commit).

lzap:
Bump, nightly is broken and I would like to fix this, opinions? Do you guys insist on ETags?

gvde:
My etag comment was meant to indicate how a "perfect" solution for http retrievals should look like, if you want to download the file only if something has changed because that's what etag does. So it seems that is more something for the future as ftp is also supposed to be supported.

So for the meantime, the modified time stamp is the next best thing and I guess it's true that it is rather unlikely to cause too many problems.

In the long run, I think there is a lot of room for improvement here. I was also thinking that the approach to check for those files could be completely different in a Katello environment compared to the scenario where you have to retrieve the files from some mirror...

ekohl:
What :point_up:  said. The risk with a mismatched timestamp is that you download it too often. As an admin I think I prefer that over corruption.

lzap:
Thank you! Let's continue then in the PR: #799

PR #786, opened 2021-03-23
Fixes #32164 - skip IPv6 link local addresses in FQDN lookups

lzap:
REMOTE_ADDR on IPv6 can contain %ident part with network interface identifier, this breaks our code.

I tried to pass this into IPAddr.new but Ruby does not understand the syntax, so I am proposing a dummy search and replace.

### DIFF
--- a/lib/proxy/helpers.rb
+++ b/lib/proxy/helpers.rb
@@ -83,6 +83,9 @@ module Proxy::Helpers
     ip = request.env['REMOTE_ADDR']
     log_halt 403, 'could not get remote address from environment' if ip.empty?
 
+    # IPv6 with NIC identifier are not resolved
+    return ip if ip.include?('%')
+
     begin
       dns = resolv
       fqdn = dns.getname(ip)


lib/proxy/helpers.rb
```diff
@@ -83,6 +83,9 @@ def remote_fqdn(forward_verify = true)
     ip = request.env['REMOTE_ADDR']
     log_halt 403, 'could not get remote address from environment' if ip.empty?
 
+    # IPv6 can contain %eno1 (NIC identifier) so strip it
+    ip = ip.sub(/%.*$/, '')
```
ekohl: I think we should return if the IP contains `%`. I don't think we'll find a reverse DNS for link local addresses and if we do, can we really trust it?

ekohl (changes requested):

lib/proxy/helpers.rb
lzap: Yeah makes sense, rebased.

lib/proxy/helpers.rb
```diff
@@ -83,6 +83,9 @@ def remote_fqdn(forward_verify = true)
     ip = request.env['REMOTE_ADDR']
     log_halt 403, 'could not get remote address from environment' if ip.empty?
 
+    # IPv6 with NIC identifier are not resolved
+    return ip if ip.sub(/%.*$/, '')
```
ekohl: This always resolves to true.
```suggestion
    return ip if ip.include?('%')
```

ekohl (changes requested):
Would you mind changing the message to say something like "skip IPv6 link local addresses in FQDN lookups".

lzap:
Sure, done!

ekohl (approved):
Thanks!

PR #788, opened 2021-04-26
Fixes #32418,34552 - foreman request timeout config option

lzap:
It is set to the default 60 seconds with no way to override this.

I am going for a simple solution of setting multiple timeouts via a single setting. I am not sure if it is worth introducing all of them.

### DIFF
--- a/config/settings.yml.example
+++ b/config/settings.yml.example
@@ -33,7 +33,11 @@
 # (default: true)
 #:forward_verify: true
 
+# URL to Foreman instance
 #:foreman_url: http://127.0.0.1:3000
+
+# Open, read, write HTTP(S) timeout (in seconds) for requests to Foreman
+#:foreman_http_timeout: 60
 
 # SSL settings for client authentication against Foreman. If undefined, the values
 # from general SSL options are used instead. Mainly useful when Foreman uses
--- a/lib/proxy/request.rb
+++ b/lib/proxy/request.rb
@@ -63,9 +63,13 @@ module Proxy::HttpRequest
     private
 
     def http_init
-      http             = Net::HTTP.new(uri.host, uri.port)
-      http.use_ssl     = uri.scheme == 'https'
+      http = Net::HTTP.new(uri.host, uri.port)
+      http.use_ssl = (uri.scheme == 'https')
       http.verify_mode = OpenSSL::SSL::VERIFY_NONE
+      http.open_timeout = Proxy::SETTINGS.foreman_http_timeout
+      http.ssl_timeout = Proxy::SETTINGS.foreman_http_timeout
+      http.read_timeout = Proxy::SETTINGS.foreman_http_timeout
+      http.write_timeout = Proxy::SETTINGS.foreman_http_timeout if http.respond_to?(:write_timeout)
 
       if http.use_ssl?
         ca_file = Proxy::SETTINGS.foreman_ssl_ca || Proxy::SETTINGS.ssl_ca_file
--- a/lib/proxy/settings/global.rb
+++ b/lib/proxy/settings/global.rb
@@ -19,6 +19,7 @@ module ::Proxy::Settings
       :ssl_disabled_ciphers => [],
       :tls_disabled_versions => [],
       :dns_resolv_timeouts => [5, 8, 13], # Ruby default is [5, 20, 40] which is a bit too much for us
+      :foreman_http_timeout => 60,
     }
 
     HOW_TO_NORMALIZE = {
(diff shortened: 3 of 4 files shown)


lib/proxy/request.rb
```diff
@@ -63,9 +63,14 @@ def http
...
 
     def http_init
-      http             = Net::HTTP.new(uri.host, uri.port)
-      http.use_ssl     = uri.scheme == 'https'
+      http = Net::HTTP.new(uri.host, uri.port)
+      http.use_ssl = uri.scheme == 'https'
       http.verify_mode = OpenSSL::SSL::VERIFY_NONE
+      http.open_timeout = Proxy::SETTINGS.foreman_http_timeout
+      http.ssl_timeout = Proxy::SETTINGS.foreman_http_timeout
+      http.continue_timeout = Proxy::SETTINGS.foreman_http_timeout
```
ekohl: Reading https://ruby-doc.org/stdlib-2.7.1/libdoc/net/http/rdoc/Net/HTTP.html#attribute-i-continue_timeout
> Seconds to wait for 100 Continue response. If the HTTP object does not receive a response in this many seconds it sends the request body. The default value is nil.

It doesn't say what nil actually means. Does that disable the timeout?

I really doubt we use HTTP 100 Continue so should we modify the default behavior? I'm leaning to be safe and not set it.
```suggestion
```

lib/proxy/request.rb
```diff
@@ -63,9 +63,14 @@ def http
     private
 
     def http_init
-      http             = Net::HTTP.new(uri.host, uri.port)
-      http.use_ssl     = uri.scheme == 'https'
+      http = Net::HTTP.new(uri.host, uri.port)
+      http.use_ssl = uri.scheme == 'https'
       http.verify_mode = OpenSSL::SSL::VERIFY_NONE
+      http.open_timeout = Proxy::SETTINGS.foreman_http_timeout
```
ekohl, line 69: I think the default value for this is nil. I doubt that works.

config/settings.yml.example
```diff
@@ -33,8 +33,12 @@
 # (default: true)
 #:forward_verify: true
 
+# URL to Foreman instance
 #:foreman_url: http://127.0.0.1:3000
 
+# Open, read, write HTTPS timeout (in seconds)
```
ekohl:
```suggestion
# Open, read, write HTTP(S) timeout (in seconds)
```

ekohl (changes requested):

lib/proxy/request.rb
lzap: Okay makes sense, I had no idea honestly what that is, now it makes sense, HTTP 100.

lib/proxy/request.rb
lzap: Right, rebased and added 60 as the default.

lzap:
Pushed.

config/settings.yml.example
```diff
@@ -33,8 +33,12 @@
 # (default: true)
 #:forward_verify: true
 
+# URL to Foreman instance
 #:foreman_url: http://127.0.0.1:3000
 
+# Open, read, write HTTP(S) timeout (in seconds)
```
ekohl:
```suggestion
# Open, read, write HTTP(S) timeout (in seconds) for requests to Foreman
```

ekohl (approved):
Other than a bit more explicit comment, I think this looks good.

lzap:
Sure, rebased.

ekohl (approved):

ekohl:
I took the liberty of editing the PR title to make it slightly more descriptive. This should also be done to the commit. Either by you (with a force push) or by squashing on merge.

ekohl:
Failures look related btw.

lzap:
I don't see any errors locally, rebased.

ekohl:
There were still test failures, but they've been rotated
[test smart-proxy]

lzap:
No failures locally, rebased it was 404.

lzap:
So there is a missing include statement in the registration API which caused:

```
undefined method `exception' for #<Rack::NullLogger:0x0000000003628b28> (NoMethodError)
/home/jenkins/workspace/smart-proxy-pr-test/modules/registration/registration_api.rb:16:in `rescue in block in <class:Api>'
/home/jenkins/workspace/smart-proxy-pr-test/modules/registration/registration_api.rb:13:in `block in <class:Api>'
...
```

Fixing this in separate commit and pushing to see what the root cause was, locally it works weird.

lzap:
Ah I think that `write_timeout` is not available for Ruby 2.5, on my Ruby 2.7 it works fine that's why it was passing locally.

lzap:
Yup, all green.

ekohl (approved):
This looks good to me, but the logger inclusion is something we may need to cherry pick. Perhaps it deserves its own issue?

lzap:
Done

PR #790, opened 2021-05-17
Fixes #32607 - drop host from x-forwarded-for

lzap:
We do not use this at all.

### DIFF
--- a/modules/templates/proxy_request.rb
+++ b/modules/templates/proxy_request.rb
@@ -27,7 +27,6 @@ module Proxy::Templates
 
     def call_template(method, path, env, params, body = '')
       template_url = Proxy::Templates::Plugin.settings.template_url
-      proxy_ip = URI.parse(template_url).host
       opts = params.clone.merge(:url => template_url)
       BLACKLIST_PARAMETERS.each do |blacklisted_parameter|
         opts.delete(blacklisted_parameter)
@@ -36,7 +35,7 @@ module Proxy::Templates
       path = path.map { |x| CGI.escape(x) }.join('/')
       logger.debug "Template: request for #{path} using #{opts.inspect} at #{uri.host}"
       proxy_headers = extract_request_headers(env)
-      proxy_headers["X-Forwarded-For"] = "#{env['REMOTE_ADDR']}, #{proxy_ip}"
+      proxy_headers["X-Forwarded-For"] = env['REMOTE_ADDR']
       proxy_headers["Content-Type"] = params["Content-Type"] if params["Content-Type"]
       if method == :get
         proxy_req = request_factory.create_get(path, opts, proxy_headers)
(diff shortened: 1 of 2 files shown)


modules/templates/proxy_request.rb
```diff
@@ -36,7 +35,7 @@ def call_template(method, path, env, params, body = '')
       path = path.map { |x| CGI.escape(x) }.join('/')
       logger.debug "Template: request for #{path} using #{opts.inspect} at #{uri.host}"
       proxy_headers = extract_request_headers(env)
-      proxy_headers["X-Forwarded-For"] = "#{env['REMOTE_ADDR']}, #{proxy_ip}"
+      proxy_headers["X-Forwarded-For"] = "#{env['REMOTE_ADDR']}"
```
ekohl:
```suggestion
      proxy_headers["X-Forwarded-For"] = env['REMOTE_ADDR']
```

ekohl (changes requested):
Last Friday I tried to read up on this and it's my reading that this is correct, just a minor inline code suggestion.

It would be even more correct to see if there already was a `X-Forwarded-For` header. If so, send that together with `REMOTE_ADDR`. Respecting `X-Forwarded-For` could be considered a feature though.

This might need a cherry pick into stable branches. If so, I'd limit the cherry pick to just what this PR is (the bugfix) and open a new PR for the feature.

lzap:
Yeah sorry, rebased and amended the change. Since this is an important bug, I want just to fix this.

Introducing passing already-forwarded entries might actually trigger the bug we have in Foreman too, so that needs probably more work.

ekohl (approved):
Thanks!

ekohl (changes requested):
Apologies, test failures are related.

lzap:
For the record I asked one of our users to test this patch in https://community.theforeman.org/t/foreman-2-4-katello-4-ipxe-not-working/23398/45

Thulium-Drake:
As also mentioned on the forum, this patch in combination with an edit of the ```foreman_proxies``` works like a charm:

```
# /etc/foreman/settings.yaml
# ... existing config
:trusted_proxies:
  - "127.0.0.0/8"
  - "::1"
  - "<IP address of the foreman server>"
```

Thulium-Drake (approved):

ekohl:
> Apologies, test failures are related.

This is still true. The actual change looks good, but I don't want to break the build.

lzap:
Fixed tests, rebased.

ekohl (approved):

Thulium-Drake:
One addition, you also need to put _all_ smart proxy IPs on the same ```trusted_proxies``` list, if you don't you get 500 errors on the client. And 409 conflicts on the Foreman server (not the proxy)

PR #791, opened 2021-06-30
Fixes #32915 - accept strings as setting names

lzap:
I do not know why we use the super-weird :setting: "xyz" syntax but it is confusing. This is not a typical YAML. It is not a typical UNIX config. It's some kind of Ruby-YAML hybrid.

This patch converts all keys to symbols, we can do a followup puppet update and write all config files without the colons.

### DIFF
--- a/lib/proxy/settings.rb
+++ b/lib/proxy/settings.rb
@@ -16,7 +16,7 @@ module Proxy::Settings
   def self.load_plugin_settings(defaults, settings_file, settings_directory = nil)
     settings = {}
     begin
-      settings = YAML.load(File.read(File.join(settings_directory || ::Proxy::SETTINGS.settings_directory, settings_file))) || {}
+      settings = read_settings_file(settings_file, settings_directory)
     rescue Errno::ENOENT
       logger.warn("Couldn't find settings file #{settings_directory || ::Proxy::SETTINGS.settings_directory}/#{settings_file}. Using default settings.")
     end
@@ -24,6 +24,6 @@ module Proxy::Settings
   end
 
   def self.read_settings_file(settings_file, settings_directory = nil)
-    YAML.load(File.read(File.join(settings_directory || ::Proxy::SETTINGS.settings_directory, settings_file))) || {}
+    YAML.load(File.read(File.join(settings_directory || ::Proxy::SETTINGS.settings_directory, settings_file))).transform_keys(&:to_sym) || {}
   end
 end


lib/proxy/plugin_initializer.rb
```diff
@@ -114,7 +114,7 @@ def configure
   end
 
   def fail_group(an_exception)
-    fail_group_with_message("Disabling all modules in the group #{printable_module_names(member_names)} due to a failure in one of them: #{an_exception}", an_exception.backtrace)
+    fail_group_with_message("Disabling all modules in the group #{printable_module_names(member_names)} due to a failure in one of them: #{an_exception}", an_exception)
```
lzap: When debugging this, backtrace was not printed correctly. It's because the method accepts exception, not backtrace array.

lib/proxy/settings.rb
```diff
@@ -16,14 +16,14 @@ def self.initialize_global_settings(settings_path = nil, argv = ARGV)
   def self.load_plugin_settings(defaults, settings_file, settings_directory = nil)
     settings = {}
     begin
-      settings = YAML.load(File.read(File.join(settings_directory || ::Proxy::SETTINGS.settings_directory, settings_file))) || {}
+      settings = read_settings_file(settings_file, settings_directory)
```
lzap, line 19: This line was just a copy of the line below, so DRY.

lib/proxy/plugin_initializer.rb
```diff
@@ -114,7 +114,7 @@ def configure
   end
 
   def fail_group(an_exception)
-    fail_group_with_message("Disabling all modules in the group #{printable_module_names(member_names)} due to a failure in one of them: #{an_exception}", an_exception.backtrace)
```
ekohl: Would you mind submitting this as a separate PR? Looks like a good fix, but an unrelated fix.

ekohl (commented):
I looked and couldn't find any settings that are a hash, but I wonder how that will happen. It'll be weird if only the first level accepts strings where the inner levels only accept symbols. Did you happen to check for this in plugins?

lib/proxy/plugin_initializer.rb
lzap: Sure, created: https://projects.theforeman.org/issues/33482

Edit: Ooops this was fixed in b94c9b07

lzap:
Rebased, @ekohl

PR #792, opened 2021-07-15
Fixes #18936 - checking server against CAs in TFTP module

avitova:


lzap:
Hello Anna and welcome to the community!

We require all commits to be in format of `Fixes #18936 - some short subject". We have a bot that enforces this. Can you please squash both commits into a single one with this commit message?

config/settings.d/tftp.yml.example
```diff
@@ -17,3 +17,7 @@
 # Defines the default dns timeout in seconds needed to download tftp artifacts
 # like initrd and vmlinuz. Default value 10 seconds
 #:tftp_dns_timeout: 10
+
+# Defines the default certificate action for certificate checking.
```
lzap, line 21: Maybe something like "Check the server certificate against available CAs when downloading boot files."

config/settings.d/tftp.yml.example
```diff
@@ -17,3 +17,7 @@
...
+# When false, the argument --no-check-certificate will be used.
+#:verify_server_cert: false
```
lzap: We are aiming for a safe default, to prevent breaking existing installations. But now that I read this, it might feels weird long-term. @ekohl do you prefer this to be true or false?

Maybe we can set it to true by default and for upgrades, we can provide a false value via the installer?

lib/proxy/http_download.rb
```diff
@@ -7,20 +7,21 @@ class HttpDownload < Proxy::Util::CommandTask
...
-      super([wget,
+      args = [wget,
              "--connect-timeout=#{connect_timeout}",
              "--dns-timeout=#{dns_timeout}",
              "--read-timeout=#{read_timeout}",
              "--tries=3",
-             "--no-check-certificate",
-             "-nv", "-c", src.to_s, "-O", dst.to_s])
+             "-nv", "-c", src.to_s, "-O", dst.to_s]
+      args << "--no-check-certificate" if not verify_server_cert
```
lzap:
```suggestion
      args << "--no-check-certificate" unless verify_server_cert
```

modules/tftp/server.rb
```diff
@@ -169,7 +169,9 @@ def self.choose_protocol_and_fetch(src, destination)
                                 destination.to_s,
                                 Proxy::TFTP::Plugin.settings.tftp_read_timeout,
                                 Proxy::TFTP::Plugin.settings.tftp_connect_timeout,
-                                Proxy::TFTP::Plugin.settings.tftp_dns_timeout).start
+                                Proxy::TFTP::Plugin.settings.tftp_dns_timeout,
+				Proxy::TFTP::Plugin.settings.verify_server_cert).start
+
```
lzap, line 174: Can you keep the indentation? Rubocop will likely fail here.

lzap:
Nice patch, thank you. I left a few nitpicks. There is an open question of the default value, let's discuss this with @ekohl and either keep it or change it.

ekohl:
Certainly a nice addition @avitova. I'm in favor of verifying certificates whenever you can. I also thought about the default. I'm torn between compatibility (current implementation) and security.

I'm leaning to merging this (after @lzap's comment have been addressed). Right now we're still not 100% certain if the next version will be 2.6 or 3.0 (to be decided in the next few days). If the next version is 3.0, then we can merge another patch that flips the default to be secure.

config/settings.d/tftp.yml.example
lzap: Let's make this `true` then.

lzap:
I do not know why tests are started: [test]

evgeni:
ok to test

ekohl:
I think something went wrong in merging it. You should rebase, not merge develop into your branch.

avitova:
Yes, I tried to rebase but unfortunately was not able to and somehow broke it. I will discuss it with lzap soon and clean all the mess hopefully.

lzap (approved):
Thank you

ekohl (approved):
Code wise :+1: but I'd like to see a bit more descriptive commit message. I already added TFTP to the PR title. https://chris.beams.io/posts/git-commit/ has a lot more tips.

lzap:
Oh yeah, if you could add "TFTP" into the commit subject that would be better.

avitova:
Thank you, next time I'll focus on the message bit more.

lzap:
For the record, I told Anna not to exceed 50 characters in commit subjects, but it's not a hard rule. If you exceed and the subject cannot be reasonably shortened, it is fine. Just do not waste space, do not use articles, just the simplest meaningful statement.

lzap:
For the record, I tested that wget accepts arguments at the very end after the URL:

    wget -O /dev/null https://downloads.theforeman.org/devconf-youtube-link.txt --no-check-certificate

Works fine, please rebase the subject change and we are merging. Thanks.

ekohl:
One thing I see a lot of people do wrong is not using the imperative mood (https://chris.beams.io/posts/git-commit/#imperative). So I'd write `Check server certs in the TFTP module`.

lzap:
Can you remove the trailing dot? Read the guide that Ewoud linked above, these are unwanted :-)

avitova:
Sure, I actually did read it, but using the dot when starting with capital, is my habit, an unpleasant one in this case. I'm sorry, it should be fixed by now.

avitova:
I pushed while on the wrong branch, nothing has changed. Hopefully it will not be a problem.

lzap (approved):
Thank you!

PR #793, opened 2021-07-26
fixes #18069 - Add ip/mac Addr. validations (DHCP)

avitova:
Add ip and mac address validations to api endpoints in dhcp api controller

Reporting an invalid mac or ip address for easier troubleshooting. I also made the Error class more self-describing, it should not be used anywhere else in the project now. I am not sure if I should have changed the MAC validation. From what I know, there is no other option than regex and it did work so far so I left it as it was.

### DIFF
--- a/lib/proxy/validations.rb
+++ b/lib/proxy/validations.rb
@@ -1,8 +1,15 @@
+require 'ipaddr'
+require 'resolv'
+
 module Proxy::Validations
   MAC_REGEXP_48BIT = /\A([a-f0-9]{1,2}:){5}[a-f0-9]{1,2}\z/i
   MAC_REGEXP_64BIT = /\A([a-f0-9]{1,2}:){19}[a-f0-9]{1,2}\z/i
 
   class Error < RuntimeError; end
+  class InvalidIPAddress < Error; end
+  class InvalidMACAddress < Error; end
+  class InvalidSubnet < Error; end
+
   private
 
   def valid_mac?(mac)
@@ -11,20 +18,36 @@ module Proxy::Validations
     false
   end
 
+  def valid_ip?(ip)
+    valid_ip4?(ip) || valid_ip6?(ip)
+  end
+
+  def valid_ip4?(ip)
+    Resolv::IPv4::Regex.match?(ip)
+  end
+
+  def valid_ip6?(ip)
+    Resolv::IPv6::Regex.match?(ip)
+  end
+
   # validates the ip address
-  def validate_ip(ip)
-    raise Error, "Invalid IP Address #{ip}" unless ip =~ /(\d{1,3}\.){3}\d{1,3}/
+  def validate_ip(ip, version = nil)
+    valid = false
+    valid = valid_ip4?(ip) if version == 4
+    valid = valid_ip6?(ip) if version == 6
+    valid = valid_ip?(ip) if version.nil?
+    raise InvalidIPAddress, "Invalid IP Address #{ip}" unless valid
     ip
   end
 
   # validates the mac
   def validate_mac(mac)
-    raise Error, "Invalid MAC #{mac}" unless valid_mac?(mac)
+    raise InvalidMACAddress, "Invalid MAC #{mac}" unless valid_mac?(mac)
     mac.downcase
   end
 
   def validate_subnet(subnet)
-    raise Error, "Invalid Subnet #{subnet}" unless subnet.is_a?(Proxy::DHCP::Subnet)
+    raise InvalidSubnet, "Invalid Subnet #{subnet}" unless subnet.is_a?(Proxy::DHCP::Subnet)
     subnet
   end
 
--- a/modules/dhcp/dhcp_api.rb
+++ b/modules/dhcp/dhcp_api.rb
@@ -1,5 +1,8 @@
+require "proxy/validations"
+
 class Proxy::DhcpApi < ::Sinatra::Base
   extend Proxy::DHCP::DependencyInjection
+  include Proxy::Validations
 
   helpers ::Proxy::Helpers
   authorize_with_trusted_hosts
@@ -16,6 +19,7 @@ class Proxy::DhcpApi < ::Sinatra::Base
   end
 
   get "/:network" do
+    server.validate_supported_address(params[:network])
     content_type :json
     {:reservations => server.all_hosts(params[:network]), :leases => server.all_leases(params[:network])}.to_json
   rescue ::Proxy::DHCP::SubnetNotFound
@@ -25,6 +29,9 @@ class Proxy::DhcpApi < ::Sinatra::Base
   end
 
   get "/:network/unused_ip" do
+    server.validate_supported_address(params[:network])
+    server.validate_supported_address(params[:from], params[:to]) if params[:from] && params[:to]
+    validate_mac(params[:mac]) if params[:mac]
     content_type :json
     {:ip => server.unused_ip(params[:network], params[:mac], params[:from], params[:to])}.to_json
   rescue ::Proxy::DHCP::SubnetNotFound
@@ -53,8 +60,8 @@ class Proxy::DhcpApi < ::Sinatra::Base
 
   # returns an array of records for an ip address
   get "/:network/ip/:ip_address" do
+    server.validate_supported_address(params[:network], params[:ip_address])
     content_type :json
-
     records = server.find_records_by_ip(params[:network], params[:ip_address])
     log_halt 404, "No DHCP records for IP #{params[:network]}/#{params[:ip_address]} found" if records.empty?
     records.to_json
@@ -67,8 +74,10 @@ class Proxy::DhcpApi < ::Sinatra::Base
   # returns a record for a mac address
   get "/:network/mac/:mac_address" do
     param_network = params[:network]
-    param_mac = params[:mac_address].downcase unless params[:mac_address].nil?
+    param_mac = params[:mac_address].downcase if params[:mac_address]
     begin
+      server.validate_supported_address(params[:network])
+      validate_mac(param_mac) if param_mac
       content_type :json
       record = server.find_record_by_mac(param_network, param_mac)
       log_halt 404, "No DHCP record for MAC #{param_network}/#{param_mac} found" unless record
@@ -82,6 +91,7 @@ class Proxy::DhcpApi < ::Sinatra::Base
 
   # create a new record in a network
   post "/:network" do
+    server.validate_supported_address(params[:network])
     content_type :json
     # NOTE: sinatra overwrites params[:network] (required by add_record call) with the :network url parameter
     server.add_record(params)
@@ -110,6 +120,7 @@ class Proxy::DhcpApi < ::Sinatra::Base
 
   # deletes all records for an ip address from a network
   delete "/:network/ip/:ip_address" do
+    server.validate_supported_address(params[:network], params[:ip_address])
     server.del_records_by_ip(params[:network], params[:ip_address])
     nil
   rescue ::Proxy::DHCP::SubnetNotFound # rubocop:disable Lint/SuppressedException
@@ -120,6 +131,8 @@ class Proxy::DhcpApi < ::Sinatra::Base
 
   # delete a record for a mac address from a network
   delete "/:network/mac/:mac_address" do
+    server.validate_supported_address(params[:network])
+    validate_mac(params[:mac_address])
     server.del_record_by_mac(params[:network], params[:mac_address].nil? ? nil : params[:mac_address].downcase)
     nil
   rescue ::Proxy::DHCP::SubnetNotFound # rubocop:disable Lint/SuppressedException
--- a/modules/dhcp_common/server.rb
+++ b/modules/dhcp_common/server.rb
@@ -29,6 +29,10 @@ module Proxy::DHCP
                          end
     end
 
+    def validate_supported_address(ip)
+      validate_ip(ip)
+    end
+
     def subnets
       service.all_subnets
     end
@@ -132,7 +136,6 @@ module Proxy::DHCP
 
       name, ip_address, mac_address, subnet_address, options = clean_up_add_record_parameters(options)
 
-      validate_ip(ip_address)
       validate_mac(mac_address)
       raise(Proxy::DHCP::Error, "Must provide hostname") unless name
 
--- a/modules/dhcp_native_ms/dhcp_native_ms_main.rb
+++ b/modules/dhcp_native_ms/dhcp_native_ms_main.rb
@@ -13,6 +13,12 @@ module Proxy::DHCP::NativeMS
       @free_ips = free_ips_service
     end
 
+    def validate_supported_address(*args)
+      args.each do |ip|
+        validate_ip(ip, 4)
+      end
+    end
+
     def del_record(record)
       logger.debug "Deleting '#{record}'"
       if record.is_a?(::Proxy::DHCP::Reservation)
@@ -25,8 +31,6 @@ module Proxy::DHCP::NativeMS
     def add_record(options)
       name, ip_address, mac_address, subnet_address, options = clean_up_add_record_parameters(options)
 
-      validate_ip(ip_address)
-      validate_mac(mac_address)
       subnet = retrieve_subnet_from_server(subnet_address)
 
       create_reservation(ip_address, subnet.netmask, mac_address, name)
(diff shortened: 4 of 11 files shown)


lib/proxy/validations.rb
```diff
@@ -11,20 +17,42 @@ def valid_mac?(mac)
     false
   end
 
+  def valid_ip?(ip)
+    !!IPAddr.new(ip) rescue false
+  end
+
+  def valid_ip4?(ip)
+    !!(ip =~ Resolv::IPv4::Regex)
```
ekohl: In modern Ruby (I think 2.5+) you can use
```suggestion
    Resolv::IPv4::Regex.match?(ip)
```

lib/proxy/validations.rb
```diff
@@ -1,8 +1,14 @@
+require 'ipaddr'
+require 'resolv'
+
 module Proxy::Validations
   MAC_REGEXP_48BIT = /\A([a-f0-9]{1,2}:){5}[a-f0-9]{1,2}\z/i
   MAC_REGEXP_64BIT = /\A([a-f0-9]{1,2}:){19}[a-f0-9]{1,2}\z/i
 
-  class Error < RuntimeError; end
```
ekohl, line 5: Please keep this one since it's the public API and used (as `Proxy::Validations::Error`). Also let the other exceptions inherit from this so you can use `rescue Proxy::Validations::Error` to catch any validation error.

modules/dhcp/dhcp_api.rb
```diff
@@ -8,6 +12,19 @@ class Proxy::DhcpApi < ::Sinatra::Base
 
   inject_attr :dhcp_provider, :server
 
+  def validate_arguments(params)
+    [:network, :ip_address, :from, :to, :ip].each do |ip|
+      validate_ip(params[ip]) unless params[ip].nil?
+    rescue
+      log_halt 500, "Invalid IP Address #{params[ip]}"
```
ekohl: I don't think we should treat invalid input as an internal server error. 5xx codes are server errors but in this case it's really a client error (see https://developer.mozilla.org/en-US/docs/Web/HTTP/Status). In core we typically return HTTP 422 but in other places in the smart-proxy I've seen 400. 422 feels semantically more correct.

ekohl (dismissed):

avitova:
Thank you for your comments, @ekohl. Hopefully I understood everything correctly. If there is anything else, please, let me know.

lib/proxy/validations.rb
```diff
@@ -11,20 +18,42 @@ def valid_mac?(mac)
     false
   end
 
+  def valid_ip?(ip)
+    !!IPAddr.new(ip) rescue false
```
ekohl: `IPAddr.new` also accepts a subnet so it'll accept `192.0.2.0/24`. Perhaps:
```suggestion
    valid_ip4?(ip) || valid_ip6?(ip)
```

lib/proxy/validations.rb
```diff
@@ -11,20 +18,42 @@ def valid_mac?(mac)
...
+  def valid_ip4?(ip)
+    Resolv::IPv4::Regex.match?(ip)
+  end
+
+  def valid_ip6?(ip)
+    Resolv::IPv6::Regex.match?(ip)
+  end
+
   # validates the ip address
   def validate_ip(ip)
```
ekohl: Given the next version is a major version, it's not too bad to change the API for this. However, I think you should replace some calls to `validate_ip` to `validate_ip4`. Our DHCP modules are currently IPv4-only. Accepting IPv6 may lead to problems.

Also another thought for the API:
```suggestion
  def validate_ip(ip, version: nil)
```

Where `nil` would be any (4 or 6) and you can also pass in `4` or `6`.

Then you can drop `validate_ip4` and `validate_ip6` methods. The calls would become:
```ruby
validate_ip(ip) # any
validate_ip(ip, version: 4) # IPv4
validate_ip(ip, version: 6) # IPv6
```

You can also think about using `Socket::AF_INET` and `Socket::AF_INET6`.

lzap: A new argument is a good idea, compatible way of preparing this for IPv6. I would use `version: 4` as the default argument not to break existing DHCP modules tho.

lzap: > Accepting IPv6 may lead to problems.

Can you elaborate? Not following?

lzap:
@ekohl can you re-review please?

lib/proxy/validations.rb
ekohl: > > Accepting IPv6 may lead to problems.
> 
> Can you elaborate? Not following?

If today something calls `validate_ip` and expects a valid IPv4 address, but the guarantee is reduced to only a valid IP address, then code later on may fail. For example, I'm not sure how well our DHCP API would work with IPv6. As the API user it would likely change from HTTP 400 Bad Request to HTTP 500 Internal Server Error.

lzap:
So I finally tested this and it does not work. I would expect this to return 422 HTTP error:

```
[lzap@nuc foreman]$ curl -H "Content-Type:application/json" -X GET http://localhost:8448/dhcp/xxx
Subnet xxx could not found
```

As well as other endpoints, can you take a look please? ok to test

modules/dhcp_common/subnet.rb
```diff
@@ -62,7 +62,7 @@ def netmask_to_i
     end
 
     def validate_subnet_range!(from_address, to_address)
-      if (from = validate_ip(from_address)) && (to = validate_ip(to_address))
+      if (from = validate_ip(from_address, 4)) && (to = validate_ip(to_address, 4))
         raise Proxy::DHCP::Error, "Range does not belong to provided subnet" unless include?(from) && include?(to)
         from = IPAddr.new(from)
         to   = IPAddr.new(to)
```
lzap: Please remove the changes to validate only for IPv4 from `modules/dhcp_common` files. See, these classes are shared by other plugins and may or may not support IPv6, we don't know (e..g. Infoblox does support it I think).

Where it is appropriate to do these checks are SP modules we know do not support IPv6, these are: MS native, libvirt and isc.

So please find `add_record` method and perform IPv4 only validation there.

modules/dhcp_native_ms/dhcp_native_ms_main.rb
```diff
@@ -25,7 +25,7 @@ def del_record(record)
     def add_record(options)
       name, ip_address, mac_address, subnet_address, options = clean_up_add_record_parameters(options)
 
-      validate_ip(ip_address)
+      validate_ip(ip_address, 4)
```
lzap: Yes, this is correct - do this for libvirt and isc, but not for the common module.

lzap (changes requested):
It works fine, just remove IPv4 validation for common module and move it into libvirt/isc modules.

lzap:
ok to test

modules/dhcp/dhcp_api.rb
```diff
@@ -1,3 +1,7 @@
+require "proxy/validations"
+
+include Proxy::Validations
```
lzap: This smells, either include this into a class or just avoid including and use full namespace.

modules/dhcp/dhcp_api.rb
```diff
@@ -8,6 +12,19 @@ class Proxy::DhcpApi < ::Sinatra::Base
...
+      validate_ip(params[ip], 4) unless params[ip].nil?
+    rescue
+      log_halt 422, "Invalid IP Address #{params[ip]}"
+    end
+    [:mac, :mac_address].each do |mac|
+      validate_mac(params[mac]) unless params[mac].nil?
+    rescue
+      log_halt 422, "Invalid MAC Address #{params[mac]}"
+    end
+  end
```
lzap: Is this even used? If not, please drop it.

modules/dhcp_common/server.rb
```diff
@@ -29,6 +29,11 @@ def initialize(name, managed_subnets, subnet_service, free_ips_service = nil)
                          end
     end
 
+    def validate_supported_addresses_ip(ip)
+      # Not implemented. Method overloading for validation.
+      ip
+    end
```
lzap, line 34: Apologies for leading you into this, please make the default implementation to validate IPv4 or IPv6.

modules/dhcp/dhcp_api.rb
```diff
@@ -17,6 +34,7 @@ class Proxy::DhcpApi < ::Sinatra::Base
 
   get "/:network" do
     content_type :json
+    server.validate_supported_addresses_ip(:ip)
```
lzap: We need to pass values here, e.g. `params[:xyz]`.

lzap (changes requested):
Few more.

modules/dhcp_common/server.rb
```diff
@@ -29,6 +29,10 @@ def initialize(name, managed_subnets, subnet_service, free_ips_service = nil)
                          end
     end
 
+    def validate_supported_address(ip)
+      ::Proxy::Validations.validate_ip(ip)
+    end
```
lzap, line 34: Namespace is not required.

modules/dhcp/dhcp_api.rb
```diff
@@ -10,13 +10,15 @@ class Proxy::DhcpApi < ::Sinatra::Base
...
     content_type :json
+    server.validate_supported_address(params[:ip])
     server.subnets.map { |s| {:network => s.network, :netmask => s.netmask, :options => s.options} }.to_json
   rescue => e
     log_halt 400, e
   end
 
   get "/:network" do
     content_type :json
+    server.validate_supported_address(params[:ip])
```
lzap: Have you tested this? :-)

```suggestion
    server.validate_supported_address(params[:network])
```

modules/dhcp/dhcp_api.rb
```diff
@@ -26,6 +28,7 @@ class Proxy::DhcpApi < ::Sinatra::Base
 
   get "/:network/unused_ip" do
     content_type :json
+    server.validate_supported_address(params[:ip])
```
lzap:
```suggestion
    server.validate_supported_address(params[:network])
```

modules/dhcp/dhcp_api.rb
```diff
@@ -53,8 +56,8 @@ class Proxy::DhcpApi < ::Sinatra::Base
 
   # returns an array of records for an ip address
   get "/:network/ip/:ip_address" do
+    server.validate_supported_address(params[:ip])
```
lzap: Here you need to test both - for this route, Sinatra stores two parameters named `:network` and `:ip_address`. Check both. There is no parameter named `:ip`.

modules/dhcp/dhcp_api.rb
```diff
@@ -66,6 +69,7 @@ class Proxy::DhcpApi < ::Sinatra::Base
 
   # returns a record for a mac address
   get "/:network/mac/:mac_address" do
+    server.validate_supported_address(params[:ip])
```
lzap: Ditto, pretty much every route (method) in this file has an incorrect. Sinatra uses parameter keys as defined in the argument of the HTTP method: http://sinatrarb.com/intro.html

lzap (changes requested):
The code looks good, there is one overlook which revealed itself once I started testing this.

avitova:
> The code looks good, there is one overlook which revealed itself once I started testing this.
@lzap 
Oh, right, I'm about to edit it. Also, if I remember this correctly, I was validating also params[:from] and params[:to], before implementing bigger changes. Is this even desirable, or does it not really help anyone?

modules/dhcp/dhcp_api.rb
```diff
@@ -26,6 +27,7 @@ class Proxy::DhcpApi < ::Sinatra::Base
 
   get "/:network/unused_ip" do
     content_type :json
+    server.validate_supported_address(params[:network])
     {:ip => server.unused_ip(params[:network], params[:mac], params[:from], params[:to])}.to_json
```
lzap, line 36: As you said, from/to/mac should be checked as well but beware it can be nil/blank.

modules/dhcp/dhcp_api.rb
```diff
@@ -66,6 +69,7 @@ class Proxy::DhcpApi < ::Sinatra::Base
 
   # returns a record for a mac address
   get "/:network/mac/:mac_address" do
+    server.validate_supported_address(params[:network])
```
lzap: Validate `:mac_address` as well. Just include the Proxy::Validations.

modules/dhcp/dhcp_api.rb
```diff
@@ -110,6 +115,8 @@ class Proxy::DhcpApi < ::Sinatra::Base
 
   # deletes all records for an ip address from a network
   delete "/:network/ip/:ip_address" do
+    server.validate_supported_address(params[:network])
+    server.validate_supported_address(params[:ip])
```
lzap:
```suggestion
    server.validate_supported_address(params[:ip_address])
```

modules/dhcp/dhcp_api.rb
```diff
@@ -120,6 +127,7 @@ class Proxy::DhcpApi < ::Sinatra::Base
 
   # delete a record for a mac address from a network
   delete "/:network/mac/:mac_address" do
+    server.validate_supported_address(params[:network])
```
lzap, line 134: Ditto.

modules/dhcp/dhcp_api.rb
```diff
@@ -25,6 +29,11 @@ class Proxy::DhcpApi < ::Sinatra::Base
   end
 
   get "/:network/unused_ip" do
+    ip_array = [params[:network]]
+    ip_array << params[:from] unless params[:from].nil?
+    ip_array << params[:to] unless params[:to].nil?
+    server.validate_supported_address(*ip_array)
```
lzap: What I meant was:

    server.validate_supported_address(params[:network])
    server.validate_supported_address(params[:from], params[:to]) if params[:from] && params[:to]

That is easier for our eyes. From and to are always sent both (or none).

modules/dhcp/dhcp_api.rb
```diff
@@ -25,6 +29,11 @@ class Proxy::DhcpApi < ::Sinatra::Base
...
+    validate_mac(params[:mac]) unless params[:mac].nil?
```
lzap:
```suggestion
    validate_mac(params[:mac]) if params[:mac]
```

is generally better. In RoR there is `present?` helper for strings, but it is not available here unfortunately.

There are couple of other places where you could reverse this.

lzap (changes requested):
Two small comments.

modules/dhcp/dhcp_api.rb
```diff
@@ -67,8 +74,10 @@ class Proxy::DhcpApi < ::Sinatra::Base
   # returns a record for a mac address
   get "/:network/mac/:mac_address" do
     param_network = params[:network]
-    param_mac = params[:mac_address].downcase unless params[:mac_address].nil?
+    param_mac = params[:mac_address].downcase if params[:mac_address]
     begin
+      server.validate_supported_address(params[:network])
+      validate_mac(params[:mac_address]) if params[:mac_address]
```
lzap: Small problem, you downcase parameter into `param_mac` variable and then validate param. Not a big deal but if you could validate `param_mac` instead, that is cleaner.

lzap:
This was my last comment, it works fine. RTM

test/dhcp/dhcp_api_test.rb
```diff
@@ -316,34 +335,40 @@ def test_delete_record
...
     @server.expects(:del_record_by_mac).with("192.168.122.0", "00:11:bb:cc:dd:ee")
+    @server.stubs(:validate_supported_address).returns(true)
     delete "/192.168.122.0/mac/00:11:BB:CC:DD:EE"
     assert_equal 200, last_response.status
     assert_empty last_response.body
   end
 
   def test_delete_records_by_mac_for_nonexistent_subnet
     @server.expects(:del_record_by_mac).with("192.168.122.0", "00:11:bb:cc:dd:ee").raises(::Proxy::DHCP::SubnetNotFound)
+    @server.stubs(:validate_supported_address).returns(true)
```
lzap: Apologies for another comment, I should have see it.

Can you move all these statements into the `setup` method above? It's being called for each test and it is exactly what intended for. You can override `stubs` with `expects` in any test if you want to test for it.

avitova: I see, that for sure looks better. Hopefully it will be correct now.

PR #794, opened 2021-08-02
Update README

ekohl:
The badges were outdated and nobody cared. It's better to remove them. We've also migrated to Libera.


xprazak2:
Looks good, failing redmine check prevents me from merging.

PR #796, opened 2021-08-06
fixes #4026 secure Windows command execution

avitova:


modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -34,6 +34,7 @@ def do_remove(name, type)
     end
 
     def execute(cmd, msg = nil, error_only = false)
+      cmd = escape_for_cmdexe(cmd)
       tsecs = 5
       response = nil
       interpreter = Proxy::SETTINGS.x86_64 ? 'c:\windows\sysnative\cmd.exe' : 'c:\windows\system32\cmd.exe'
```
ekohl: So the line below this should become an array to which arguments are appended. It does mean the value of `cmd` should also be an array, but callers are the only place that actually know what are the various arguments.

ekohl (changes requested):
I'm not sure this is the right implementation. Rather I'd suggest changing `command` and all callers to be arrays.

test/util_test.rb
```diff
@@ -16,6 +16,15 @@ def test_util_shell_escape
     assert_equal test_class.new.escape_for_shell("vm.test.com physical.test.com"), 'vm.test.com\ physical.test.com'
   end
 
+  def test_util_escape_for_cmdexe
+    assert Proxy::Util.instance_methods.include? :escape_for_cmdexe
+
+    test_class = eval "class ProxyUtilTestHelper; include Proxy::Util; end"
+    assert_equal test_class.new.escape_for_cmdexe("Test text &! continues"), ["Test", "text", "&!", "continues"]
```
ares: small nit, assert_equal expects the first argument to be the one that we expect (in your case the one you use as the last), that way it constructs more meaningful output when the test fails. The same applies to the line below.

test/util_test.rb
```diff
@@ -16,6 +16,15 @@ def test_util_shell_escape
...
+    assert_equal test_class.new.escape_for_cmdexe("Test text&!continues"), ["Test", "text&!continues"]
+    assert_equal test_class.new.escape_for_cmdexe('vm.test.com,"physical.test.com'), ["vm.test.com,\"physical.test.com"]
```
ares: does this actually work? The `escape_for_cmdexe` only splits the command into array by empty space. The actual espacing is performed by passing command as an array on line 40 if I understand.

ares (commented):
ok to test

lib/proxy/util.rb
```diff
@@ -75,6 +75,10 @@ def escape_for_shell(command)
     Shellwords.escape(command)
   end
 
+  def escape_for_cmdexe(command)
+    command.split(" ")
+  end
```
ekohl: This itn't proper escaping. If you want to keep it, it should be a private method in `dns_cmd_main.rb` but we shouldn't make it a general utility.

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -37,8 +37,7 @@ def execute(cmd, msg = nil, error_only = false)
...
-      command = interpreter + ' /c c:\Windows\System32\dnscmd.exe ' + "#{@server} #{cmd}"
-
+      command = [interpreter, ' /c c:\Windows\System32\dnscmd.exe ', @server] + escape_for_cmdexe(cmd)
```
ekohl: I think each individual argument should be an entry in the array
```suggestion
      command = [interpreter, '/c', 'c:\Windows\System32\dnscmd.exe', @server] + escape_for_cmdexe(cmd)
```

It should also deal with `cmd` being an array. That way you can fix every call to `execute` to properly pass all arguments.

lib/proxy/util.rb
ares: I think this was added here for the consistency with the method above (though I agree it's not the best place).

I did some testing on windows and realized `spawn` can't escape reliably everything even when used with splatted array. We must escape `^`, `&`, `|`, `(`, `)` with `^`. So that's should be the goal for this method and that's what should be tested.

modules/dns_dnscmd/dns_dnscmd_main.rb
ares: I wouldn't mind having `/c` as another array item, however I don't think it's necessary to support also array for `cmd`. Given it is added to the command array, `popen3` take it as an argument and takes care of any special chars, it just won't format whitespaces and `cmd` can contain multiple spaces between args. The important thing is, it's added to the single command array.

edit: or better to say, it's ok to support `cmd` being an array as well, but I don't think we need to modify all the existing commands in such way

lib/proxy/util.rb
ares: for the record, the problem was if the argument was `'"&'`, spawn method cuts it after escaped `"`.

modules/dns_dnscmd/dns_dnscmd_main.rb
ekohl: I don't think combining arguments in a long string is even valid. Just to be clear, I'd like to avoid using a shell at all. This would be the second form in https://docs.ruby-lang.org/en/2.4.0/Process.html#method-c-spawn. By avoiding a shell, there is no need to escape anything.

To me it's odd to combine multiple arguments in a string only to split it later. That's a step that can only introduce problems.

ares: Oh I think I understand where we misunderstood each other.

1) Yes, let's use splatted array form, meaning `cmdname, arg1, ...` because that performs some escaping and does not use shell.

2) The arg1 could contain more arguments, it could be `"/c c:\Windows\System32\dnscmd.exe"` and it AFAICT does not cause any harm. I agree it feels odd but is valid. I think that converting all existing command strings to arrays just because of it would result in very hard to read snippets. But it's not a strong opinion. The most important thing is to always have a command in the array form splitting at least the cmd name from the arguments.

Still, if you test the example from above and pass `'echo', 'hello', "& whatever` you'll see it cuts off everything after `&`. To avoid that, the characters need to be escaped. And that should be the content of `escape_for_cmdexe`, it should imho not be splitting strings by spaces.

ezr-ondrej:
[test smart-proxy]

lib/proxy/util.rb
lzap: I confirm, we talked about this with Anna and I named exactly these characters Marek mentions. These must be escaped. Example: all occurences of `&` character must be escaped as `^&`. Etc.

modules/dns_dnscmd/dns_dnscmd_main.rb
lzap: I want to avoid regressions, this code is poorly tested and I asked Anna only to do a simple approach by escaping cmd.exe special characters. Please let's not go into bigger refactoring, honestly, I was hoping to merge this patch without testing it as I have no idea how an array passed into `popen3` would work on Windows and I don't have Windows DNS setup available.

lzap:
Apologies for the delay caused by my PTO. Let me explain this.

I asked you to do replacement for special Windows Shell characters as the most simple solution. Ewoud would like to see better solution, which is to use arrays. Let me explain.

Ruby methods `popen3` and `spawn` (as well as few others) accept several variants. If its a string (the current implementation), new shell is executed. This is causing the problem we are trying to solve - in shell there are special characters which can be interpreted and used for an attack.

https://docs.ruby-lang.org/en/2.5.0/Process.html#method-c-spawn

There are two other options to start a new process, the second one is to provide each argument as a method argument. In this case, no shell is started (except some edge cases). We need to make sure full path to the executable is provided (`C:\xyz\abc.exe` on windows) and then shell will be never started. This is the proper solution of the problem.

To get there, you need to construct an array with arguments and executable:

```ruby
executable = "c:\Windows\System32\dnscmd.exe"
args = ["localhost", "/RecordAdd", "example.com", "myhost.", "A", "xyz"]
Open3.popen(executable, *args)
```

Note the special Ruby syntax to pass an array and unwrap it as arguments: https://ruby-doc.org/core-2.0.0/doc/syntax/calling_methods_rdoc.html#label-Array+to+Arguments+Conversion

Therefore, you need to refactor the code, the `cmd` no longer can be string but array. Ideally you change the `execute` method into something like this:

```ruby
def execute(*args, **keyword_args)
  # args contains arguments array
  # and keyword_args contains hash with named arguments
end

execute("/RecordAdd", "xyz", "xyz", "xyz", msg: "Optional message", error_only: true)
```

For more info see: https://www.justinweiss.com/articles/fun-with-keyword-arguments/

Then refactor all uses of the method in the file (possibly whole project as it is not private and probably should be a private method) and test. Hope it is clear.

modules/dns_dnscmd/dns_dnscmd_main.rb
ekohl: I that case I withdraw myself as a reviewer. I don't see the point of making a change in the first place then.

Right now I'd argue there is already no real risk at all for a shell escape. This class has a well defined API and all the inputs are validated (in [dns_api](https://github.com/theforeman/smart-proxy/blob/develop/modules/dns/dns_api.rb)).

If we don't have a setup to validate this and there is no real risk for exploitation, we're only introducing risk for regression. I thought this was about implementing a proper solution.

I'll also leave a note that `escape_for_shell` in util right now is unused within this codebase, but it's only there compatibility for other modules. I'd strongly be in favor of deprecating it and eventually removing it, but this PR takes it in the opposite direction.

ares: Maybe I should have been more explicit, we have a setup to validate this and that's what I actually did. I ran it on Windows. The things I wrote above is a result on the actual test of the `popen3` behavior on Windows. Array aproach is good still needed some escaping. But I'm totally fine with fixing just part of it, still better than what we have now. Leaving up to @lzap to find the right scope of this PR.

lib/proxy/util.rb
```diff
@@ -75,6 +75,7 @@ def escape_for_shell(command)
     Shellwords.escape(command)
   end
 
+
   def shell_command(cmd, wait = true)
```
lzap: Unnecessary blank line.

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -12,16 +12,18 @@ def initialize(a_server, a_ttl)
     def do_create(name, value, type)
       zone = match_zone(name, enum_zones)
       msg = "Added #{type} entry #{name} => #{value}"
+      keyword_args = {msg: msg, error_only: false}
       value = "#{value}." if type == "PTR"
-      cmd = "/RecordAdd #{zone} #{name}. #{type} #{value}"
-      execute(cmd, msg)
+      cmd = ["/RecordAdd", zone, "#{name}.", type, value]
+      execute(*cmd, keyword_args)
```
lzap: You can write this as:

    execute("/RecordAdd", zone, "#{name}.", type, value, msg: msg, error_only: false)

That is a tad cleaner :-)

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -12,16 +12,18 @@ def initialize(a_server, a_ttl)
...
       nil
     end
 
     def remove_specific_record_from_zone(zone_name, node_name, record, type)
       msg = "Removed #{record} #{type} record #{node_name} from #{zone_name}"
-      cmd = "/RecordDelete #{zone_name} #{node_name}. #{type} #{record} /f"
-      execute(cmd, msg)
+      keyword_args = {msg: msg, error_only: false}
+      cmd = ["/RecordDelete", zone_name, "#{node_name}.", type, "#{record} /f"]
+      execute(*cmd, msg)
```
lzap: Ditto.

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -33,13 +35,13 @@ def do_remove(name, type)
...
     end
 
-    def execute(cmd, msg = nil, error_only = false)
+    def execute(*args, **keyword_args)
       tsecs = 5
       response = nil
       interpreter = Proxy::SETTINGS.x86_64 ? 'c:\windows\sysnative\cmd.exe' : 'c:\windows\system32\cmd.exe'
-      command = interpreter + ' /c c:\Windows\System32\dnscmd.exe ' + "#{@server} #{cmd}"
-
+      command = interpreter + ' /c c:\Windows\System32\dnscmd.exe ' + "#{@server}" + args.join(" ")
```
lzap: You no longer pass "/c xxxxx\dnscmd.exe" so remove this. I think you can drop the command variable and just do this below:

    logger.debug "executing: #{interpreter} #{args.join(" ")}"

Or better

    logger.debug { "executing: #{interpreter} #{args.join(" ")}" }

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -33,13 +35,13 @@ def do_remove(name, type)
...
       std_in = std_out = std_err = nil
+      Open3.popen3(interpreter, *args)
```
lzap: I believe this is incorrect, this belong to the line 46, this method returns array of three values.

For easier testing, I suggest you you create a private method named `popen3` so you can easily mock it without mocking the system method from Open3 module (this can sometimes be painful).

test/dns_dnscmd/dnscmd_test.rb
```diff
@@ -21,25 +22,33 @@ def setup
...
+    Proxy::Dns::Dnscmd::Record.any_instance.expects(:execute).with('/RecordAdd', 'bar.domain.local', 'alias.foo.bar.domain.local.', 'CNAME', 'host.foo.bar.domain.local', anything).returns(true)
     assert_nil @server.do_create('alias.foo.bar.domain.local', 'host.foo.bar.domain.local', 'CNAME')
   end
 
+  def test_popen_arguments
+    interpreter = Proxy::SETTINGS.x86_64 ? 'c:\windows\sysnative\cmd.exe' : 'c:\windows\system32\cmd.exe'
+    args = ['/RecordAdd', 'bar.domain.local', 'alias.foo.bar.domain.local.', 'CNAME', 'host.foo.bar.domain.local']
+    command = interpreter + ' /c c:\Windows\System32\dnscmd.exe ' + "#{@server}" + args.join(" ")
+    Open3.expects(:popen3).with(command, *args).once.returns(true)
+    Open3.popen3(command, *args)
```
lzap: Yeah I suggest to create a helper method `popen3` which you can easily mock like:

    @server.expects(:popen3).with(xyz).once.returns([nil, nil, nil])

Then in the main method you also need to add an extra code that is missing so nil exception is not thrown:

```ruby
          response = std_out&.readlines
          response += std_err&.readlines
```

So the method does nothing and returns response which should be `nil`.

lzap (changes requested):
All right, almost there!

test/dns_dnscmd/dnscmd_test.rb
```diff
@@ -1,6 +1,7 @@
 require 'test_helper'
 require 'dns_common/dns_common'
 require 'dns_dnscmd/dns_dnscmd_main'
+require 'open3'
```
lzap: This should no be required, you do not want to call this actually.

test/dns_dnscmd/dnscmd_test.rb
```diff
@@ -21,25 +22,33 @@ def setup
...
   end
 
   def test_create_cname_record
-    Proxy::Dns::Dnscmd::Record.any_instance.expects(:execute).with('/RecordAdd bar.domain.local alias.foo.bar.domain.local. CNAME host.foo.bar.domain.local', anything).returns(true)
+    Proxy::Dns::Dnscmd::Record.any_instance.expects(:execute).with('/RecordAdd', 'bar.domain.local', 'alias.foo.bar.domain.local.', 'CNAME', 'host.foo.bar.domain.local', anything).returns(true)
     assert_nil @server.do_create('alias.foo.bar.domain.local', 'host.foo.bar.domain.local', 'CNAME')
   end
 
+  def test_popen_arguments
+    interpreter = Proxy::SETTINGS.x86_64 ? 'c:\windows\sysnative\cmd.exe ' : 'c:\windows\system32\cmd.exe '
```
lzap: Here is a tailing space that is failing your tests.

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -5,6 +5,12 @@ class Record < ::Proxy::Dns::Record
     include Proxy::Log
     include Proxy::Util
 
+    private
+    def popen3(*args)
+      Open3.popen3(*args)
+    end
+
+    public
```
lzap: It is a common pattern to start with public methods first and move private section towards the end of the class, so you do not need to use `public` keyword at all.

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -33,19 +37,17 @@ def do_remove(name, type)
...
-
       std_in = std_out = std_err = nil
       begin
         timeout(tsecs) do
-          logger.debug "executing: #{command}"
-          std_in, std_out, std_err = Open3.popen3(command)
-          response = std_out.readlines
-          response += std_err.readlines
+          logger.debug { "executing: #{interpreter} #{args.join(" ")}" }
+          std_in, std_out, std_err = popen3(interpreter, *args)
```
lzap: Here is a bug, previously it was: `cmd.exe` calls `dnscmd.exe` + arguments.

Now it is: `cmd.exe` + arguments that will not be understood by this program.

How it should be: `dnscmd.exe` + arguments:

```suggestion
          std_in, std_out, std_err = popen3('c:\Windows\System32\dnscmd.exe', @server, *args)
```

You can completely drop helper interpreter string, I suggest you create a helper array called `real_cmd` and then put `dnscmd.exe` + `@server` into it and append the `cmd` argument. That you can pass into `popen3` as well as into logging methods.

test/dns_dnscmd/dnscmd_test.rb
```diff
@@ -21,25 +21,33 @@ def setup
   end
 
   def test_create_a_record_with_longest_zone_match
-    Proxy::Dns::Dnscmd::Record.any_instance.expects(:execute).with('/RecordAdd bar.domain.local host.foo.bar.domain.local. A 192.168.33.33', anything).returns(true)
+    Proxy::Dns::Dnscmd::Record.any_instance.expects(:execute).with('/RecordAdd', 'bar.domain.local', 'host.foo.bar.domain.local.', 'A', '192.168.33.33', anything).returns(true)
```
lzap: Please change all tests in the following way:

```suggestion
    @server.expects(:popen3).with('c:\Windows\System32\dnscmd.exe', '_msdcs.bar.domain.local', '/RecordAdd', 'bar.domain.local', 'alias.foo.bar.domain.local.', 'CNAME', 'host.foo.bar.domain.local').once.returns([nil, nil, nil])
```

Not sure about

lzap (changes requested):
Almost there!

lzap:
[ok to test]

lzap:
[ok to test]

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -33,19 +31,17 @@ def do_remove(name, type)
...
 
-    def execute(cmd, msg = nil, error_only = false)
+    def execute(*args, **keyword_args)
       tsecs = 5
       response = nil
-      interpreter = Proxy::SETTINGS.x86_64 ? 'c:\windows\sysnative\cmd.exe' : 'c:\windows\system32\cmd.exe'
-      command = interpreter + ' /c c:\Windows\System32\dnscmd.exe ' + "#{@server} #{cmd}"
-
       std_in = std_out = std_err = nil
+      real_cmd = ['c:\Windows\System32\dnscmd.exe' + "#{@server} #{args.join(' ')}"]
```
lzap: Okay, now it's an array, but just with one element:

    ["command arg arg arg"]

It needs to be literally:

    ["command", "arg", "arg", "arg"]

lzap (changes requested):
Still not there, but it is just a detail and a small final change.

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -33,19 +31,17 @@ def do_remove(name, type)
...
 
-    def execute(cmd, msg = nil, error_only = false)
+    def execute(*args, **keyword_args)
       tsecs = 5
       response = nil
-      interpreter = Proxy::SETTINGS.x86_64 ? 'c:\windows\sysnative\cmd.exe' : 'c:\windows\system32\cmd.exe'
-      command = interpreter + ' /c c:\Windows\System32\dnscmd.exe ' + "#{@server} #{cmd}"
-
       std_in = std_out = std_err = nil
+      real_cmd = ['c:\Windows\System32\dnscmd.exe' + "#{@server}"] + args
```
lzap: This needs to be:

```suggestion
      real_cmd = ['c:\Windows\System32\dnscmd.exe', @server] + args
```

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -33,19 +31,17 @@ def do_remove(name, type)
...
       begin
         timeout(tsecs) do
-          logger.debug "executing: #{command}"
-          std_in, std_out, std_err = Open3.popen3(command)
-          response = std_out.readlines
-          response += std_err.readlines
+          logger.debug { "executing: #{real_cmd}" }
+          std_in, std_out, std_err = popen3(real_cmd)
```
lzap: Also, you cannot pass array into popen3 method, it will pass it as an array, that will not work:

```
>> x, y, z = Open3.popen3(["/usr/bin/bash", "-c", "echo hello"])
Traceback (most recent call last):
        8: from /home/lzap/.rbenv/versions/2.7.2/bin/irb:23:in `<main>'
        7: from /home/lzap/.rbenv/versions/2.7.2/bin/irb:23:in `load'
        6: from /home/lzap/.rbenv/versions/2.7.2/lib/ruby/gems/2.7.0/gems/irb-1.2.6/exe/irb:11:in `<top (required)>'
        5: from (irb):9
        4: from (irb):10:in `rescue in irb_binding'
        3: from /home/lzap/.rbenv/versions/2.7.2/lib/ruby/2.7.0/open3.rb:101:in `popen3'
        2: from /home/lzap/.rbenv/versions/2.7.2/lib/ruby/2.7.0/open3.rb:213:in `popen_run'
        1: from /home/lzap/.rbenv/versions/2.7.2/lib/ruby/2.7.0/open3.rb:213:in `spawn'
ArgumentError (wrong first argument)
```

You need to unwrap the array here:

```
>> x, y, z = Open3.popen3("/usr/bin/bash", "-c", "echo hello")
=> [#<IO:fd 8>, #<IO:fd 10>, #<IO:fd 12>, #<Process::Waiter:0x0000000000e53ee0 run>]
```

This can be done with:

```suggestion
          std_in, std_out, std_err = popen3(*real_cmd)
```

lzap (changes requested):
Few more.

lzap:
There are relevant test failures.

ares:
seems unrelated [test] [test smart-proxy]

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -33,19 +31,17 @@ def do_remove(name, type)
...
       begin
         timeout(tsecs) do
-          logger.debug "executing: #{command}"
-          std_in, std_out, std_err = Open3.popen3(command)
-          response = std_out.readlines
-          response += std_err.readlines
+          logger.debug { "executing: #{real_cmd}" }
+          std_in, std_out, std_err = popen3(*real_cmd)
+          response = std_out&.readlines
+          response.nil? ? response = std_err&.readlines : response += std_err&.readlines
```
lzap: While this is correct, it is a bit unreadable. I would not recommend to use this operator, something slower but more readable:

```
response = [std_out&.readlines, std_err&.readlines]
...
response.flatten.compact
```

lzap (changes requested):
Styling nitpick but looks great.

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -13,15 +13,13 @@ def do_create(name, value, type)
...
-      execute(cmd, msg)
+      execute("/RecordAdd", zone, "#{name}.", type, value, msg: msg, error_only: false)
       nil
     end
 
     def remove_specific_record_from_zone(zone_name, node_name, record, type)
       msg = "Removed #{record} #{type} record #{node_name} from #{zone_name}"
-      cmd = "/RecordDelete #{zone_name} #{node_name}. #{type} #{record} /f"
-      execute(cmd, msg)
+      execute("/RecordDelete", zone_name, "#{node_name}.", type, "#{record} /f", msg: msg, error_only: false)
```
lzap: One more thing I noticed, every argument must be a separate array element. Here the `/f` argument is passed with a record string. There is difference if you call an executable with:

    executable RECORD /f

or

    executable "RECORD /f"

The first one is two arguments (that is expected) however the second form is a single argument with "RECORD /f" string, that would fail actually.

test/dns_dnscmd/dnscmd_test.rb
```diff
@@ -54,12 +54,12 @@ def test_remove_ptr_records
   end
 
   def test_remove_specific_a_record_from_zone
-    Proxy::Dns::Dnscmd::Record.any_instance.expects(:execute).with('/RecordDelete domain.local host.domain.local. A 192.168.33.33 /f', anything).returns(true)
+    Proxy::Dns::Dnscmd::Record.any_instance.expects(:execute).with('/RecordDelete', 'domain.local', 'host.domain.local.', 'A', '192.168.33.33 /f', anything).returns(true)
```
lzap: Here it needs to be updated as well, "/f" must be an array element.

test/dns_dnscmd/dnscmd_test.rb
```diff
@@ -54,12 +54,12 @@ def test_remove_ptr_records
...
     assert_nil @server.remove_specific_record_from_zone('domain.local', 'host.domain.local', '192.168.33.33', 'A')
   end
 
   def test_remove_specific_ptr_record_from_zone
-    Proxy::Dns::Dnscmd::Record.any_instance.expects(:execute).with('/RecordDelete 33.168.192.in-addr.arpa 33.33.168.192.in-addr.arpa. PTR host.domain.local /f', anything).returns(true)
+    Proxy::Dns::Dnscmd::Record.any_instance.expects(:execute).with('/RecordDelete', '33.168.192.in-addr.arpa', '33.33.168.192.in-addr.arpa.', 'PTR', 'host.domain.local /f', anything).returns(true)
```
lzap: Ditto.

modules/dns_dnscmd/dns_dnscmd_main.rb
avitova: I see. I misunderstood the meaning of that parameter. I am about to fix it.

lzap (approved):
Thanks for the patience, merged.

lzap:
For the record, we had a user hitting a bug in 3.1, I have missed during the review. Here is a fix: https://github.com/theforeman/smart-proxy/pull/811

PR #797, opened 2021-08-26
Fixes #33356 - APP_ROOT defined for tests

lzap:
OpenSCAP plugin class requires this to be defined, tests do fail when openscap plugin is enabled.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ pkg/
 test/tmp/*
 /tmp/
 /vendor
+/data/*
--- a/data/README
+++ b/data/README
@@ -1,0 +1,6 @@
+This directory is reserved for data in development mode. To access it, use
+DATA_DIR constant. Plugins should create subdirectories within the directory.
+
+The constant is set to /var/lib/foreman-proxy, if such directory exists.
+Otherwise this directory is used. When running tests, DATA_DIR is set to
+../tmp/data.
--- a/lib/smart_proxy_for_testing.rb
+++ b/lib/smart_proxy_for_testing.rb
@@ -26,6 +26,10 @@ require 'launcher'
 require 'sinatra/base'
 require 'sinatra/authorization'
 
+APP_ROOT = File.absolute_path(File.expand_path("#{__dir__}/.."))
+DATA_ROOT = File.absolute_path(File.expand_path("#{__dir__}/../tmp/data"))
+FileUtils.mkdir_p(DATA_ROOT) unless Dir.exist?(DATA_ROOT)
+
 Proxy::SETTINGS = ::Proxy::Settings::Global.new(:log_file => './logs/test.log', :log_level => 'DEBUG')
 Proxy::VERSION = File.read(File.join(__dir__, '..', 'VERSION')).chomp
 
--- a/lib/smart_proxy_main.rb
+++ b/lib/smart_proxy_main.rb
@@ -1,4 +1,9 @@
-APP_ROOT = "#{__dir__}/.."
+APP_ROOT = File.absolute_path(File.expand_path("#{__dir__}/.."))
+if Dir.exist?('/var/lib/foreman-proxy')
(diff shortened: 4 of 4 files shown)


lzap:
Context: https://github.com/theforeman/smart_proxy_openscap/pull/86

test/test_helper.rb
```diff
@@ -2,6 +2,8 @@
 require "test/unit"
 require 'fileutils'
 
+APP_ROOT = "#{File.dirname(__FILE__)}/.."
```
ekohl:
```suggestion
APP_ROOT = "#{__dir__}/.."
```

ekohl (changes requested):
Please don't rush this. I've noticed that a lot of plugins are starting to build their own constants for this. In particular, I want to introduce something that in production should point to `/var/lib/foreman-proxy`. It looks like OpenSCAP needs that, but there are many others. Let's think about that properly.

lzap:
Rebased, I cannot wait with this, I am unable to continue on OpenSCAP refactoring due to this. I fully support resolving the common storage problem, but it has nothing to do with this - this constant is simply not available when we load plugins from unit tests without loading full smart proxy main stack.

ekohl (changes requested):
Let's start with the background. The `APP_ROOT` is defined here:
https://github.com/theforeman/smart-proxy/blob/d9b05020cb30c196e247213af482ed9b3075da90/lib/smart_proxy_main.rb#L1

This is an issue because you have smart_proxy_openscap in your bundle.

For testing, we have an alternative to smart_proxy_main, namely [smart_proxy_for_testing](https://github.com/theforeman/smart-proxy/blob/develop/lib/smart_proxy_for_testing.rb). I wonder if that  would be the correct place because it would then also be present in modules that don't use the test_helper here but define their own. However, then the root would be the Smart Proxy root, not the plugin root itself.

For reference:
https://github.com/theforeman/smart_proxy_openscap/blob/1af66df4083c5968f4e6489fed7fb6e4f7eb6c21/test/test_helper.rb#L9-L13

Overall I'd say this isn't so urgent because it only happens if smart_proxy_openscap is in your bundle.

lzap:
I moved this into `smart_proxy_for_testing`. It still resolves my problem.

> Overall I'd say this isn't so urgent because it only happens if smart_proxy_openscap is in your bundle.

In other words, you don't care about me needing to comment out OpenSCAP plugin everytime I work on smart-proxy core PR. Well, how cool is that.

ekohl:
There are other considerations, like what it means. `APP_ROOT` should really be the application root, but in the case of smart_proxy_openscap what's needed is a `DATA_ROOT`.

So putting it in `smart_proxy_for_testing` is correct from a naming perspective. On the other hand, the way it's used in smart_proxy_openscap is different. It means that if you have a git checkout installed via bundler your data files can end up somewhere in `~/.gem`.



> > Overall I'd say this isn't so urgent because it only happens if smart_proxy_openscap is in your bundle.
> 
> In other words, you don't care about me needing to comment out OpenSCAP plugin everytime I work on smart-proxy core PR. Well, how cool is that.

There's other workarounds. For example, use this in your `bundler.d` file:
```ruby
unless ENV['SKIP_OPENSCAP']
  gem ...
end
```

Then in your shell you can `export SKIP_OPENSCAP=1`.

lzap:
Can you just re-review the current version? There is nothing wrong with this proposal, APP_ROOT is smart proxy application root being defined in a testing ruby helper file.

lzap:
Ok, how about this - I created DATA_DIR constant and an empty directory in git named `./data/` which has contents in gitignore. It has a README file with explanation:

```
[lzap@nuc smart-proxy]$ cat data/README
This directory is reserved for data in development mode. To access it, use
DATA_DIR constant. Plugins should create subdirectories within the directory.

The constant is set to /var/lib/foreman-proxy, if such directory exists.
Otherwise this directory is used. When running tests, DATA_DIR is set to
../tmp/data.
```

PR #799, opened 2021-09-02
Refs #2412 - use curl for downloads

lzap:
This is continuation of an incorrect implementation we merged as https://github.com/theforeman/smart-proxy/pull/785

Unfortunately, `wget` in EL7 seems to be a bit old and required features are no there, however, `curl` has all the features and supports both HTTP and FTP. It can be configured also to print a nice message after download is finished which looks nice. And it correctly supports redownloading of newer files, it performs comparsion of last modification correctly as needed.

The downside is it only supports connect timeout, read/dns timeouts cannot be supported. It is a detail IMHO.

I am creating a draft, please test this. I did testing with:

* CentOS main mirror
* a random Czechia CentOS mirror
* Katello (Satellite 6.9)

Example curl to call proxy API:

```
curl -X POST -d "" "http://localhost:8448/tftp/fetch_boot_file?prefix=test&path=http://mirror.centos.org/centos/8-stream/BaseOS/x86_64/os/LICENSE"
```

And how it looks like in logs:

```
2021-09-02T10:11:36 bae83840 [I] [1270504] Started task /usr/bin/curl --silent --show-error --connect-timeout 10 --retry 3 --retry-delay 10 --max-time 3600 --remote-time --time-cond /var/lib/tftpboot/test-kernel.img --write-out "Task done, result: %{http_code}, size downloaded: %{size_download}b, speed: %{speed_download}b/s, time: %{time_total}ms" -o /var/lib/tftpboot/test-kernel.img --location http://zzzap.xxx.redhat.com/pulp/repos/Default_Organization/Library/content/dist/rhel8/8.0/s390x/baseos/kickstart/images/kernel.img
2021-09-02T10:11:37 bae83840 [D] [1270504] "Task done, result: 200, size downloaded: 4911272b, speed: 9390577b/s, time: 0.523823ms"
```

Tests are failing because I will fix them only if we agree to proceed with this solution.

There was a proposal to use HTTP ETag to do comparsion instead of last-modified, however, that would not work for FTP protocol.

### DIFF
--- a/lib/proxy/http_download.rb
+++ b/lib/proxy/http_download.rb
@@ -3,35 +3,57 @@ require 'proxy/file_lock'
 module Proxy
   class HttpDownload < Proxy::Util::CommandTask
     include Util
-    DEFAULT_READ_TIMEOUT = 60
     DEFAULT_CONNECT_TIMEOUT = 10
-    DEFAULT_DNS_TIMEOUT = 10
 
     def initialize(src, dst, read_timeout = nil, connect_timeout = nil, dns_timeout = nil, verify_server_cert = false)
       @dst = dst
-      wget = which("wget")
-      read_timeout ||= DEFAULT_READ_TIMEOUT
-      dns_timeout ||= DEFAULT_CONNECT_TIMEOUT
-      connect_timeout ||= DEFAULT_DNS_TIMEOUT
+      logger.warn('Deprecated: HttpDownload read_timeout is deprecated and will be removed in 4.0') if read_timeout
+      logger.warn('Deprecated: HttpDownload dns_timeout is deprecated and will be removed in 4.0') if dns_timeout
+      connect_timeout ||= DEFAULT_CONNECT_TIMEOUT
+      args = [which('curl')]
 
-      args = ["--connect-timeout=#{connect_timeout}",
-              "--dns-timeout=#{dns_timeout}",
-              "--read-timeout=#{read_timeout}",
-              "--tries=3",
-              "--timestamping", # turn on timestamping to prevent redownloads
-              "--no-if-modified-since", # but use HTTP HEAD instead If-Modified-Since
-              "-nv", src.to_s, "-O", dst.to_s]
-      args.unshift("--no-check-certificate") unless verify_server_cert
-      args.unshift(wget)
+      # no cert verification if set
+      args << "--insecure" unless verify_server_cert
+      # print nothing
+      args << "--silent"
+      # except errors
+      args << "--show-error"
+      # timeout (others were supported by wget but not by curl)
+      args += ["--connect-timeout", connect_timeout.to_s]
+      # try several times
+      args += ["--retry", "3"]
+      # with a short delay
+      args += ["--retry-delay", "10"]
+      # but exit after one hour
+      args += ["--max-time", "3600"]
+      # keep last changed file attribute
+      args << "--remote-time"
+      # only download newer files
+      args += ["--time-cond", dst.to_s]
+      # print stats in the end
+      args += [
+        "--write-out",
+        'Task done, result: %{http_code}, size downloaded: %{size_download}b, speed: %{speed_download}b/s, time: %{time_total}ms',
+      ]
+      # output file
+      args += ["--output", dst.to_s]
+      # follow redirects
+      args << "--location"
+      # and the url to download
+      args << src.to_s
+
       super(args)
     end
 
     def start
-      lock = Proxy::FileLock.try_locking(@dst)
+      lock = Proxy::FileLock.try_locking(File.join(File.dirname(@dst), ".#{File.basename(@dst)}.lock"))
       if lock.nil?
         false
       else
-        super { Proxy::FileLock.unlock(lock) }
+        super do
+          Proxy::FileLock.unlock(lock)
+          File.unlink(lock)
+        end
       end
     end
   end
--- a/lib/proxy/util.rb
+++ b/lib/proxy/util.rb
@@ -23,7 +23,8 @@ module Proxy::Util
       @task = Thread.new(@command, @input) do |cmd, input|
         status = nil
         Open3.popen3(*cmd) do |stdin, stdout, stderr, thr|
-          logger.info "[#{thr.pid}] Started task #{cmd}"
+          cmdline_string = Shellwords.escape(cmd.is_a?(Array) ? cmd.join(' ') : cmd)
+          logger.info "[#{thr.pid}] Started task #{cmdline_string}"
           stdin.write(input) if input
           stdin.close
           stdout.each do |line|
--- a/config/settings.d/tftp.yml.example
+++ b/config/settings.d/tftp.yml.example
@@ -6,17 +6,9 @@
 # Defines the TFTP Servername to use, overrides the name in the subnet declaration
 #:tftp_servername: tftp.domain.com
 
-# Defines the default read timeout in seconds needed to download tftp artifacts
-# like initrd and vmlinuz. Default value 60 seconds
-#:tftp_read_timeout: 60
-
 # Defines the default connection timeout in seconds needed to download tftp artifacts
 # like initrd and vmlinuz. Default value 10 seconds
 #:tftp_connect_timeout: 10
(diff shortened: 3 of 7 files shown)


lib/proxy/http_download.rb
```diff
@@ -3,35 +3,55 @@
...
+      args << "--location"
+      # and the url to download
+      args << src.to_s
+
+      super(args.flatten)
     end
 
     def start
-      lock = Proxy::FileLock.try_locking(@dst)
+      lock = Proxy::FileLock.try_locking("#{@dst}.lock")
```
lzap: This made me a ton of headache this morning when I was debugging it, it confused curl so it would not download anything as the lock created was always newer than the source file. So now it creates an explicit lock file and it is deleted after download is finished.

lib/proxy/util.rb
```diff
@@ -23,7 +23,7 @@ def start(&ensured_block)
       @task = Thread.new(@command, @input) do |cmd, input|
         status = nil
         Open3.popen3(*cmd) do |stdin, stdout, stderr, thr|
-          logger.info "[#{thr.pid}] Started task #{cmd}"
+          logger.info "[#{thr.pid}] Started task #{cmd.join(' ')}"
```
lzap: This allows you to copy and paste commands from logs to simulate the command (when something goes wrong). Pretty handy.

lib/proxy/http_download.rb
```diff
@@ -3,35 +3,55 @@
...
+      args << ["--max-time", "3600"]
+      # keep last changed file attribute
+      args << "--remote-time"
+      # only download newer files
+      args << ["--time-cond", dst.to_s]
+      # print stats in the end
+      args << [
+        "--write-out",
+        '"Task done, result: %{http_code}, size downloaded: %{size_download}b, speed: %{speed_download}b/s, time: %{time_total}ms"'
+      ]
```
lzap, line 37: Noticed unnecessary double quotes, will remove.

lzap:
Can you take a look before I update tests? I am in hurry because it looks like the regression caused by my previous patch is getting into 3.0, we want to fix this ASAP and cherry pick as well.

ares:
Well, nightlies are still broken on el7, wget does not work at all because of unknown attribute `--no-if-modified-since` argument and this is still a draft. If this is going to take more time, should we revert the previous patch? Or do we feel we're close?

lzap:
I am looking for confirmation this is acceptable solution from @ekohl before I proceed with any test changes.

lib/proxy/http_download.rb
```diff
@@ -3,35 +3,55 @@
...
+      # output file
+      args << ["-o", dst.to_s]
```
ekohl: Slight nit, but I'd recommend long arguments. Then the comment is redundant and it's clear it's not a typo. For example, you could have meant `-O` (which is `--remote-name`):
```suggestion
      args << ["--output", dst.to_s]
```

lib/proxy/http_download.rb
```diff
@@ -3,35 +3,55 @@
...
-      args.unshift(wget)
-      super(args)
+      # no cert verification if set
+      args << "--insecure" unless verify_server_cert
+      # print nothing
+      args << "--silent"
+      # except errors
+      args << "--show-error"
+      # timeout (others were supported by wget but not by curl)
+      args << ["--connect-timeout", connect_timeout.to_s]
```
ekohl: I'm not a fan of adding arrays and then flattening it. How about:
```suggestion
      args += ["--connect-timeout", connect_timeout.to_s]
```

lib/proxy/http_download.rb
ekohl: I'm not strictly opposed, but I wonder if we can make it a hidden file. Also wondering if it should use the target directory or rather some temporary other place.

ekohl (commented):
I think overall this makes sense. Some coding style issues inline but I'd be OK with this.

We should update packaging after this is merged. Right now it depends on wget and that should be changed to curl.

Also tests need to be updated, but given this is a draft I'm sure you're asking about the general direction of this, not the specific implementation.

lib/proxy/util.rb
ekohl: Should it use `Shellwords.escape` to deal with spaces etc?

lib/proxy/http_download.rb
lzap: Sure no problem.

lib/proxy/util.rb
lzap: Sure.

lib/proxy/http_download.rb
lzap: Done

lib/proxy/http_download.rb
lzap: Good point, makes sense.

lzap:
Rebased, incorporated your comments, fixed tests. Can I ask for a quick re-review @ekohl ? Reason being we actually merged my patch which broke TFTP in nightlies, this fixes it properly.

lib/proxy/http_download.rb
```diff
@@ -3,35 +3,56 @@
...
-              "--no-if-modified-since", # but use HTTP HEAD instead If-Modified-Since
-              "-nv", src.to_s, "-O", dst.to_s]
-      args.unshift("--no-check-certificate") unless verify_server_cert
-      args.unshift(wget)
       super(args)
     end
 
     def start
-      lock = Proxy::FileLock.try_locking(@dst)
+      lock = Proxy::FileLock.try_locking("#{File.dirname(@dst)}/.#{File.basename(@dst)}.lock")
```
ekohl: I'm always a fan of using `File.join()`.
```suggestion
      lock = Proxy::FileLock.try_locking(File.join(File.dirname(@dst), ".#{File.basename(@dst)}.lock"))
```

lib/proxy/http_download.rb
```diff
@@ -3,35 +3,56 @@
 module Proxy
   class HttpDownload < Proxy::Util::CommandTask
     include Util
-    DEFAULT_READ_TIMEOUT = 60
     DEFAULT_CONNECT_TIMEOUT = 10
-    DEFAULT_DNS_TIMEOUT = 10
 
     def initialize(src, dst, read_timeout = nil, connect_timeout = nil, dns_timeout = nil, verify_server_cert = false)
```
ekohl, line 8: It looks like read_timeout and dns_timeout are no longer used. Shouldn't we drop those instead? Or at least issue a deprecation warning?

ekohl (commented):
Overall this looks good. Some inline comments.

lib/proxy/http_download.rb
lzap: Hmm I just realized these are exposed as settings, this is from 2018 so it is likely in active use. Deprecation is then the best way to do this.

lzap:
Rebased, deprecated those settings and arguments.

I will followup with a patch that will further refactor the HttpDownload - I think what this asks for are named arguments just in case we would like to extend it again in the future.

ekohl (approved):
Thanks!

lzap:
Fixing Rubocop.

lzap:
All green.

ekohl (approved):

lzap:
Thanks, I don't see the 3.1 branch yet so I made it, no cherrypicking needed. That was close!

lzap:
For the record, here it is: https://github.com/theforeman/smart-proxy/pull/804

ares:
phew, thanks! this is killing me the whole dev phase, I'm glad it will be resolved for 3.1

ekohl:
> We should update packaging after this is merged. Right now it depends on wget and that should be changed to curl.

RPM: https://github.com/theforeman/foreman-packaging/pull/7257
Deb: https://github.com/theforeman/foreman-packaging/pull/7258

lzap:
> phew, thanks! this is killing me the whole dev phase, I'm glad it will be resolved for 3.1

I was literally falling asleep with this on my mind: "Oh and I need to fix broken TFTP downloads in the nightlies before branching..."

Thanks Ewoud for quick action and packaging which I forgot!

PR #801, opened 2021-09-20
Fixes #33520 - Silence bundler

evgeni:


evgeni:
contrary to https://github.com/theforeman/foreman/commit/f25a21e1b0c9642209df91558a9b51e24e820af9 this doesn't hook into `if production` as I didn't know where such a thing is defined in the Proxy without Rails and Sinatra is initialized too late:

https://github.com/theforeman/smart-proxy/blob/603ce4c4ae4b1be149fe510136d2f4e4cfc91e7c/lib/smart_proxy_main.rb#L48

lib/bundler_helper.rb
```diff
@@ -19,9 +19,11 @@ def self.require_groups(*groups)
             exit 1
           end
         end
+        Bundler.ui = Bundler::UI::Silent.new
         BundlerExt.system_require(File.expand_path('../Gemfile.in', __dir__), *groups)
       else
         require 'bundler' unless defined?(Bundler)
+        Bundler.ui = Bundler::UI::Silent.new
```
ekohl: In production we should have `Gemfile.in` and use `bundler_ext` and this would be the development case. I think we should leave it out here.
```suggestion
```

ekohl (changes requested):

ekohl (approved):

ekohl:
@lzap want to take a look as well?

lib/bundler_helper.rb
```diff
@@ -19,6 +19,7 @@ def self.require_groups(*groups)
             exit 1
           end
         end
+        Bundler.ui = Bundler::UI::Silent.new
```
ezr-ondrej, line 22: Why do we silence only on the `BundlerExt` environments?
Do I get this correctly this is only RedHat family production by default?

evgeni: https://github.com/theforeman/smart-proxy/pull/801#discussion_r712922320

ekohl: >  Do I get this correctly this is only RedHat family production by default?

https://github.com/theforeman/foreman-packaging/blob/58266ab62271af0bd8be57092868577e8a0212b9/debian/buster/foreman-proxy/control#L15

ruby-bundler-ext is in there so I think if bundler_ext is present, we can assume it's production

lzap:
Yeah sorry no comments, thanks.

PR #802, opened 2021-10-20
Fixes #33756 - Require 'templates' for 'registration' module

stejskalleos:
Related to https://github.com/theforeman/foreman/pull/8855

### DIFF
--- a/modules/registration/registration_plugin.rb
+++ b/modules/registration/registration_plugin.rb
@@ -4,5 +4,6 @@ module Proxy::Registration
     https_rackup_path File.expand_path("http_config.ru", File.expand_path(__dir__))
 
     plugin :registration, ::Proxy::VERSION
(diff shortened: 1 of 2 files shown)


ekohl:
Could you open a new Redmine issue for this? We don't allow cross project refs.
The test failure is also related and shows exactly what is supposed to happen. Perhaps you can have a test to assert the failure and a test where templates is enabled and assert success?

Edit: this should probably also be reflected in the installer. There it should either error out early or automatically enable the feature.

ekohl (commented):
Could you also add a test case where templates isn't loaded and assert that it fails? Other than that :+1:

stejskalleos:
> Could you also add a test case where templates isn't loaded and assert that it fails? Other than that +1

Test added

ekohl (approved):
Thanks!

Edit: tests are still running.

PR #803, opened 2021-10-25
Fixes #33772 - improve DHCP parser binding test

lzap:
In the https://community.theforeman.org/t/dhcp-entries-never-deleted/25688 converstaion I wanted to ensure Foreman does not remove/override reservation with expired leases (in the `free` binding state). So I wrote a test.

### DIFF
--- a/test/dhcp/dhcp_isc_subnet_service_initialization_test.rb
+++ b/test/dhcp/dhcp_isc_subnet_service_initialization_test.rb
@@ -138,7 +138,7 @@ class DhcpIscSubnetServiceInitializationTest < Test::Unit::TestCase
     @initialization.load_leases_file(File.read("./test/fixtures/dhcp/dhcp.leases"))
     parsed = @subnet_service.all_hosts + @subnet_service.all_leases
 
-    assert_equal 12, parsed.size
+    assert_equal 13, parsed.size
     assert_equal 'bravo.example.com', @subnet_service.find_host_by_hostname("bravo2.example.com").hostname
   end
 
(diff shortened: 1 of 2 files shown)


test/dhcp/dhcp_isc_subnet_service_initialization_test.rb
```diff
@@ -138,7 +138,7 @@ def test_parse_config_and_leases
     @initialization.load_leases_file(File.read("./test/fixtures/dhcp/dhcp.leases"))
     parsed = @subnet_service.all_hosts + @subnet_service.all_leases
 
-    assert_equal 12, parsed.size
+    assert parsed.size > 0
```
ekohl: This feels like a much weaker assertion. Can't we use a specific number?

test/fixtures/dhcp/dhcp.leases
```diff
@@ -204,20 +204,38 @@ host static.example.com {
...
+# Prevent override of a reserved record via free lease by MAC
+host quax.example.org {
+  dynamic;
+  hardware ethernet 52:54:00:31:a3:98;
+  fixed-address 192.168.122.53;
+  supersede server.filename = "pxelinux.0";
+  supersede server.next-server = ac:14:0a:01;
+  supersede host-name = "quax.example.org";
+}
+lease 192.168.122.55 {
```
ekohl, line 234: Is it intentional that this address differs from the `fixed-address`? I'm not sure what this example is saying since I'm not too familiar with the DHCP lease file format.

ekohl (commented):
This is tough for me to review since I'm not that familiar with the DHCP lease file format.

test/fixtures/dhcp/dhcp.leases
lzap: That is actually the bug the user was running into - a node requests a lease (.55) and THEN it is provisioned in Foreman and reservation is created (.53). Note the MAC address is the same. Now, after a lease expire, typically after 10 minutes, DHCP server will append new record that says `binding state free`. And I thought we have a bug in our parser, created the test but it turned out to be something else. I thought it is a good idea to add this use case.

lzap:
Rebased.

ekohl (approved):
I'm not too familiar with the DHCP lease file and I'm going to trust you on this.

PR #804, opened 2021-11-09
Fixes #33881 - deprecations for 3.2

lzap:
Ton of deprecations, I searched the codebase and here are several commits which either deprecate some stuff or keeps the code in forever. Most of this was created by Dmitry when he was writing the DI plugin framework. I think some are still useful and we should not deprecate them, some are obvious.

The one I am not sure about is the DNS common module one, whole no core plugin use this, this might break others.

I suggest to review commit by commit. Do not merge before 3.1 branching! This is post-branch material.

### DIFF
--- a/lib/proxy/plugins.rb
+++ b/lib/proxy/plugins.rb
@@ -51,10 +51,6 @@ class ::Proxy::Plugins
     end
   end
 
-  #
-  # below are methods that are going to be removed/deprecated
-  #
-
   def enabled_plugins
     loaded.select { |p| p[:state] == :running && p[:class].ancestors.include?(::Proxy::Plugin) }.map { |p| p[:class] }
   end
--- a/modules/dhcp/dhcp_api.rb
+++ b/modules/dhcp/dhcp_api.rb
@@ -42,22 +42,6 @@ class Proxy::DhcpApi < ::Sinatra::Base
     log_halt 400, e
   end
 
-  # Deprecated, returns a single record
-  get "/:network/:record" do
-    content_type :json
-
-    logger.warn('GET dhcp/:network/:record endpoint has been deprecated and will be removed in future versions. '\
-                'Please use GET dhcp/:network/mac/:mac_address or GET dhcp/:network/ip/:ip_address instead.')
-
-    record = server.find_record(params[:network], params[:record])
-    log_halt 404, "No DHCP record for #{params[:network]}/#{params[:record]} found" unless record
-    {:hostname => (record.hostname rescue record.name), :ip => record.ip, :mac => record.mac }.to_json
-  rescue ::Proxy::DHCP::SubnetNotFound
-    log_halt 404, "Subnet #{params[:network]} could not found"
-  rescue => e
-    log_halt 400, e
-  end
-
   # returns an array of records for an ip address
   get "/:network/ip/:ip_address" do
     server.validate_supported_address(params[:network], params[:ip_address])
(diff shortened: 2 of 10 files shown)


lib/proxy/plugins.rb
```diff
@@ -51,10 +51,6 @@ def select
     end
   end
 
-  #
-  # below are methods that are going to be removed/deprecated
-  #
-
   def enabled_plugins
     loaded.select { |p| p[:state] == :running && p[:class].ancestors.include?(::Proxy::Plugin) }.map { |p| p[:class] }
```
ekohl, line 55: I think these are deprecated because they can easily be replaced with the select method here. Surprisingly, nothing actually calls these so I wonder why you think it's useful. For example, the `root` module can't use them due the `.map {}` at the end. The same is true for the other methods as well, except `find_provider`.

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -37,7 +38,7 @@ def execute(*args, **keyword_args)
       std_in = std_out = std_err = nil
       real_cmd = ['c:\Windows\System32\dnscmd.exe', @server] + args
       begin
-        timeout(tsecs) do
+        Timeout::timeout(tsecs) do
```
ekohl: This line looks good, but a bit odd in the commit that it's in.

modules/dhcp/dhcp_api.rb
```diff
@@ -42,22 +42,6 @@ class Proxy::DhcpApi < ::Sinatra::Base
     log_halt 400, e
   end
 
-  # Deprecated, returns a single record
```
ekohl, line 45: The commit message for this should say "remove deprecated methods in DHCP API"

lzap:
Quick grep for my plugins that I have:

```
[lzap@nuc work]$ ag HttpDownload smart*proxy*/
smart_proxy_discovery_image/lib/smart_proxy_discovery_image/power_api.rb
86:      download_thread = ::Proxy::HttpDownload.new(url, File.join(prefix, name)).start
```

This is safe, it does not use arguments at all. Which reminds me - I will need to check if `curl` is installed on the FDI.

I checked the deprecated methods from DNS common module and it is all fine. It should be all good.

modules/dns_common/dependency_injection/dependencies.rb
```diff
@@ -3,9 +3,8 @@ module DependencyInjection
     class Dependencies
       extend Proxy::Log
       extend Proxy::DependencyInjection::Wiring
+
       def self.container_instance
-        logger.warn('Proxy::Dns::DependencyInjection::Dependencies class has been deprecated and will be removed in future versions. '\
-          'Please use ::Proxy::Dns::YourPlugin#load_dependency_injection_wirings instead.')
```
ekohl: Are you sure it's in heavy use? Can you point me to some use of it?

test/http_download_test.rb
```diff
@@ -25,6 +25,23 @@ def test_regular_call
     assert_equal expected, Proxy::HttpDownload.new('src', 'dst').command
   end
 
+  def test_regular_call_with_ssl_verify
```
ekohl, line 28: Was this test case intended for the other PR that was already merged?

ekohl:
> This is safe, it does not use arguments at all. Which reminds me - I will need to check if `curl` is installed on the FDI.

That reminds me, we need to update foreman-packaging to reflect it as well. We can drop the wget dependency and replace it with curl.

lib/proxy/plugins.rb
lzap: Ansible used them, this was actually a reason why I started this patch. I am hesitant to remove these as I see nothing wrong with these methods.

modules/dns_common/dependency_injection/dependencies.rb
lzap: Well not heavy use, that was a bad grep. Smart Proxy IPAM plugin does appear to use it somehow, but I am not sure the DI stuff is unnecessary complex. I have no intentions to be fixing code that makes use of it, so I am removing this from my patchset.

test/http_download_test.rb
lzap: No, it is meant to test optional arguments. In the current test file there were zero tests for this.

lzap:
Dropped the commit in question, made suggested changes.

lib/proxy/plugins.rb
ekohl: Where? I can't find an example.

modules/dns_common/dependency_injection/dependencies.rb
ekohl: No, that's just regular dependency injection. IPAM doesn't use this class and you can safely drop it.

lzap:
Rebased.

lib/proxy/plugins.rb
ekohl: Please answer this. IMHO they should be removed instead.

ekohl:
I hope you don't mind I opened https://github.com/theforeman/smart-proxy/pull/840. I kept the commits I agreed with (slightly reworded) and added more.

lzap:
Thank you very much for this, feel free to carry over any my open PR and finish it, or completely redesign it :) Closing.

PR #807, opened 2021-12-16
Fixes #34162 - Allow calling load_test_settings without settings

ekohl:
This allows loading only the default settings:

```ruby
Proxy::MyPlugin::Plugin.load_test_settings
```

Or

```ruby
Proxy::MyPlugin::Plugin.load_test_settings()
```

Previously this had to be written as

```ruby
Proxy::MyPlugin::Plugin.load_test_settings({})
```


lzap (approved):
Sure.

PR #808, opened 2021-12-16
Fixes #34163 - Allow clearing the log buffer

ekohl:
In testing it can be useful to reset the log buffer. This is not
intended to be used as a public API but tests can use:

```ruby
Proxy::LogBuffer::Buffer.instance.send(:clear)
```

### DIFF
--- a/test/bmc/integration_test.rb
+++ b/test/bmc/integration_test.rb
@@ -4,14 +4,7 @@ require 'root/root_v2_api'
 require 'bmc/bmc_plugin'
 require 'bmc/ipmi'
 
-class BmcApiFeaturesTest < Test::Unit::TestCase
-  include Rack::Test::Methods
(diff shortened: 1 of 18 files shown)


test/bmc/integration_test.rb
```diff
@@ -7,6 +7,10 @@
 class BmcApiFeaturesTest < Test::Unit::TestCase
   include Rack::Test::Methods
 
+  def setup
+    Proxy::LogBuffer::Buffer.instance.send(:clear)
+  end
```
lzap: Sure, can you introduce SmartProxyTestCase instead and do it there, then change parent class of all tests.

lzap (changes requested):
Also rebase.

ehelms (approved):

PR #810, opened 2022-01-03
Fixes #34205 - External IPAM Integration

grizzthedj:
@lzap

This long overdue PR is for merging the External IPAM features from the [smart_proxy_ipam](https://github.com/grizzthedj/smart_proxy_ipam) plugin (https://github.com/grizzthedj/smart_proxy_ipam) into Smart Proxy Core. 

This currently adds support for the below IPAM providers:

  1. phpIPAM
  2. Netbox

Tests for the External IPAM module have been written and are green. Please note that I am unable to run the entire test suite due to some mac install issues with the `rkerberos` gem.

Looking forward to feedback!

### DIFF
--- a/config/settings.d/externalipam.yml.example
+++ b/config/settings.d/externalipam.yml.example
@@ -1,0 +1,8 @@
+---
+:enabled: false
+
+# Built-in providers:
+#   1. phpIPAM: externalipam_phpipam
+#   2. Netbox:  externalipam_netbox
+
+:use_provider: externalipam_netbox
--- a/config/settings.d/externalipam_netbox.yml.example
+++ b/config/settings.d/externalipam_netbox.yml.example
@@ -1,0 +1,6 @@
+---
+# url is the hostname and path of the Netbox instance
+:url: 'https://netbox.example.com'
+
+# token is the Netbox API token
+:token: 'netbox_token'
--- a/config/settings.d/externalipam_phpipam.yml.example
+++ b/config/settings.d/externalipam_phpipam.yml.example
@@ -1,0 +1,12 @@
+---
+# url is the hostname and path of the phpIPAM instance.
+:url: 'https://phpipam.example.com'
+
+# The phpIPAM user name for authentication. Please note that an API Key also needs to be 
+# setup with the exact same name as the user name configured here. When setting up the API 
+# Key in phpIPAM, "User token" must be used for the "App Security" setting.
+:user: 'ipam_user'
+
+# The password for above user account. Note that this is the password of the user, and not 
+# the API Key itself.
+:password: 'ipam_password'
--- a/lib/proxy/validations.rb
+++ b/lib/proxy/validations.rb
@@ -9,6 +9,8 @@ module Proxy::Validations
   class InvalidIPAddress < Error; end
   class InvalidMACAddress < Error; end
   class InvalidSubnet < Error; end
+  class InvalidCidr < Error; end
+  class IpNotInCidr < Error; end
 
   private
 
@@ -59,6 +61,23 @@ module Proxy::Validations
   def validate_subnet(subnet)
     raise InvalidSubnet, "Invalid Subnet #{subnet}" unless subnet.is_a?(Proxy::DHCP::Subnet)
     subnet
+  end
+
+  def validate_cidr(address, prefix)
+    cidr = "#{address}/#{prefix}"
+    network = IPAddr.new(cidr)
+    if network != IPAddr.new(address)
+      raise InvalidCidr, "Network address #{address} should be #{network} with prefix #{prefix}"
+    end
+    cidr
+  rescue IPAddr::Error => e
+    raise Proxy::Validations::Error, e.to_s
+  end
+
+  def validate_ip_in_cidr(ip, cidr)
+    raise IpNotInCidr, "IP #{ip} is not in #{cidr}" unless IPAddr.new(cidr).include?(IPAddr.new(ip))
+  rescue IPAddr::Error => e
+    raise Proxy::Validations::Error, e.to_s
   end
 
   def validate_server(server)
--- a/modules/externalipam/api_resource.rb
+++ b/modules/externalipam/api_resource.rb
@@ -1,0 +1,55 @@
+require 'yaml'
+require 'json'
+require 'net/http'
+require 'uri'
+require 'externalipam/ipam_helper'
+
+module Proxy::Ipam
+  # Class to handle authentication and HTTP transactions with External IPAM providers
+  class ApiResource
+    include ::Proxy::Log
+    include Proxy::Ipam::IpamHelper
+
+    def initialize(params = {})
+      @api_base = params[:api_base]
+      @token = params[:token]
+      @auth_header = params[:auth_header] || 'Authorization'
+    end
+
+    def get(path, params = nil)
+      url = @api_base + path
+      url += "?#{URI.encode_www_form(params)}" if params
+      uri = URI(url)
+      request = Net::HTTP::Get.new(uri)
+      request[@auth_header] = @token
+      request['Accept'] = 'application/json'
+      request(request, uri)
+    end
+
+    def delete(path)
+      uri = URI(@api_base + path)
+      request = Net::HTTP::Delete.new(uri)
+      request[@auth_header] = @token
+      request['Accept'] = 'application/json'
+      request(request, uri)
+    end
+
+    def post(path, body = nil)
+      uri = URI(@api_base + path)
+      request = Net::HTTP::Post.new(uri)
+      request.body = body
+      request[@auth_header] = @token
+      request['Accept'] = 'application/json'
+      request['Content-Type'] = 'application/json'
+      request(request, uri)
+    end
+
+    private
+
+    def request(request, uri)
+      Net::HTTP.start(uri.hostname, uri.port, use_ssl: uri.scheme == 'https') do |http|
+        http.request(request)
+      end
+    end
+  end
+end
--- a/modules/externalipam/externalipam_plugin.rb
+++ b/modules/externalipam/externalipam_plugin.rb
@@ -1,0 +1,11 @@
+require 'externalipam/phpipam/phpipam_plugin'
+require 'externalipam/netbox/netbox_plugin'
+
+module Proxy::Ipam
+  class Plugin < ::Proxy::Plugin
+    plugin :externalipam, ::Proxy::VERSION
+    uses_provider
+    default_settings use_provider: nil
+    rackup_path File.expand_path('http_config.ru', __dir__)
+  end
+end
--- a/modules/externalipam/ip_cache.rb
+++ b/modules/externalipam/ip_cache.rb
@@ -1,0 +1,133 @@
+require 'yaml'
+require 'json'
+require 'monitor'
+require 'concurrent'
+require 'time'
+require 'externalipam/ipam_helper'
+require 'singleton'
+
+module Proxy::Ipam
+  # Class for managing temp in-memory cache to prevent same IP's being suggested in race conditions
+  class IpCache
+    include Singleton
+    include Proxy::Log
+    include Proxy::Ipam::IpamHelper
+
+    DEFAULT_CLEANUP_INTERVAL = 180
+
+    def initialize
+      @m = Monitor.new
+      @ip_cache = {'': {}}
+      start_cleanup_task
+    end
+
+    def provider_name(provider)
+      @provider = provider
+    end
+
+    def get_cidr(group_name, cidr)
+      @ip_cache.dig(group_name, cidr)
+    end
+
+    def get_ip(group_name, cidr, mac)
+      @ip_cache.dig(group_name, cidr, mac, :ip)
+    end
+
+    def ip_exists?(group_name, cidr, ip)
+      subnet_hash = get_cidr(group_name, cidr)
+      return false if subnet_hash.nil?
+      subnet_hash&.any? { |mac, cached_ip| cached_ip[:ip] == ip }
+    end
+
+    def ip_expired?(group_name, cidr, ip)
+      return true unless ip_exists?(group_name, cidr, ip)
+      subnet_hash = get_cidr(group_name, cidr)
+      subnet_hash&.any? { |mac, cached_ip| cached_ip[:ip] == ip && expired(cached_ip[:timestamp]) }
+    end
+
+    def cleanup_interval
+      DEFAULT_CLEANUP_INTERVAL
+    end
+
+    def add(group_name, cidr, ip, mac = nil)
+      logger.debug("Adding IP '#{ip}' to cache for subnet '#{cidr}' in group '#{group_name}' for IPAM provider #{@provider}")
+      @m.synchronize do
+        mac_addr = mac.nil? || mac.empty? ? SecureRandom.uuid : mac
+        group_hash = @ip_cache[group_name]
+
+        if group_hash&.key?(cidr)
+          @ip_cache[group_name][cidr][mac_addr] = { ip: ip.to_s, timestamp: Time.now }
+        else
+          @ip_cache[group_name] = { cidr => { mac_addr => { ip: ip.to_s, timestamp: Time.now }}}
+        end
+      end
+    end
+
+    private
+
+    def expired(ip_expiration)
+      Time.now - ip_expiration > DEFAULT_CLEANUP_INTERVAL
+    end
+
+    def start_cleanup_task
+      logger.info("Starting ip cache maintenance for External IPAM provider, used by /next_ip.")
+      @timer_task = Concurrent::TimerTask.new(execution_interval: DEFAULT_CLEANUP_INTERVAL) { clean_cache }
+      @timer_task.execute
+    end
+
+    # @ip_cache structure
+    #
+    # Groups of subnets are cached under the External IPAM Group name. For example,
+    # "IPAM Group Name" would be the section name in phpIPAM. All IP's cached for subnets
+    # that do not have an External IPAM group specified, they are cached under the "" key. IP's
+    # are cached using one of two possible keys:
+    #    1). Mac Address
+    #    2). UUID (Used when Mac Address not specified)
+    #
+    # {
+    #   "": {
+    #     "192.0.2.0/24":{
+    #       "00:0a:95:9d:68:10": {"ip": "192.0.2.1", "timestamp": "2019-09-17 12:03:43 -D400"},
+    #       "906d8bdc-dcc0-4b59-92cb-665935e21662": {"ip": "192.0.2.2", "timestamp": "2019-09-17 11:43:22 -D400"}
+    #     },
+    #   },
+    #   "IPAM Group Name": {
+    #     "198.51.100.0/24":{
+    #       "00:0a:95:9d:68:33": {"ip": "198.51.100.1", "timestamp": "2019-09-17 12:04:43 -0400"},
+    #       "00:0a:95:9d:68:34": {"ip": "198.51.100.2", "timestamp": "2019-09-17 12:05:48 -0400"},
+    #       "00:0a:95:9d:68:35": {"ip": "198.51.100.3", "timestamp:: "2019-09-17 12:06:50 -0400"}
+    #     }
+    #   },
+    #   "Another IPAM Group": {
+    #     "203.0.113.0/24":{
+    #       "00:0a:95:9d:68:55": {"ip": "203.0.113.1", "timestamp": "2019-09-17 12:04:43 -0400"},
+    #       "00:0a:95:9d:68:56": {"ip": "203.0.113.2", "timestamp": "2019-09-17 12:05:48 -0400"}
+    #     }
+    #   }
+    # }
+    def clean_cache
+      @m.synchronize do
+        entries_deleted = 0
+        total_entries = 0
+
+        @ip_cache.each do |group, subnets|
+          subnets.each do |cidr, macs|
+            macs.each do |mac, ip|
+              if expired(ip[:timestamp])
+                @ip_cache[group][cidr].delete(mac)
+                entries_deleted += 1
+              end
+              total_entries += 1
+            end
+            @ip_cache[group].delete(cidr) if @ip_cache[group][cidr].nil? || @ip_cache[group][cidr].empty?
+            @ip_cache.delete(group) if @ip_cache[group].nil? || @ip_cache[group].empty?
+          end
+        end
+
+        cache_count = total_entries - entries_deleted
+        logger.debug("Removing #{entries_deleted} entries from IP cache for IPAM provider #{@provider}") if entries_deleted > 0
+        logger.debug("Current count of IP cache entries for IPAM provider #{@provider}: #{cache_count}") if entries_deleted > 0
+      end
+    end
+  end
+end
--- a/modules/externalipam/ipam_api.rb
+++ b/modules/externalipam/ipam_api.rb
@@ -1,0 +1,358 @@
+require 'proxy/validations'
+require 'externalipam/ipam_helper'
+require 'externalipam/ipam_validator'
+require 'externalipam/dependency_injection'
+
+module Proxy::Ipam
+  # Generic API for External IPAM interactions
+  class Api < ::Sinatra::Base
+    extend Proxy::Ipam::DependencyInjection
+    include ::Proxy::Log
+    include ::Proxy::Validations
+    helpers ::Proxy::Helpers
+    include Proxy::Ipam::IpamHelper
+    include Proxy::Ipam::IpamValidator
+
+    inject_attr :externalipam_client, :client
+
+    # Gets the next available IP address based on a given External IPAM subnet
+    #
+    # Inputs:   1. address:          Network address of the subnet(e.g. 100.55.55.0)
+    #           2. prefix:           Network prefix(e.g. 24)
+    #           3. group(optional):  The External IPAM group
+    #
+    # Returns:
+    #   Response if success:
+    #   ======================
+    #     Http Code:     200
+    #     JSON Response:
+    #       "100.55.55.3"
+    #
+    #   Response if missing parameter(e.g. 'mac')
+    #   ======================
+    #     Http Code:     400
+    #     JSON Response:
+    #       {"error": "A 'mac' address must be provided(e.g. 00:0a:95:9d:68:10)"}
+    #
+    #   Response if no free ip's available
+    #   ======================
+    #     Http Code:     404
+    #     JSON Response:
+    #       {"error": "There are no free IP's in subnet 100.55.55.0/24"}
+    get '/subnet/:address/:prefix/next_ip' do
+      content_type :json
+
+      validate_required_params!([:address, :prefix], params)
+      mac_param = params[:mac]
+      mac = validate_mac(params[:mac]) unless mac_param.nil? || mac_param.empty?
+      cidr = validate_cidr(params[:address], params[:prefix])
+      group_name = get_request_group(params)
+      next_ip = provider.get_next_ip(mac, cidr, group_name)
+      halt 404, { error: ERRORS[:no_free_ips] }.to_json if next_ip.nil?
+      { data: next_ip }.to_json
+    rescue Proxy::Validations::Error => e
+      logger.exception(ERRORS[:proxy_validation_error], e)
+      halt 400, { error: e.to_s }.to_json
+    rescue RuntimeError => e
+      logger.exception(ERRORS[:runtime_error], e)
+      halt 500, { error: e.to_s }.to_json
+    rescue Errno::ECONNREFUSED, Errno::ECONNRESET
+      logger.exception(ERRORS[:no_connection], e)
+      halt 500, { error: ERRORS[:no_connection] }.to_json
+    end
+
+    # Gets the subnet from External IPAM
+    #
+    # Inputs:   1. address:          Network address of the subnet
+    #           2. prefix:           Network prefix(e.g. 24)
+    #           3. group(optional):  The name of the External IPAM group
+    #
+    # Returns:
+    #   Response if subnet exists:
+    #   ===========================
+    #     Http Code:     200
+    #     JSON Response:
+    #       {"data": {
+    #         "id": "33",
+    #         "subnet": "10.20.30.0",
+    #         "description": "Subnet description",
+    #         "mask": "29"}
+    #       }
+    #
+    #   Response if subnet does not exist:
+    #   ===========================
+    #     Http Code:     404
+    #     JSON Response:
+    #       {"error": "No subnets found"}
+    get '/subnet/:address/:prefix' do
+      content_type :json
+
+      validate_required_params!([:address, :prefix], params)
+      cidr = validate_cidr(params[:address], params[:prefix])
+      group_name = get_request_group(params)
+      subnet = provider.get_ipam_subnet(cidr, group_name)
+      halt 404, { error: ERRORS[:no_subnet] }.to_json if subnet.nil?
+      subnet.to_json
+    rescue Proxy::Validations::Error => e
+      logger.exception(ERRORS[:proxy_validation_error], e)
+      halt 400, { error: e.to_s }.to_json
+    rescue RuntimeError => e
+      logger.exception(ERRORS[:runtime_error], e)
+      halt 500, { error: e.to_s }.to_json
+    rescue Errno::ECONNREFUSED, Errno::ECONNRESET
+      logger.exception(ERRORS[:no_connection], e)
+      halt 500, { error: ERRORS[:no_connection] }.to_json
+    end
+
+    # Get a list of groups from External IPAM
+    #
+    # Returns:
+    #   Response if success:
+    #   ===========================
+    #     Http Code:     200
+    #     JSON Response:
+    #       {"data": [
+    #         {"id": "1", "name": "Group 1", "description": "This is group 1"},
+    #         {"id": "2", "name": "Group 2", "description": "This is group 2"}
+    #       ]}
+    #
+    #   Response if no groups exist:
+    #   ===========================
+    #     Http Code:     404
+    #     JSON Response:
+    #       {"error": "Groups are not supported"}
+    #
+    #   Response if groups are not supported:
+    #   ===========================
+    #     Http Code:     422
+    #     JSON Response:
+    #       {"error": "Groups are not supported"}
+    get '/groups' do
+      content_type :json
+
+      halt 422, { error: ERRORS[:groups_not_supported] }.to_json unless provider.groups_supported?
+      groups = provider.ipam_groups
+      groups.to_json
+    rescue Proxy::Validations::Error => e
+      logger.exception(ERRORS[:proxy_validation_error], e)
+      halt 400, { error: e.to_s }.to_json
+    rescue RuntimeError => e
+      logger.exception(ERRORS[:runtime_error], e)
+      halt 500, { error: e.to_s }.to_json
+    rescue Errno::ECONNREFUSED, Errno::ECONNRESET
+      logger.exception(ERRORS[:no_connection], e)
+      halt 500, { error: ERRORS[:no_connection] }.to_json
+    end
+
+    # Get a group from External IPAM
+    #
+    # Inputs:   1. group:       The name of the External IPAM group
+    #
+    # Returns:
+    #   Response if success:
+    #   ===========================
+    #     Http Code:     200
+    #     JSON Response:
+    #       {"data": {"id": "1", "name": "Group 1", "description": "This is group 1"}}
+    #
+    #   Response if group not found:
+    #   ===========================
+    #     Http Code:     404
+    #     JSON Response:
+    #       {"error": "Group not Found"}
+    #
+    #   Response if groups are not supported:
+    #   ===========================
+    #     Http Code:     500
+    #     JSON Response:
+    #       {"error": "Groups are not supported"}
+    get '/groups/:group' do
+      content_type :json
+
+      validate_required_params!([:group], params)
+      group_name = get_request_group(params)
+      group = provider.get_ipam_group(group_name)
+      halt 404, { error: ERRORS[:no_group] }.to_json if group.nil?
+      group.to_json
+    rescue Proxy::Validations::Error => e
+      logger.exception(ERRORS[:proxy_validation_error], e)
+      halt 400, { error: e.to_s }.to_json
+    rescue RuntimeError => e
+      logger.exception(ERRORS[:runtime_error], e)
+      halt 500, { error: e.to_s }.to_json
+    rescue Errno::ECONNREFUSED, Errno::ECONNRESET
+      logger.exception(ERRORS[:no_connection], e)
+      halt 500, { error: ERRORS[:no_connection] }.to_json
+    end
+
+    # Get a list of subnets for a given External IPAM group
+    #
+    # Input:   1. group:         The name of the External IPAM group
+    #
+    # Returns:
+    #   Response if success:
+    #   ===========================
+    #     Http Code:     200
+    #     JSON Response:
+    #       {"data":[
+    #         {"subnet":"10.20.30.0","mask":"29","description":"This is a subnet"},
+    #         {"subnet":"40.50.60.0","mask":"29","description":"This is another subnet"}
+    #       ]}
+    #
+    #   Response if no subnets exist in group:
+    #   ===========================
+    #     Http Code:     404
+    #     JSON Response:
+    #       {"error": "No subnets found in External IPAM group"}
+    #
+    #   Response if groups are not supported:
+    #   ===========================
+    #     Http Code:     500
+    #     JSON Response:
+    #       {"error": "Groups are not supported"}
+    get '/groups/:group/subnets' do
+      content_type :json
+
+      validate_required_params!([:group], params)
+      group_name = get_request_group(params)
+      subnets = provider.get_ipam_subnets(group_name)
+      halt 404, { error: ERRORS[:no_subnets_in_group] }.to_json if subnets == []
+      subnets.to_json
+    rescue Proxy::Validations::Error => e
+      logger.exception(ERRORS[:proxy_validation_error], e)
+      halt 400, { error: e.to_s }.to_json
+    rescue RuntimeError => e
+      logger.exception(ERRORS[:runtime_error], e)
+      halt 500, { error: e.to_s }.to_json
+    rescue Errno::ECONNREFUSED, Errno::ECONNRESET
+      logger.exception(ERRORS[:no_connection], e)
+      halt 500, { error: ERRORS[:no_connection] }.to_json
+    end
+
+    # Checks whether an IP address has already been taken in External IPAM
+    #
+    # Inputs: 1. address:         The network address of the IPv4 or IPv6 subnet.
+    #         2. prefix:          The subnet prefix(e.g. 24)
+    #         3. ip:              IP address to be queried
+    #         4. group(optional): The name of the External IPAM Group, containing the subnet to check
+    #
+    # Returns:
+    #   Response if exists:
+    #   ===========================
+    #     Http Code:  200
+    #     Response:   true
+    #
+    #   Response if not exists:
+    #   ===========================
+    #     Http Code:      200
+    #     JSON Response:  false
+    get '/subnet/:address/:prefix/:ip' do
+      content_type :json
+
+      validate_required_params!([:address, :prefix, :ip], params)
+      ip = validate_ip(params[:ip])
+      cidr = validate_cidr(params[:address], params[:prefix])
+      group_name = get_request_group(params)
+      subnet = provider.get_ipam_subnet(cidr, group_name)
+      halt 404, { error: ERRORS[:no_subnet] }.to_json if subnet.nil?
+      validate_ip_in_cidr(ip, cidr)
+      ip_exists = provider.ip_exists?(ip, subnet[:id], group_name)
+      halt 200, { result: ip_exists }.to_json
+    rescue Proxy::Validations::Error => e
+      logger.exception(ERRORS[:proxy_validation_error], e)
+      halt 400, { error: e.to_s }.to_json
+    rescue RuntimeError => e
+      logger.exception(ERRORS[:runtime_error], e)
+      halt 500, { error: e.to_s }.to_json
+    rescue Errno::ECONNREFUSED, Errno::ECONNRESET
+      logger.exception(ERRORS[:no_connection], e)
+      halt 500, { error: ERRORS[:no_connection] }.to_json
+    end
+
+    # Adds an IP address to the specified subnet for the specified IPAM provider
+    #
+    # Params: 1. address:         The network address of the IPv4 or IPv6 subnet
+    #         2. prefix:          The subnet prefix(e.g. 24)
+    #         3. ip:              IP address to be added
+    #         4. group(optional): The name of the External IPAM Group, containing the subnet to add ip to
+    #
+    # Returns:
+    #   Response if added successfully:
+    #   ===========================
+    #     Http Code:  201
+    #     Response:   Empty
+    #
+    #   Response if not added successfully:
+    #   ===========================
+    #     Http Code:  500
+    #     JSON Response:
+    #       {"error": "Unable to add IP to External IPAM"}
+    post '/subnet/:address/:prefix/:ip' do
+      content_type :json
+
+      validate_required_params!([:address, :ip, :prefix], params)
+      ip = validate_ip(params[:ip])
+      cidr = validate_cidr(params[:address], params[:prefix])
+      group_name = get_request_group(params)
+      subnet = provider.get_ipam_subnet(cidr, group_name)
+      halt 404, { error: ERRORS[:no_subnet] }.to_json if subnet.nil?
+      add_ip_params = { cidr: cidr, subnet_id: subnet[:id], group_name: group_name }
+      validate_ip_in_cidr(ip, cidr)
+      ip_added = provider.add_ip_to_subnet(ip, add_ip_params) # Returns nil on success
+      halt 500, ip_added.to_json unless ip_added.nil?
+      status 201
+    rescue Proxy::Validations::Error => e
+      logger.exception(ERRORS[:proxy_validation_error], e)
+      halt 400, { error: e.to_s }.to_json
+    rescue RuntimeError => e
+      logger.exception(ERRORS[:runtime_error], e)
+      halt 500, { error: e.to_s }.to_json
+    rescue Errno::ECONNREFUSED, Errno::ECONNRESET
+      logger.exception(ERRORS[:no_connection], e)
+      halt 500, { error: ERRORS[:no_connection] }.to_json
+    end
+
+    # Deletes IP address from a given subnet
+    #
+    # Params: 1. address:         The network address of the IPv4 or IPv6 subnet
+    #         2. prefix:          The subnet prefix(e.g. 24)
+    #         3. ip:              IP address to be deleted
+    #         4. group(optional): The name of the External IPAM Group, containing the subnet to delete ip from
+    #
+    # Returns:
+    #   Response if deleted successfully:
+    #   ===========================
+    #     Http Code:  200
+    #     Response:   Empty
+    #
+    #   Response if not added successfully:
+    #   ===========================
+    #     Http Code:  500
+    #     JSON Response:
+    #       {"error": "Unable to delete IP from External IPAM"}
+    delete '/subnet/:address/:prefix/:ip' do
+      content_type :json
+
+      validate_required_params!([:address, :ip, :prefix], params)
+      ip = validate_ip(params[:ip])
+      cidr = validate_cidr(params[:address], params[:prefix])
+      group_name = get_request_group(params)
+      subnet = provider.get_ipam_subnet(cidr, group_name)
+      halt 404, { error: ERRORS[:no_subnet] }.to_json if subnet.nil?
+      del_ip_params = { cidr: cidr, subnet_id: subnet[:id], group_name: group_name }
+      validate_ip_in_cidr(ip, cidr)
+      ip_deleted = provider.delete_ip_from_subnet(ip, del_ip_params) # Returns nil on success
+      halt 500, ip_deleted.to_json unless ip_deleted.nil?
+      halt 204
+    rescue Proxy::Validations::Error => e
+      logger.exception(ERRORS[:proxy_validation_error], e)
+      halt 400, { error: e.to_s }.to_json
+    rescue RuntimeError => e
+      logger.exception(ERRORS[:runtime_error], e)
+      halt 500, { error: e.to_s }.to_json
+    rescue Errno::ECONNREFUSED, Errno::ECONNRESET
+      logger.exception(ERRORS[:no_connection], e)
+      halt 500, { error: ERRORS[:no_connection] }.to_json
+    end
+  end
+end
--- a/modules/externalipam/ipam_helper.rb
+++ b/modules/externalipam/ipam_helper.rb
@@ -1,0 +1,113 @@
+require 'proxy/validations'
+require 'externalipam/externalipam'
+
+# Module containing helper methods for use by all External IPAM provider implementations
+module Proxy::Ipam::IpamHelper
+  include ::Proxy::Validations
+
+  MAX_IP_RETRIES = 10
+  ERRORS = {
+    cidr: "A 'cidr' parameter for the subnet must be provided(e.g. IPv4: 100.10.10.0/24, IPv6: 2001:db8:abcd:12::/124)",
+    mac: "A 'mac' address must be provided(e.g. 00:0a:95:9d:68:10)",
+    ip: "Missing 'ip' parameter. An IPv4 or IPv6 address must be provided(e.g. IPv4: 100.10.10.22, IPv6: 2001:db8:abcd:12::3)",
+    group_name: "A 'group_name' must be provided",
+    no_ip: 'IP address not found',
+    no_free_ips: 'No free addresses found',
+    no_connection: 'Unable to connect to External IPAM server',
+    no_group: 'Group not found in External IPAM',
+    no_groups: 'No groups found in External IPAM',
+    no_subnet: 'Subnet not found in External IPAM',
+    no_subnets_in_group: 'No subnets found in External IPAM group',
+    provider: "The IPAM provider must be specified(e.g. 'externalipam_phpipam' or 'externalipam_netbox')",
+    groups_not_supported: 'Groups are not supported',
+    add_ip: 'Error adding IP to External IPAM',
+    bad_mac: 'Mac address is invalid',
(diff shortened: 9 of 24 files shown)


config/settings.d/externalipam.yml.example
```diff
@@ -0,0 +1,8 @@
+---
+:enabled: false
+
+# Supported providers:
+#   1. phpIPAM: externalipam_phpipam
+#   2. Netbox:  externalipam_netbox
+
+:use_provider: externalipam_netbox
```
lzap, line 8: Note for myself / @ekohl - I am fine with bringing this feature into core, after all External IPAM is already in Foreman core, however, I would suggest to ship providers as RPM/DEB subpackages so products owners can decide what to ship and what not. Opinion?

config/settings.d/externalipam_netbox.yml.example
```diff
@@ -0,0 +1,3 @@
+---
+:url: 'http://my-netbox-url.com'
+:token: 'netbox_token'
```
lzap: Missing newline, also is the URL really in the format of hostname without any path? Configuration files often do not explain this properly, a full path would be good to have if it is actually required. The same for phpIPAM.

modules/externalipam/externalipam_plugin.rb
```diff
@@ -0,0 +1,11 @@
+require 'externalipam/phpipam/phpipam_plugin'
+require 'externalipam/netbox/netbox_plugin'
+
+module Proxy::Ipam
+  class Plugin < ::Proxy::Plugin
+    plugin :externalipam, ::Proxy::VERSION
+    uses_provider
+    default_settings use_provider: 'externalipam_phpipam'
```
lzap: Default setting should be probably nil - error out during startup, the proxy stack will automatically disable the plugin. We have a validation framework for that, just look on how other modules are implemented.

modules/externalipam/ip_cache.rb
```diff
@@ -0,0 +1,133 @@
...
+    #     "185.45.39.0/24":{
+    #       "00:0a:95:9d:68:55": {"ip": "185.45.39.1", "timestamp": "2019-09-17 12:04:43 -0400"},
+    #       "00:0a:95:9d:68:56": {"ip": "185.45.39.2", "timestamp": "2019-09-17 12:05:48 -0400"}
+    #     }
+    #   }
+    # }
+    def init_cache
+      @m.synchronize do
+        if @ip_cache && !@ip_cache.empty?
+          logger.debug("Processing ip cache for IPAM provider #{@provider}")
```
lzap: This method is called every minute, it will flood logs, can you use `logger.trace` which is a special level we added for these noisy messages? Please update `clean_up_allocated_ips` in the same way, this is driving me crazy let's fix this finally.

lzap: Also if you could revert this log message, instead of "starting something" just add a message at the end of the process in the form of "finished something" and please show counts of:

* how many entries removed
* how many entries are still in the cache

modules/externalipam/ipam_api.rb
```diff
@@ -0,0 +1,389 @@
...
+        validate_required_params!([:address, :prefix], params)
+
+        mac_param = params[:mac]
+        mac = validate_mac!(params[:mac]) unless mac_param.nil? || mac_param.empty?
+        cidr = validate_cidr!(params[:address], params[:prefix])
+        group_name = get_request_group(params)
+
+        next_ip = provider.get_next_ip(mac, cidr, group_name)
+        halt 404, { error: ERRORS[:no_free_ips] }.to_json if next_ip.nil?
+        next_ip.to_json
```
lzap: One nasty thing I missed during my initial review - it is not a good practice to return any other values than hash in JSON. It is not extensible, if you want to add another element with this design (when it is actually a JSON-encoded string) you cannot do that easily. I would suggest something like this:

```suggestion
        {ip: next_ip}.to_json
```

A change in Foreman core will be required to fix this.

modules/externalipam/ipam_api.rb
```diff
@@ -0,0 +1,389 @@
...
+
+        ip = validate_ip!(params[:ip])
+        cidr = validate_cidr!(params[:address], params[:prefix])
+        group_name = get_request_group(params)
+        subnet = provider.get_ipam_subnet(cidr, group_name)
+
+        halt 404, { error: ERRORS[:no_subnet] }.to_json if subnet.nil?
+        validate_ip_in_cidr!(ip, cidr)
+        ip_exists = provider.ip_exists?(ip, subnet[:id], group_name)
+        halt 200, ip_exists.to_json
```
lzap: Ditto:

```suggestion
        halt 200, {result: ip_exists}.to_json
```

lzap: This is a typical problem - later on we would like to extend this with an error message explaining maybe a reason why the IP was taken (where it belongs to etc).

modules/externalipam/phpipam/phpipam_client.rb
```diff
@@ -0,0 +1,188 @@
...
+        data.push(
+          {
+            id: group['id'],
+            name: group['name'],
+            description: group['description'],
+          }
+        )
+      end
+
+      return data if json_body['data']
```
ekohl: You already have a guard earlier so the conditional for the return is redundant. You can also map it to transform it,removing the need for the additional variable.
```suggestion
      json_body['data'].map do |group|
        {
          id: group['id'],
          name: group['name'],
          description: group['description'],
        }
      end
```

Note that this is a repeated pattern in various of these methods.

modules/externalipam/api_resource.rb
```diff
@@ -0,0 +1,54 @@
...
+
+    def get(path)
+      uri = URI(@api_base + path)
+      request = Net::HTTP::Get.new(uri)
+      request[@auth_header] = @token
+      request['Accept'] = 'application/json'
+
+      Net::HTTP.start(uri.hostname, uri.port, use_ssl: uri.scheme == 'https') do |http|
+        http.request(request)
+      end
```
ekohl: This part is repeated for every request. I'd recommend a private helper like `def request(request)`

modules/externalipam/ip_cache.rb
```diff
@@ -0,0 +1,133 @@
...
+    def get_group(group)
+      @ip_cache[group.to_sym]
+    end
+
+    def get_cidr(group, cidr)
+      @ip_cache[group.to_sym][cidr.to_sym]
+    end
+
+    def get_ip(group_name, cidr, mac)
+      @ip_cache[group_name.to_sym][cidr.to_sym][mac.to_sym][:ip]
```
ekohl: Wouldn't this be safer?
```suggestion
      @ip_cache.dig(group_name.to_sym, cidr.to_sym, mac.to_sym, :ip)
```

If for some reason the group (or cidr or mac) is not found, your code will result in `nil[]` which raises an exception where `dig()` will return nil.

modules/externalipam/ip_cache.rb
```diff
@@ -0,0 +1,133 @@
...
+    end
+
+    def cleanup_interval
+      DEFAULT_CLEANUP_INTERVAL
+    end
+
+    def ip_exists(ip, cidr, group_name)
+      cidr_key = @ip_cache[group_name.to_sym][cidr.to_sym]&.to_s
```
ekohl: Why `&.to_s`? Would it be better to use [IPAddr](https://ruby-doc.org/stdlib-2.5.9/libdoc/ipaddr/rdoc/IPAddr.html) instance and check if the IP is present in the subnet?

modules/externalipam/ip_cache.rb
```diff
@@ -0,0 +1,133 @@
...
+      cidr_key.include?(ip.to_s)
+    end
+
+    def add(ip, mac, cidr, group_name)
+      logger.debug("Adding IP '#{ip}' to cache for subnet '#{cidr}' in group '#{group_name}' for IPAM provider #{@provider}")
+      @m.synchronize do
+        mac_addr = mac.nil? || mac.empty? ? SecureRandom.uuid : mac
```
ekohl, line 55: if `mac` is an optional parameter, I'd expect the method to be
```ruby
def add(ip, cidr, group_name, mac = nil)
```

Something else that bugs me is the wildly inconsistent ordering of parameters. `get_ip`, `add`, `ip_exists` should be symmetrical. So I'd expect something like
```ruby
def get_ip(group_name, cidr, mac)
def ip_exist(group_name, cidr, ip)
def add(group_name, cidr, ip, mac = nil)
```

That also makes me wonder: what does `get_ip` do if `mac` is `nil`. The `SecureRandom` ensures that the entry can never be retrieved anymore.

modules/externalipam/ip_cache.rb
```diff
@@ -0,0 +1,133 @@
...
+          @ip_cache.each do |group, subnets|
+            subnets.each do |cidr, macs|
+              macs.each do |mac, ip|
+                if Time.now - Time.parse(ip[:timestamp]) > DEFAULT_CLEANUP_INTERVAL
+                  @ip_cache[group][cidr].delete(mac)
+                end
+              end
+              @ip_cache[group].delete(cidr) if @ip_cache[group][cidr].nil? || @ip_cache[group][cidr].empty?
+            end
+          end
```
ekohl: I think this is equivalent:

```suggestion
          @ip_cache.each do |group, subnets|
            subnets.each do |cidr, macs|
              macs.delete_if do |mac, ip|
                Time.now - Time.parse(ip[:timestamp]) > DEFAULT_CLEANUP_INTERVAL
              end
            end
            subnets.delete_if { |cidr, macs| macs.empty? }
          end
          @ip_cache.delete_if { |group, subnets| subnets.empty? }
```

That said, you are parsing all timestamps every time you go through this. Why don't you parse it when adding it? That's a lot more efficient.

modules/externalipam/ip_cache.rb
```diff
@@ -0,0 +1,133 @@
...
+    #     }
+    #   },
+    #   "Another IPAM Group": {
+    #     "185.45.39.0/24":{
+    #       "00:0a:95:9d:68:55": {"ip": "185.45.39.1", "timestamp": "2019-09-17 12:04:43 -0400"},
+    #       "00:0a:95:9d:68:56": {"ip": "185.45.39.2", "timestamp": "2019-09-17 12:05:48 -0400"}
+    #     }
+    #   }
+    # }
+    def init_cache
```
ekohl: I don't like that this is both pruning the cache and initializing. The naming is confusing.

modules/externalipam/ip_cache.rb
```diff
@@ -0,0 +1,133 @@
...
+                if Time.now - Time.parse(ip[:timestamp]) > DEFAULT_CLEANUP_INTERVAL
+                  @ip_cache[group][cidr].delete(mac)
+                end
+              end
+              @ip_cache[group].delete(cidr) if @ip_cache[group][cidr].nil? || @ip_cache[group][cidr].empty?
+            end
+          end
+        else
+          logger.debug("Clearing ip cache for IPAM provider #{@provider}")
+          @ip_cache = {'': {}}
```
ekohl: Won't this entry be removed the next time `init_cache` runs? So it'd be:
```suggestion
          @ip_cache = {}
```

modules/externalipam/ip_cache.rb
```diff
@@ -0,0 +1,133 @@
...
+          if values.key? mac_addr.to_sym
+            @ip_cache[group_name.to_sym][key].delete(mac_addr.to_sym)
+          end
+          @ip_cache[group_name.to_sym].delete(key) if @ip_cache[group_name.to_sym][key].nil? || @ip_cache[group_name.to_sym][key].empty?
+        end
+
+        if group_hash.key?(cidr.to_sym)
+          @ip_cache[group_name.to_sym][cidr.to_sym][mac_addr.to_sym] = {ip: ip.to_s, timestamp: Time.now.to_s}
+        else
+          @ip_cache = @ip_cache.merge({group_name.to_sym => {cidr.to_sym => {mac_addr.to_sym => {ip: ip.to_s, timestamp: Time.now.to_s}}}})
```
ekohl: Why do you do this? This means you have a new instance for `@ip_cache` all the time (because `merge()` returns a new instance). Also, if you don't use `.to_s` you don't need here you don't need to parse it all the time.

modules/externalipam/ip_cache.rb
```diff
@@ -0,0 +1,133 @@
...
+      @m.synchronize do
+        mac_addr = mac.nil? || mac.empty? ? SecureRandom.uuid : mac
+        group_hash = @ip_cache[group_name.to_sym]
+
+        group_hash.each do |key, values|
+          if values.key? mac_addr.to_sym
+            @ip_cache[group_name.to_sym][key].delete(mac_addr.to_sym)
+          end
+          @ip_cache[group_name.to_sym].delete(key) if @ip_cache[group_name.to_sym][key].nil? || @ip_cache[group_name.to_sym][key].empty?
+        end
```
ekohl: Why do you have deletion code in `add()`? You already have a regular task to prune the cache.

modules/externalipam/ipam_api.rb
```diff
@@ -0,0 +1,389 @@
...
+        logger.warn(e.message)
+        halt 400, { error: e.to_s }.to_json
+      rescue RuntimeError => e
+        logger.warn(e.message)
+        halt 500, { error: e.message }.to_json
+      rescue Errno::ECONNREFUSED, Errno::ECONNRESET
+        logger.warn(ERRORS[:no_connection])
+        halt 500, { error: ERRORS[:no_connection] }.to_json
+      end
+    end
```
ekohl: Now that we're on Ruby 2.5 we can use `rescue` at the method level.
```suggestion
      validate_required_params!([:group], params)

      group_name = get_request_group(params)
      group = provider.get_ipam_group(group_name)

      halt 404, { error: ERRORS[:no_group] }.to_json if group.nil?
      group.to_json
    rescue Proxy::Validations::Error => e
      logger.warn(e.message)
      halt 400, { error: e.to_s }.to_json
    rescue RuntimeError => e
      logger.warn(e.message)
      halt 500, { error: e.message }.to_json
    rescue Errno::ECONNREFUSED, Errno::ECONNRESET
      logger.warn(ERRORS[:no_connection])
      halt 500, { error: ERRORS[:no_connection] }.to_json
    end
  end
```

modules/externalipam/ipam_helper.rb
```diff
@@ -0,0 +1,113 @@
...
+    return ip if found_ip.nil?
+
+    found_ip
+  end
+
+  # Checks the cache for existing ip, and returns it if it exists. If not exists, it will
+  # find a new ip (using find_new_ip), and it is added to the cache.
+  def cache_next_ip(ip_cache, ip, mac, cidr, subnet_id, group_name)
+    group = group_name.nil? ? '' : group_name
+    ip_cache.set_group(group, {}) if ip_cache.get_group(group).nil?
```
ekohl: This should not be needed. The `ip_cache` should just handle that for us (for example, by using `.dig()` internally.

modules/externalipam/ipam_helper.rb
```diff
@@ -0,0 +1,113 @@
...
+    subnet_hash = ip_cache.get_cidr(group, cidr)
+    next_ip = nil
+
+    if mac && subnet_hash&.key?(mac.to_sym)
+      next_ip = ip_cache.get_ip(group, cidr, mac)
+    else
+      new_ip = ip
+      ip_not_in_cache = subnet_hash.nil? ? true : !subnet_hash.to_s.include?(new_ip.to_s)
```
ekohl: You can simplify the logic here:
```suggestion
      ip_not_in_cache = subnet_hash.nil? || !subnet_hash.to_s.include?(new_ip.to_s)
```

However, I would stay away from negatively named variables. So stick to `ip_in_cache`. I'd also check for existence without relying on `.to_s`. That's just unreliable.
```suggestion
      ip_in_cache = subnet_hash&.any? { |_mac, cached_ip| cached_ip[:ip] == new_ip }
```

Whether it's in the cache should really be part of the `ip_cache` implementation though and not in this helper.

Technically it should also check if the cache hasn't expired, but that **really** belongs in the cache implementation.

modules/externalipam/ipam_helper.rb
```diff
@@ -0,0 +1,113 @@
...
+    next_ip
+  end
+
+  def increment_ip(ip)
+    IPAddr.new(ip.to_s).succ.to_s
+  end
+
+  def usable_ip(ip, cidr)
+    network = IPAddr.new(cidr)
+    network.include?(IPAddr.new(ip)) && network.to_range.last != ip
```
ekohl: I suppose `.to_range.last` is to check the broadcast address. Technically that's IPv4 only and not relevant for IPv6. For IPv4 you would also need to check the network address.

modules/externalipam/netbox/netbox_client.rb
```diff
@@ -0,0 +1,201 @@
...
+    def get_ipam_subnets(group_name)
+      if group_name.nil?
+        params = URI.encode_www_form({ status: 'active' })
+      else
+        group_id = get_group_id(group_name)
+        params = URI.encode_www_form({ status: 'active', vrf_id: group_id })
+      end
+
+      response = @api_resource.get("ipam/prefixes/?#{params}")
+      json_body = JSON.parse(response.body)
```
ekohl: Shouldn't you check if there was a valid response code before attempting to parse the body? If it returned a 5xx code it probably isn't valid.

modules/externalipam/netbox/netbox_client.rb
```diff
@@ -0,0 +1,201 @@
...
+      end
+
+      return subnets if json_body['results']
+    end
+
+    def ip_exists?(ip, subnet_id, group_name)
+      group_id = get_group_id(group_name)
+      url = "ipam/ip-addresses/?#{URI.encode_www_form({ address: ip })}"
+      url += "&#{URI.encode_www_form({ prefix_id: subnet_id })}" unless subnet_id.nil?
+      url += "&#{URI.encode_www_form({ vrf_id: group_id })}" unless group_id.nil?
```
ekohl: What I would write is:
```suggestion
      params = {address: ip}
      params[:prefix_id] = subnet_id unless subnet_id.nil?
      params[:vrf_id] = group_id unless group_id.nil?
      url = "ipam/ip-addresses/?#{URI.encode_www_form(params)}"
```

modules/externalipam/netbox/netbox_client.rb
```diff
@@ -0,0 +1,201 @@
...
+      response = @api_resource.get(url)
+      json_body = JSON.parse(response.body)
+      return false if json_body['count'].zero?
+      true
```
ekohl:
```suggestion
      !json_body['count'].zero?
```

modules/externalipam/netbox/netbox_client.rb
```diff
@@ -0,0 +1,201 @@
...
+          }
+        )
+      end
+
+      groups
+    end
+
+    def get_ipam_group(group_name)
+      raise ERRORS[:groups_not_supported] unless groups_supported?
+      response = @api_resource.get("ipam/vrfs/?name=#{group_name}")
```
ekohl, line 59: I would escape `group_name` in case there's a weird character.

modules/externalipam/netbox/netbox_client.rb
```diff
@@ -0,0 +1,201 @@
...
+
+    def get_ipam_subnets(group_name)
+      if group_name.nil?
+        params = URI.encode_www_form({ status: 'active' })
+      else
+        group_id = get_group_id(group_name)
+        params = URI.encode_www_form({ status: 'active', vrf_id: group_id })
+      end
+
+      response = @api_resource.get("ipam/prefixes/?#{params}")
```
ekohl: I would first make a hash of params and then later use that
```suggestion
      params = {status: 'active}
      params[:vrf_id] = get_group_id(group_name) if group_name

       response = @api_resource.get("ipam/prefixes/?#{URI.encode_www_form(params)}") 
```

modules/externalipam/api_resource.rb
```diff
@@ -0,0 +1,54 @@
...
+    include Proxy::Ipam::IpamHelper
+
+    def initialize(params = {})
+      @api_base = params[:api_base]
+      @token = params[:token]
+      @auth_header = params[:auth_header] || 'Authorization'
+    end
+
+    def get(path)
+      uri = URI(@api_base + path)
```
ekohl: Given the repetition in how often you call `URI.encode_www_form()` with this method, should it be solved here?
```suggestion
    def get(path, params = nil)
      url = @api_base + path
      url += "?#{URI.encode_www_form(params)}" if params
      uri = URI(url)
```

modules/externalipam/netbox/netbox_client.rb
```diff
@@ -0,0 +1,201 @@
+require 'yaml'
+require 'json'
+require 'net/http'
+require 'uri'
+require 'sinatra'
```
ekohl: Why is this here?

modules/externalipam/netbox/netbox_client.rb
```diff
@@ -0,0 +1,201 @@
...
+      end
+    end
+
+    def get_ipam_subnet_by_group(cidr, group_id)
+      params = URI.encode_www_form({ status: 'active', prefix: cidr, vrf_id: group_id })
+      response = @api_resource.get("ipam/prefixes/?#{params}")
+      json_body = JSON.parse(response.body)
+      return nil if json_body['count'].zero?
+      subnet = subnet_from_result(json_body['results'][0])
+      return subnet if json_body['results']
```
ekohl: if `json_body['results']` was `nil`, the previously line would already have raised an exception
```suggestion
      subnet_from_result(json_body['results'][0])
```

This is a repeated pattern so I'm not commenting it on every instance.

ekohl (changes requested):
I went through it, but there's a bunch of repeated patterns. It may feel a bit like an inconsistent review and I think we may have a long back and forth on this for a while. That said, I am in favor of getting this into the Smart Proxy so please don't take it as being against it. Quite the opposite: Foreman itself has the functionality in core so the Smart Proxy should too.

config/settings.d/externalipam.yml.example
ekohl: It does not introduce additional dependencies and the Smart Proxy is already quite smart in not loading the module if it's inactive so I'd just ship it in the same RPM/deb. Subpackages only make sense to reduce additional dependencies or reducing runtime overhead at the cost of more complex installation. In this case I think it's not worth it but I appreciate the consideration.

modules/externalipam/ipam_validator.rb
```diff
@@ -0,0 +1,49 @@
...
+    end
+  end
+
+  def validate_mac!(mac)
+    unless /^([0-9a-fA-F]{2}[:]){5}[0-9a-fA-F]{2}$/i =~ mac
+      raise Proxy::Validations::Error.new, ERRORS[:bad_mac]
+    end
+    mac
+  end
+end
```
lzap, line 36: We already have `validate_mac` and ip in the codebase, would you mind extracting the current usage into some helper and then reuse that helper here too? No CIDR validations yet.

config/settings.d/externalipam.yml.example
lzap: Okay let me rephrase it this way - I do not want to support this in Red Hat Satellite at this moment. Perhaps we can disable it via the installer scenario?

ekohl: Currently we don't have good facilities for that. It would need some thought.

modules/externalipam/ipam_api.rb
grizzthedj: Will be submitting a PR on the foreman side as well

modules/externalipam/ip_cache.rb
grizzthedj: We are checking if the IP is in the cache, not in the subnet

modules/externalipam/netbox/netbox_client.rb
grizzthedj: Still working on this

modules/externalipam/netbox/netbox_client.rb
grizzthedj: If I escape, it returns group not found. Leaving as is for now, but will look into further

modules/externalipam/ipam_validator.rb
grizzthedj: Have not done this yet, but will in the next push

grizzthedj:
Pushed most of these changes, and still working through the last few. I will squash the commits once everything is in alignment

modules/externalipam/netbox/netbox_client.rb
grizzthedj: `group_name` is already escaped upstream in the `get_request_group` method

config/settings.d/externalipam.yml.example
lzap: Ok this is an internal discussion, I do not want block this effort.

lzap:
I have no further comments, @ekohl ?

config/settings.d/externalipam.yml.example
```diff
@@ -0,0 +1,8 @@
+---
+:enabled: false
+
+# Supported providers:
```
ekohl: Technically any gem can add supported providers. Perhaps we should rephrase this to
```suggestion
# Built-in providers:
```

config/settings.d/externalipam_netbox.yml.example
```diff
@@ -0,0 +1,6 @@
+---
+# url is the hostname and path of the Netbox instance
+:url: 'http://netbox-example.com'
```
ekohl: Slight nit, but I'd use the `example.com` domain and recommend HTTPS.
```suggestion
:url: 'https://netbox.example.com'
```

config/settings.d/externalipam_phpipam.yml.example
```diff
@@ -0,0 +1,12 @@
+---
+# url is the hostname and path of the phpIPAM instance.
+:url: 'http://phpipam-example.com'
```
ekohl:
```suggestion
:url: 'https://phpipam.example.com'
```

modules/dhcp_common/free_ips.rb
```diff
@@ -29,11 +29,11 @@ def mark_ip_as_allocated(ip_address)
     def clean_up_allocated_ips
       @m.synchronize do
         loop do
-          logger.debug("Starting allocated ip addresses cleanup pass...")
+          logger.trace("Starting allocated ip addresses cleanup pass for DHCP ...")
```
ekohl: This looks like an unrelated change. It'd be better to separate that out to a different PR.

modules/externalipam/ipam_validator.rb
```diff
@@ -0,0 +1,49 @@
...
+    err = []
+    required_params.each do |param|
+      unless params[param.to_sym]
+        err.push errors[param.to_sym]
+      end
+    end
+    raise Proxy::Validations::Error, err unless err.empty?
+  end
+
+  def validate_ip!(ip)
```
ekohl: We already have one:
https://github.com/theforeman/smart-proxy/blob/6b75029e01767cad7873a5c58971fbbde5aabb65/lib/proxy/validations.rb#L33-L41

modules/externalipam/ipam_validator.rb
```diff
@@ -0,0 +1,49 @@
...
+    good_ip = ip =~ Regexp.union([Resolv::IPv4::Regex, Resolv::IPv6::Regex])
+    raise Proxy::Validations::Error, ERRORS[:bad_ip] if good_ip.nil?
+    ip
+  end
+
+  def validate_cidr!(address, prefix)
+    cidr = "#{address}/#{prefix}"
+    network = IPAddr.new(cidr).to_s
+    if IPAddr.new(cidr).to_s != IPAddr.new(address).to_s
```
ekohl: You can reuse the `network` variable:
```suggestion
    if network != IPAddr.new(address).to_s
```

modules/externalipam/ipam_helper.rb
```diff
@@ -0,0 +1,115 @@
...
+  # not actually persisted in External IPAM yet. This method will increment the IP, up to
+  # MAX_IP_RETRIES times, and check if it is available in External IPAM each iteration. It
+  # will return the original IP(the 'ip' param) if no new IP's are found after MAX_IP_RETRIES
+  # iterations.
+  def find_new_ip(ip_cache, subnet_id, ip, mac, cidr, group_name)
+    found_ip = nil
+    temp_ip = ip
+    retry_count = 0
+
+    loop do
```
ekohl: I'm not a big fan of `loop`, especially if there's a condition. Can't this be explicit?
```suggestion
    while retry_count < MAX_IP_RETRIES
```

modules/externalipam/ipam_helper.rb
```diff
@@ -0,0 +1,115 @@
...
+
+  def usable_ip(ip, cidr)
+    ip_addr = IPAddr.new(ip)
+    return true if ip_addr.ipv6?
+    network = IPAddr.new(cidr)
+    network.include?(ip_addr) && network.to_range.last != ip && network.to_range.first != ip
+  end
+
+  def get_request_group(params)
+    group = params[:group] ? CGI.escape(CGI.unescape(params[:group])) : nil
```
ekohl: Shouldn't `CGI.escape(CGI.unescape())` be a noop? What's the reason for this?

modules/externalipam/ipam_helper.rb
```diff
@@ -0,0 +1,115 @@
...
+
+      if ip_in_cache && !ip_expired
+        next_ip = find_new_ip(ip_cache, subnet_id, new_ip, mac, cidr, group)
+      else
+        next_ip = new_ip.to_s
+        ip_cache.add(group, cidr, new_ip, mac)
+      end
+
+      unless usable_ip(next_ip, cidr)
+        return { error: "No free addresses found in subnet #{cidr}. Some available ip's may be cached. Try again in #{@ip_cache.cleanup_interval} seconds after cache is cleared." }
```
ekohl: Shouldn't this be an exception or return `nil`?

modules/externalipam/netbox/netbox_client.rb
```diff
@@ -0,0 +1,167 @@
...
+
+    def subnets_from_result(result)
+      result.map do |subnet|
+        {
+          subnet: subnet['prefix'].split('/').first,
+          mask: subnet['prefix'].split('/').last,
+          description: subnet['description'],
+          id: subnet['id'],
+        }
+      end
```
ekohl: Isn't this the same?
```suggestion
      result.map { |subnet| subnet_from_result(subnet) }
```

modules/externalipam/netbox/netbox_client.rb
```diff
@@ -0,0 +1,167 @@
...
+      group[:id]
+    end
+
+    def get_ipam_subnets(group_name)
+      params = { status: 'active' }
+      params[:vrf_id] = get_group_id(group_name) if group_name
+      response = @api_resource.get("ipam/prefixes/", params)
+      json_body = JSON.parse(response.body)
+      return nil if json_body['count'].zero?
+      subnets_from_result(json_body['results'])
```
ekohl: What's the reason to return `nil` when it's empty? Isn't an empty array just fine?

modules/externalipam/netbox/netbox_client.rb
```diff
@@ -0,0 +1,167 @@
...
+
+    def ipam_groups
+      response = @api_resource.get('ipam/vrfs/')
+      json_body = JSON.parse(response.body)
+      groups = []
+      return groups if json_body['count'].zero?
+      json_body['results'].each do |group|
+        groups.push({ name: group['name'], description: group['description'] })
+      end
+      groups
```
ekohl: Personally I'd write this as:
```suggestion
      return [] if json_body['count'].zero?
      json_body['results'].map do |group|
        { name: group['name'], description: group['description'] }
      end
```

And if `json_body['results']` is always an array, even if the count is 0 then the separate `return` statement isn't needed.

modules/externalipam/phpipam/phpipam_client.rb
```diff
@@ -0,0 +1,168 @@
...
+      return nil if subnets.nil?
+      subnet_id = nil
+
+      subnets.each do |subnet|
+        subnet_cidr = "#{subnet[:subnet]}/#{subnet[:mask]}"
+        subnet_id = subnet[:id] if subnet_cidr == cidr
+      end
+
+      return nil if subnet_id.nil?
+      response = @api_resource.get("subnets/#{subnet_id}/")
```
ekohl: I think this should also work and provide the same guarantees:
```suggestion
      subnet = subnets&.find { |subnet| cidr == "#{subnet[:subnet]}/#{subnet[:mask]}" }
      return nil if subnet.nil? || subnet[:id].nil?

      response = @api_resource.get("subnets/#{subnet[:id]}/")
```

modules/externalipam/phpipam/phpipam_client.rb
```diff
@@ -0,0 +1,168 @@
...
+      json_body = JSON.parse(response.body)
```
ekohl, line 42: Looking at this, is there ever a case where you call `@api_resource.get()` and not call `JSON.parse`? Should the API abstraction actually parse it for you?

modules/externalipam/api_resource.rb
```diff
@@ -0,0 +1,55 @@
...
+      request = Net::HTTP::Delete.new(uri)
+      request[@auth_header] = @token
+      request['Accept'] = 'application/json'
+      request(request, uri)
+    end
+
+    def post(path, body = nil)
+      uri = URI(@api_base + path)
+      request = Net::HTTP::Post.new(uri)
+      request.body = body
```
ekohl, line 40: Should this actually call `.to_json`? By sending `Content-Type` you are already giving that guarantee so it would save effort on the caller side.

test/externalipam/externalipam_api_test.rb
```diff
@@ -0,0 +1,273 @@
+require 'test_helper'
+require 'externalipam/dependency_injection'
+
+ENV['RACK_ENV'] = 'test'
+
+# Test IPAM provider
+class ExternalIpamTestProvider
+  def get_next_ip(mac, cidr, group_name)
+    { data: "1.2.3.4" }
```
ekohl: I've been trying to get people to use the RFC defined ranges for this. Could you please follow https://community.theforeman.org/t/using-ips-and-domains-in-tests-documentation/16600 throughout the tests?

modules/externalipam/ip_cache.rb
```diff
@@ -0,0 +1,131 @@
...
+    end
+
+    def ip_exists(group_name, cidr, ip)
+      subnet_hash = get_cidr(group_name, cidr)
+      return false if subnet_hash.nil?
+      subnet_hash&.any? { |mac, cached_ip| cached_ip[:ip] == ip }
+    end
+
+    def ip_expired(group_name, cidr, ip)
+      return true unless ip_exists(group_name, cidr, ip)
```
ekohl: This is rather expensive. You're essentially going to do the same lookup twice. Would it make sense to have a parameter on `ip_exists` to allow retrieving expired IPs? Looking at the caller you never call `ip_exists` without also calling `ip_expired`.

modules/externalipam/ip_cache.rb
```diff
@@ -0,0 +1,131 @@
...
+
+    def get_cidr(group_name, cidr)
+      @ip_cache.dig(group_name.to_sym, cidr.to_sym)
+    end
+
+    def get_ip(group_name, cidr, mac)
+      @ip_cache.dig(group_name.to_sym, cidr.to_sym, mac.to_sym, :ip)
+    end
+
+    def ip_exists(group_name, cidr, ip)
```
ekohl: Ruby convention suggests you could name this:
```suggestion
    def ip_exists?(group_name, cidr, ip)
```

modules/externalipam/ip_cache.rb
```diff
@@ -0,0 +1,131 @@
...
+      subnet_hash = get_cidr(group_name, cidr)
+      return false if subnet_hash.nil?
+      subnet_hash&.any? { |mac, cached_ip| cached_ip[:ip] == ip }
+    end
+
+    def ip_expired(group_name, cidr, ip)
```
ekohl: And here too
```suggestion
    def ip_expired?(group_name, cidr, ip)
```

modules/externalipam/ip_cache.rb
```diff
@@ -0,0 +1,131 @@
...
+      @ip_cache = {'': {}}
+      start_cleanup_task
+    end
+
+    def provider_name(provider)
+      @provider = provider
+    end
+
+    def get_cidr(group_name, cidr)
+      @ip_cache.dig(group_name.to_sym, cidr.to_sym)
```
ekohl: Looking through this file, it looks like you're converting everything to symbols with `to_sym`. Do you ever provide it as non-strings? Wouldn't it make more sense to always use strings?

modules/externalipam/ipam_api.rb
```diff
@@ -0,0 +1,357 @@
...
+      group_name = get_request_group(params)
+      next_ip = provider.get_next_ip(mac, cidr, group_name)
+      halt 404, { error: ERRORS[:no_free_ips] }.to_json if next_ip.nil?
+      { data: next_ip }.to_json
+    rescue Proxy::Validations::Error => e
+      logger.warn(e.message)
+      halt 400, { error: e.to_s }.to_json
+    rescue RuntimeError => e
+      logger.warn(e.message)
+      halt 500, { error: e.message }.to_json
```
ekohl: I'd be careful with this and not use `e.message`. RuntimeErrors can be anything and could leak internal data. That could turn into security risks, for example if it contained some private data. Either catch safe exceptions or return some generic error message.

modules/externalipam/ipam_api.rb
```diff
@@ -0,0 +1,357 @@
...
+      cidr = validate_cidr!(params[:address], params[:prefix])
+      group_name = get_request_group(params)
+      next_ip = provider.get_next_ip(mac, cidr, group_name)
+      halt 404, { error: ERRORS[:no_free_ips] }.to_json if next_ip.nil?
+      { data: next_ip }.to_json
+    rescue Proxy::Validations::Error => e
+      logger.warn(e.message)
+      halt 400, { error: e.to_s }.to_json
+    rescue RuntimeError => e
+      logger.warn(e.message)
```
ekohl: Please use `logger.exception('My Context', e)` instead. That provides a proper backtrace in the logs.

ekohl (changes requested):

config/settings.d/externalipam.yml.example
lzap: I like this wording, "support" creates so much confusion.

lzap:
I think this is close to merging, please rebase and we can move forward with this.

grizzthedj:
@lzap Almost finished with most of the recommended changes. Will try to rebase and push at some point next week

grizzthedj:
@lzap @ekohl Sorry for the delay. Had to rebuild my local Foreman dev env. Changes are done and rebased. Let me know if anything else is needed here.

Here is the Foreman PR that is also needed for this integration to work

https://github.com/theforeman/foreman/pull/9174

lzap:
Overall I am okay with the patch, Ewoud had more comments than I anyway I will let him to finish the review. Thanks for this feature!

grizzthedj:
@ekohl Just did a fresh rebase and also wanted to follow-up on this PR. Changes are done and tested. Let me know if you need anything else. Will be great to have this as part of the Foreman/Proxy core!

lib/proxy/validations.rb
```diff
@@ -61,6 +63,23 @@ def validate_subnet(subnet)
     subnet
   end
 
+  def validate_cidr(address, prefix)
+    cidr = "#{address}/#{prefix}"
+    network = IPAddr.new(cidr).to_s
+    if network != IPAddr.new(address).to_s
```
ekohl: Is it needed to do a string-based comparison? Can't the `IPAddr` class be compared?
```suggestion
    network = IPAddr.new(cidr)
    if network != IPAddr.new(address)
```

modules/dhcp_common/free_ips.rb
```diff
@@ -33,7 +33,7 @@ def clean_up_allocated_ips
           break if @allocation_timestamps.first.nil? || @allocation_timestamps.first[1] > time_now
           ip, _ = @allocation_timestamps.shift
           @allocated_ips.delete(ip)
-          logger.debug("#{ip} marked as free.")
+          logger.trace("#{ip} marked as free for DHCP.")
```
ekohl: Is this a trace or should it stay at debug?

modules/externalipam/api_resource.rb
```diff
@@ -0,0 +1,55 @@
+require 'yaml'
+require 'json'
+require 'net/http'
+require 'uri'
+require 'externalipam/ipam_helper'
+
+module Proxy::Ipam
+  # Class to handle authentication and HTTP transactions with External IPAM providers
+  class ApiResource
```
ekohl, line 9: Not something to be fixed in this PR, but I believe this is the third or so implementation of some HTTP client inside smart-proxy. We should really standardize that at some point. Perhaps even use some library which allows connection reuse.

modules/externalipam/ip_cache.rb
```diff
@@ -0,0 +1,133 @@
...
+    # Groups of subnets are cached under the External IPAM Group name. For example,
+    # "IPAM Group Name" would be the section name in phpIPAM. All IP's cached for subnets
+    # that do not have an External IPAM group specified, they are cached under the "" key. IP's
+    # are cached using one of two possible keys:
+    #    1). Mac Address
+    #    2). UUID (Used when Mac Address not specified)
+    #
+    # {
+    #   "": {
+    #     "100.55.55.0/24":{
```
ekohl: I strongly believe we shouldn't use public IP ranges for this, but rather the [RFC5735](https://tools.ietf.org/html/rfc5735) documented example ranges. See https://community.theforeman.org/t/using-ips-and-domains-in-tests-documentation/16600 for more.

modules/externalipam/ip_cache.rb
```diff
@@ -0,0 +1,133 @@
...
+    def add(group_name, cidr, ip, mac = nil)
+      logger.debug("Adding IP '#{ip}' to cache for subnet '#{cidr}' in group '#{group_name}' for IPAM provider #{@provider}")
+      @m.synchronize do
+        mac_addr = mac.nil? || mac.empty? ? SecureRandom.uuid : mac
+        group_hash = @ip_cache[group_name]
+
+        if group_hash&.key?(cidr)
+          @ip_cache[group_name][cidr][mac_addr.to_sym] = { ip: ip.to_s, timestamp: Time.now }
+        else
+          @ip_cache[group_name] = { cidr => { mac_addr.to_sym => { ip: ip.to_s, timestamp: Time.now }}}
```
ekohl: What is the benefit of storing mac addresses as symbols? I think I'd store them as strings. Symbols are intended for developers as something to reuse. IIRC there have been some historical problems with symbols from user input.

modules/externalipam/ipam_api.rb
```diff
@@ -0,0 +1,358 @@
...
+    #   ======================
+    #     Http Code:     200
+    #     JSON Response:
+    #       "100.55.55.3"
+    #
+    #   Response if missing parameter(e.g. 'mac')
+    #   ======================
+    #     Http Code:     400
+    #     JSON Response:
+    #       {"error": ["A 'mac' address must be provided(e.g. 00:0a:95:9d:68:10)"]}
```
ekohl: Looking at the code I'd expect this to be a string
```suggestion
    #       {"error": "A 'mac' address must be provided(e.g. 00:0a:95:9d:68:10)"}
```

modules/externalipam/ipam_api.rb
```diff
@@ -0,0 +1,358 @@
...
+    #
+    #   Response if no groups exist:
+    #   ===========================
+    #     Http Code:     404
+    #     JSON Response:
+    #       {"error": "Groups are not supported"}
+    #
+    #   Response if groups are not supported:
+    #   ===========================
+    #     Http Code:     500
```
ekohl: Internal server errors (which is what HTTP 500 is) shouldn't happen and clients generally don't process them as something to expect. In the Foreman API we often use [HTTP 422 Unprocessable Entity](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422) in these situations.

ekohl (commented):
I think this is looking pretty good. Some minor comments inline.

modules/externalipam/ipam_api.rb
grizzthedj: Understood

modules/externalipam/ip_cache.rb
grizzthedj: Makes sense

modules/externalipam/api_resource.rb
grizzthedj: Agreed

modules/dhcp_common/free_ips.rb
grizzthedj: This should stay at debug - will revert

lib/proxy/validations.rb
grizzthedj: Tested this suggestion locally, works as intended

modules/externalipam/ip_cache.rb
grizzthedj: @ekohl It has been awhile, but I remember there being a reason for using symbols here instead of string. I will switch to string and regression test both providers, and report back here

grizzthedj:
@ekohl - Changes are done and tested. Sorry I squashed the commit already but probably should have done that after your review

grizzthedj:
@ekohl Just a friendly ping on this. I hoping the plan is still to merge this into core. I have completed the requested changes - Let me know if there is anything else needed to move this forward

dmgeurts:
@ekohl Where are things regarding the integration of External IPAM? It would be great to finally have this integrated.

athompson-a2:
@ekohl I ran across this PR while looking to implement a connection to our IPAM with smart proxies.  This is pretty old, so I'm not sure if it's mergeable but it would be very nice to have this feature for our proxies.

wardhus:
Can this get merged? @ekohl @lzap

modules/externalipam/ipam_validator.rb
```diff
@@ -0,0 +1,36 @@
...
+# Module containing validation methods for use by all External IPAM provider implementations
+module Proxy::Ipam::IpamValidator
+  include ::Proxy::Validations
+  include Proxy::Ipam::IpamHelper
+
+  def validate_required_params!(required_params, params)
+    err = []
+    required_params.each do |param|
+      unless params[param.to_sym]
+        err.push errors[param.to_sym]
```
lzap, line 14: `errors` should be `ERRORS` here — there is no `errors` method in scope, so this will raise `NoMethodError` when validation fails on a missing parameter.

modules/externalipam/ipam_helper.rb
```diff
@@ -0,0 +1,113 @@
...
+      ip_expired = ip_cache.ip_expired?(group, cidr, new_ip)
+
+      if ip_in_cache && !ip_expired
+        next_ip = find_new_ip(ip_cache, subnet_id, new_ip, mac, cidr, group)
+      else
+        next_ip = new_ip.to_s
+        ip_cache.add(group, cidr, new_ip, mac)
+      end
+
+      halt 404, { error: "No free addresses found in subnet #{cidr}. Some available ip's may be cached. Try again in #{@ip_cache.cleanup_interval} seconds after cache is cleared." }.to_json unless usable_ip(next_ip, cidr)
```
lzap, line 92: Should be `ip_cache.cleanup_interval` — `ip_cache` is the method parameter here, not an instance variable. This will raise when the usable-IP check fails.

modules/externalipam/ipam_helper.rb
```diff
@@ -0,0 +1,113 @@
...
+
+  def usable_ip(ip, cidr)
+    ip_addr = IPAddr.new(ip)
+    return true if ip_addr.ipv6?
+    network = IPAddr.new(cidr)
+    network.include?(ip_addr) && network.to_range.last != ip && network.to_range.first != ip
+  end
+
+  def get_request_group(params)
+    halt 500, { error: errors[:groups_not_supported] }.to_json if params[:group] && !provider.groups_supported?
```
lzap, line 110: Same as `ipam_validator.rb`: `errors` should be `ERRORS`.

modules/externalipam/ipam_api.rb
```diff
@@ -0,0 +1,358 @@
...
+require 'externalipam/ipam_validator'
+require 'externalipam/dependency_injection'
+
+module Proxy::Ipam
+  # Generic API for External IPAM interactions
+  class Api < ::Sinatra::Base
+    extend Proxy::Ipam::DependencyInjection
+    include ::Proxy::Log
+    include ::Proxy::Validations
+    helpers ::Proxy::Helpers
```
lzap, line 12: Other API modules (DHCP, DNS, Realm, etc.) call `authorize_with_trusted_hosts` and `authorize_with_ssl_client` here. This module can allocate/reserve/delete IPs in external systems and should do the same.

modules/externalipam/ipam_api.rb
```diff
@@ -0,0 +1,358 @@
...
+      group_name = get_request_group(params)
+      next_ip = provider.get_next_ip(mac, cidr, group_name)
+      halt 404, { error: ERRORS[:no_free_ips] }.to_json if next_ip.nil?
+      { data: next_ip }.to_json
+    rescue Proxy::Validations::Error => e
+      logger.exception(ERRORS[:proxy_validation_error], e)
+      halt 400, { error: e.to_s }.to_json
+    rescue RuntimeError => e
+      logger.exception(ERRORS[:runtime_error], e)
+      halt 500, { error: e.to_s }.to_json
```
lzap, line 58: Returning `e.to_s` to the client can leak internal/upstream error details. Prefer a generic message in the response and use `logger.exception` for the real error. Same pattern is repeated on every route in this file.

modules/externalipam/ipam_api.rb
```diff
@@ -0,0 +1,358 @@
...
+    rescue Errno::ECONNREFUSED, Errno::ECONNRESET
```
lzap, line 59: `e` is not bound in this rescue clause — needs `rescue Errno::ECONNREFUSED, Errno::ECONNRESET => e`. Repeated on every route.

modules/externalipam/phpipam/phpipam_client.rb
```diff
@@ -0,0 +1,161 @@
...
+      return nil if json_body['success']
+      { error: 'Unable to delete IP from External IPAM' }
+    end
+
+    def get_next_ip(mac, cidr, group_name)
+      subnet = get_ipam_subnet(cidr, group_name)
+      raise ERRORS[:no_subnet] if subnet.nil?
+      response = @api_resource.get("subnets/#{subnet[:id]}/first_free/")
+      json_body = JSON.parse(response.body)
+      return { error: json_body['message'] } if json_body['message']
```
lzap, line 103: Returning `{ error: ... }` here is not handled by the API layer (it only checks for `nil`), so phpIPAM errors end up as `{"data":{"error":"..."}}` with HTTP 200. Raise or return `nil` and let the API map it to an appropriate error response.

modules/externalipam/phpipam/phpipam_client.rb
```diff
@@ -0,0 +1,161 @@
...
+    end
+
+    def ip_exists?(ip, subnet_id, _group_name)
+      ip = @api_resource.get("subnets/#{subnet_id}/addresses/#{ip}/")
+      json_body = JSON.parse(ip.body)
+      json_body['success']
+    end
+
+    def add_ip_to_subnet(ip, params)
+      data = { subnetId: params[:subnet_id], ip: ip, description: 'Address auto added by Foreman' }
```
lzap, line 84: No hostname is passed when adding an address. Foreman typically has the FQDN available at orchestration time — without it, entries show up unnamed in phpIPAM (see grizzthedj/smart_proxy_ipam#61).

modules/externalipam/netbox/netbox_client.rb
```diff
@@ -0,0 +1,158 @@
...
+      response = @api_resource.get(url, params)
+      json_body = JSON.parse(response.body)
+      !json_body['count'].zero?
+    end
+
+    def add_ip_to_subnet(ip, params)
+      desc = 'Address auto added by Foreman'
+      address = "#{ip}/#{params[:cidr].split('/').last}"
+      group_name = params[:group_name]
+      data = { address: address, nat_outside: 0, description: desc }
```
lzap, line 96: `nat_outside: 0` may not be valid on newer Netbox versions — worth verifying against supported Netbox releases.

modules/externalipam/netbox/netbox_client.rb
```diff
@@ -0,0 +1,158 @@
...
+      ip = json_body[0]['address'].split('/').first
+      cache_next_ip(@ip_cache, ip, mac, cidr, subnet[:id], group_name)
+    end
+
+    def groups_supported?
+      true
+    end
+
+    def authenticated?
+      !@token.nil?
```
lzap, line 132: `authenticated?` only checks that the token is non-nil; it never validates the token against the Netbox API. A bad token will only fail on the first real request.

modules/externalipam/api_resource.rb
```diff
@@ -0,0 +1,55 @@
...
+      request[@auth_header] = @token
+      request['Accept'] = 'application/json'
+      request['Content-Type'] = 'application/json'
+      request(request, uri)
+    end
+
+    private
+
+    def request(request, uri)
+      Net::HTTP.start(uri.hostname, uri.port, use_ssl: uri.scheme == 'https') do |http|
```
lzap, line 50: No read/open timeout, no TLS options for custom CAs, and no HTTP status check before callers `JSON.parse` the body. A 401/500 HTML response will surface as a parse error rather than a clear auth/server failure.

modules/externalipam/api_resource.rb
```diff
@@ -0,0 +1,55 @@
...
+require 'json'
+require 'net/http'
+require 'uri'
+require 'externalipam/ipam_helper'
+
+module Proxy::Ipam
+  # Class to handle authentication and HTTP transactions with External IPAM providers
+  class ApiResource
+    include ::Proxy::Log
+    include Proxy::Ipam::IpamHelper
```
lzap, line 11: `ApiResource` probably does not need `IpamHelper` — that pulls in Sinatra `halt` semantics and the heavy `externalipam/externalipam` require chain into a plain HTTP client class.

modules/externalipam/ip_cache.rb
```diff
@@ -0,0 +1,133 @@
...
+      @ip_cache.dig(group_name, cidr, mac, :ip)
+    end
+
+    def ip_exists?(group_name, cidr, ip)
+      subnet_hash = get_cidr(group_name, cidr)
+      return false if subnet_hash.nil?
+      subnet_hash&.any? { |mac, cached_ip| cached_ip[:ip] == ip }
+    end
+
+    def ip_expired?(group_name, cidr, ip)
```
lzap, line 42: `ip_expired?` calls `ip_exists?` first, then repeats the same cache scan. Callers always use both together — consider a single lookup or an `include_expired` flag to avoid the double walk.

modules/externalipam/ip_cache.rb
```diff
@@ -0,0 +1,133 @@
...
+    include Singleton
+    include Proxy::Log
+    include Proxy::Ipam::IpamHelper
+
+    DEFAULT_CLEANUP_INTERVAL = 180
+
+    def initialize
+      @m = Monitor.new
+      @ip_cache = {'': {}}
+      start_cleanup_task
```
lzap, line 21: The cleanup `TimerTask` starts in `initialize` via the Singleton, so it runs even when the externalipam module is disabled. Consider starting it lazily on first use, and add a `stop` hook like `Proxy::DHCP::FreeIps`.

modules/externalipam/ip_cache.rb
```diff
@@ -0,0 +1,133 @@
...
+      Time.now - ip_expiration > DEFAULT_CLEANUP_INTERVAL
+    end
+
+    def start_cleanup_task
+      logger.info("Starting ip cache maintenance for External IPAM provider, used by /next_ip.")
+      @timer_task = Concurrent::TimerTask.new(execution_interval: DEFAULT_CLEANUP_INTERVAL) { clean_cache }
+      @timer_task.execute
+    end
+
+    # @ip_cache structure
```
lzap, line 78: MAC addresses are used as hash keys without normalizing case/format. `00:0A:95:...` and `00:0a:95:...` would be treated as different cache entries and could get different suggested IPs.

modules/externalipam/ipam_validator.rb
```diff
@@ -0,0 +1,36 @@
...
+      end
+    end
+    raise Proxy::Validations::Error, err unless err.empty?
+  end
+
+  # def validate_cidr!(address, prefix)
```
lzap, line 20: Commented-out dead code — please remove before merge.

modules/externalipam/ipam_helper.rb
```diff
@@ -0,0 +1,113 @@
...
+    bad_cidr: 'The network cidr is invalid',
+    proxy_validation_error: 'Proxy validation error',
+    runtime_error: 'Runtime error',
+  }.freeze
+
+  def provider
+    @provider ||=
+      begin
+        unless client.authenticated?
+          halt 500, { error: 'Invalid credentials for External IPAM' }.to_json
```
lzap, line 35: Invalid credentials returns HTTP 500. Foreman often uses 422 for expected client/config errors rather than 500.

test/externalipam/externalipam_api_test.rb
```diff
@@ -0,0 +1,273 @@
+require 'test_helper'
+require 'externalipam/dependency_injection'
+
+ENV['RACK_ENV'] = 'test'
+
+# Test IPAM provider
+class ExternalIpamTestProvider
+  def get_next_ip(mac, cidr, group_name)
+    { data: "192.0.2.1" }
```
lzap, line 9: The test provider returns `{ data: "192.0.2.1" }` but the API wraps the provider result again as `{ data: next_ip }`, so the real response is double-wrapped. Mocks should return plain values (e.g. `"192.0.2.1"`) like the real providers do.

modules/externalipam/ipam_helper.rb
```diff
@@ -0,0 +1,113 @@
+require 'proxy/validations'
+require 'externalipam/externalipam'
```
lzap, line 2: This require pulls in the full plugin chain (`externalipam/externalipam` → all providers). That creates a circular/heavy dependency for a helper module — consider moving shared constants (e.g. `ERRORS`) to a smaller file.

lzap:
Hello, I joined the team back recently and I can help to push this forward. This will need a rebase for sure.

The biggest problem is that every other Smart Proxy API module uses trusted-host and SSL client verification:

```ruby
helpers ::Proxy::Helpers
authorize_with_trusted_hosts
authorize_with_ssl_client
```

Please extend tests for providers, cache, auth, integration and IPv6. Testing is really slim for such a feature.

PR #811, opened 2022-01-06
Fixes #34226 - dnscmd nil string error

lzap:
This was introduced in

https://github.com/theforeman/smart-proxy/commit/9522669b7287b62a8bc073f6f8abb5554184e57d

I missed that there are few more uses of the `execute` methods which should have been also modified for the change.


ekohl (approved):

lzap:
Thanks, CP: https://github.com/theforeman/smart-proxy/pull/812

PR #813, opened 2022-01-11
Refs #34226 - dnscmd nil string error

lzap:
There was one more problem which I missed. Some arguments were appended into argument array as space-separated strings which indeed fails when executed.


ekohl (approved):

lzap:
Damn, I should have stated that I don't have Windows setup so I was waiting for the user to confirm. There is still one problem, I will do another PR.

PR #816, opened 2022-02-01
Fixes #34359 - fix DHCP IP validation arguments

lzap:
This sneaked into develop and 3.1 because the method is stubbed in all tests.

This is a blocker bug, let's put it into 3.1 too.

### DIFF
--- a/modules/dhcp_common/server.rb
+++ b/modules/dhcp_common/server.rb
@@ -29,8 +29,10 @@ module Proxy::DHCP
                          end
     end
 
-    def validate_supported_address(ip)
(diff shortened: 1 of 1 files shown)


evgeni:
Can we get a test so it doesn't happen again?

lzap:
Let's please merge, backport and I will leave the test on @avitova :)

evgeni:
No.

lzap:
I don't have time tomorrow and I am not working for the rest of the week, well, I am leaving this on @avitova then.

PR #817, opened 2022-02-01
Fixes #34359 - fix DHCP IP validation arguments tests

avitova:
I was inspecting my PR where this bug was "created". As well as there were not test for server.rb file, I managed to not write any test for 3 similar methods. Somehow only one of them was problematic, but I believe that I should add these test for all methods.


ekohl (commented):
For reference, this fixes eba7d3b463fca6a37ae8d047cdf8e981a96b5a10.

You could also write an integration test. For example in `test/dhcp_isc/integration_test.rb` you could test `get /:network/unused_ip` which calls the method with both a single and multiple parameters (if you pass from and to). It's also easy to stub out the `unused_ip` call to validate the normal flow.

avitova:
> For reference, this fixes [eba7d3b](https://github.com/theforeman/smart-proxy/commit/eba7d3b463fca6a37ae8d047cdf8e981a96b5a10).
> 
> You could also write an integration test. For example in `test/dhcp_isc/integration_test.rb` you could test `get /:network/unused_ip` which calls the method with both a single and multiple parameters (if you pass from and to). It's also easy to stub out the `unused_ip` call to validate the normal flow.

We have validate_supported_address method stubbed in test right now, so it would always return true. I believe it should be enough to not do this and instead change the expected return values in current tests. I get the idea about integration test, but since we have different flows of calling the method for validation, would it not be safer to test as it is but without the stub?

Edit: Nevermind this comment, I think I figured it out.

avitova:
I was also thinking whether we should really return string with IP address in this method `validate_supported_address`. 
I left the return value for `validate_ip` because it is actually used somewhere. But for this specific method it seems kind of random to me, to return only the last parameter.

avitova:
I deleted few lines in `dhcp_api_test` file, as it could probably cause false assumption that the `validate_supported_address` is tested there. Return values of the method itself are tested as well as its invocation in `get /:network/unused_ip`.

test/dhcp_isc/integration_test.rb
```diff
@@ -39,3 +40,22 @@ def test_features
...
+    app = Proxy::DhcpApi.new
+    app.helpers.server = @server
+    app
+  end
+
+  def test_validate_ip
+    @server = mock()
+    @server.stubs(:unused_ip).returns(true)
+    @server.expects(:validate_supported_address).with("192.168.122.0").returns("192.168.122.0")
+    @server.expects(:validate_supported_address).with("192.168.122.10", "192.168.122.20").returns("192.168.122.20")
```
ekohl: I think this is not really an integration test. I'd suggest to not stub this.

It would also be a good test to perform a request with an invalid IP address.

test/dhcp_isc/integration_test.rb
```diff
@@ -39,3 +40,22 @@ def test_features
...
+  include Rack::Test::Methods
+
+  def app
+    app = Proxy::DhcpApi.new
+    app.helpers.server = @server
+    app
+  end
+
+  def test_validate_ip
+    @server = mock()
```
ekohl: I'd avoid a mock in an integration test. Rather test the real thing.

test/dhcp_isc/integration_test.rb
```diff
@@ -39,3 +40,22 @@ def test_features
...
+    @server.stubs(:unused_ip).returns(true)
```
ekohl: This is still a good idea, but I'd rather stub this on the actual implementation.

test/dhcp_isc/integration_test.rb
```diff
@@ -39,3 +40,22 @@ def test_features
...
+    @server.expects(:validate_supported_address).with("192.168.122.0").returns("192.168.122.0")
+    @server.expects(:validate_supported_address).with("192.168.122.10", "192.168.122.20").returns("192.168.122.20")
+    get "/192.168.122.0/unused_ip?mac=01:02:03:04:05:06&from=192.168.122.10&to=192.168.122.20"
```
ekohl: While the repository is littered with bad examples, please use the example ranges according to https://community.theforeman.org/t/using-ips-and-domains-in-tests-documentation/16600.

avitova:
The only stubbed method is now unused_ip.

test/dhcp/dhcp_api_test.rb
```diff
@@ -104,14 +101,12 @@ def test_get_network_unused_ip
   end
 
   def test_get_unused_ip_for_nonexistent_network
-    @server.expects(:validate_supported_address).returns(true)
     @server.expects(:unused_ip).with('192.168.122.0', "01:02:03:04:05:06", "192.168.122.10", "192.168.122.20").raises(::Proxy::DHCP::SubnetNotFound)
     get "/192.168.122.0/unused_ip?mac=01:02:03:04:05:06&from=192.168.122.10&to=192.168.122.20"
     assert_equal 404, last_response.status
   end
 
   def test_get_unused_when_not_inmplemented
```
ekohl, line 109: Unrelated, but
```suggestion
  def test_get_unused_when_not_implemented
```

ekohl (approved):
Thanks!

PR #822, opened 2022-03-05
Fixes #34563 - Drop rdoc from test dependencies

ekohl:
By allowing the Rakefile to gracefully handle rdoc's absense it can be dropped from the test dependencies.

This also fixes the task so it actually runs in an additional commit. Given that it's been broken for so long it does suggest nobody uses it.


lzap (approved):
Thanks

PR #827, opened 2022-03-15
Fixes #34613 - Verify SSL CA file is readable on startup

adamruzicka:


lib/launcher.rb
```diff
@@ -102,13 +102,19 @@ def https_app(https_port, plugins = https_plugins)
...
         :SSLOptions => ssl_options,
         :SSLCiphers => CIPHERS - Proxy::SETTINGS.ssl_disabled_ciphers,
         :daemonize => false,
       }
     end
 
+    def verify_ssl_ca_file(path)
+     return path if File.readable?(path)
+
+     logger.error "Unable to load CA SSL certificate. Are the values correct in settings.yml and do permissions allow reading?"
```
adamruzicka: Author of the original issue expected an explicit error in the logs, but letting the service start. I'd be inclined to raise an exception the same way as we do when private key or certificate are invalid

lzap: I would probably avoid such naming for X509 cert, this can make impression that the method actually validates the cert while I read the code as it checks if file is readable.

I also doubt if `File.readable?` really needs an explicit function, I would prefer doing this before the return statement of the previous method, but this is just a preference.

Also the wording is a bit rough, I would also add the filepath in the error message too: "Unable to read #{path}, check settings.yml and permissions" or something like that.

lib/launcher.rb
```diff
@@ -102,13 +102,19 @@ def https_app(https_port, plugins = https_plugins)
         :SSLVerifyClient => OpenSSL::SSL::VERIFY_PEER,
         :SSLPrivateKey => load_ssl_private_key(settings.ssl_private_key),
         :SSLCertificate => load_ssl_certificate(settings.ssl_certificate),
-        :SSLCACertificateFile => settings.ssl_ca_file,
+        :SSLCACertificateFile => verify_ssl_ca_file(settings.ssl_ca_file),
```
ekohl: Would it make sense to do it earlier? Like after line 67?

lib/launcher.rb
lzap: To elaborate "naming", what I mean is the name of the method.

lib/launcher.rb
lzap: Yeah I tend to prefer to do just a single if statement in this method, unless there are plans to add additional verification.

lzap (approved):
Thanks!

PR #828, opened 2022-03-16
Fixes #34624 - Set MALLOC_ARENA_MAX to counter memory bloating

adamruzicka:


ekohl (approved):
We already do this in other areas. I'd mostly consider not having this here is an oversight. Thanks!

ekohl:
@adamruzicka since you asked on IRC while I was out, I would consider picking this into stable branches.

PR #829, opened 2022-03-19
Add required timeout import

StrangeQuark87:
Add missing import causing errors with Windows DNS

### DIFF
--- a/modules/dns_dnscmd/dns_dnscmd_main.rb
+++ b/modules/dns_dnscmd/dns_dnscmd_main.rb
@@ -1,4 +1,5 @@
 require 'open3'
+require 'timeout'
 
 module Proxy::Dns::Dnscmd
   class Record < ::Proxy::Dns::Record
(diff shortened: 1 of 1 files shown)


ezr-ondrej:
ok to test

ekohl:
Looks like this was introduced in 168ee4e64718eed50e3e140d0fd2d54db0999b56. Even Ruby 2.0.0 documentation states:
> Previous versions didn't use a module for namespacing, however timeout is provided for backwards compatibility. You should prefer Timeout#timeout instead.

So this looks correct. We do require an issue to be created. Could you [create one](https://projects.theforeman.org/projects/smart-proxy/issues/new)?

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -37,7 +38,7 @@ def execute(*args, **keyword_args)
       std_in = std_out = std_err = nil
       real_cmd = ['c:\Windows\System32\dnscmd.exe', @server] + args
       begin
-        timeout(tsecs) do
+        Timeout.timeout(tsecs) do
```
ekohl, line 41: This was addressed in aa25357d4404908d5e5707e767ecd798b754d7d1.

modules/dns_dnscmd/dns_dnscmd_main.rb
```diff
@@ -1,4 +1,5 @@
 require 'open3'
+require 'timeout'
```
ekohl, line 2: Done in https://github.com/theforeman/smart-proxy/pull/842

PR #832, opened 2022-05-02
Fixes #34844 - load SSL keys using OpenSSL::PKey.read

evgeni:
This allows using non-RSA keys as PKey.read can figure out the type of key automatically


lzap (approved):
Looks sane, I haven't tested tho.

evgeni:
I did verify that it does still work with RSA keys and expect OpenSSL to do the right thing if fed non-RSA keys.

PR #837, opened 2022-07-25
Fixes #35270 - Enable boot image download

bastian-src:
Manage ISO images as Installation media. Refers to [this](https://github.com/theforeman/foreman/pull/9318) PR.

Includes the following changes:
* Implement fetch and extract boot image
* Apply correct file permissions
* Implement classes for file extraction and permission changes

*ToDo:* Add tests.


lib/proxy/filesystem_permission.rb
```diff
@@ -0,0 +1,19 @@
+module Proxy
+  class FilesystemPermission < Proxy::Util::CommandTask
```
ekohl: Please don't shell out for chmod. There is `FileUtils.chmod` and `FileUtils.chmod_R`.

lib/proxy/archive_extract.rb
```diff
@@ -0,0 +1,21 @@
...
+      args = [which('7z')]
+
+      # extract command
+      args << "x"
+      # source file
+      args << src.to_s
+      # skip existing files
+      args << "-aos" if skip_existing
+      # destination directory
+      args << "-o#{dst}"
```
ekohl: I don't know if 7z supports this, but if there are long options I'd always use them in scripts. That way they're self documenting.

modules/tftp/server.rb
```diff
@@ -150,6 +150,22 @@ def pxeconfig_file(mac)
...
+    dst_file = Pathname.new(dst).cleanpath
+    dst_path = Pathname.new(dst.delete_suffix(".iso"))
+
+    FileUtils.mkdir_p dst_path.parent
+    choose_protocol_and_fetch(src, dst_file).join
+    # extract iso
+    extract_task = ::Proxy::ArchiveExtract.new(dst_file, dst_path).start
+    raise "TFTP image extraction error" unless extract_task.join == 0
+    # adapt extracted file permission
+    permission_task = ::Proxy::FilesystemPermission.new(dst_path, "755", true).start
```
ekohl: This implies you're making the files executable. Is that really what you want?

lib/proxy/archive_extract.rb
```diff
@@ -0,0 +1,21 @@
+module Proxy
+  class ArchiveExtract < Proxy::Util::CommandTask
+    include Util
+
+    def initialize(src, dst, skip_existing = true)
+
+      args = [which('7z')]
```
ekohl: 7z is not commonly available. Is this really the only tool that can extract it? For example, it's not in CentOS 8 (which is the only EL platform now that we've dropped EL7). It is in EPEL, but at least right now Foreman is incompatible with EPEL8 (https://github.com/theforeman/foreman/pull/9317 may change that).

Looking at https://www.tecmint.com/extract-files-from-iso-files-linux/ there's also isoinfo (in the genisoimage package in EL8).

Also note that this should be reflected in packaging so that the command is actually available.

lib/proxy/filesystem_permission.rb
bastian-src: Done! Makes totally sense.

modules/tftp/server.rb
bastian-src: Yes, if I don't make the executable, they did not show up in the pub folder / were not downloadable.

lib/proxy/archive_extract.rb
bastian-src: Thanks for your idea with `isoinfo`. I tried to setup an alternative which makes use of it instead of `7z`. Unfortunately, the fact that `isoinfo` does not have an option to directly export the file makes some trouble:
It seems to be best practice with `isoinfo` to forward stdout to a file. AFAIK, this can't simply be adapted to the `CommandTask` class. It logs the stdout but I didn't find an easy way to store the stdout from [here](https://github.com/theforeman/smart-proxy/pull/838/files#diff-c2b0ceb586ecbff7db402cf8b53d18cec5ab6a5dac16e48ded1b194f34ac342cR32) to a file. 

I've created a "prototype" PR which summarizes my approach with `isoinfo` here: #838

modules/tftp/server.rb
ekohl: On Unix systems directories need to be executable, but files don't. However, I'd expect it to already have that by default.

modules/tftp/server.rb
```diff
@@ -150,6 +150,21 @@ def pxeconfig_file(mac)
     end
   end
 
+  def self.fetch_boot_image(dst, src)
+
```
ekohl: No need for the empty line.
```suggestion
```

modules/tftp/tftp_api.rb
```diff
@@ -34,6 +34,10 @@ def create_default(variant)
       end
     end
 
+    post "/fetch_boot_image" do
+      log_halt(400, "TFTP: Failed to fetch boot file: ") { Proxy::TFTP.fetch_boot_image(params[:path], params[:url]) }
```
ekohl: What is the reason to present this as a HTTP 400 Bad Request? I see it was presented in the old code too, but I'm not sure it makes sense. HTTP 400 errors means the client did something wrong but isn't it really on the server side that there's a problem if it fails? So HTTP 5xx?

lib/proxy/archive_extract.rb
```diff
@@ -0,0 +1,21 @@
+module Proxy
+  class ArchiveExtract < Proxy::Util::CommandTask
+    include Util
+
+    def initialize(src, dst, skip_existing = true)
+
```
ekohl:
```suggestion
```

test/tftp/tftp_test.rb
```diff
@@ -5,7 +5,8 @@
 class TftpTest < Test::Unit::TestCase
   def setup
     @tftp = Proxy::TFTP::Server.new
-    Proxy::TFTP::Plugin.load_test_settings(:tftproot => "/some/root")
+    Proxy::TFTP::Plugin.load_test_settings(:tftproot => "/some/root",
+                                           :tftp_image_path => "/another/root")
```
ekohl: I don't see this setting anywhere else. Is this a leftover or something you intend to implement?

modules/tftp/server.rb
```diff
@@ -150,6 +150,21 @@ def pxeconfig_file(mac)
...
+    # Verify dst is a valid directory
+    dst_file = Pathname.new(dst).cleanpath
+    dst_path = Pathname.new(dst.delete_suffix(".iso"))
+
+    FileUtils.mkdir_p dst_path.parent
```
ekohl: What if the parent is already present? Will you overwrite files?

ekohl (commented):
We should also expose this as a capability. In https://theforeman.org/2019/04/smart-proxy-capabilities-explained.html I've explained this, but this allows the Foreman side to know if the API is supported. You could even look up the presence of `7z` and only present the capability if it's present. That's another way of dealing with it.

Of course the Foreman side also needs to be modified to respect this.

modules/tftp/server.rb
```diff
@@ -150,6 +150,21 @@ def pxeconfig_file(mac)
...
+    choose_protocol_and_fetch(src, dst_file).join
+    # extract iso
+    extract_task = ::Proxy::ArchiveExtract.new(dst_file, dst_path).start
```
lzap: This IMHO introduces command execution flaw, Pathname and cleanpath only cleans pathnames, but any arbitrary path can be passed:

```
>> Pathname.new("/anything/can/be/passed").cleanpath
=> #<Pathname:/anything/can/be/passed>
```

I do not have any attack vector in my hands now, but clearly passing `/etc/passwd` into any extracting utility is not what we want to allow. I think the path must be enforced to be relative, then cleaned (stripped of any `..`) and then appended to some kind of base dir constant.

modules/tftp/server.rb
m-bucher: As far as the directory is concerned, this is handled by `mkdir_p`, which `causes to make following directories, if it does not exist.` (https://ruby-doc.org/stdlib-2.5.1/libdoc/fileutils/rdoc/FileUtils.html#method-c-mkdir_p)

The `dst_file` itself will be overwritten, I guess.

modules/tftp/tftp_api.rb
m-bucher: I guess this could be discussed, since it might as well be that an incorrect URL has been specified, which would then mean 404, can't find what you are looking for.

This would probably need a more sophisticated differentiation on what error actually occurred.

modules/tftp/server.rb
ekohl: > The dst_file itself will be overwritten, I guess.

Will you have the case where 2 concurrent hosts are provisioning and one reads the file while it's being overwritten?

modules/tftp/server.rb
m-bucher: I see what you mean.

Maybe it might be better to set the base-path (here: `/var/www/html/pub/installation_media`) in the foreman-proxy configuration rather than by the request.
This has the downside that we cannot customize it at runtime, but I am not sure if this really is necessary.

lib/proxy/archive_extract.rb
bastian-src: Replaced with `isoinfo`.

modules/tftp/server.rb
bastian-src: Tested without additional permission setting - you are right, it's unnecessary.

lib/proxy/archive_extract.rb
bastian-src: Changed to `isoinfo` implementation.

test/tftp/tftp_test.rb
bastian-src: You are right. Removed it.

modules/tftp/server.rb
bastian-src: Fixed with implementing the locking from the `http_download` module.

lib/proxy/util.rb
```diff
@@ -27,8 +28,13 @@ def start(&ensured_block)
...
           stdin.close
-          stdout.each do |line|
-            logger.debug "[#{thr.pid}] #{line}"
+          unless @output.nil?
+            File.binwrite(@output, stdout.read)
+            stdout.close
+          else
+            stdout.each do |line|
+              logger.debug "[#{thr.pid}] #{line}"
+            end
```
ekohl: I think the debug logging behavior should remain, regardless of the presence of `@output`. If not, at least the comments above the method should reflect the implementation.

It may be better to introduce a new method to "pipe" data.

modules/tftp/server.rb
```diff
@@ -150,6 +150,26 @@ def pxeconfig_file(mac)
...
+
+    files.each do |file|
+      file_path = Pathname.new file
+      extr_file_path = Pathname.new(File.join(extr_image_dir, file_path)).cleanpath
+
+      # Create destination directory
+      FileUtils.mkdir_p extr_file_path.parent
+      # extract iso
+      extract_task = ::Proxy::ArchiveExtract.new(image_path, file_path, extr_file_path).start
+      raise "TFTP image file extraction error: #{file_path}" unless extract_task.join == 0
```
ekohl: This does mean that if a single file couldn't be extracted, you leave somewhere in the middle. Not sure how you would do this well though. Making this transactional is hard.

lib/proxy/archive_extract.rb
```diff
@@ -0,0 +1,30 @@
+module Proxy
+  class ArchiveExtract < Proxy::Util::CommandTask
+    include Util
+
+    def initialize(image_path, file_in_image, dst_path)
+      args = [which('isoinfo')]
+
+      # read the file
+      args << "-R"
```
ekohl: Is this correct?
```console
$ isoinfo --help 2>&1 | grep -- -R
	-R		Print information from Rock Ridge extensions
	dev=target	SCSI target to use as CD/DVD-Recorder
```

lib/proxy/archive_extract.rb
```diff
@@ -0,0 +1,30 @@
...
+      # set image path
+      args += ["-i", image_path.to_s]
+      # set file path within the image
+      args += ["-x", file_in_image.to_s]
```
ekohl: Performance wise it's probably better to instantiate complete. You could do this:
```ruby
args = [
  which('isoinfo'),
  # Print information from Rock Ridge extensions
  '-R',
  # Filename to read ISO-9660 image from
  '-i', image_path.to_s,
  # Extract specified file to stdout
  '-x', file_in_image.to_s,
]
```

lib/proxy/util.rb
```diff
@@ -27,8 +28,13 @@ def start(&ensured_block)
           logger.info "[#{thr.pid}] Started task #{cmdline_string}"
           stdin.write(input) if input
           stdin.close
-          stdout.each do |line|
-            logger.debug "[#{thr.pid}] #{line}"
+          unless @output.nil?
+            File.binwrite(@output, stdout.read)
```
ekohl: Note this reads the whole file to memory and then writes it at once. It means that if there are large files to be extracted in the ISO the Smart Proxy's memory usage will spike. In the `Open3` module there are other methods which allow streaming.

ekohl (commented):
I left some comments about the command task. Perhaps `ArchiveExtract` shouldn't call `super` at all in `start` but rather implement its own logic.

modules/tftp/server.rb
bastian-src: I've implemented a constant base-path to minimise the amount of parameters for this feature. 
Moreover, a constantly defined path makes it easier to know the exact path on the foreman-side. This important to render the templates correctly.

lib/proxy/archive_extract.rb
bastian-src: You're right, I applied this from a tutorial and I think we don't need it here.

lib/proxy/archive_extract.rb
bastian-src: Done.

modules/tftp/server.rb
bastian-src: Do you think it would be better to continue with extracting the other files? We could check in the end whether all files have been extracted and return an error message depending on that.

lib/proxy/util.rb
bastian-src: What output would you like to print here? Even if we pipe data into the file, there won't be any kind of special output since we want all of it to be written to the file. Moreover, it's a binary file so I wouldn't recommend attaching this to some debug log.

lib/proxy/util.rb
bastian-src: I guess this also what you refer to in your comment here?
> Perhaps ArchiveExtract shouldn't call super at all in start but rather implement its own logic.

Can you give a litte more detail what you have in mind with `own logic`? Should it be a class different from CommandTask? Or should there just be a different `initialize` when you want to pipe the output to a file? I guess we would also need a different (or at least more comprehensive) `start` method which differentiates between piping the output or just logging it.
Do you think about a more general implementation which pipes the output to another command or just into a file? I would favour the latter.

Thanks for your help on this, highly appreciated!

bastian-src:
> We should also expose this as a capability. In https://theforeman.org/2019/04/smart-proxy-capabilities-explained.html I've explained this, but this allows the Foreman side to know if the API is supported. You could even look up the presence of `7z` and only present the capability if it's present. That's another way of dealing with it.
> 
> Of course the Foreman side also needs to be modified to respect this.

I had a look at your comment and I think it makes totally sense to have such a `capability` handling for this feature. I just added [this](https://github.com/theforeman/smart-proxy/pull/837/commits/4ec8d4def512892ce1bba88a1301f5d072360c4e) commit from what I see from other modules (like the bmc plugin [here](https://github.com/theforeman/smart-proxy/blob/develop/modules/bmc/bmc_plugin.rb#L10)) and your blog post. Can you maybe give me an example how the capability is checked on the Foreman-side? I would implement it then before triggering the iso image download API.

ekohl:
I've been super busy, so wasn't able to get back to you. In the SmartProxy model there are both `capabilities(feature)` and `capability?(feature, capability)` (https://github.com/theforeman/foreman/blob/c122b06d124401fa7766caaa04d289054a9a6591/app/models/smart_proxy.rb#L100-L106) to check. An example in the Foreman codebase is here: https://github.com/theforeman/foreman/blob/c122b06d124401fa7766caaa04d289054a9a6591/app/models/concerns/orchestration/dhcp.rb#L81-L89

It is up to you to deal with this. In the future I would like an implementation of a feature on the Foreman side to be able to declare required capabilities. That way we can let the feature registration fail, but we're not there today. And in this particular case it isn't right anyway since you should support an overlap period where you gracefully handle the capability not being there.

Does that answer your question?

modules/tftp/tftp_plugin.rb
```diff
@@ -9,6 +9,9 @@ class Plugin < ::Proxy::Plugin
                      :verify_server_cert => true
     validate :verify_server_cert, boolean: true
 
+    # Expose automatic iso handling capability
+    capability 'extraction'
```
ekohl: You could make this dynamic and see if `isoinfo` is in the path. Upside is that it won't fail at runtime, downside is that you need to refresh the features from Foreman after installing `isofind`. For that downside we can compensate by handling the installation either in packaging or in the installer.

bastian-src: Sounds like a good idea!
When you talk about making it dynamic, should I take the BMC plugin as an example and how it's done here for the IPMI part?
https://github.com/theforeman/smart-proxy/blob/a48da738dd0f41cb28c7042e5f2773a8d9a29293/modules/bmc/bmc_plugin.rb#L9-L14

I would suggest doing a similar implementation which checks the availability of `isoinfo` in this file:
https://github.com/theforeman/smart-proxy/blob/a48da738dd0f41cb28c7042e5f2773a8d9a29293/modules/tftp/tftp.rb#L1-L2

What do you think? Or should I create a similar class as it's done for IPMI instead of using the "base" class `tftp.rb`?

ekohl: > I would suggest doing a similar implementation which checks the availability of `isoinfo` in this file:

This indeed

bastian-src: Just pushed an example implementation. I'm not sure about [this](https://github.com/theforeman/smart-proxy/pull/837/files#diff-d919198de16f6e911bb0e4e47ec7f39cd03a5ccce51a32cb8b4be69c99c64d44R7) part for two reasons tho:

* Should I be more cautious about the `system` call here?
* In the `CommandTask` implementations like `HttpDownload`, they just have a statement like `args = [which('url')]` and I don't understand why `which` is not a string here. Is it an actual function? If yes, where does it come from?

modules/tftp/tftp_plugin.rb
```diff
@@ -9,6 +9,9 @@ class Plugin < ::Proxy::Plugin
                      :verify_server_cert => true
     validate :verify_server_cert, boolean: true
 
+    # Expose automatic iso handling capability
+    capability -> { Proxy::TFTP::Capabilities.has_extraction }
```
ekohl: There isn't a strict need to call a method on a different class. You can just write
```suggestion
    capability -> { which(Proxy::ArchiveExtract.SHELL_COMMAND) ? 'extraction' : nil }
```

modules/tftp/tftp.rb
```diff
@@ -1,2 +1,10 @@
 require 'tftp/tftp_plugin'
-module Proxy::TFTP; end
+module Proxy::TFTP
+  class Capabilities
+
+    # check availability of isoinfo command
+    def self.has_extraction
+      system(which(Proxy::ArchiveExtract.SHELL_COMMAND)) ? 'extraction' : nil
```
ekohl: `system()` means you actually execute it. I think that's too much and would think it's sufficient to check it's there.

modules/tftp/tftp_api.rb
```diff
@@ -34,6 +34,11 @@ def create_default(variant)
       end
     end
 
+    post "/fetch_boot_image" do
+      log_halt(400, "TFTP: Wrong input parameters given. ") unless [params[:path], params[:url], params[:files]].all?
```
ekohl: Is the empty space intended?

Nit: I'd leave an empty line after it since essentially this is a return statement
```suggestion
      log_halt(400, "TFTP: Wrong input parameters given.") unless [params[:path], params[:url], params[:files]].all?

```

modules/tftp/tftp_api.rb
```diff
@@ -34,6 +34,11 @@ def create_default(variant)
...
+      log_halt(500, "TFTP: Failed to fetch boot file: ") { Proxy::TFTP.fetch_boot_image(params[:path], params[:url], params[:files]) }
```
ekohl: I'd use a block for readability.
```suggestion
      log_halt(500, "TFTP: Failed to fetch boot file: ") do
        Proxy::TFTP.fetch_boot_image(params[:path], params[:url], params[:files])
      end
```

modules/tftp/server.rb
```diff
@@ -150,6 +150,36 @@ def pxeconfig_file(mac)
     end
   end
 
+  def self.fetch_boot_image(image_dst, url, files)
+    # Verify image_dst does not contain ".." (switch folder)
+    raise "TFTP image file extraction error: #{file_path}" unless image_dst.split("..").size <= 1
+
+    base_path = Pathname.new("/var/www/html/pub/installation_media")
```
ekohl: Hardcoded paths are always bad. This must at the very least be a setting. But there's more. `/var/www/html` is a Red Hat-ism and `/var/www/html/pub` is a Katello-ism. In fact, there being some HTTP server at all is something that's not guaranteed at all since only Katello does that, but vanilla Foreman doesn't.

I'd say that we should not do this and the Smart Proxy itself should serve the files. Otherwise the feature isn't available on non-Katello and this is a core Smart Proxy module.

Having said that, today we don't have a setting for Foreman Proxy's state directory (essential `/var/lib/foreman-proxy` in production). We've ran into that a few times and would be a good feature. Perhaps this is the excuse to start introducing that.

modules/tftp/server.rb
```diff
@@ -150,6 +150,36 @@ def pxeconfig_file(mac)
...
+    image_path = Pathname.new(File.expand_path(image_dst, base_path)).cleanpath
+    extr_image_dir = Pathname.new(image_path.to_s.delete_suffix(".iso"))
```
ekohl: This assumes the suffix is `.iso`. How about using `Pathname`:
```suggestion
    extr_image_dir = image_path.sub_ext('')
```

modules/tftp/server.rb
```diff
@@ -150,6 +150,36 @@ def pxeconfig_file(mac)
...
+
+    FileUtils.mkdir_p image_path.parent
```
ekohl:
```suggestion
    image_path.parent.mkpath
```

modules/tftp/server.rb
```diff
@@ -150,6 +150,36 @@ def pxeconfig_file(mac)
...
+    download_task = choose_protocol_and_fetch(url, image_path)
+    # Wait for concurrent processes
+    raise "Cannot download boot image at the moment - is another process downloading it already?" if download_task.is_a?(FalseClass)
+    # wait for download completion
+    download_task.join
+
+    files.each do |file|
+      raise "File to extract from image contains up-directory: #{file_path}" unless file.split("..").size <= 1
```
ekohl: Isn't this the same as
```suggestion
      raise "File to extract from image contains up-directory: #{file_path}" unless file.include?("..")
```

This can actually be a false positive. Take `/path/to/some...file.iso'` for example. This is a valid path without exploits.

A better way would be to see if the absolute path of `extr_file_path` is a subdirectory of `extr_image_dir`. I think this works:
```
if extr_file_path.expand_path.relative_path_from(extr_image_dir).to_s.start_with('..')
  raise "File to extract from image contains up-directory: #{file_path}"
end
```

There may be a shorter or more efficient way.

modules/tftp/server.rb
```diff
@@ -150,6 +150,36 @@ def pxeconfig_file(mac)
...
+      file_path = Pathname.new file
+      extr_file_path = Pathname.new(File.join(extr_image_dir, file_path)).cleanpath
+
+      # Create destination directory
+      FileUtils.mkdir_p extr_file_path.parent
```
ekohl:
```suggestion
      extr_file_path.parent.mkpath
```

modules/tftp/tftp_api.rb
bastian-src: Change to 500 http error. Add another check on input parameters which raises http error 400.

modules/tftp/server.rb
bastian-src: Done.

modules/tftp/server.rb
bastian-src: Done.

modules/tftp/server.rb
bastian-src: Done.

modules/tftp/tftp_api.rb
bastian-src: Done.

modules/tftp/tftp_api.rb
bastian-src: Done.

modules/tftp/tftp.rb
bastian-src: Moved it directly to `tftp_plugin.rb`.

modules/tftp/tftp_plugin.rb
bastian-src: Done.

modules/tftp/server.rb
```diff
@@ -150,6 +150,36 @@ def pxeconfig_file(mac)
     end
   end
 
+  def self.fetch_boot_image(image_dst, url, files)
+    # Verify image_dst does not contain ".." (switch folder)
+    raise "TFTP image file extraction error: #{file_path}" unless image_dst.split("..").size <= 1
```
ekohl: Same note as https://github.com/theforeman/smart-proxy/pull/837/files#r977674171

modules/tftp/server.rb
bastian-src: Done, thanks for the hint!

modules/tftp/server.rb
bastian-src: Done.

modules/tftp/server.rb
bastian-src: Alright, thanks for this comment. It makes totally sense to have this as a core Smart Proxy feature to guarantee that it works independent of the environment (Katello installed or not).
Can you maybe elaborate a bit more on how the Smart Proxy itself can serve the files? And how is this connected to the state directory? I assume you don't want to run a web server offering the files in `/var/lib/foreman-proxy/`.

Moreover, if the directory where the files are hosted is a dynamic setting on the Smart Proxy, the corresponding address might be dynamic as well (not just `http://my.proxy/installation_media/..`, but depending on the Smart Proxy configuration). How can we sync this information to the Foreman side of the plugin? This is necessary because the boot template needs the full URL of the ISO image. 
So, is there any tool to sync Smart Proxy settings with the Foreman and make them accessible during runtime?

Thanks for all ur help on this btw. It's highly appreciated! :raised_hands:

ekohl: > Can you maybe elaborate a bit more on how the Smart Proxy itself can serve the files? And how is this connected to the state directory? I assume you don't want to run a web server offering the files in /var/lib/foreman-proxy/.

The Smart Proxy **is** a webserver. That's how it also serves httpboot files. See https://github.com/theforeman/smart-proxy/blob/develop/modules/httpboot/httpboot_api.rb for an example.

> Moreover, if the directory where the files are hosted is a dynamic setting on the Smart Proxy, the corresponding address might be dynamic as well (not just http://my.proxy/installation_media/.., but depending on the Smart Proxy configuration). How can we sync this information to the Foreman side of the plugin? This is necessary because the boot template needs the full URL of the ISO image.

I'd suggest that the HTTP call to download them has the resulting paths in the HTTP response. There's no way to know the filenames from an archive.

An interesting thought is idempotency and concurrency. What if the system tries to download the same ISO multiple times. Does it download it again or reuses? If it keeps files, how does it know if it's still up to date? This is also a problem we have with the downloaded kernel files so it may be fine to ignore for now and assume ISOs never change. Just want you to be aware of it.

>  So, is there any tool to sync Smart Proxy settings with the Foreman and make them accessible during runtime?

Settings can be exposed. Again, https://github.com/theforeman/smart-proxy/blob/develop/modules/httpboot/httpboot_plugin.rb is an interesting example.

bastian-src: Thanks for all these examples! I've mostly taken the additional API endpoint as it's implemented in `httpboot` and added the endpoint `/tftp/boot_image/`. You can see the "most important" diff [here](https://github.com/theforeman/smart-proxy/compare/2fa76995ae193158e6d59d3659661ecadb5c1c70..a0ee76d1fbf7fc7f2e2c9d6b8648e9f1e2dd3e9f).

I think it is important to mention that the plugin setting in `/etc/foreman-proxy/settings.d/tftp.yml` must be:
```
:enabled: true
...
```
Instead of just `https` since both protocols are used from now on.

I think idempotency and concurrency is an important topic here, since the ISO image download takes a while which increases the probability that another download process is triggered while the first one is not finished. As of now, the second process is just going to fail if it identifies that the ISO is currently being downloaded/extracted (have a look [here](https://github.com/theforeman/smart-proxy/pull/837/files#diff-db3f0d8d32c77f2719de93fd56902295e1d21a1a1c0dc1e19f398cf94d69426eR167)).
In terms of ISO image "version" (do we have the latest ISO downloaded?), there is currently no specific strategy. If an image has been downloaded, this one going to be used. A re-download can be "triggered" by deleting the image manually.

bastian-src:
@ekohl coming back to this. I've added tests for the `tftp_boot_image_api` with respect to the tests of `httpboot`. The tests run fine when executing them on-by-one. Unfortunately, my tests fail by checking the capabilites [here](https://github.com/theforeman/smart-proxy/pull/837/files#diff-32207ad0860bb9f050527059eaf4c0c2623f1f1be0f4ba3428249e2a52205d6cR14) when I want to execute all tests at once. I think the `which` function is not available in the testing context. Is there an approach to mock this for testing?

Moreover, I'd like to know whether we could also trigger the `bootimage_download` API endpoint within a Foreman job task. I think this might make it more easy to handle concurrency?

bastian-src:
@ekohl tests are green.

The `/fetch_system_image` API endpoint is not a dangling API call anymore. It triggers another thread which downloads the OS image file and extracts the boot files accordingly. Moreover, I tried to make the API endpoint concurrency safe by including a lock. The endpoint returns a status accordingly:
* 200: image and boot files exist already - no need to do anything
* 202: any of the files does not exist - process triggered
* 423: another download/extraction process is still running (resource locked)
* 500: any other error (input error for example)

What do you think about it? Can we continue with the foreman-side of this feature?

bastian-src:
@ekohl I've also added PRs for `foreman-installer` and `foreman-packaging` due to the new settings variable and package dependency.

bastian-src:
@ekohl do you know why there are no test results? Did I forgot anything, maybe?

lzap:
@bastian-src hey can you take a look at https://github.com/theforeman/smart-proxy/pull/958 if it solves your problem? I built it for Red Hat and Debian initially, but I am adding Ubuntu as well into the test shell script (should work but I want to validate it).

PR #844, opened 2022-08-31
Fixes #35468 - Verify boolean settings in modules

ekohl:
Since debffb85b60a5dfc2d5347075d8f2089573f289f it's possible to validate boolean settings. Validating settings makes it harder to misconfigure modules.


ehelms:
While I would say `settins` in my every day speech, I suppose it should read `settings` in the commit message.

ehelms (approved):

PR #846, opened 2022-09-09
Fixes #35510 - Depend on webrick for Ruby 3

ekohl:
This has always been a dependency, but Ruby 3 has dropped the code. It's now a regular gem.

It is set in the gemspec since that allows plugins to also benefit from it. That does mean that it's also required on older Ruby versions and should be packaged.


adamruzicka:
Will this need an accompanying packaging change?

ekohl:
Took me a while to get back to it, but I've opened https://github.com/theforeman/foreman-packaging/pull/8580

evgeni (commented):
> Took me a while to get back to it, but I've opened [theforeman/foreman-packaging#8580](https://github.com/theforeman/foreman-packaging/pull/8580)

Which is now merged, so ACK

evgeni (approved):

PR #847, opened 2022-09-14
Implement a verification API

ekohl:
The goal of this is to have an end-to-end verification of compatibility or provide concrete pointers to what failed.

### DIFF
--- a/lib/proxy/verify_runtime.rb
+++ b/lib/proxy/verify_runtime.rb
@@ -1,0 +1,52 @@
+module Proxy
+  class VerifyRuntime
+    class << self
+      # RFC5737 declares 192.0.2.0/24 as TEST-NET-1
+      MOCK_IP = '192.0.2.42'
+
+      def settings
+        Proxy::SETTINGS
+      end
+
+      def verify
+        {
+          reverse_proxy: verify_reverse_proxy,
+        }
+      end
+
+      def verify_reverse_proxy
+        # It's valid if there is no Foreman URL
+        return true unless settings.foreman_url
+
+        # Only needed for templates / registration
+        # TODO: make this more generic
+        return true unless ::Proxy::Plugins.instance.any? { |p| p[:state] == :running && ['templates', 'registration'].include?(p[:name]) }
+
+        foreman = Proxy::HttpRequest::ForemanRequest.new
+        request = foreman.request_factory.create_get('/api/status', headers: {'X-Forwarded-For': MOCK_IP})
+        response = foreman.send_request(request)
+
+        if response.status != '200'
+          logger.info("Foreman status API returned #{response.status}")
+          return false
+        end
+
+        status = JSON.parse(response.body)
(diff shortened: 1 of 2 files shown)


lib/proxy/verify_runtime.rb
```diff
@@ -0,0 +1,52 @@
...
+        response = foreman.send_request(request)
+
+        if response.status != '200'
+          logger.info("Foreman status API returned #{response.status}")
+          return false
+        end
+
+        status = JSON.parse(response.body)
+        unless status.key?('remote_ip')
+          message = if ::Gem::Dependency.new('', '>= 3.5.0').match?('', status['version'])
```
ekohl, line 36: Depends on https://github.com/theforeman/foreman/pull/9423

lib/proxy/verify_runtime.rb
```diff
@@ -0,0 +1,52 @@
...
+        }
+      end
+
+      def verify_reverse_proxy
+        # It's valid if there is no Foreman URL
+        return true unless settings.foreman_url
+
+        # Only needed for templates / registration
+        # TODO: make this more generic
+        return true unless ::Proxy::Plugins.instance.any? { |p| p[:state] == :running && ['templates', 'registration'].include?(p[:name]) }
```
ekohl, line 23: templates has a `template_url` setting while registration uses the global `foreman_url` setting.

lib/proxy/verify_runtime.rb
```diff
@@ -0,0 +1,52 @@
...
+
+      def settings
+        Proxy::SETTINGS
+      end
+
+      def verify
+        {
+          reverse_proxy: verify_reverse_proxy,
+        }
+      end
```
ekohl, line 15: I'm thinking about the format. Perhaps it should have some global status and then a hash for modules.
```json
{
  "global": {},
  "modules": {
    "templates": {
      "status": true,
      "reverse_proxy": true,
    },
    "registration": {
      "status": true,
    }
  }
}
```

Perhaps also a summary?

What if it doesn't? Should we use error codes? Human readable strings? both?

ehelms:
How do you see this being different than if we were to implement a status API which is a common endpoint?

ekohl:
I started with this because it was hard to verify if `request.remote_ip` inside Foreman was set correctly. That's how I came to `/verify`. Hadn't considered a status API.

I think a status API shouldn't reach out to external sources unless they're critical. Reaching out can be expensive.

That does make me think about more things. For example, if a plugin (or provider) has an external connection (think Puppet -> Puppetserver or isc_dhcp -> dhcpd) that should be verified. So this endpoint should be pluggable and the plugin API should provide some DSL to verify.

But this PR is the start of a discussion IMHO: what do we need to verify to say "this Smart Proxy works" and how can we provide hints if it doesn't.

We should also consider if this endpoint should be authenticated. Probably yes if it reaches out to external services. Otherwise it may be used for DoS attacks.

PR #850, opened 2022-10-19
Fixes #35639 - registration_url setting for Registration module

stejskalleos:
RFC: https://community.theforeman.org/t/rfc-host-registration-and-load-balancers/30462
Foreman: https://projects.theforeman.org/issues/35626
Katello: https://projects.theforeman.org/issues/35627

### DIFF
--- a/config/settings.d/registration.yml.example
+++ b/config/settings.d/registration.yml.example
@@ -1,2 +1,3 @@
 ---
 :enabled: false
+# :registration_url: https://loadbalancer.example.com
--- a/lib/proxy/plugin_validators.rb
+++ b/lib/proxy/plugin_validators.rb
@@ -57,6 +57,12 @@ module ::Proxy::PluginValidators
     end
   end
 
+  class OptionalUrl < Url
+    def validate!(settings)
+      settings[@setting_name].nil? || super
+    end
+  end
+
   class Boolean < Base
     def validate!(settings)
       setting_value = settings[@setting_name]
--- a/test/plugins/validator_test.rb
+++ b/test/plugins/validator_test.rb
@@ -134,6 +134,40 @@ class UrlValidatorTest < Test::Unit::TestCase
   end
 end
 
+class OptionalUrlValidatorTest < Test::Unit::TestCase
+  class OptionalUrlValidatorTestPlugin < ::Proxy::Plugin
+    default_settings url: 'http://example.com'
+  end
+
+  def validator
+    ::Proxy::PluginValidators::OptionalUrl.new(OptionalUrlValidatorTestPlugin, 'url', nil, nil)
+  end
+
+  def test_required_parameter_with_a_value_passes_validation
+    assert validator.validate!(url: 'http://example.com')
+  end
+
+  def test_empty_string_treated_as_missing_value
+    error = assert_raises ::Proxy::Error::ConfigurationError do
+      validator.validate!(url: '')
+    end
+
+    assert_match(/expected to contain a url/, error.message)
+  end
+
+  def test_without_a_value
+    assert validator.validate!(url: nil)
+  end
+
+  def test_required_parameter_without_scheme_fails_validation
+    error = assert_raises ::Proxy::Error::ConfigurationError do
+      validator.validate!(url: 'example.com')
+    end
+
+    assert_match(/missing a scheme/, error.message)
+  end
+end
+
 class BooleanValidatorTest < Test::Unit::TestCase
   class BooleanValidatorTestPlugin < ::Proxy::Plugin
     default_settings :a_settting => true
--- a/test/registration/integration_test.rb
+++ b/test/registration/integration_test.rb
@@ -22,6 +22,7 @@ class RegistrationApiFeaturesTest < Test::Unit::TestCase
 
     refute_nil(mod)
     assert_equal('running', mod['state'], Proxy::LogBuffer::Buffer.instance.info[:failed_modules][:registration])
+    assert_equal({'registration_url' => nil}, response['registration']['settings'])
   end
 
   def test_without_templates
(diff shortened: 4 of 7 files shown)


config/settings.d/registration.yml.example
```diff
@@ -1,2 +1,3 @@
 ---
 :enabled: false
+:registration_url: https://loadbalancer.example.com
```
ekohl: The example file is deployed on every fresh installation via packaging. I'd make it a comment.
```suggestion
# :registration_url: https://loadbalancer.example.com
```

ekohl (commented):
Minor comment inline.

I also wonder about normalization with or without a trailing slash. That can also happen on the Foreman side, but something to consider. It may be sufficient to suggest users enter it without a trailing slash. The current code suggests it's without a trailing slash.

stejskalleos:
> I also wonder about normalization with or without a trailing slash. 

Trailing slash will cause an issue, it will generate final URL like this: `https://load.example.com//register`

ehelms:
@stejskalleos Would you mind squashing the commits to make Redmine check happy?

ekohl:
> Trailing slash will cause an issue, it will generate final URL like this: `https://load.example.com//register`

That's something you should at least document, but it may be better to strip it. I'm pretty sure you can use `load_programmable_settings` to also modify existing settings (but be sure to verify it). Here's an example:
https://github.com/theforeman/smart-proxy/blob/6a9159f9d4d6ca1eb03ea71f0b139db008b88bd0/modules/httpboot/httpboot_plugin.rb#L7-L11

It's probably something like:
```ruby
load_programmable_settings do |settings|
  settings[:registration_url]&.chomp!('/')
  settings
end
```

test/registration/integration_test.rb
```diff
@@ -13,15 +13,17 @@ def app
   end
 
   def test_with_templates
+    registration_url = 'https://loadbalancer.example.com'
```
ekohl: Please make this a separate test. The setting should remain optional and that should be tested.

ekohl (changes requested):
I tested this out, but if you don't provide a registration URL it fails with:
> 2022-10-26T12:02:14  [W] Error details for Disabling all modules in the group ['registration'] due to a failure in one of them: Setting 'registration_url' is expected to contain a url: <Proxy::Error::ConfigurationError>: Setting 'registration_url' is expected to contain a url

I'm wondering what is the best course of action now. I'm leaning to introducing an `OptionalURL` validator. https://github.com/theforeman/smart-proxy/commit/685c026e30ade098f1748fb8c5a0dd16e993efec is a commit where a validator was added.

ekohl:
Thanks! Could you split this in two commits: first one that adds the `OptionalUrlValidator` (with `Refs #35639 - ...` as the commit message) and then another that's the actual `registration_url` change? Then I'll merge it.

stejskalleos:
> Thanks! Could you split this in two commits: first one that adds the `OptionalUrlValidator` (with `Refs #35639 - ...` as the commit message) and then another that's the actual `registration_url` change? Then I'll merge it.

Done

ekohl:
Could you order them the other way around? Otherwise there's a point in history where the software is in an invalid state: d63f5dad27215abb40a8045fd5b6882eff302111 depends on 45bf25e9d8c66b5e747ed446a26c362cfb060099, not the other way around. It's not often that I run `git bisect`, but once you do, you want your history to be correct.

ekohl (approved):
Thanks!

lib/proxy/plugin_validators.rb
```diff
@@ -57,6 +57,21 @@ def validate!(settings)
...
+      return true if setting_value.nil?
+
+      parsed = URI.parse(setting_value)
+      raise ::Proxy::Error::ConfigurationError, "Setting '#{@setting_name}' is missing a scheme" if parsed.scheme.nil? || parsed.scheme.empty?
+
+      true
+    rescue URI::InvalidURIError
+      raise ::Proxy::Error::ConfigurationError.new("Setting '#{@setting_name}' contains an invalid url")
+    end
+  end
```
ekohl: I'd suggest this should inherit from `Url`
```suggestion
  class OptionalUrl < Url
    def validate!(settings)
      settings[@setting_name].nil? || super
    end
  end
```

test/plugins/validator_test.rb
```diff
@@ -134,6 +134,44 @@ def test_optional_parameter_without_a_value_fails_validation
...
+    assert_match(/expected to contain a url/, error.message)
+  end
+
+  def test_without_a_value_fails_validation
+    error = assert_raises ::Proxy::Error::ConfigurationError do
+      validator.validate!(url: nil)
+    end
+
+    assert_match(/expected to contain a url/, error.message)
+  end
```
ekohl: I'd have expected this to actually verify nil was accepted somewhere. If this fails validation, it doesn't sound like it's working.

test/registration/integration_test.rb
```diff
@@ -34,4 +34,16 @@ def test_without_templates
...
+  def test_chomp_trailing_slash
+    registration_url = 'https://loadbalancer.example.com'
+
+    Proxy::LegacyModuleLoader.any_instance.expects(:load_configuration_file).with('templates.yml').returns(enabled: true, template_url: 'http://smart-proxy.example.com:8000')
+    Proxy::DefaultModuleLoader.any_instance.expects(:load_configuration_file).with('registration.yml').returns(enabled: true, registration_url: "#{registration_url}/")
+
+    get '/features'
+    response = JSON.parse(last_response.body)
+
+    assert_equal({'registration_url' => registration_url}, response['registration']['settings'])
```
ekohl, line 48: Since you're verifying the exposed settings here, can you verify it in `test_with_templates` too?

ekohl (changes requested):
Sorry, bit hasty approval.

stejskalleos:
@ekohl Rebased & updated

ekohl (approved):
In the interest of speed I took the liberty to squash the changes to the validator in the validator commit. This leads to a better commit history where you don't introduce changes in one commit and then modify it in the next.

PR #851, opened 2022-10-24
Update rubocop to 0.89

evgeni:


evgeni:
I was lazy, and didn't actually enable any of the new cops *whistle*

ehelms (approved):
Works for me as an initial update

PR #854, opened 2023-01-11
Fixes #35921 - Unify use of AccessLog

adamruzicka:
across http and https applications

### DIFF
--- a/lib/launcher.rb
+++ b/lib/launcher.rb
@@ -48,7 +48,6 @@ module Proxy
         :app => app,
         :Port => http_port, # only being used to correctly log http port being used
         :Logger => ::Proxy::LogBuffer::TraceDecorator.instance,
-        :AccessLog => [],
       }
       base_app_settings.merge(http_settings)
     end
@@ -180,6 +179,7 @@ module Proxy
         :DoNotListen => true,
         :ServerSoftware => "foreman-proxy/#{Proxy::VERSION}",
         :daemonize => false,
+        :AccessLog => [],
       }
     end
   end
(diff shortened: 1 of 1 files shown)


adamruzicka:
Should we unify on on access log being turned off or on? Alternatively we could do it conditionally based on logging configuration

lib/launcher.rb
```diff
@@ -106,6 +106,7 @@ def https_app(https_port, plugins = https_plugins)
         :SSLOptions => ssl_options,
         :SSLCiphers => CIPHERS - Proxy::SETTINGS.ssl_disabled_ciphers,
         :daemonize => false,
+        :AccessLog => [],
```
ekohl: It took me a moment to understand how this was to unify it. Would it make sense to have a `base_app_settings` that returns all common options between HTTP and HTTPS? Things like `Logger`, `ServerSoftware` etc should be there and then merged with HTTP(S) specific options (`Port`, SSL settings).

adamruzicka: I wanted to make the changes as small as possible, but sure, done

ekohl (commented):
If you want to make it clear, you can introduce a commit that introduces `base_app_settings` without `AccessLog` (as `Refs #35921`) and then one that sets it in base_app_settings.

I also wonder about a clear description for those who don't know what `AccessLog` does. Perhaps it should say something like "Prior to this change http access logs weren't logged while https logs were. After this, neither logs access logs." At least, that's if I properly understood the change. Did I?

adamruzicka:
Ping

ekohl (approved):
Thanks!

PR #856, opened 2023-02-23
Fixes #36138 - Invalid syntax for curl --time-cond

stejskalleos:
Fixes: 3d87c6f
According to Curl man page, syntax for `--time-cond` attribute should be `--time-cond file <path>`, not `--time-cond <path>`.

This can cause issues with outdated vmlinuz and initrd.img files,
once the files are downloaded, they are never replaced with never versions.

**How to test**
* Provision host (baremetal, networking)
* `touch -a -m -t 201512180130.09 initrd.img & vmlinuz` in `/var/lib/tftpboot/boot`
* Provision another host with same OS
* Check that files have been updated

[0] https://curl.se/docs/manpage.html

### DIFF
--- a/lib/proxy/http_download.rb
+++ b/lib/proxy/http_download.rb
@@ -26,7 +26,7 @@ module Proxy
       # keep last changed file attribute
       args << "--remote-time"
       # only download newer files
-      args += ["--time-cond", dst.to_s]
+      args += ["--time-cond", "file", dst.to_s]
       # print stats in the end
       args += [
         "--write-out",
--- a/test/http_download_test.rb
+++ b/test/http_download_test.rb
@@ -17,7 +17,7 @@ class HttpDownloadsTest < Test::Unit::TestCase
       "--retry-delay", "10",
       "--max-time", "3600",
       "--remote-time",
-      "--time-cond", "dst",
+      "--time-cond", "file", "dst",
       "--write-out", "Task done, result: %{http_code}, size downloaded: %{size_download}b, speed: %{speed_download}b/s, time: %{time_total}ms",
       "--output", "dst",
       "--location", "src"
@@ -34,7 +34,7 @@ class HttpDownloadsTest < Test::Unit::TestCase
       "--retry-delay", "10",
       "--max-time", "3600",
       "--remote-time",
-      "--time-cond", "dst",
+      "--time-cond", "file", "dst",
       "--write-out", "Task done, result: %{http_code}, size downloaded: %{size_download}b, speed: %{speed_download}b/s, time: %{time_total}ms",
       "--output", "dst",
       "--location", "src"


stejskalleos:
And I guess we should CP it to 3.5 stable too.

lib/proxy/http_download.rb
```diff
@@ -26,7 +26,7 @@ def initialize(src, dst, connect_timeout: DEFAULT_CONNECT_TIMEOUT, verify_server
       # keep last changed file attribute
       args << "--remote-time"
       # only download newer files
-      args += ["--time-cond", dst.to_s]
+      args += ["--time-cond", "file #{dst}"]
```
ekohl: Looking at the man page I wonder if it really is a single argument, or should be two. The manpage has these examples:
```
Examples:
 curl -z "Wed 01 Sep 2021 12:18:00" https://example.com
 curl -z "-Wed 01 Sep 2021 12:18:00" https://example.com
 curl -z file https://example.com
```

When I run it manually to simulate your version:
```console
$ curl -z "file initrd.img" https://mirror.stream.centos.org/9-stream/BaseOS/x86_64/os/isolinux/initrd.img -o initrd.img 
Warning: Illegal date format for -z, --time-cond (and not a file name). 
Warning: Disabling time condition. See curl_getdate(3) for valid date syntax.
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 95.3M  100 95.3M    0     0  90.7M      0  0:00:01  0:00:01 --:--:-- 90.8M
```

Which makes me think it should be:
```suggestion
      args += ["--time-cond", "file", dst.to_s]
```

ekohl (changes requested):
Thanks for providing so much context. Made it really easy to review.

> And I guess we should CP it to 3.5 stable too.

I always find out in which commit it was introduced. `git blame -L 29,29 lib/proxy/http_download.rb` quickly tells me it was last modified in 3d87c6feaa8caaf9b8c040f96061c3b8c33a39ef. Always verify it wasn't some patch that just changed formatting, but in this case it's obvious.

Then I add `Fixes: $COMMIT` to the commit message, which I saw in the Linux kernel. While looking for a link to their description I see they also include a description. https://www.kernel.org/doc/html/latest/process/submitting-patches.html#using-reported-by-tested-by-reviewed-by-suggested-by-and-fixes & https://www.kernel.org/doc/html/latest/process/submitting-patches.html#describe-changes have more.

So in this case it's 3d87c6feaa8caaf9b8c040f96061c3b8c33a39ef which introduced the bug, so then I check where it's included:
```console
$ git tag --contains 3d87c6feaa8caaf9b8c040f96061c3b8c33a39ef
3.1.0
3.1.0-rc1
3.1.0-rc2
3.1.1
3.1.1.1
3.1.1.2
3.1.1.3
3.1.2
3.1.3
3.2.0
3.2.0-rc1
3.2.0-rc2
3.2.1
3.3.0
3.3.0-rc1
3.3.0-rc2
3.3.1
3.4.0
3.4.0-rc1
3.4.0-rc2
3.4.1
3.5.0
3.5.0-rc1
3.5.0-rc2
3.5.1
```

So this certainly needs to be picked to 3.5-stable, but also 3.4-stable and since we branched 3.6-stable yesterday that too.

ekohl (commented):
Thanks! Small nit: it's `Fixes: COMMIT` instead of `Fixes COMMIT`. https://git-scm.com/docs/git-interpret-trailers isn't very clear, but the core is that they're supposed be parseable. A more concrete example is 22282fc7483f7a2439cc69c57d361860ea221fc9, which you can then parse:
```console
$ git show 22282fc7483f7a2439cc69c57d361860ea221fc9 --format="%(trailers)" --quiet
Fixes: aa25357d4404908d5e5707e767ecd798b754d7d1

```

We don't have it today, but this could then be used in automation to determine cherry picks. That's still on my wish list.

I believe it is also needed to put it at the end, separated by a newline from the main body.

Ron-Lavi:
@ekohl can we merge this?

ekohl (approved):
Thanks!

PR #860, opened 2023-03-16
Update RuboCop to 1.56.x

archanaserver:
This PR updates the RuboCop version from 1.28.0 to 1.56.0.

#### Requires:

The following PRs were reviewed and adjusted to ensure compatibility with the updated RuboCop/Ruby version:

- [x] https://github.com/theforeman/smart_proxy_omaha/pull/24
  - [x] https://github.com/theforeman/smart_proxy_omaha/pull/25
- [x] https://github.com/theforeman/smart_proxy_dynflow/pull/123
  - [x] https://github.com/theforeman/smart_proxy_dynflow/pull/122
  - [x] https://github.com/theforeman/smart_proxy_dynflow/pull/124
- [x] https://github.com/theforeman/smart_proxy_remote_execution_ssh/pull/101
- [x] https://github.com/theforeman/smart_proxy_salt/pull/83
- [x] https://github.com/theforeman/smart_proxy_monitoring/pull/32
- [x] https://github.com/theforeman/jenkins-jobs/pull/317
   - [x] https://github.com/theforeman/jenkins-jobs/pull/324
     - [x] https://github.com/theforeman/jenkins-jobs/pull/326

For the rest of the smart-proxy plugins (`ansible`, `openscap`, `dhcp_infoblox`, `dhcp_remote_isc`, `discovery`, `dns_infoblox`, `pulp`), no changes were required to support Ruby v3.


ekohl:
FYI: there is https://projects.theforeman.org/issues/35374 open as an issue. There I chose version 1.35.x. 1.28.x is the latest version to support Ruby 2.5, but if we move to 2.7+ I think it's safe to go with a newer version. At the time that was 1.35.x, but it's probably even safe to move to 1.48.x which is the current latest.

ehelms:
@archanaserver When you get time, can you come back around to this and do the update to 1.35 (or if there is a newer latest) ?

bundler.d/test.rb
```diff
@@ -4,7 +4,7 @@
   gem 'test-unit'
   gem 'benchmark-ips'
   gem 'ruby-prof', '< 1.4'
-  gem 'rubocop', '~> 0.89.0'
+  gem 'rubocop', '~> 1.28.0'
```
ekohl: We can update to the latest version:
```suggestion
  gem 'rubocop', '~> 1.55.0'
```

archanaserver:
@ekohl, could you please review this?

.rubocop.yml
```diff
@@ -137,3 +137,6 @@ Style/TrailingCommaInHashLiteral:
 #Allow both ['a', 'b'], %w[a b] and %w(a b) style arrays
 Style/WordArray:
   Enabled: false
+
+Style/RedundantCondition:
+  Enabled: false
```
ekohl: I would not disable this cop just for one thing that is hard to rewrite. Instead I'd fix the particular case. You can write:
```ruby
type = params.fetch(:type) { name.end_with?('.in-addr.arpa', '.ip6.arpa') ? 'PTR' : 'A' }
```

ekohl (changes requested):
I'd also enable `rubocop-rake` as a plugin.

Gemfile
```diff
@@ -3,6 +3,7 @@ source 'https://rubygems.org'
 gemspec
 
 gem 'concurrent-ruby', '~> 1.0', require: 'concurrent'
+gem 'rubocop-rake'
```
ekohl: Please keep it in the right place: If you place it here then it will also be required in production which isn't true for `bundler.d/test.rb`.

tasks/jenkins.rake
```diff
@@ -3,17 +3,24 @@ require 'English'
...
-      task :minitest  => [:pre_ci, "ci:setup:minitest"]
+
+      desc 'Sets up pre-CI environment for MiniTest'
+      task :minitest => [:pre_ci, "ci:setup:minitest"]
+
+      desc 'Sets up pre-CI environment for Test::Unit'
       task :test_unit => [:pre_ci, "ci:setup:testunit"]
     end
 
+    desc 'Run RuboCop analysis'
```
ekohl: We no longer use this task so it's better to drop it. This is how we really invoke it now:

https://github.com/theforeman/jenkins-jobs/blob/a44d192c1549be46b3f0d85c254f4247c4ec9331/theforeman.org/pipelines/test/smart-proxy.groovy#L47

Note that if you drop this task, you no longer need the `require 'English'` at the top.

tasks/jenkins.rake
```diff
@@ -3,17 +3,24 @@ require 'English'
...
     namespace :setup do
+      desc 'Sets up pre-CI environment for Minitest'
       task :pre_ci do
         ENV["CI_REPORTS"] = 'jenkins/reports/unit/'
         gem 'ci_reporter'
       end
-      task :minitest  => [:pre_ci, "ci:setup:minitest"]
+
+      desc 'Sets up pre-CI environment for MiniTest'
+      task :minitest => [:pre_ci, "ci:setup:minitest"]
```
ekohl: This repository doesn't use minitest so we can drop this. It's a left over from when we copied it over to multiple repositories.

tasks/jenkins.rake
```diff
@@ -3,17 +3,24 @@ require 'English'
 begin
   require "ci/reporter/rake/test_unit"
   namespace :jenkins do
+    desc 'Sets up pre-CI environment for testing'
     task :unit => ["jenkins:setup:test_unit", 'rake:test']
 
     namespace :setup do
+      desc 'Sets up pre-CI environment for Minitest'
       task :pre_ci do
         ENV["CI_REPORTS"] = 'jenkins/reports/unit/'
```
ekohl: While you're at it, I think you can actually drop this. We now set it in CI: https://github.com/theforeman/jenkins-jobs/blob/a44d192c1549be46b3f0d85c254f4247c4ec9331/theforeman.org/pipelines/test/smart-proxy.groovy#L97

tasks/jenkins.rake
```diff
@@ -3,17 +3,24 @@ require 'English'
...
         gem 'ci_reporter'
```
ekohl: I wonder if this is still needed. We have an explicit require on line 4 already. I think the whole `pre_ci` task can be dropped.

tasks/jenkins.rake
```diff
@@ -3,17 +3,24 @@ require 'English'
 begin
   require "ci/reporter/rake/test_unit"
   namespace :jenkins do
+    desc 'Sets up pre-CI environment for testing'
     task :unit => ["jenkins:setup:test_unit", 'rake:test']
```
ekohl: You can probably drop most tasks and reduce this to
```suggestion
    task :unit => ['ci:setup:testunit', 'rake:test']
```

tasks/jenkins.rake
```diff
@@ -3,17 +3,24 @@ require 'English'
...
         ENV["CI_REPORTS"] = 'jenkins/reports/unit/'
         gem 'ci_reporter'
       end
-      task :minitest  => [:pre_ci, "ci:setup:minitest"]
+
+      desc 'Sets up pre-CI environment for MiniTest'
+      task :minitest => [:pre_ci, "ci:setup:minitest"]
+
+      desc 'Sets up pre-CI environment for Test::Unit'
       task :test_unit => [:pre_ci, "ci:setup:testunit"]
```
ekohl: If you're dropping `pre_ci`, this also becomes redundant.

tasks/jenkins.rake
```diff
@@ -3,17 +3,24 @@ require 'English'
 begin
   require "ci/reporter/rake/test_unit"
   namespace :jenkins do
+    desc 'Sets up pre-CI environment for testing'
```
ekohl: This doesn't just set it up: it requires both the setup and then runs the tests.

.rubocop_todo.yml
```diff
@@ -1,280 +1,551 @@
 # This configuration was generated by
-# `rubocop --auto-gen-config --exclude-limit 0`
-# on 2020-02-26 21:20:38 +0100 using RuboCop version 0.80.0.
+# `rubocop --auto-gen-config --no-offense-counts --no-auto-gen-timestamp`
+# using RuboCop version 1.28.2.
```
ekohl: Looks like you didn't update this with 1.55.x.

bundler.d/test.rb
```diff
@@ -4,7 +4,7 @@
   gem 'test-unit'
   gem 'benchmark-ips'
   gem 'ruby-prof', '< 1.4'
-  gem 'rubocop', '~> 0.89.0'
+  gem 'rubocop', '~> 1.55.0'
```
ekohl: By now 1.56 is also out.
```suggestion
  gem 'rubocop', '~> 1.56.0'
```

ekohl (changes requested):

archanaserver:
@ekohl the changes is updated now, as per the comments. Thanks for the review!

bundler.d/test.rb
```diff
@@ -4,11 +4,12 @@
   gem 'test-unit'
   gem 'benchmark-ips'
   gem 'ruby-prof', '< 1.4'
-  gem 'rubocop', '~> 0.89.0'
+  gem 'rubocop', '~> 1.56.0'
   gem 'rubocop-performance', '~> 1.5.2'
   gem 'rack-test'
   gem 'rake'
   gem 'rubocop-checkstyle_formatter', '~> 0.2'
```
ekohl: We actually no longer use this so it can be dropped
```suggestion
```

tasks/jenkins.rake
```diff
@@ -1,28 +1,7 @@
-require 'English'
-
 begin
```
ekohl: If you don't catch the `LoadError` you don't need this `begin` and `end`.

ekohl (changes requested):

archanaserver:
@ekohl I looked into the failure but can't find the issue, can you help what is the issue here? https://ci.theforeman.org/blue/organizations/jenkins/smart-proxy-pr-test/detail/smart-proxy-pr-test/275/pipeline

modules/dns/dns_api.rb
```diff
@@ -47,11 +47,7 @@ class Api < ::Sinatra::Base
 
     delete '/:value/?:type?' do
       name = params[:value]
-      if params[:type]
-        type = params[:type]
-      else
-        type = (name =~ /\.(in-addr|ip6)\.arpa$/) ? "PTR" : "A"
-      end
+      type = params.fetch(:type) { name.end_with?('.in-addr.arpa', '.ip6.arpa') ? 'PTR' : 'A' }
```
ekohl: This fails because `params` does contain the key `type` but has the value `nil`. `fetch` only works if the key isn't present. So this isn't going to work.

A simpler way may be
```suggestion
      type = params[:type]
      type ||= name.end_with?('.in-addr.arpa', '.ip6.arpa') ? 'PTR' : 'A'
```

bundler.d/test.rb
```diff
@@ -4,11 +4,11 @@
   gem 'test-unit'
   gem 'benchmark-ips'
   gem 'ruby-prof', '< 1.4'
-  gem 'rubocop', '~> 0.89.0'
+  gem 'rubocop', '~> 1.56.0'
   gem 'rubocop-performance', '~> 1.5.2'
   gem 'rack-test'
   gem 'rake'
-  gem 'rubocop-checkstyle_formatter', '~> 0.2'
+  gem 'rubocop-rake'
```
ekohl: I know it isn't completely sorted, but I'd at least make sure the rubocop dependencies are sorted.

ekohl (approved):

PR #861, opened 2023-03-22
Fixes #36209 - tftp initrd/vmlinux generation: curl malformed

stejskalleos:
This reverts commit fba49ab1eb9bbe383572575309558ff628f5de60.

https://github.com/theforeman/smart-proxy/pull/856 was fixing invalid curl `--time-cont` condition, but it actually break it and now files are not downloaded at all.

I'm still looking into the details what happened and why I didn't catch it during development and reviews.
Revert tested on Fedora 37 & RHEL 8 and it works correctly as before did.

https://community.theforeman.org/t/curl-url-malformed-during-initrd-and-vmlinux-tftp-copy/32850/4

### DIFF
--- a/lib/proxy/http_download.rb
+++ b/lib/proxy/http_download.rb
@@ -26,7 +26,7 @@ module Proxy
       # keep last changed file attribute
       args << "--remote-time"
       # only download newer files
-      args += ["--time-cond", "file", dst.to_s]
+      args += ["--time-cond", dst.to_s]
       # print stats in the end
       args += [
         "--write-out",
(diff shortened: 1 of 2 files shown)


lib/proxy/http_download.rb
```diff
@@ -26,7 +26,7 @@ def initialize(src, dst, connect_timeout: DEFAULT_CONNECT_TIMEOUT, verify_server
       # keep last changed file attribute
       args << "--remote-time"
       # only download newer files
-      args += ["--time-cond", "file", dst.to_s]
+      args += ["--time-cond", dst.to_s]
```
ekohl, line 29: Looking at how it actually runs, perhaps this should be guarded with `if File.exist?(dst.to_s)` to avoid generating a warning.

ekohl (approved):
Kind of weird I didn't spot it either, because I certainly read the docs but also misinterpreted them.

Let me know if you want to address my comment, or merge it as is now.

PR #864, opened 2023-03-29
Templates - Return Foreman's API status code instead of 500

stejskalleos:


Ron-Lavi (approved):
Thanks @stejskalleos ! code wise looks good to me,
but didn't test it yet.

Ron-Lavi:
Our QE tested it and confirmed it fixes the issue, merging!

lib/proxy/helpers.rb
```diff
@@ -20,7 +20,7 @@ def log_halt(code = nil, exception_or_msg = nil, custom_msg = nil)
     rescue => e
       exception = e
       message += e.message
-      code ||= 400
+      code ||= e.status_code
```
ekohl, line 23: Please revert this. It's wrong to change the API this way since not every exception has a status code, which in turn will cause internal server errors:
```ruby
[1] pry(main)> IndexError.status_code
NoMethodError: undefined method `status_code' for IndexError:Class
from (pry):2:in `__pry__'
```

modules/templates/templates_unattended_api.rb
```diff
@@ -13,19 +13,19 @@ class Proxy::TemplatesUnattendedApi < Sinatra::Base
   end
 
   get "/:kind/:template/:hostgroup" do |kind, template, hostgroup|
-    log_halt(500, "Failed to retrieve #{kind} hostgroup template for #{params.inspect}: ") do
+    log_halt(nil, "Failed to retrieve #{kind} hostgroup template for #{params.inspect}: ") do
```
ekohl, line 16: If you don't want it to halt, then don't use `log_halt`. Instead, implement your own exception handling.

ekohl (commented):
I think I'm fine with the idea, but not on the implementation. Changing `log_halt` this was it not ok.

Instead, if you don't want to to halt, then don't.

Additionally, you haven't considered HTTP redirects. Will those actually make sense? Probably not.

Ron-Lavi:
Sorry about that, probably I was too happy this actually fixed the issue,
reverting PR was opened in #868

lib/proxy/helpers.rb
stejskalleos: Fixed in https://github.com/theforeman/smart-proxy/pull/869

modules/templates/templates_unattended_api.rb
stejskalleos: I want to use log_halt, I just want to set my own status code, checking the `log_halt` `code` is optional:
https://github.com/theforeman/smart-proxy/blob/70073c07e991cbe85a332a00e90d4196d8cbcaaf/lib/proxy/helpers.rb#L10

stejskalleos:
> Additionally, you haven't considered HTTP redirects. Will those actually make sense? Probably not.

Good point, I didn't consider them TBH.
Good thing is that `template_url` is set by users so it can be easily changed and we didn't care about redirects before, if I'm not mistaken.

stejskalleos:
Follow up PR: https://github.com/theforeman/smart-proxy/pull/869

PR #865, opened 2023-03-29
Fix typo / fix link markup

bittner:
Makes a MarkDown link in the README work as, probably, originally planned.


ekohl (approved):
Thanks!

PR #871, opened 2023-04-24
Fixes #36325 - Support key-algorithm in omshell

ekohl:
EL 8.2 [introduced support](https://access.redhat.com/errata/RHBA-2021:1623) for specifying key-algorithm in omshell and Debian stable also supports this. If unspecified, it defaults to the insecure HMAC-MD5. Especially on FIPS (where MD5 is forbidden) this is problematic.

### DIFF
--- a/modules/dhcp_common/isc/omapi_provider.rb
+++ b/modules/dhcp_common/isc/omapi_provider.rb
@@ -6,11 +6,12 @@ module Proxy::DHCP::CommonISC
     include Proxy::Util
     attr_reader :omapi_port, :key_name, :key_secret
 
-    def initialize(server, omapi_port, subnets = nil, key_name = nil, key_secret = nil, service = nil, free_ips_service = nil)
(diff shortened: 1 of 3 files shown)


modules/dhcp_common/isc/omapi_provider.rb
```diff
@@ -6,11 +6,12 @@ class IscOmapiProvider < ::Proxy::DHCP::Server
     include Proxy::Util
     attr_reader :omapi_port, :key_name, :key_secret
 
-    def initialize(server, omapi_port, subnets = nil, key_name = nil, key_secret = nil, service = nil, free_ips_service = nil)
+    def initialize(server, omapi_port, subnets = nil, key_name = nil, key_secret = nil, service = nil, free_ips_service = nil, key_algorithm = nil)
```
ekohl, line 9: We should really refactor this to keyword arguments, but I wanted to stay compatible.

evgeni (approved):

PR #876, opened 2023-09-26
Enable new cops by default & fix many cops

ekohl:


extra/migrate_settings.rb
```diff
@@ -91,7 +91,7 @@ def create_migration_dirs
     end
 
     def underscore(src)
-      src = src.gsub(/::/, '/')
+      src = src.gsub("::", '/')
```
evgeni, line 94: would `'::'` also work? (I would have expected :police_officer: to yell at you)

bin/smart-proxy-dhcp-isc
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env ruby
```
evgeni: why was this file added and how does it fix the `Style/RedundantConstantBase` cop?

not that I mind adding it, just puzzled how it fits into this PR

ekohl: The wonders of `git add` with some local file that you didn't intend to.

extra/migrate_settings.rb
ekohl: In this repo that cop is disabled. I'm not sure why RuboCop would prefer `"`.

evgeni: Ah, very well, then something for the future.

PR #878, opened 2023-11-02
Fixes #36899 - Add Dependabot Configuration

archanaserver:
This PR adds the Dependabot configuration for automated dependency management. Dependabot will help us keep smart-proxy bundler dependencies up to date.

The configuration is located in `.dependabot/config.yml`.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: 'bundler'
+    directory: '/'
+    schedule:
+      interval: 'daily'
--- a/Gemfile
+++ b/Gemfile
@@ -5,5 +5,5 @@ gemspec
 gem 'concurrent-ruby', '~> 1.0', require: 'concurrent'
 
 Dir[File.join(__dir__, 'bundler.d', '*.rb')].each do |bundle|
-  instance_eval(Bundler.read_file(bundle))
+  eval_gemfile(bundle)
 end


archanaserver:
@ekohl Do you want me to add any specific config here?

.github/dependabot.yml
```diff
@@ -0,0 +1,7 @@
+version: 2
+updates:
+  - package-ecosystem: 'bundler'
+    directory: '/'
+    schedule:
+      interval: 'daily'
+    open-pull-requests-limit: 10
```
evgeni: is that necessary? we don't set it in the other repos and so far didn't have any PR floods.

evgeni (approved):
I wonder if this will work with our
https://github.com/theforeman/smart-proxy/blob/54d4f0f10866db9c1fa630c79c774223f1b55119/Gemfile#L7-L9

But Bundler has an [`eval_gemfile` function since 1.3](https://github.com/rubygems/bundler/commit/c6100f6bd246a8259f159285967f075b592e80c9) which we could use instead, and that is (according to tests in https://github.com/dependabot/dependabot-core/tree/main/bundler/spec/fixtures/projects/bundler2) supported by dependabot.

Let's try this!

.github/dependabot.yml
ekohl: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#open-pull-requests-limit says the default limit is 5. I don't think we need to increase that to 10 so :+1: for removing this part.

Gemfile
```diff
@@ -5,5 +5,5 @@ gemspec
 gem 'concurrent-ruby', '~> 1.0', require: 'concurrent'
 
 Dir[File.join(__dir__, 'bundler.d', '*.rb')].each do |bundle|
-  instance_eval(Bundler.read_file(bundle))
+  Bundler.eval_gemfile(bundle)
```
ekohl:
```
[!] There was an error parsing `Gemfile`: undefined method `eval_gemfile' for Bundler:Module. Bundler cannot continue.
```

You should be able to reproduce this locally with `bundle install`. This works for me
```suggestion
  eval_gemfile(bundle)
```

ekohl (changes requested):

Gemfile
archanaserver: Yes, Thank you.

ekohl (approved):
Let's try this.

PR #883, opened 2024-01-05
add matrix.json specifying our Ruby versions

evgeni:


.github/matrix.json
```diff
@@ -0,0 +1,3 @@
+{
+  "ruby": ["2.7", "3.0"]
```
evgeni, line 2: jenkins also runs 3.1, should we add that?

ehelms: Yes.

ekohl (changes requested):
I'd like to derive this from a place we already have, the gemspec:
https://github.com/theforeman/smart-proxy/blob/e8fe59f83843ca619264e8d4757bc99fcde93fa9/smart_proxy.gemspec#L15

It may be harder to reuse it in plugins, but it avoids duplication. Remember that the Smart Proxy can actually be a real gem.

ekohl:
I'd still be in favor of using our [gem testing workflow](https://github.com/theforeman/actions?tab=readme-ov-file#gem-test). It's simpler and testing of Smart Proxy & plugins is usually fast anyway. We'll automatically test on Ruby 3.2 as well, which allows us to move forward faster. I just need to update https://github.com/ekohl/ruby-version/ to also test Ruby 3.3 (which helps with EL 10).

Shall we close this?

.github/matrix.json
evgeni: given we decided this file only indicates "minimal tested versions", I think 3.1 doesn't (yet) belong in here

ekohl (approved):

PR #884, opened 2024-01-05
Use RUBY_PLATFORM to decide between rb-inotify and rb-kqueue

ekohl:
This matches what modules/dhcp_isc/configuration_loader.rb does in load_programmable_settings.

Currently untested.


ekohl:
Note that `journald.rb` probably deserves a similar treatment in making it Linux-only. `windows.rb` may also be in the same boat.

ekohl:
Looks like the last commit doesn't work on older Bundler because the `:windows` platform was introduced in Ruby 3.1.

PR #885, opened 2024-02-06
Fixes #37147 - Pass --fail option to curl

ekohl:
This prevents curl from writing out error pages, for example when the URL returns a HTTP 404.

It still doesn't return any error to Foreman since the TFTP /fetch_boot_file only starts a thread and there's no way to monitor it.

Fixes: 3d87c6feaa8c ("Refs #2412 - use curl for downloads")

### DIFF
--- a/test/http_download_test.rb
+++ b/test/http_download_test.rb
@@ -12,6 +12,7 @@ class HttpDownloadsTest < Test::Unit::TestCase
       "/usr/bin/curl",
       "--insecure",
       "--silent", "--show-error",
+      '--fail',
       "--connect-timeout", "10",
       "--retry", "3",
       "--retry-delay", "10",
@@ -29,6 +30,7 @@ class HttpDownloadsTest < Test::Unit::TestCase
     expected = [
       "/usr/bin/curl",
       "--silent", "--show-error",
+      '--fail',
       "--connect-timeout", "10",
       "--retry", "3",
       "--retry-delay", "10",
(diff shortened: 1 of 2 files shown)


test/http_download_test.rb
```diff
@@ -42,6 +44,12 @@ def test_regular_call_with_ssl_verify
     assert_equal expected, Proxy::HttpDownload.new('src', 'dst', verify_server_cert: true).command
   end
 
+  def test_retrieving_404
+    thread = Proxy::HttpDownload.new('http://example.com/404', '/tmp/404', verify_server_cert: true).start
```
ekohl: This reaches out to the internet, which is a bad idea. It would be better to start some HTTP server locally and use that.

lzap:
LGTM

Another improvement would be downloading files into temporary files and then renaming them after successful download atomically.

ekohl:
@evgeni mentioned he recalled we tried this in the past and caused problems. I don't recall, but perhaps I missed something.

evgeni:
> @evgeni mentioned he recalled we tried this in the past and caused problems. I don't recall, but perhaps I missed something.

Nah, I dug my memory, and it actually was the time condition stuff that we "recently" broke. LGTM!

test/http_download_test.rb
```diff
@@ -49,3 +51,22 @@ def test_should_skip_download_if_one_is_in_progress
     end
   end
 end
+
+class HttpDownloadsIntegrationTest < Test::Unit::TestCase
+  include Proxy::IntegrationTestCase
```
ekohl: Thinking more about this: this launches a full Smart Proxy but really we need a (minimal) webserver. I'm tempted to only launch a minimal webrick server so curl can work. Then we can also add some more integration tests like verifying that HTTP 200 works.

ekohl:
I've rewritten the tests to use plain webrick and that's MUCH faster. I also asked it to listen on a random port instead of the current code which manually implements port randomization.

ekohl:
@evgeni @lzap is the LGTM an approval?

evgeni (approved):

PR #886, opened 2024-02-12
Wait for TFTP downloads

ekohl:
The /fetch_boot_file API is used to download TFTP files. The HttpDownload task is implemented as a thread. Previously it was fired off and there was no way to see the result. This implements code that waits for some time before telling the client the task is still running. There is no way for the client to find out if it really did finish.

Currently untested, but this at least shows the principle.

### DIFF
--- a/modules/tftp/tftp_api.rb
+++ b/modules/tftp/tftp_api.rb
@@ -35,7 +35,16 @@ module Proxy::TFTP
     end
 
     post "/fetch_boot_file" do
-      log_halt(400, "TFTP: Failed to fetch boot file: ") { Proxy::TFTP.fetch_boot_file(params[:prefix], params[:path]) }
+      log_halt(400, "TFTP: Failed to fetch boot file: ") do
+        result = Proxy::TFTP.fetch_boot_file(params[:prefix], params[:path])
+
+        # fetch_boot_file may start a thread
+        # This waits for a bit before telling the client we've accepted it
+        # TODO: implement a mechanism to poll for completion
+        if result.respond_to(:join)
+          202 unless result.join(5)
+        end
+      end
     end
 
     post "/:variant/create_default" do |variant|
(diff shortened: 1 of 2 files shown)


ekohl:
Related: https://github.com/theforeman/smart-proxy/pull/472 implemented a mechanism to poll the status. Probably worth looking at that downloader implementation even if we continue to wrap curl.

modules/tftp/tftp_api.rb
```diff
@@ -35,7 +35,16 @@ def create_default(variant)
...
     post "/fetch_boot_file" do
-      log_halt(400, "TFTP: Failed to fetch boot file: ") { Proxy::TFTP.fetch_boot_file(params[:prefix], params[:path]) }
+      log_halt(400, "TFTP: Failed to fetch boot file: ") do
+        result = Proxy::TFTP.fetch_boot_file(params[:prefix], params[:path])
+
+        # fetch_boot_file may start a thread
+        # This waits for a bit before telling the client we've accepted it
+        # TODO: implement a mechanism to poll for completion
+        if result.respond_to(:join)
+          202 unless result.join(5)
```
adamruzicka, line 45: Assuming webrick is single threaded, does this mean that this will block for up to five seconds, preventing any other requests from being processed during that time?

ekohl: Certainly a concern. I was debating this, how to make it configurable. But I also assumed we at least had some threads in Webrick. Yet I can't find proof for it. That may invalidate the whole approach.

Perhaps a mechanism similar to #472 to return a HTTP 202 with a task ID to poll for is a better approach.

ekohl: For an unrelated item I looked into this. Webrick has a single thread to accept clients, but then starts a thread for every client.

https://github.com/ruby/webrick/blob/cd0b1dd2286492287876773e719bb50a533d2698/lib/webrick/server.rb#L173-L194

So that would mean this just runs a sleep in its own thread, I don't see how it would be a problem.

Still, a proper task ID that can be monitored is a much better API. Both approaches aren't mutually exclusive though.

PR #890, opened 2024-04-15
Fixes #37355 - Explicitly query for IPv4 address in omapi code

ekohl:
The omapi API is IPv4-only so the DNS lookup should also only ask for IPv4 addresses. Otherwise it may return an IPv6 address which ip2hex can't deal with.

### DIFF
--- a/modules/dhcp_common/isc/omapi_provider.rb
+++ b/modules/dhcp_common/isc/omapi_provider.rb
@@ -205,9 +205,9 @@ module Proxy::DHCP::CommonISC
     rescue
       begin
         logger.info "Next-server option not IPv4, trying to resolve '#{server}'"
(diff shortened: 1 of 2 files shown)


evgeni (approved):
You need to adjust the tests/mocks, but otherwise LGTM

ekohl:
Further interesting observation: if it's not resolved by the Smart Proxy, something else does. I didn't check if it's dhcpd or the client, but something resolved the address (which is ipv6-only). Then I can only assume the address is truncated to fit in ipv4. So you may see an unexpected value used for your tftp server

evgeni:
[test smart-proxy]

evgeni:
[test smart-proxy]

PR #891, opened 2024-05-07
Fixes #37413 - Use ip address instead of localhost

dosas:
when testing in a docker container 'localhost'
cannot be assigned but '127.0.0.1' works

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -39,15 +39,16 @@ module Proxy::IntegrationTestCase
   end
 
   def launch(protocol: 'https', plugins: [], settings: {})
+    ip = "127.0.0.1"
     port = 1024 + rand(63_000)
     @settings = Proxy::Settings::Global.new(settings.merge("#{protocol}_port" => port))
     @t = Thread.new do
       launcher = Proxy::Launcher.new(@settings)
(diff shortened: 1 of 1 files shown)


adamruzicka:
How does it fail in docker?

dosas:
> How does it fail in docker?

I put it into the [redmine ticket](https://projects.theforeman.org/issues/37413).
Should I add a more detailed traceback?

ekohl (commented):
In e8fe59f83843ca619264e8d4757bc99fcde93fa9 I used plain webrick to listen on port 0 and then extracted the actual port from the config. I suspect it's a better approach for this integration test too because it avoids picking a random port that may or may not be available. Could you try to see if that works?

dosas:
> In [e8fe59f](https://github.com/theforeman/smart-proxy/commit/e8fe59f83843ca619264e8d4757bc99fcde93fa9) I used plain webrick to listen on port 0 and then extracted the actual port from the config. I suspect it's a better approach for this integration test too because it avoids picking a random port that may or may not be available. Could you try to see if that works?

I made an attempt here https://github.com/theforeman/smart-proxy/pull/892

dosas:
Closed in favor of https://github.com/theforeman/smart-proxy/pull/893

PR #892, opened 2024-05-07
Fixes #37413 - Refactor ssl verification tests

dosas:
by using webrick instead of custom proxy launcher

possible replacement for https://github.com/theforeman/smart-proxy/pull/891

### DIFF
--- a/test/sinatra/ssl_client_verification_integration_test.rb
+++ b/test/sinatra/ssl_client_verification_integration_test.rb
@@ -1,43 +1,64 @@
 require 'test_helper'
 require 'net/http'
 
-class SSLClientVerificationIntegrationTest < Test::Unit::TestCase
-  include Proxy::IntegrationTestCase
-
-  class TestAPIWithSSLClientAuth < ::Sinatra::Base
-    helpers ::Proxy::Helpers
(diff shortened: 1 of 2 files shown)


ekohl (changes requested):
This isn't what I meant. I meant to keep launching the real server (because that's what it's testing: the client SSL verification code), but replace the `port = 1024 + rand(63_000)` bit with `port = 0` and then later figure out which port it really started to listen on. I don't know how well that's possible with `webrick_server` given how we add additional listen statements.

The fact this continues passing means we don't have good coverage. We should have a test case where it has a cert from the right authority, but the host is not listed in `trusted_hosts` and then should be denied. On a plain webrick server that should fail.

dosas:
> This isn't what I meant. I meant to keep launching the real server (because that's what it's testing: the client SSL verification code), but replace the `port = 1024 + rand(63_000)` bit with `port = 0` and then later figure out which port it really started to listen on. I don't know how well that's possible with `webrick_server` given how we add additional listen statements.
I don't understand what you mean and how to implement it.

ekohl:
https://github.com/theforeman/smart-proxy/pull/893

PR #893, opened 2024-05-08
Fixes #37413 - Let the kernel assign a TCP port in tests

ekohl:
This binds on port 0, which lets the kernel decide which port to use. The port is then read back from the listener. Because the kernel knows which ports it's allowed to bind to, it removes the risk of a failure due to collisions or policies.

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -39,19 +39,24 @@ module Proxy::IntegrationTestCase
   end
 
   def launch(protocol: 'https', plugins: [], settings: {})
-    port = 1024 + rand(63_000)
+    port = 0
     @settings = Proxy::Settings::Global.new(settings.merge("#{protocol}_port" => port))
     @t = Thread.new do
       launcher = Proxy::Launcher.new(@settings)
       app = launcher.public_send("#{protocol}_app", port, plugins)
-      launcher.webrick_server(app.merge(AccessLog: [Logger.new('/dev/null')]), ['localhost'], port).start
+      server = launcher.webrick_server(app.merge(AccessLog: [Logger.new('/dev/null')]), ['localhost'], port)
+      # Read back the actual port it bound to
+      @settings["#{protocol}_port"] = server.listeners[0].addr[1]
+      server.start
     end
     Timeout.timeout(2) do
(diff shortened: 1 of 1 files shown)


dosas (approved):
Very nice, thanks. Fixes the issue with docker.

test/test_helper.rb
```diff
@@ -39,19 +39,24 @@ def setup
...
 
   def launch(protocol: 'https', plugins: [], settings: {})
-    port = 1024 + rand(63_000)
+    port = 0
     @settings = Proxy::Settings::Global.new(settings.merge("#{protocol}_port" => port))
     @t = Thread.new do
       launcher = Proxy::Launcher.new(@settings)
       app = launcher.public_send("#{protocol}_app", port, plugins)
-      launcher.webrick_server(app.merge(AccessLog: [Logger.new('/dev/null')]), ['localhost'], port).start
+      @server = server = launcher.webrick_server(app.merge(AccessLog: [Logger.new('/dev/null')]), ['localhost'], port)
```
ekohl: Oops, debug left over. Will update:
```suggestion
      server = launcher.webrick_server(app.merge(AccessLog: [Logger.new('/dev/null')]), ['localhost'], port)
```

test/test_helper.rb
```diff
@@ -39,19 +39,24 @@ def setup
...
+      # Read back the actual port it bound to
+      @settings["#{protocol}_port"] = server.listeners[0].addr[1]
+      server.start
     end
     Timeout.timeout(2) do
-      sleep(0.1) until can_connect?('localhost', @settings.send("#{protocol}_port"))
+      sleep(0.1) until can_connect?('localhost', @settings["#{protocol}_port"])
```
ekohl, line 53: Technically this change isn't needed, but it feels more consistent if we're also setting it as a hash on line 49.

test/test_helper.rb
```diff
@@ -39,19 +39,24 @@ def setup
...
     end
   end
 
   def can_connect?(host, port)
+    return false if port == 0
```
ekohl, line 58: This took me a while to figure out, but the server is started in a thread, so the port isn't immediately set.

PR #894, opened 2024-05-14
Fixes #37450 - start free_ips service for native ms dhcp

sbernhard:


ekohl (approved):
Pretty much the same as https://github.com/theforeman/smart_proxy_dhcp_infoblox/commit/bb75263c653d04d36a4692d89d0f7116acc39f9c.

You can see it indeed uses the free_ips service:
https://github.com/theforeman/smart-proxy/blob/bf8b44f708b852bcff20af19b63951c936854178/modules/dhcp_native_ms/plugin_configuration.rb#L12

PR #895, opened 2024-05-17
fixes #37486 - Request BMC using Redfish without reaching sessions limit

neomilium:
As described in issue, sessions limit is reach quickly when using current _Redfish_ implementation.

Here, redefining the `RedfishClient::Root#session_path` allow us to change the way `redfish_client` process the authentication in order to force to use only `basic_auth` and never retrieve a session that Foreman never closes.

Issue: https://projects.theforeman.org/issues/37486

### DIFF
--- a/bundler.d/bmc.rb
+++ b/bundler.d/bmc.rb
@@ -1,4 +1,4 @@
 group :bmc do
   gem 'rubyipmi', '>= 0.10.0'
-  gem 'redfish_client', '>= 0.5.1'
+  gem 'redfish_client', '>= 0.6.0'
 end
--- a/modules/bmc/redfish.rb
+++ b/modules/bmc/redfish.rb
@@ -16,7 +16,7 @@ module Proxy
       end
 
       def connect(args = { })
-        connection = RedfishClient.new("https://#{args[:host]}/", verify: @redfish_verify_ssl)
+        connection = RedfishClient.new("https://#{args[:host]}/", verify: @redfish_verify_ssl, use_session: false)
         connection.login(args[:username], args[:password])
         connection
       end


evgeni:
ok to test

neomilium:
I just force-pushed to fix rubocop offence related to comment.

evgeni:
Is this the same issue as described in https://community.theforeman.org/t/redfish-leaks-connections-and-results-in-many-active-connections-to-bmc-interface/30559?

evgeni:
cc @dgoetz @laugmanuel @gardar

laugmanuel:
Without really looking into it any deepter, it looks plausible.
However, I can't test it as we stopped using the BMC feature in Foreman altogether.

laugmanuel:
cc @ochnerd, @thomas-merz

ekohl:
I'd consider an alternative and that is cleaning up sessions by logging out. https://github.com/theforeman/smart-proxy/pull/896 is that. I'm not sure yet what's best. Perhaps avoiding sessions in the first place fits the model better, depending on how many API calls within a session we perform.

A more complex approach would be to reuse connections, but I'd be hesitant given the complexity of that.

gardar:
Looks good at first glance, I should be able to test this next week

neomilium:
@ekohl About how many call we made, I can say that a simple look on host's _Overview_ page took one session token while browsing to _BMC_ tab in legacy UI took 5 sessions.

As in iDRAC/8 there are 8 session slots, a regular browsing can easily reach the limit even the sessions stored in iDRAC expire automatically within hours.
Note: I did not count it exactly nor found this delay in iDRAC's web UI, just saw that an iDRAC with all sessions allocated will be reachable after a night.

neomilium:
> Is this the same issue as described in https://community.theforeman.org/t/redfish-leaks-connections-and-results-in-many-active-connections-to-bmc-interface/30559?

Yes, exactly.

ekohl:
@neomilium It's Foreman -> Smart Proxy -> BMC. With https://github.com/theforeman/smart-proxy/pull/896 one API call to the Smart Proxy would be a single session. So if that call performs multiple requests, a single session **could** be a benefit, but if basic auth is just as fast then your PR is preferred (because my session cleanup could fail while your will never open a session).

neomilium:
> [...] but if basic auth is just as fast then your PR is preferred (because my session cleanup could fail while your will never open a session).

I show some benchmarks at #896

ekohl (changes requested):
@neomilium my patch was merged and 0.6.0 was released. Could you rebase your patch to use the new parameter and raise the minimum version to 0.6.0?

neomilium:
@ekohl Done. Thank you.

ekohl (approved):
Thanks!

ekohl:
https://github.com/theforeman/foreman-packaging/pull/11028 & https://github.com/theforeman/foreman-packaging/pull/11029 update the dependency.

PR #899, opened 2024-07-02
Fixes #37621 - Correctly resolve the IPv4 address

ekohl:
In 3f079df67820053dcf626f0dd4f908cfc3bc9830 the lookup code was changed to explicitly query IPv4, but it assumed a string was returned instead of a Resolv::DNS::Resource::IN::A instance.

Fixes: 3f079df67820 ("Fixes #37355 - Explicitly query for IPv4 address in omapi code")

Currently untested, but it was reported on Matrix.


evgeni (approved):

h00tomsu:
Just applied the patch and it solved my issue

PR #900, opened 2024-07-04
Fixes #37624 - Pin FFI to < 1.17 on Ruby 2.x

ekohl:
Version 1.17 no longer works on Ruby 2 and Bundler isn't smart enough to figure this out.


adamruzicka (approved):
I ran into this just a moment ago, this did the trick

adamruzicka:
Thank you @ekohl !

PR #901, opened 2024-07-05
Fixes #37631 - Allow nested data structures as request parameters

nadjaheitmann:
If you go to the Host Registration form in foreman and choose a smart-proxy and multiple repositories, the registration fails with an internal server error. This is due to the new format of the repo_data in the API which supports an array of repository and corresponding gpg key pairs (https://github.com/theforeman/foreman/blob/develop/app/controllers/api/v2/registration_controller.rb#L29).

### DIFF
--- a/lib/proxy/request.rb
+++ b/lib/proxy/request.rb
@@ -10,7 +10,7 @@ module Proxy::HttpRequest
     end
 
     def query_string(input = {})
-      input.compact.map { |k, v| "#{CGI.escape(k.to_s)}=#{CGI.escape(v)}" }.join("&")
+      Rack::Utils.build_nested_query(input.compact)
     end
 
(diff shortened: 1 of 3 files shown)


sbernhard (approved):
LGTM. Thanks!

m-bucher (approved):
Nice, makes sense to use an existing method to build the query-string.

sbernhard:
the "query_string" method is used from "create_get" and "create_post" which are tested in https://github.com/theforeman/smart-proxy/blob/beb0147354797bd1c2fe1f74cb12a43476d2dfce/test/request_test.rb#L10 

Can you please add more tests to make sure this method works as expected.

sbernhard:
@adamruzicka Any concerns that I merge this?

adamruzicka:
None from me, looks good

PR #905, opened 2024-09-09
Add HSTS middleware

ekohl:
This header is entirely useless for us because the header is aimed at browsers, but some scanners still think this is needed.

This PR is not intended to be merged, but only serve as an example for how it could be implemented.

Link: https://www.tenable.com/plugins/nessus/142960

### DIFF
--- a/lib/proxy/hsts_middleware.rb
+++ b/lib/proxy/hsts_middleware.rb
@@ -1,0 +1,19 @@
+module Proxy
+  # Add the HSTS header if not present. This header is entirely useless for us
+  # because the header is aimed at browsers, but some scanners still think this
+  # is needed.
+  # https://www.tenable.com/plugins/nessus/142960
+  class HstsMiddleware
+    def initialize(app)
+      @app = app
+    end
+
+    def call(env)
+      status, headers, body = @app.call(env)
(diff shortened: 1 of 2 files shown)


evgeni (dismissed):

evgeni:
Why not merging this? If it will make people shut up, I'm all for it.

ehelms:
> Why not merging this? If it will make people shut up, I'm all for it.

I am, sadly, in the same boat as this. While I detest that scanners force us into these situations, I also feel for our users and a small change that improves their experience gets my vote.

willdarton:
I appreciate you taking the time to create a PR as a potential solution.  As others have mentioned it seems like a simple fix to make the customer experience better.

ekohl:
I'm still tempted to detect the Nessus user agent and simply reject them to silence it. Another option is to only send it to Nessus, just to shut it up.

If we want merge this, we need to be compliant with the RFC and only send it when the connection is over TLS ([RFC 6797 section 5.1](https://www.rfc-editor.org/rfc/rfc6797#section-5.1)). It's also good to add some tests to this.

ekohl:
Updated to inspect the env to detect HTTPS and added a Redmine issue.

ehelms (approved):

PR #910, opened 2025-01-22
Fixes #38157 - Declare rexml as dependency

adamruzicka:
Apparently rexml (used in the realm module) used to be a default gem up until ruby 3. Starting with ruby 3, it has to be treated as any other dependency.

### DIFF
--- a/smart_proxy.gemspec
+++ b/smart_proxy.gemspec
@@ -16,6 +16,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'json'
   s.add_dependency 'logging'
   s.add_dependency 'rack', '>= 1.3'
+  s.add_dependency 'rexml', '~> 3.2'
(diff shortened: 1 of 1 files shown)


smart_proxy.gemspec
```diff
@@ -16,6 +16,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'json'
   s.add_dependency 'logging'
   s.add_dependency 'rack', '>= 1.3'
+  s.add_dependency 'rexml', '~> 3.4'
```
ekohl: If we leave it at 3.2 we can match what's bundled in Ruby 3.0: https://stdgems.org/rexml/
```suggestion
  s.add_dependency 'rexml', '~> 3.2'
```

ekohl (changes requested):

ekohl (approved):

PR #912, opened 2025-02-25
Fixes #38241 - Extend userdata API by MAC address endpoint

bastian-src:
Extend the proxy API to match the userdata API of foreman: https://github.com/theforeman/foreman/pull/9213

### DIFF
--- a/modules/templates/templates_userdata_api.rb
+++ b/modules/templates/templates_userdata_api.rb
@@ -8,4 +8,10 @@ class Proxy::TemplatesUserdataApi < Sinatra::Base
       Proxy::Templates::UserdataProxyRequest.new.get(kind, request.env, params)
     end
   end
+
+  get "/:mac/:kind" do |mac, kind|
+    log_halt(500, "Failed to retrieve #{kind} userdata template for #{params.inspect}: ") do
+      Proxy::Templates::UserdataProxyRequest.new.get([mac, kind], request.env, params)
+    end
+  end
 end
--- a/test/templates/templates_userdata_api_test.rb
+++ b/test/templates/templates_userdata_api_test.rb
@@ -25,4 +25,13 @@ class TemplatesApiTest < Test::Unit::TestCase
     assert last_response.ok?, "Last response was ok"
     assert_match("A user-data template", last_response.body)
   end
+
+  def test_api_can_ask_for_a_cloud_init_template_by_mac
+    stub_request(:get, "#{@foreman_url}/userdata/00:a1:b2:c3:d4:f5/user-data")
+      .with(query: {"url" => @template_url, "mac" => "00:a1:b2:c3:d4:f5"})
+      .to_return(:body => 'A user-data template')
+    get "/00:a1:b2:c3:d4:f5/user-data"
+    assert last_response.ok?, "Last response was ok"
+    assert_equal("A user-data template", last_response.body)
+  end
 end


test/templates/templates_userdata_api_test.rb
```diff
@@ -25,4 +25,11 @@ def test_api_can_ask_for_a_cloud_init_template
     assert last_response.ok?, "Last response was ok"
     assert_match("A user-data template", last_response.body)
   end
+
+  def test_api_can_ask_for_a_cloud_init_template_by_mac
+    stub_request(:get, "#{@foreman_url}/userdata/00:a1:b2:c3:d4:f5/user-data").with(query: {"url" => @template_url}).to_return(:body => 'A user-data template')
+    get "00:a1:b2:c3:d4:f5/user-data", {}
```
ekohl: Test failure is related. I think this needs to be
```suggestion
    get "/00:a1:b2:c3:d4:f5/user-data", {}
```

ekohl (dismissed):
> Refs #34887

That issue has been closed and already a `Fixed in release` version that went out so you can't use `Refs` for that. Please open a new issue.

test/templates/templates_userdata_api_test.rb
```diff
@@ -25,4 +25,11 @@ def test_api_can_ask_for_a_cloud_init_template
     assert last_response.ok?, "Last response was ok"
     assert_match("A user-data template", last_response.body)
   end
+
+  def test_api_can_ask_for_a_cloud_init_template_by_mac
+    stub_request(:get, "#{@foreman_url}/userdata/00:a1:b2:c3:d4:f5/user-data").with(query: {"url" => @template_url}).to_return(:body => 'A user-data template')
+    get "/00:a1:b2:c3:d4:f5/user-data", {}
```
ekohl: GET requests don't have data so this should suffice
```suggestion
    get "/00:a1:b2:c3:d4:f5/user-data"
```

bastian-src:
Thanks a lot for the fast responses! @ekohl

bastian-src:
@ekohl do you have an idea why I gotta define the additional parameter [here](https://github.com/theforeman/smart-proxy/pull/912/files#diff-9037532b127679768b6b03eb0a51acd83ccb58294f4f43965aa162c6f598515cR31) while it is not needed in the unattended API [here](https://github.com/theforeman/smart-proxy/blob/develop/test/templates/templates_unattended_api_test.rb#L44)?

test/templates/templates_userdata_api_test.rb
```diff
@@ -25,4 +25,13 @@ def test_api_can_ask_for_a_cloud_init_template
...
     assert_match("A user-data template", last_response.body)
   end
+
+  def test_api_can_ask_for_a_cloud_init_template_by_mac
+    stub_request(:get, "#{@foreman_url}/userdata/00:a1:b2:c3:d4:f5/user-data")
+      .with(query: {"url" => @template_url, "mac" => "00:a1:b2:c3:d4:f5"})
+      .to_return(:body => 'A user-data template')
+    get "/00:a1:b2:c3:d4:f5/user-data"
+    assert last_response.ok?, "Last response was ok"
+    assert_match("A user-data template", last_response.body)
```
ekohl: I think `assert_match` is intended for regular expressions. In this case I think you can use the more precise `assert_equal`
```suggestion
    assert_equal("A user-data template", last_response.body)
```

modules/templates/templates_userdata_api.rb
```diff
@@ -8,4 +8,10 @@ class Proxy::TemplatesUserdataApi < Sinatra::Base
       Proxy::Templates::UserdataProxyRequest.new.get(kind, request.env, params)
     end
   end
+
+  get "/:mac/:kind" do |mac, kind|
+    log_halt(500, "Failed to retrieve #{kind} userdata template for #{params.inspect}: ") do
+      Proxy::Templates::UserdataProxyRequest.new.get([mac, kind], request.env, params)
```
ekohl, line 14: Looking at the method this feels like an abuse of the API:
https://github.com/theforeman/smart-proxy/blob/6624f2c22cf263dd49025a4af28d893a9e22ff52/modules/templates/userdata_proxy_request.rb#L5

Technically it may work since it calls `flatten`, but I really don't like `flatten` because it ends up being really unpredictable what gets called.

But that's probably not the real issue. If you look at `ProxyRequest` you can see it copies the parameters while removing a blacklist:
https://github.com/theforeman/smart-proxy/blob/6624f2c22cf263dd49025a4af28d893a9e22ff52/modules/templates/proxy_request.rb#L30-L33

Looking at that blacklist, we can see `kind` is in there:
https://github.com/theforeman/smart-proxy/blob/6624f2c22cf263dd49025a4af28d893a9e22ff52/modules/templates/proxy_request.rb#L9

I suspect both `mac` and `kind` are in `params`. Not entirely sure what the correct thing to do here, but this whole API feels awkward.

test/templates/templates_userdata_api_test.rb
bastian-src: Applied, thanks!

modules/templates/templates_userdata_api.rb
bastian-src: Alright, I see why `hostgroup` and `template` don't need to be added in the test - thanks!

However, to make this API feel less awkward I see two improvements currently:

a.) Add another function in `UserdataProxyRequest` to mitigate the `flatten` [here](https://github.com/theforeman/smart-proxy/blob/develop/modules/templates/userdata_proxy_request.rb#L5)?
Like a `def get_mac` and then put the `mac` as mandatory function parameter? But, this makes it less generic..

b.) Overwrite BLACKLIST_PARAMETERS when inheriting `ProxyRequest` and add `mac` there?

What do you think?

sbernhard: @ekohl thoughts on the comment from Bastian?

dgoetz:
As I am currently running into this problem and the patch at least works, may I ask if we can get it in as a fix or even as a workaround we can later improve on?

test/templates/templates_userdata_api_test.rb
```diff
@@ -25,4 +25,13 @@ def test_api_can_ask_for_a_cloud_init_template
     assert last_response.ok?, "Last response was ok"
     assert_match("A user-data template", last_response.body)
   end
+
+  def test_api_can_ask_for_a_cloud_init_template_by_mac
+    stub_request(:get, "#{@foreman_url}/userdata/00:a1:b2:c3:d4:f5/user-data")
+      .with(query: {"url" => @template_url, "mac" => "00:a1:b2:c3:d4:f5"})
+      .to_return(:body => 'A user-data template')
+    get "/00:a1:b2:c3:d4:f5/user-data"
```
stejskalleos, line 33: I'm not sure what I doing wrong, but when I try:
```
curl "http://foreman.local.lan:8080/00:aa:aa:10:10:01/user-data"

=>

2025-07-30T10:07:07 c1c036b7 [I] Started GET /00:aa:aa:10:10:01/user-data 
2025-07-30T10:07:07 c1c036b7 [I] Finished GET /00:aa:aa:10:10:01/user-data with 404 (0.52 ms)
```

No forwarded requests to Foreman at all.

```
curl -H 'Content-Type: text/plain' 'http://foreman.local.lan:8080/features'
["ansible","dhcp","discovery","dynflow","facts","logs","puppet","registration","script","templates","tftp"]
```

bastian-src: Does it work when you adapt your endpoint a little, like this?

```
curl "http://foreman.local.lan:8080/userdata/00:aa:aa:10:10:01/user-data"
```

The `get`-call in the test should put the module path `userdata` automatically which is why it's not explicitly needed [here](https://github.com/theforeman/smart-proxy/pull/912/files#diff-9037532b127679768b6b03eb0a51acd83ccb58294f4f43965aa162c6f598515cR33).

stejskalleos: Oh I see, I got confused by the `userdata` being twice in the URL.

stejskalleos (approved):
:green_apple: **LGTM**

Tested with Foreman and it works, all the comments have been addressed, let's get this in.

sbernhard:
Thanks @stejskalleos  and @bastian-src :-)

PR #913, opened 2025-04-09
Fixes #38357 - add base64 runtime dependency

bastelfreak:
lib/proxy/util.rb does a `require 'base64'`. This gem won't be part of Ruby core with Ruby 3.4. It needs to be listed explicitly in the gemspec.


bastelfreak:
This is required to get the CI in https://github.com/theforeman/actions/pull/65 fixed

evgeni (approved):

ekohl:
I suspect we'll need to patch packaging for this.

evgeni:
This will most probably be correct yeah

ekohl:
https://github.com/theforeman/foreman-packaging/pull/11929

PR #914, opened 2025-04-14
Refs #38357 - Support of ruby version 3.4

sbernhard:
Support of ruby version 3.4.

Therefore we need to add the dependency to logger here. See https://github.com/ruby-concurrency/concurrent-ruby/pull/1062


ekohl (commented):
Looking at https://stdgems.org/logger/ it states:

> The gem type will be changed to bundled gem in a future version of Ruby

So it will be dropped (possibly Ruby 3.5). We also require the `logging` gem in `lib/proxy/log.rb` so I wonder if we can align on that.

Reading the concurrent-ruby source code I see you can assign the global logger so perhaps we should do that instead? You can see they deprecated the `create_stdlib_logger` so they're moving away from it as well.

Having said that, I do think in some places (`log_buffer` at least) we directly rely on the logger gem.

What do you think we should do?

sbernhard:
I don't think its about the Logger rubygem - the logger is part of ruby itself.
In Ubuntu deliverd by libruby3.2. In EL world deliverd by ruby-libs

The same change was done in https://github.com/rails/rails/pull/54621, too

ekohl:
> I don't think its about the Logger rubygem - the logger is part of ruby itself.
> In Ubuntu deliverd by libruby3.2. In EL world deliverd by ruby-libs

This is the key thing: **for now** it is, but in Ruby 3.5 it will be removed: https://github.com/ruby/ruby/commit/d7e558e3c48c213d0e8bedca4fb547db55613f7c. We can expect Ruby 3.5 in December and I like to look forward. If we can get rid of it altogether now that'll save work in the future.

sbernhard:
> > I don't think its about the Logger rubygem - the logger is part of ruby itself.
> > In Ubuntu deliverd by libruby3.2. In EL world deliverd by ruby-libs
> 
> This is the key thing: **for now** it is, but in Ruby 3.5 it will be removed: [ruby/ruby@d7e558e](https://github.com/ruby/ruby/commit/d7e558e3c48c213d0e8bedca4fb547db55613f7c). We can expect Ruby 3.5 in December and I like to look forward. If we can get rid of it altogether now that'll save work in the future.

Well, during a installation of the smart-proxy on Windows we got this issue because of the missing 'require logger'. 
Therefore, if it does not break anything right now, I would suggest to add this fix.

ekohl:
It concerns me we don't see it in CI now and I question the location. We also have https://github.com/theforeman/smart-proxy/blob/develop/lib/smart_proxy_for_testing.rb. Does it mean we simply don't hit the code paths in our testing?

Related: right now we're only testing on Ruby 3.0 but we should expand our matrix with Ruby 3.1 (Debian 12), 3.2 (Ubuntu 24.04) and 3.3 (EL 10 & Debian 13). Ideally also 3.4 for the future. Perhaps then we'd probably see it ourselves in CI?

sbernhard:
> Related: right now we're only testing on Ruby 3.0 but we should expand our matrix with Ruby 3.1 (Debian 12), 3.2 (Ubuntu 24.04) and 3.3 (EL 10 & Debian 13). Ideally also 3.4 for the future. Perhaps then we'd probably see it ourselves in CI?

I added these versions. Only 3.4 failed to test:
```
"test/tftp/tftp_api_test.rb" "test/tftp/tftp_server_test.rb" "test/tftp/tftp_test.rb" "test/util_test.rb" "test/validations_test.rb" 
/home/runner/work/smart-proxy/smart-proxy/vendor/bundle/ruby/3.4.0/gems/rack-2.2.13/lib/rack/show_exceptions.rb:3: warning: ostruct was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.
You can add ostruct to your Gemfile or gemspec to silence this warning.
/home/runner/work/smart-proxy/smart-proxy/vendor/bundle/ruby/3.4.0/gems/logging-2.4.0/lib/logging.rb:10: warning: syslog was loaded from the standard library, but is not part of the default gems starting from Ruby 3.4.0.
You can add syslog to your Gemfile or gemspec to silence this warning.
/home/runner/work/smart-proxy/smart-proxy/vendor/bundle/ruby/3.4.0/gems/rubyipmi-0.11.1/lib/rubyipmi/observablehash.rb:1: warning: observer was loaded from the standard library, but is not part of the default gems starting from Ruby 3.4.0.
You can add observer to your Gemfile or gemspec to silence this warning.
/opt/hostedtoolcache/Ruby/3.4.3/x64/lib/ruby/3.4.0/bundled_gems.rb:[8](https://github.com/theforeman/smart-proxy/actions/runs/14521274599/job/40742594648#step:7:9)2:in 'Kernel.require': cannot load such file -- observer (LoadError)
	from /opt/hostedtoolcache/Ruby/3.4.3/x64/lib/ruby/3.4.0/bundled_gems.rb:82:in 'block (2 levels) in Kernel#replace_require'
	```
=> observer rubygem
But we should also have a look at ostruct.

**Comment by sbernhard** [comment]
I tried without the "require logger" - but this wasn't found by our testing. I wonder why...

**Comment by sbernhard** [comment]
Fixed the tests. 

How to continue with this @ekohl ?

File: Gemfile [hunk 1]
```diff
@@ -7,6 +7,12 @@ gem 'concurrent-ruby', '~> 1.0', require: 'concurrent'
 # FFI 1.17 needs rubygems 3.3.22+, which is Ruby 3.0+ only
 gem "ffi", "<1.17" if RUBY_VERSION < '3.0'
 
+# these gems will be moved out of the default ruby gems in >=3.4
```
ekohl: 3.4 is already GA so future tense isn't really correct

lib/smart_proxy_main.rb
```diff
@@ -1,6 +1,7 @@
 APP_ROOT = "#{__dir__}/.."
 
 require 'smart_proxy'
+require 'logger'
```
ekohl: I'd prefer this to live in `lib/proxy/log.rb`

Gemfile
```diff
@@ -7,6 +7,12 @@ gem 'concurrent-ruby', '~> 1.0', require: 'concurrent'
...
+if RUBY_VERSION >= '3.4'
+  gem 'syslog'
+  gem 'ostruct'
```
ekohl: This really needs to live in the gemspec because it's a core dependency. If you don't then all plugins will break anyway.

.github/matrix.json
```diff
@@ -1,3 +1,3 @@
 {
-  "ruby": ["2.7", "3.0"]
+  "ruby": ["2.7", "3.0", "3.1", "3.2", "3.3", "3.4"]
```
ekohl, line 2: This is a bigger change. It's very welcome, but perhaps we should change the PR description (and issue description) to Ruby 3.4 support?

Ideally you'd also verify plugins wouldn't break. https://github.com/theforeman/actions/pull/68 adds documentation on how to do that.

ekohl (commented):
Some minor nits but overall I like where this is heading.

.github/matrix.json
sbernhard: So, you mean to change this PR and rename it to something like "support ruby 3.4" ?

smart_proxy.gemspec
```diff
@@ -16,10 +16,12 @@ Gem::Specification.new do |s|
   s.add_dependency 'base64'
   s.add_dependency 'json'
   s.add_dependency 'logging'
+  s.add_dependency 'ostruct'
   s.add_dependency 'rack', '>= 1.3'
   s.add_dependency 'rexml', '~> 3.2'
   s.add_dependency 'sd_notify', '~> 0.1'
   s.add_dependency 'sinatra', '~> 2.0'
+  s.add_dependency 'syslog'
```
ekohl: I debated this one and ended up not posting a comment because it was in `Gemfile`. Technically this is a soft dependency:
https://github.com/theforeman/smart-proxy/blob/3e4ca4e408e108b5abfe83d76196396e5f85157a/lib/proxy/log.rb#L15-L20

However, the probing triggers the deprecation warning. I'm unsure how to proceed with this. My feeling says to keep it in `Gemfile` since it's a soft dependency. Even then, only pull it in on Ruby >= 3.4

bundler.d/dhcp_isc.rb
```diff
@@ -7,3 +7,8 @@
 group :dhcp_isc_kqueue do
   gem 'rb-kqueue'
 end
+
+# these gems were moved out of the default ruby gems in >=3.4
```
ekohl: I'm not sure we need to keep this block, but if we do, would this be a better description?
```suggestion
# Changed from a default gem to a bundled gem in Ruby 3.4
# https://stdgems.org/new-in/3.4/
```

bundler.d/dhcp_isc.rb
```diff
@@ -7,3 +7,8 @@
...
+if RUBY_VERSION >= '3.4'
+  gem 'observer'
```
ekohl: I can't find where we use the `observer` gem. https://github.com/ruby/observer says you use it with `include Observable` but AFAIK dhcp_isc only depends on inotify or kqueue to get file updates.

ekohl (commented):
Looking at https://projects.theforeman.org/issues/38357 this should probably be `Refs #38357` and be cherry picked into 3.15.0.

sbernhard:
Tests broken. Oh, this lovely rubygems:

image
(excon updated 26min ago :-) )

bundler.d/dhcp_isc.rb
sbernhard: Ah. Thanks. Found out which module observer needs. Its bmc:

```
/home/runner/work/smart-proxy/smart-proxy/vendor/bundle/ruby/3.4.0/gems/rubyipmi-0.11.1/lib/rubyipmi/observablehash.rb:1: warning: observer was loaded from the standard library, but is not part of the default gems starting from Ruby 3.4.0.
You can add observer to your Gemfile or gemspec to silence this warning.
/opt/hostedtoolcache/Ruby/3.4.4/x64/lib/ruby/3.4.0/bundled_gems.rb:[8](https://github.com/theforeman/smart-proxy/actions/runs/15279421183/job/42974399220?pr=914#step:7:9)2:in 'Kernel.require': cannot load such file -- observer (LoadError)
	from /opt/hostedtoolcache/Ruby/3.4.4/x64/lib/ruby/3.4.0/bundled_gems.rb:82:in 'block (2 levels) in Kernel#replace_require'
[...]
	from /home/runner/work/smart-proxy/smart-proxy/modules/bmc/bmc_api.rb:1:in '<top (required)>'
	from /opt/hostedtoolcache/Ruby/3.4.4/x64/lib/ruby/3.4.0/bundled_gems.rb:82:in 'Kernel.require'
	from /opt/hostedtoolcache/Ruby/3.4.4/x64/lib/ruby/3.4.0/bundled_gems.rb:82:in 'block (2 levels) in Kernel#replace_require'
	from /home/runner/work/smart-proxy/smart-proxy/test/bmc/bmc_api_shell_test.rb:3:in '<top (required)>'
```

sbernhard:
Pinned excon <= 1.2.5 for now. 
Excon fixed it already in https://github.com/excon/excon/commit/2f5654b34388ea1e81ddcd4c954412cf45362d03 - just waiting for a new version. 

Update: Fixed. new 1.2.7 of excon was just released.

bundler.d/dhcp_isc.rb
ekohl: Ah, it's [rubyipmi](https://github.com/logicminds/rubyipmi) that actually needs it. I've opened https://github.com/logicminds/rubyipmi/pull/60.

bundler.d/bmc.rb
```diff
@@ -1,4 +1,7 @@
 group :bmc do
   gem 'rubyipmi', '>= 0.10.0'
   gem 'redfish_client', '>= 0.6.0'
+  # Changed from a default gem to a bundled gem in Ruby 3.4
+  # See: https://stdgems.org/new-in/3.4/
```
ekohl: Perhaps reflect that it was really needed for rubyipmi instead of our codebase?
```suggestion
  # https://github.com/logicminds/rubyipmi/pull/60
```

ekohl (dismissed):
Thanks for updating this. What do you think about the last inline suggestion?

@lhellebr I'd like to include this in RC2 or GA if possible.

ekohl:
@sbernhard could you also squash the commits into 1 so Redmine issues is pleased and we'll also cherry pick correctly?

Also a note to https://community.theforeman.org/c/development/9 will be nice for all developers.

sbernhard:
@ekohl rebase / squash done.
would be good to finish this soonish. Looks like https://github.com/theforeman/smart_proxy_salt/actions/runs/15293534588/job/43017475837?pr=106#step:9:53 does fail because of this PR, too.

bundler.d/bmc.rb
ekohl: I see you marked it as resolved, but haven't applied it. Does that mean you don't think we should or missed it in your update?

sbernhard: Sorry. I have changed it like this and thought its even more clear why we did this change:

```
  # observer is a transitive dependency of rubyipmi
  # Changed from a default gem to a bundled gem in Ruby 3.4
  # See: https://stdgems.org/new-in/3.4/
```

See: https://github.com/theforeman/smart-proxy/pull/914/files#diff-05c54079449e49b795cfbdb6b546fa1e4ca9c3388cccbb44a9890928eb671379R4

ekohl: Ah yes, but my point of adding the link to the PR was to indicate it's a workaround and it really doesn't belong here.

ekohl: And I just noticed it should have been https://github.com/logicminds/rubyipmi/pull/61.

sbernhard: Make sense. I have updated the comment again. Hope this solves it now :)

ekohl (approved):
Thanks!

ekohl:
@sbernhard thanks a lot for picking this up. Would you mind creating the Discourse post to mention we'll do this? For developers the impact is that their Smart Proxy CI will start running on Ruby 3.4 now and that can cause failures.

sbernhard:
> Would you mind creating the Discourse post to mention we'll do this? For developers the impact is that their Smart Proxy CI will start running on Ruby 3.4 now and that can cause failures.

I hope this is ok: https://community.theforeman.org/t/smart-proxy-ci-now-running-on-ruby-3-4/43409

PR #917, opened 2025-06-13
Fixes #38498 - Add Reboot power operation to Redfish provider

vanou:
Bug #3073 is caused by 3 things:
1. Foreman core doesn't call BMC proxy's API correctly
   - Changing power state to 'Reboot' via Web GUI calls BMC proxy's 'soft' power action
   - Changing power state to 'Reset' via Web GUI calls BMC proxy's 'cycle' power action
2. Foreman BMC proxy doesn't support reboot at all BMC provider
3. Foreman BMC proxy doesn't support reset(powerreset) at Redfish provider

2nd & 3rd item are mentioned in #38498. And this PR fixes 2nd item.

## Modifications in this PR

This PR does 2 things to make BMC smart proxy support reboot power action:

* Modify BMC smart proxy API to accept `reboot` power action and call BMC provider's `powerreboot` method
* Add `powerreboot` method to Redfish provider class which makes Redfish server do `GracefulRestart` reset action
  (According to IPMI v2.0 specification, IPMI doesn't support Reboot)

PR #916, opened 2025-06-13
Fixes #38498 - Add support for Reset (powerreset) via Redfish provider

spesnova717:
Bug [#3073](https://projects.theforeman.org/issues/3073?tab=history) is caused by 3 things:
1. Foreman core doesn't call BMC proxy's API correctly
   - Changing power state to 'Reboot' via Web GUI calls BMC proxy's 'soft'
     power action
   - Changing power state to 'Reset' via Web GUI calls BMC proxy's 'cycle'
     power action
2. Foreman BMC proxy doesn't support reboot at all BMC provider
3. Foreman BMC proxy doesn't support reset(powerreset) at Redfish provider
 
2nd & 3rd item are mentioned in [#38498](https://projects.theforeman.org/issues/38498). And this PR fixes 3rd item.

### Enable proper handling of "Reset" power action via Redfish

This PR enables the powerreset operation to be processed correctly by the Smart Proxy when using the Redfish provider.
Internally, this maps the "Reset" request from Foreman core to the appropriate Redfish ForceRestart operation, aligning the behavior with user expectations in the Web UI. 
To support this, a new powerreset API endpoint is introduced, and the Redfish BMC class is extended to implement the corresponding logic.

PR #919, opened 2025-07-25
Fixes #38498 - Fix incorrect Cycle power operation in Redfish provider

hiroshi-ikuta:
Bug #38498 mentions 3 power operation bugs in Redfish provider. This PR fixes wrong implementation of Cycle power operatoin: When user pushes Cycle button, Redfish provider does ForceRestart instead of PowerCycle.

### Fix incorrect "Cycle" power operation implementation in Redfish provider
The powercycle method defined in `modules/bmc/redfish.rb` of smart-proxy does 'ForceRestart' instead of 'PowerCycle'. This PR makes powercycle method does 'PowerCycle'.

### DIFF
--- a/test/bmc/redfish_test_helper.rb
+++ b/test/bmc/redfish_test_helper.rb
@@ -1,0 +1,51 @@
+require 'json'
+
+module RedfishTestHelper
+  ROOT_DATA = {
+    "Name" => "Redfish Server Test Mock",
+    "Systems" => {
+      "@odata.id" => "/redfish/v1/Systems",
+    },
+    "Managers" => {
+      "@odata.id" => "/redfish/v1/Managers",
+    },
+    "@odata.id" => "/redfish/v1",
+  }.freeze
+
+  SYSTEMS_DATA = {
+    "@odata.id" => "/redfish/v1/Systems",
+    "Members" => [
+      { "@odata.id" => "/redfish/v1/Systems/TESTSYSTEM" },
+    ],
+  }.freeze
+
+  SYSTEM_DATA = {
+    "@odata.id" => "/redfish/v1/Systems/TESTSYSTEM",
+    "Manufacture" => "Server Vendor",
+    "IndicatorLED" => "Off",
+    "PowerState" => "On",
+    "Boot" => {
+      "BootSourceOverrideTarget" => "Pxe",
+    },
+    "Model" => "TestServer A",
+    "SerialNumber" => "AA645",
+    "AssetTag" => "ax",
+    "Actions" => {
+      "#ComputerSystem.Reset" => {
+        "target" => "/redfish/v1/Systems/TESTSYSTEM/Actions/ComputerSystem.Reset",
+      },
+    },
+  }.freeze
+
+  def mask_redfish_acceess(protocol: "https", host: "host")
+    # root
+    stub_request(:get, "#{protocol}://#{host}#{ROOT_DATA['@odata.id']}").
+      to_return(status: 200, body: JSON.generate(ROOT_DATA))
+    # root.systems
+    stub_request(:get, "#{protocol}://#{host}#{SYSTEMS_DATA['@odata.id']}").
+      to_return(status: 200, body: JSON.generate(SYSTEMS_DATA))
+    # root.system
+    stub_request(:get, "#{protocol}://#{host}#{SYSTEM_DATA['@odata.id']}").
+      to_return(status: 200, body: JSON.generate(SYSTEM_DATA))
+  end
+end
--- a/modules/bmc/redfish.rb
+++ b/modules/bmc/redfish.rb
@@ -71,7 +71,7 @@ module Proxy
       end
 
       def powercycle
-        poweraction('ForceRestart')
+        poweraction('PowerCycle')
       end
 
       def poweron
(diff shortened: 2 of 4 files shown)


modules/bmc/bmc_plugin.rb
```diff
@@ -11,6 +11,7 @@ class Plugin < Proxy::Plugin
     capability 'shell'
     capability 'ssh'
     capability -> { Proxy::BMC::IPMI.providers_installed }
+    capability 'poweraction_fix_3_17'
```
stejskalleos: Is exposing this capability necessary here in this PR?
I understand it's needed for #916 & #917, but here I don't see the relation. Or do I miss anything?

vanou: Thanks for comment.
After considering your comment a while, I feel you are right & capability is not necessary in this PR. So we'll remove things related to capability from this PR.

test/bmc/redfish_test_helper.rb
```diff
@@ -0,0 +1,101 @@
...
+    # root.manager
+    stub_request(:get, "#{protocol}://#{host}#{MANAGER_DATA['@odata.id']}").
+      to_return(status: 200, body: JSON.generate(MANAGER_DATA))
+    # root.manager.ethernetinterfaces
+    stub_request(:get, "#{protocol}://#{host}#{ETHERNET_INTERFACES_DATA['@odata.id']}").
+      to_return(status: 200, body: JSON.generate(ETHERNET_INTERFACES_DATA))
+    # bmcnic
+    stub_request(:get, "#{protocol}://#{host}#{BMC_NIC_DATA['@odata.id']}").
+      to_return(status: 200, body: JSON.generate(BMC_NIC_DATA))
+  end
```
stejskalleos:
```suggestion
    def mask_redfish_acceess(protocol: "https", host: "host")
    # root
    stub_request(:get, "#{protocol}://#{host}#{ROOT_DATA['@odata.id']}").
      to_return(status: 200, body: JSON.generate(ROOT_DATA))
    # root.systems
    stub_request(:get, "#{protocol}://#{host}#{SYSTEMS_DATA['@odata.id']}").
      to_return(status: 200, body: JSON.generate(SYSTEMS_DATA))
    # root.system
    stub_request(:get, "#{protocol}://#{host}#{SYSTEM_DATA['@odata.id']}").
      to_return(status: 200, body: JSON.generate(SYSTEM_DATA))
  end
```

Looks like not all of these are required

test/bmc/redfish_test_helper.rb
```diff
@@ -0,0 +1,101 @@
...
+      { "Address" => "192.168.124.123",
+        "Gateway" => "192.168.124.1",
+        "SubnetMask" => "255.255.255.0",
+        "AddressOrigin" => "DHCP" },
+    ],
+    "VLAN" => {
+      "VLANEnable" => true,
+      "VLANId" => 10,
+    },
+  }.freeze
```
stejskalleos:
```suggestion
```

These constants are not used in the test

stejskalleos (changes requested):
:green_apple: Tested with [Redfish-Mockup-Server](https://github.com/DMTF/Redfish-Mockup-Server), and it seems to be working.

```shell
curl --user admin:changeme -X PUT -H "Content-Type: application/json" -d '{}' "http://localhost:8080/bmc/localhost:8000/chassis/power/cycle"
```

Red fish server log:

```
   POST: Content: type=application/json and params={}
   POST: Data: {'ResetType': 'PowerCycle'}
127.0.0.1 - - [12/Sep/2025 09:01:03] "POST /redfish/v1/Systems/BladeServer-1/Actions/ComputerSystem.Reset HTTP/1.1" 204 -
```

A few comments regarding unused code, otherwise LGTM.

test/bmc/redfish_test_helper.rb
vanou: Yes, there is non-called mock. However the reason to write unused mock in this PR is to create common Redfish test code.
This same unit test preparation code is to be used at https://github.com/theforeman/smart-proxy/pull/916 & https://github.com/theforeman/smart-proxy/pull/917 and (with hope) future Redfish related PRs.

So unused mock is included here.

test/bmc/redfish_test_helper.rb
vanou: Same reason as above for including unused constants.

test/bmc/redfish_test_helper.rb
stejskalleos: @vanou I would argue with the same reasoning as in the [capability comment](https://github.com/theforeman/smart-proxy/pull/919#discussion_r2340232396). 

PR should introduce code that is only relevant to the implemented feature. It's easier for cherry-picking, debugging, and investigating the git log.

If there are constants needed for other PRs, they should be added from those PRs.

vanou: @stejskalleos 
Thanks for reply. I understand what you concern. I agree it's easier for cherry-picking, debugging, investigating, etc. if we remove non-relevant code.

stejskalleos (approved):
:green_apple: **LGTM**

Works as expected

stejskalleos:
Thank you, @hiroshi-ikuta and @vanou, for your contribution!

PR #920, opened 2025-09-22
Fixes #38766 - Baremetal DEV env provisioning guide

stejskalleos:
Smart proxy of continuation of the Foreman guide
for local provisioning development setup.

https://github.com/theforeman/foreman/pull/10692


arvind4501:
Hi @stejskalleos , it would be great if you can also include the `config/settings.yml` file, other than that everythings looks good.

Ack, thanks

PR #923, opened 2025-10-09
PXE & iPXE configs in :tftp_root/host_config directory

stejskalleos:
Before
```
/var/lib/tftpboot
├── grub2
│   ├── grub.cfg
│   ├── grub.cfg-00:aa:aa:10:10:ee
│   ├── grub.cfg-01-00-aa-aa-10-10-ee
├── host-config
│   └── 00-aa-aa-10-10-ee
│       ├── grub2
│       │   ├── grub.cfg
│       │   ├── grub.cfg-00:aa:aa:10:10:ee
│       │   ├── grub.cfg-01-00-aa-aa-10-10-ee
└── pxelinux.cfg
    ├── 01-00-aa-aa-10-10-ee
    └── 01-00-aa-aa-10-10-ee.ipxe

```

After
```
/var/lib/tftpboot
├── grub2
│   ├── ...
├── host-config
│   └── 00-aa-aa-10-10-ee
│       ├── grub2
│       │   ├── ...
│       ├── ipxe
│       │   └── 01-00-aa-aa-10-10-ee -> ../../../pxelinux.cfg/01-00-aa-aa-10-10-ee.ipxe
│       └── pxe
│           └── 01-00-aa-aa-10-10-ee -> ../../../pxelinux.cfg/01-00-aa-aa-10-10-ee
└── pxelinux.cfg
    ├── 01-00-aa-aa-10-10-ee
    └── 01-00-aa-aa-10-10-ee.ipxe
```

### DIFF
--- a/modules/tftp/server.rb
+++ b/modules/tftp/server.rb
@@ -12,6 +12,17 @@ module Proxy::TFTP
       pxeconfig_file(mac).each do |file|
         write_file file, config
       end
+
+      host_pxe_dir = File.join(path, 'host-config', dashed_mac(mac).downcase, host_pxe_dir_name)
+      FileUtils.mkdir_p(host_pxe_dir)
+
+      pxeconfig_file(mac).each do |path|
+        ln_from = File.join(host_pxe_dir, path.split('/').last)
+        ln_to = Pathname.new(path).relative_path_from(host_pxe_dir)
+
+        FileUtils.ln_s ln_to, ln_from, force: true
+      end
+
       true
     end
 
@@ -20,6 +31,9 @@ module Proxy::TFTP
       pxeconfig_file(mac).each do |file|
         delete_file file
       end
+
+      delete_host_dir mac
+
       true
     end
 
@@ -66,6 +80,8 @@ module Proxy::TFTP
 
     def delete_host_dir(mac)
       host_dir = File.join(path, 'host-config', dashed_mac(mac).downcase)
+      return unless Dir.exist?(host_dir)
+
       logger.debug "TFTP: Removing directory '#{host_dir}'."
       FileUtils.rm_rf host_dir
     end
@@ -80,7 +96,7 @@ module Proxy::TFTP
 
   class Syslinux < Server
     def pxeconfig_dir
-      "#{path}/pxelinux.cfg"
+      File.join(path, 'pxelinux.cfg')
     end
 
     def pxe_default
@@ -89,6 +105,10 @@ module Proxy::TFTP
 
     def pxeconfig_file(mac)
       ["#{pxeconfig_dir}/01-" + dashed_mac(mac).downcase]
+    end
+
+    def host_pxe_dir_name
+      'pxe'
     end
   end
   class Pxelinux < Syslinux; end
@@ -116,8 +136,6 @@ module Proxy::TFTP
     end
 
     def default_symlinks(bootfile_suffix, pxeconfig_dir_mac)
-      pxeconfig_dir = pxeconfig_dir()
-
       grub_source = "grub#{bootfile_suffix}.efi"
       shim_source = "shim#{bootfile_suffix}.efi"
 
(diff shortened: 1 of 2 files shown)


stejskalleos:
@evgeni 

With the current implementation, the config files are duplicated. Thinking about changing it into symlinks pointing to the default directories (`grub2` and `pxelinux.cfg`)

stejskalleos:
Test failures are related to changes; I will address them once the decision on unifying the approach is made.

evgeni:
cc @jloeser -- this structure was influenced by you in 1345bb430e2ed42dc326c4916fd348b9741662c9 and I think you're the most knowledgeable here to have an opinion  :)

evgeni:
Also @goarsna

jloeser:
Why are we doing this?

So technically it's unnecessary to store them in the host specific locations. AFAIK GRUB2 is the only bootloader reading it's origin from EFI firmware in order to use it for fetching its configuration files.

For iPXE I would need to have a look, PXE is legacy BIOS so it must be hardcoded.

However, from the user's point of view and consistency may make sense to have all bootloader configurations together in one place for one specific host. But then it should be definitely symlinks so that there is only one source of truth.

stejskalleos:
> Why are we doing this?

To centralize configs in one place for each host.

> But then it should be definitely symlinks so that there is only one source of truth.

Yeah, I was thinking about it during the weekend, and I also think it should be done via symlinks.
Will update the PR.

stejskalleos:
Updated, now it creates symlinks (relative paths) to the configs.

stejskalleos:
@evgeni the CI is :green_apple: 

* Added tests
* Now only symlinks are created, no duplicate conf files

ShimShtein:
Maybe I am a bit late to the party, but creating symlinks that are not really used by the bootloaders feels a bit overwhelming. I would prefer leaving only "functional" files. The only links that make sense for me are those that link synonyms - for example if we want the same file to be served for `grub.cfg-00:aa:aa:10:10:ee` and `grub.cfg-01-00-aa-aa-10-10-ee`. This will make a single source of truth scenario, which is a good thing.
I think having the files linked/copied to an exotic directory will cause confusion when trying to reverse engineer the boot flows.

stejskalleos:
> I think having the files linked/copied to an exotic directory will cause confusion when trying to reverse engineer the boot flows.

This was the exact issue for me when I came back to the feature, but the opposite direction. I was "annoyed" that some configs are in the host_config directory, and some are not.

jloeser:
> I think having the files linked/copied to an exotic directory will cause confusion when trying to reverse engineer the boot flows.

Understood. But I agree with @stejskalleos and I also like the idea of having one place for all.

Depending on how you work/debug there is also the minimal benefit of choosing the correct host by MAC _once_ (cd into it) and then you can simply edit the loader configuration you are interested in. ATM you always need to provide/know the MAC address for every single file you want to edit.

If a user wants to explicitly modify something he can do so and it will always be applied due to sym-linking - independently of the location (that's currently not true for GRUB2 and need to be fixed!).

If a user is interested in the boot flow we maybe should explain in one two sentences how its actually working (=which paths are involved) and why we do this.

ShimShtein:
> If a user is interested in the boot flow we maybe should explain in one two sentences how its actually working (=which paths are involved) and why we do this.

Well, the general answer here is that we depend on the bootloader and configuration (e.g. DHCP bootfileurl option) here. If the bootloader expects a specific folder structure, we will have to have it in our TFTP folder. 

>If a user wants to explicitly modify something he can do so and it will always be applied due to sym-linking - independently of the location (that's currently not true for GRUB2 and need to be fixed!).

Since this PR fixes the functional part of the requirement (as you have mentioned) I have no objections to merging it. Having an extra symling doesn't feel as too much, especially if it helps even a smalll group of sysadmins.

ShimShtein:
@stejskalleos One question: why don't you also link the grub.cfg* files?
Now we have two copies of the same file: one in the `./grub2` folder and one in the `./host_config/*/grub2` one:
image
image

stejskalleos:
Updated, now **every** config is linked to the original one.
Tested Grub2 provisioning and it worked.

ShimShtein (approved):

ShimShtein:
Merged, thanks @stejskalleos !

PR #924, opened 2025-11-04
Fixes #38883 - Drop support for Ruby 2.7

evgeni:


stejskalleos (approved):
Can't test it with the Packit since it's repeatedly failing with
```
Error: 
 Problem: cannot install the best candidate for the job
  - nothing provides systemd-rpm-macros = 252-51.el9_6.3 needed by systemd-252-51.el9_6.3.x86_64 from baseos
  - 
  ```

On my DEV setup, everything works fine. I have been running Smart Proxy on `3.0.7` for some time, and did not have any errors related to the Ruby 3 version

stejskalleos:
thanks @evgeni

evgeni:
Yeah, RHEL repos in Copr are broken, for unrelated reasons, so things don't build.

But there is also nothing in this PR that would affect the packit build result, so no test required, IMHO

PR #927, opened 2025-11-17
Fixes #38909 - Use syslog gem not in windows

sbernhard:


ekohl (commented):
Could you open a Redmine issue for this?

I'd also love to have some CI on GHA that uses a Windows runner to test if it works, but lack the specific knowledge (and time). Do you have any idea how hard that would be?

.github/workflows/windows.yml
```diff
@@ -0,0 +1,100 @@
...
+        run: |
+          $ErrorActionPreference = 'Stop'
+          $repo = 'theforeman/smart-proxy'
+
+          if ('${{ github.event.inputs.version }}') {
+            $tag = '${{ github.event.inputs.version }}'
+          } else {
+            $tag = '${{ matrix.proxytag }}'
+          }
+          "tag=$tag" | Out-File -FilePath $env:GITHUB_OUTPUT -Encoding utf8 -Append
```
ekohl: Doesn't `actions/checkout` work on Windows? This really isn't all that useful if it checks out a static tag

.github/workflows/windows.yml
```diff
@@ -0,0 +1,100 @@
...
+          ruby-version: ${{ matrix.rubyversion }}
+          bundler-cache: false
+
+      - name: Install Bundler
+        run: |
+          gem install bundler --no-document
+
+      - name: Install rake
+        run: |
+          gem install rake --no-document
```
ekohl: Why does a regular `bundle install` not work? I'd hope it does
```suggestion
          bundler-cache: true
```

.github/workflows/windows.yml
```diff
@@ -0,0 +1,100 @@
...
+        run: |
+          $repo = 'theforeman/smart-proxy'
+          $tag  = '${{ steps.tag.outputs.tag }}'
+          $zip  = "$env:RUNNER_TEMP\smart-proxy-$tag.zip"
+          $url  = "https://github.com/$repo/archive/refs/tags/$tag.zip"
+          Invoke-WebRequest -Uri $url -OutFile $zip
+          Expand-Archive -Path $zip -DestinationPath $env:RUNNER_TEMP -Force
+          # Move to a stable path
+          $extracted = Get-ChildItem "$env:RUNNER_TEMP\smart-proxy-*" -Directory | Select-Object -First 1
+          Move-Item $extracted.FullName "$env:GITHUB_WORKSPACE\smart-proxy"
```
ekohl: Can't it work directly from a git checkout?

.github/workflows/windows.yml
```diff
@@ -0,0 +1,100 @@
...
+          ruby -v
+          gem -v
+          ridk version
+
+      - name: Bundle install smart-proxy
+        working-directory: smart-proxy
+        run: |
+          # Optional: keep gems in repo-local path
+          bundle config set path vendor/bundle
+          bundle config set without 'test krb5 libvirt'
```
ekohl: You can also export a variable at the top level
```yaml
env:
  BUNDLE_WITHOUT: test krb5 libvirt
```

maximiliankolb:
@sbernhard Please create a Redmine issue & have a look at the failing tests. Please ping Ewoud after GHA are :green_apple: for a final ACK.

edit: apologies; Redmine issue is open but GHA still fails. There must be another reason for that.

sbernhard:
> @sbernhard Please create a Redmine issue & have a look at the failing tests. Please ping Ewoud after GHA are 🍏 for a final ACK.
> 
> edit: apologies; Redmine issue is open but GHA still fails. There must be another reason for that.

everything done.

ekohl:
@sbernhard the Redmine check is still failing: the commit must reference the issue.

sbernhard:
Fixed it. Thanks  @ekohl

.github/workflows/windows.yml
```diff
@@ -0,0 +1,50 @@
...
+      - name: Show Ruby env (sanity)
+        run: |
+          ruby -v
+          gem -v
+          ridk version
+
+      - name: Bundle install smart-proxy
+        run: |
+          bundle config set path vendor/bundle
+          bundle install --jobs 4
```
ekohl: This is redundant if you use `bundler-cache: true` above
```suggestion
```

.github/workflows/windows.yml
```diff
@@ -0,0 +1,50 @@
...
+
+jobs:
+  bundle-smart-proxy:
+    runs-on: windows-latest
+    defaults:
+      run:
+        shell: pwsh
+    strategy:
+      matrix:
+        rubyversion: ['3.3', '3.4']
```
ekohl: I'd really like it if this build used the same matrix we support for other plaforms, which we do here:
https://github.com/theforeman/smart-proxy/blob/28f49c6c60743c8e3ef731b254225e629d6291d7/.github/workflows/smart-proxy.yml#L19-L27

Then we use that output later:
https://github.com/theforeman/smart-proxy/blob/28f49c6c60743c8e3ef731b254225e629d6291d7/.github/workflows/smart-proxy.yml#L51-L53

ekohl (commented):
Looking at https://github.com/ruby/syslog/commit/ffbfaa60f521920260da8ae23ae9b54269933e3e I wonder if it's really needed. I'd expect the gem (>= 0.2.0) to install on Windows and be a noop.

ekohl (commented):
By now this really evolved into just testing on Windows, which is still a good thing. Do you have any idea what is needed to actually make the tests work?

ekohl:
```
C:/hostedtoolcache/windows/Ruby/3.3.10/x64/lib/ruby/3.3.0/bundled_gems.rb:69:in `require': cannot load such file -- libvirt (LoadError)
```

I think this (and the krb5) related tests should be skipped. The modules themselves (both `dns_libvirt` and `dhcp_libvirt`) both correctly guard behind `load_classes` so it should be possible.

The biggest risk I see is that we silently disable the tests without knowing it, but I'm sure we can find a way around that.

sbernhard:
> I think this (and the krb5) related tests should be skipped. The modules themselves (both `dns_libvirt` and `dhcp_libvirt`) both correctly guard behind `load_classes` so it should be possible.

I don't know how to skip the tests for these plugins. Can you please let me know more what you mean regarding `load_classes`??

ekohl:
> Can you please let me know more what you mean regarding `load_classes`??

In normal operation the Smart Proxy loads things in some order:
* The entry point is the "binary": [smart-proxy](https://github.com/theforeman/smart-proxy/blob/develop/bin/smart-proxy) or [smart-proxy-win-service](https://github.com/theforeman/smart-proxy/blob/develop/bin/smart-proxy-win-service)
* This loads [smart_proxy_main](https://github.com/theforeman/smart-proxy/blob/develop/lib/smart_proxy_main.rb)

The main file is interesting because it loads all the plugins that are built in, but there's also the [smart_proxy_for_testing](https://github.com/theforeman/smart-proxy/blob/develop/lib/smart_proxy_for_testing.rb) which does NOT do that. I'll admit the exact details here are not entirely clear to me since it's been a while. I think for plugins from gems we rely on Bundler to include them, but load order is a bit vague.

However, those plugins are only the plugin definitions. Taking DNS as an example: [dns/dns](https://github.com/theforeman/smart-proxy/blob/develop/modules/dns/dns.rb) only loads [dns/configuration_loader](https://github.com/theforeman/smart-proxy/blob/develop/modules/dns/configuration_loader.rb) and [dns/dns_plugin](https://github.com/theforeman/smart-proxy/blob/develop/modules/dns/dns_plugin.rb). Zooming in on `Proxy::Dns::ConfigurationLoader` we see it defines the `load_classes` method which loads the real implementation.

It's in the [Proxy::PluginInitializer](https://github.com/theforeman/smart-proxy/blob/develop/lib/proxy/plugin_initializer.rb) that actually really initializes the plugin. This is not a trivial class because it uses dependency injection, but I still think it makes a lot of sense. It's [initialize_plugins](https://github.com/theforeman/smart-proxy/blob/867f19ec0f6c226af3a8a9f1e1e2bc19ffdd54fe/lib/proxy/plugin_initializer.rb#L159-L197) that does a bunch of steps that eventually calls [load_classes](https://github.com/theforeman/smart-proxy/blob/867f19ec0f6c226af3a8a9f1e1e2bc19ffdd54fe/lib/proxy/plugin_initializer.rb#L336-L338).

It allows for loading plugin definitions but having the plugin itself disabled via configuration. This is more efficient (less memory used) but can also help if the actual implementation is unavailable (like Kerberos or libvirt bindings that aren't installed.

But in the end this is all what runtime does. Testing is different: it loads many things unconditionally. Taking the failing libvirt test: the test library loads `test/**/*_test.rb` which in turn loads [test/dhcp_libvirt/libvirt_dhcp_network_test.rb]https://github.com/theforeman/smart-proxy/blob/develop/test/dhcp_libvirt/libvirt_dhcp_network_test.rb) which has this line:
https://github.com/theforeman/smart-proxy/blob/867f19ec0f6c226af3a8a9f1e1e2bc19ffdd54fe/test/dhcp_libvirt/libvirt_dhcp_network_test.rb#L3
At runtime this class is only loaded if it's really needed:
https://github.com/theforeman/smart-proxy/blob/867f19ec0f6c226af3a8a9f1e1e2bc19ffdd54fe/modules/dhcp_libvirt/configuration_loader.rb#L24-L25

That part is really loading the libvirt bindings:
https://github.com/theforeman/smart-proxy/blob/867f19ec0f6c226af3a8a9f1e1e2bc19ffdd54fe/modules/dhcp_libvirt/libvirt_dhcp_network.rb#L1-L2

This is all a long intro into the loading mechanism. The real challenge is to modify the test suite so that it:
* Will test libvirt bindings if available
* Skip tests if bindings are unavailable
* Not accidentally skip the tests unconditionally

I don't have a good answer on that right now.

My had a thought about looking at which bundler groups were available. I couldn't find a good API for that, but didn't dig very deep. You can probably be blunt and use:

```ruby
begin
  Bundler.require(:libvirt)
rescue LoadError # or whatever it throws
  LIBVIRT_BUNDLER_GROUP_AVAILABLE = false
else
  LIBVIRT_BUNDLER_GROUP_AVAILABLE = true
  require 'dhcp_libvirt/libvirt_dhcp_network'
end
```

Then based on https://test-unit.github.io/test-unit/en/Test/Unit/TestCaseOmissionSupport.html I'd try enhancing each `test_` method with:
```ruby
def test_dump_xml
  omit_unless(LIBVIRT_BUNDLER_GROUP_AVAILABLE) do
    a_xml = '<xml></xml>'
    @network.expects(:xml_desc).returns(a_xml)
    assert_equal a_xml, @subject.dump_xml
  end
end
```

And perhaps there's a nicer way. https://test-unit.github.io/test-unit/en/Test/Unit/TestCase.html may be helpful in seeing if that's an option.

Looking at that particular `setup` I see that also refers to `::Libvirt` so that will also fail. So I'd also see if the following works:
```ruby
class LibvirtDHCPNetworkTest < Test::Unit::TestCase
  def setup
    omit_unless(LIBVIRT_BUNDLER_GROUP_AVAILABLE)

    # rest of the existing code here
  end
end
```

Perhaps I'm also thinking too complex and perhaps [Test::Unit](https://test-unit.github.io/) already has something built in.

sbernhard:
Running into more and more error - and the syslog issue if you have a look at the end @ekohl

bundler.d/krb5.rb
```diff
@@ -2,3 +2,6 @@
   gem 'rkerberos', '>= 0.1.1'
   gem 'gssapi'
 end
+
+tests = %w[dns realm]
+ENV['SKIP_TEST_FILES'] = ((ENV['SKIP_TEST_FILES'] || "").split(",") + tests).uniq.join(",")
```
ekohl, line 7: This feels very magical and I'm not sure this is wise. bundler files are also used at runtime and the tests leaking in there is something I'd prefer to avoid.

.github/workflows/smart-proxy.yml
```diff
@@ -69,12 +69,49 @@ jobs:
...
+    runs-on: windows-latest
+    needs:
+      - setup_matrix
+    defaults:
+      run:
+        shell: pwsh
+    strategy:
+      matrix: ${{ fromJson(needs.setup_matrix.outputs.matrix) }}
+    env:
+      BUNDLE_WITHOUT: "krb5 libvirt"
```
ekohl, line 83: https://github.com/theforeman/smart-proxy/pull/884 may be a better way to disable installing the gems depending on the platform. Also makes the real deployment easier.

ekohl (commented):
Looking at the tests I see we hardcode `/tmp` a lot in our tests and we should really avoid that. At least 1 of the logger failures is because of that.

I also took a stab at implementing the changes I suggested and https://github.com/ekohl/smart-proxy/commit/f449ee027ed48f73dc27a6625048172dae69a17e is sort of that. It's incomplete, but I need to head out and wanted to at least share the concept.

PR #928, opened 2025-12-28
Support Ruby 4.0

ekohl:
Merging this will start testing Ruby 4.0 in all our plugins. It would be good to open PRs against some of the repos in https://theforeman.github.io/foreman-plugin-overview/#foreman-proxy and focus on the ones included in the installer since those are the most active.

Because I haven't done that, this is still a draft.


bastelfreak:
This is passing on smart_proxy_openbolt: https://github.com/overlookinfra/smart_proxy_openbolt/pull/26. And on smart_proxy_ansible: https://github.com/theforeman/smart_proxy_ansible/pull/110. On how many plugins do we want to test this before merging it?

ekohl:
Foreman is in the stabilization week (https://community.theforeman.org/t/foreman-3-18-branching-process/45265) and I'd prefer to wait with this change until after 3.18.0 RC1 is out.

smart_proxy.gemspec
```diff
@@ -15,6 +15,7 @@ Gem::Specification.new do |s|
   s.required_ruby_version = '>= 3.0'
   s.add_dependency 'base64'
   s.add_dependency 'json'
+  s.add_dependency 'logger'
```
ogajduse, line 18: To satisfy the logger dependency with Ruby 4, we'll need to package it since it became a bundled gem in Ruby 4.
https://stdgems.org/logger/

I've attempted to add the package in packaging, but I realized we don't need it yet.
https://github.com/theforeman/foreman-packaging/pull/13055

Is there a Redmine issue to track Ruby 4 support changes?

ekohl: > To satisfy the logger dependency with Ruby 4, we'll need to package it since it became a bundled gem in Ruby 4.
> https://stdgems.org/logger/

We don't need to until we package for Ruby 4. https://github.com/theforeman/foreman-packaging/tree/rpm/develop?tab=readme-ov-file#handle-gems-that-are-default-ruby-gems describes how we work around it. By the time we do move to Ruby 4 we'll automatically get a repoclosure failure.

> Is there a Redmine issue to track Ruby 4 support changes?

Not yet. This was a test right after Ruby 4.0 was released to verify how compatible we are. Now that we're beyond branching I think we could look at enabling so we at least run CI.

PR #929, opened 2026-01-09
document how to setup vbmc

evgeni:


docs/bmc.md
```diff
@@ -18,6 +18,32 @@ This is a set of simple simulation tools aimed at supporting the development and
...
+```
+$ vbmc add --libvirt-uri qemu:///system libvirt_domain_name
+```
+
+The username and password of the BMC can be changed using `--username` and `--password`.
+The port can't be changed with `--port` as Foreman doesn't allow configuring any other port than `623/udp`
+
+You might need to allow `623/udp` in your firewall: `sudo firewall-cmd --zone=libvirt --add-port=623/udp`.
+
+If you're running `vbmcd` as an unprivileged user, you'll need to allow it to bind to 623: `sudo sysctl net.ipv4.ip_unprivileged_port_start=0`.
```
ekohl, line 47: Fedora ships a [vbmcd.service](https://src.fedoraproject.org/rpms/python-virtualbmc/blob/rawhide/f/vbmcd.service) that runs as `vbmcd` but I don't see a permission to start on a different port. How does it bind to port 623?

ekohl (approved):

docs/bmc.md
evgeni: it doesn't by default I guess?

evgeni: vbmcd itself runs on 50891 for client communications, and it's up to the user to make it bind the right BMC port once the bmc was defined

PR #932, opened 2026-01-23
Add logger as a dependency

bastelfreak:
This gem was vendored in Ruby until version 4.

```
$ git grep require.*logger
lib/proxy/log.rb:require 'logger'
lib/proxy/log.rb:      require 'syslog/logger'
test/log_buffer/decorator_test.rb:    require 'syslog/logger'
```

### DIFF
--- a/smart_proxy.gemspec
+++ b/smart_proxy.gemspec
@@ -15,6 +15,7 @@ Gem::Specification.new do |s|
   s.required_ruby_version = '>= 3.0'
   s.add_dependency 'base64'
   s.add_dependency 'json'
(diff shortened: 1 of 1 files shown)


ekohl (commented):
I'd prefer to include it in https://github.com/theforeman/smart-proxy/pull/928 if this is needed (and I think you're right it is).

bastelfreak:
Feel free to pull it into your PR :)

ekohl:
Will do. And it's precisely the reason why testing all the dependencies is a good step :)

PR #935, opened 2026-04-01
Fixes #39208 - Cache global registration script in smart-proxy

pablomh:
## Problem

`GET /register` returns the same shell script for all hosts sharing the same registration parameters (org, location, hostgroup, activation keys). Under bulk registration — 100+ hosts hitting the same capsule simultaneously — this endpoint is called once per host, each time proxying to Foreman and waiting for the ERB template to render (~103ms in profiling). With WEBrick's MaxClients=100, all threads can be held waiting for identical Foreman responses.

## Solution

Cache the rendered script in memory (5-minute TTL) keyed on a canonical form of the request query string (parameters sorted alphabetically, so requests differing only in parameter order share the same cache entry).

Per-key double-checked locking prevents the thundering herd on a cold cache while allowing concurrent requests for genuinely different keys (e.g. different activation keys) to fetch from Foreman in parallel without blocking each other:

- Fast path: unsynchronised Concurrent::Map read, no lock acquired
- Slow path: key-specific Mutex serialises only competing threads; re-checks under lock before fetching from Foreman

The per-key Mutex is evicted from KEY_MUTEXES immediately after the script is written to cache, so KEY_MUTEXES is empty under steady state and bounded to in-flight fetches only.

Only 2xx responses are cached — errors never poison the cache. SCRIPT_CACHE uses Concurrent::Map (already a smart-proxy dependency) for lock-free reads that are safe on all Ruby VMs.

## Test plan

- [x] Cache hit: Foreman called once for repeated identical requests
- [x] Per-key isolation: different parameter sets cached independently
- [x] Parameter order independence: requests differing only in param order share the same cache entry
- [x] TTL expiry: expired entries are not served; Foreman is re-called
- [x] Mutex eviction: KEY_MUTEXES is empty after a successful cache write
- [x] Error non-caching: Foreman is called on every request when it errors

Fixes #39208

### DIFF
--- a/modules/registration/registration_api.rb
+++ b/modules/registration/registration_api.rb
@@ -1,9 +1,51 @@
 require 'registration/proxy_request'
 
 class Proxy::Registration::Api < ::Sinatra::Base
+  # Cache for the global registration script (GET /register).
+  #
+  # The script is identical for all hosts sharing the same registration
+  # parameters (org, location, hostgroup, activation keys), making it safe
+  # to serve from an in-memory cache during concurrent bulk registration.
+  #
+  # Per-key double-checked locking prevents thundering herd while allowing
+  # genuinely independent cache keys (e.g. different activation keys) to
+  # fetch from Foreman in parallel.
+  #
+  # Only HTTP 200 responses are cached — errors are raised out of the cache
+  # block so they never poison the cache. The per-key mutex is evicted
+  # immediately after caching so KEY_MUTEXES stays bounded to keys currently
+  # being fetched for the first time (zero under steady state).
+  REGISTRATION_SCRIPT_CACHE_TTL = 5 * 60 # seconds
+  KEY_MUTEXES                   = Concurrent::Map.new
+  SCRIPT_CACHE                  = Concurrent::Map.new
+
+  class ScriptFetchError < StandardError
+    attr_reader :response
+
+    def initialize(response)
+      super()
+      @response = response
+    end
+  end
+
+  class << self
+    def registration_script_cache
+      SCRIPT_CACHE
+    end
+
+    def key_mutex(cache_key)
+      KEY_MUTEXES.compute_if_absent(cache_key) { Mutex.new }
+    end
+
+    def evict_key_mutex(cache_key)
+      KEY_MUTEXES.delete(cache_key)
+    end
+  end
+
   get '/' do
-    response = Proxy::Registration::ProxyRequest.new.global_register(request)
-    handle_response(response)
+    registration_script
+  rescue ScriptFetchError => e
+    handle_response(e.response)
   rescue StandardError => e
     logger.exception "Error when rendering Global Registration Template", e
     render_error(default_error_msg)
@@ -19,11 +61,41 @@ class Proxy::Registration::Api < ::Sinatra::Base
 
   private
 
+  def registration_script
+    cache_key = Rack::Utils.build_query(
+      Rack::Utils.parse_nested_query(request.query_string).sort_by { |k, _| k }
+    )
+    cache(cache_key) do
+      response = Proxy::Registration::ProxyRequest.new.global_register(request)
+      raise ScriptFetchError, response unless response.code == '200'
+      response.body
+    end
+  end
+
+  def cache(key, &block)
+    value = read_registration_cache(key)
+    return value if value
+
+    self.class.key_mutex(key).synchronize do
+      value = read_registration_cache(key)
+      return value if value
+
+      result = yield
+      self.class.registration_script_cache[key] = { body: result, at: Time.now }
+      self.class.evict_key_mutex(key)
+      result
+    end
+  end
+
+  def read_registration_cache(cache_key)
+    entry = self.class.registration_script_cache[cache_key]
+    entry[:body] if entry && (Time.now - entry[:at]) < REGISTRATION_SCRIPT_CACHE_TTL
+  end
+
   def handle_response(response)
-    if response.code.start_with? '2'
+    if response.code.start_with?('2')
       response.body
     else
-      # Return error message only if it is not HTML.
       message = response["content-type"].include?('text/plain') ? response.body : default_error_msg
       render_error(message, code: response.code)
     end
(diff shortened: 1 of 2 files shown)


modules/registration/registration_api.rb
```diff
@@ -19,11 +64,35 @@ class Proxy::Registration::Api < ::Sinatra::Base
...
+  def fetch_and_cache_script(cache_key)
+    self.class.key_mutex(cache_key).synchronize do
+      # Re-check under lock: another thread with the same key may have
+      # already populated the cache while we were waiting.
+      cached = read_registration_cache(cache_key)
+      return cached if cached
+
+      response = Proxy::Registration::ProxyRequest.new.global_register(request)
+
+      if response.code.start_with?('2')
```
ekohl: If this is HTTP 201 or 202 I wouldn't expect it to be cached. Only HTTP 200.

ekohl (commented):
In Ruby you can pass blocks and that creates a nice abstraction mechanism. In somewhat pseudo-code:
```ruby
CACHE = {}
MUTEX = Mutex.new

def cache(key, &block)
  if (value = CACHE[key])
    value
  else
    MUTEX.synchronize do
      CACHE[key] = block.call
    end
  end
end 

def get_value(query)
  cache(query) do
    real_call_here(query)
  end
end
```

You may need to do something like raising an exception to handle errors from Foreman code (in `handle_response`) so it doesn't get cached.

modules/registration/registration_api.rb
pablomh: Already handled — line `raise ScriptFetchError, response unless response.code == '200'` ensures only 200 responses enter the cache. Non-200 responses raise out of the cache block and are returned directly via `handle_response`. Covered by `test_global_register_does_not_cache_errors`.

pablomh:
Thanks — we followed exactly this pattern. The `cache(key, &block)` method uses double-checked locking with per-key mutexes via `Concurrent::Map`, and errors raise out of the block so they never poison the cache:

```ruby
def cache(key, &block)
  value = read_registration_cache(key)
  return value if value

  self.class.key_mutex(key).synchronize do
    value = read_registration_cache(key)
    return value if value

    result = yield
    self.class.registration_script_cache[key] = { body: result, at: Time.now }
    self.class.evict_key_mutex(key)
    result
  end
end
```

The caller:
```ruby
cache(cache_key) do
  response = Proxy::Registration::ProxyRequest.new.global_register(request)
  raise ScriptFetchError, response unless response.code == '200'
  response.body
end
```

Non-200 responses raise `ScriptFetchError` which is caught outside the cache block and passed to `handle_response`. Covered by tests including `test_global_register_does_not_cache_errors`.

pablomh:
I pushed a follow-up refinement on top of this branch.

Main changes:
- cache key now includes **auth context**, not just normalized query params
- TTL now uses **monotonic time** instead of `Time.now`
- per-key mutexes are retained instead of evicted after successful writes
- expired cache entries are removed on read
- added tests for auth-aware cache partitioning and stale-entry cleanup

The main motivation was that two requests with the same query string but different auth can produce different rendered scripts, since Foreman embeds a user-derived token in the registration script. With a query-only cache key, those could incorrectly share a cache entry.

For the auth partitioning, I currently used the **raw `Authorization` header** for exact separation and zero hash cost. I left the SHA-256 alternative commented right below the helper so we can choose which tradeoff we prefer in review:

- raw header: exact and cheap
- SHA-256: avoids raw credentials in cache keys

Happy to switch that part if there’s a preference.

pablomh:
Ping?

pablomh:
Added identity-safe mutex eviction: after the cache entry is written, the per-key mutex is removed from `KEY_MUTEXES` since the cache entry itself prevents redundant fetches until TTL expiry. The `ensure` block guarantees cleanup even if the fetch raises, and `equal?` prevents accidentally deleting a replacement mutex created by a concurrent miss for the same key.

Without this, `KEY_MUTEXES` grows with every unique cache key and never shrinks. Practically bounded for typical deployments (few activation key combinations), but prevents unbounded growth if auth-aware cache keys produce high cardinality.

jeremylenz (commented):
I don't know if this is a pattern that's followed elsewhere, but could this also be ~proxied~ cached via apache rather than the application?

pablomh:
> I don't know if this is a pattern that's followed elsewhere, but could this also be proxied via apache rather than the application?

Proxied or cached?

jeremylenz:
oh I meant cached, yeah

pablomh:
I haven't seen it done at the proxy level yet.

pablomh:
Friendly ping when you have a moment: I addressed the earlier feedback and the branch is ready for another look. I'd really appreciate a review when convenient. Thanks!

pablomh:
Regarding your question and an "extended" answer: I know we can do it at the Apache level, but we haven't done it neither in smart-proxy nor in the main foreman server, so I'm hesitant to introduce a change in the architecture at that level.

jeremylenz:
> Regarding your question and an "extended" answer: I know we can do it at the Apache level, but we haven't done it neither in smart-proxy nor in the main foreman server, so I'm hesitant to introduce a change in the architecture at that level.

@ehelms I would like to get your thoughts on this (caching the global reg script, and where that caching should live.)

pablomh:
...and when I find content for which we need to have authentication, that usually means that we'd need to replicate all the authentication at the Apache level, so in those cases I prefer to add the caching at the layer that manages the content.

ehelms:
> `GET /register` returns the same shell script for all hosts sharing the same registration parameters (org, location, hostgroup, activation keys).

What happens if the registration contains other configuration parameters beyond those 4?

pablomh:
Good catch, that sentence is imprecise. The cache key is not limited to those 4 fields; the implementation uses the full normalized query string, so any additional registration configuration passed as query params gets its own cache entry as well.

On top of that, the cache is also partitioned by auth context, since the rendered script can differ there too. Those 4 params were just the motivating examples from the bulk-registration case, not the full key definition.

I can update the PR description to make that clearer.

ehelms:
I think this is a good performance improvement for those scenarios mentioned, and that the 5 minute TTL feels right.

ehelms (approved):
Design wise from me, I gave the code a cursory review.

jeremylenz (approved):
LGTM 👍

PR #936, opened 2026-04-13
Fixes #39229 - Add foreman_request_timeout setting to control Net::HTTP read timeout

pablomh:
## Problem

`ForemanRequest` uses a bare `Net::HTTP` instance whose `read_timeout` defaults to Ruby's hardcoded 60s. Under high-concurrency registration load this causes 500 errors when Foreman takes longer than 60s to process `POST /register`.

## Fix

Introduces `:foreman_request_timeout` and documents it in `settings.yml.example`, then applies it to `http.read_timeout` in `ForemanRequest#http_init`.

`subscription-manager`'s own client-side `server_timeout` already defaults to 180s, so smart-proxy cutting the connection at 60s was strictly worse than useless — it turned a slow-but-recoverable request into a hard failure the client wasn't even asking for yet. The setting now defaults to **180s** when unset, matching subscription-manager, instead of silently inheriting Ruby's unrelated 60s default. `:foreman_request_timeout: 0` is kept as an escape hatch back to that 60s behavior.

Validated with an isolated A/B in a lab environment, ramping concurrent registrations up, on both an RPM/puppet-based deployment and a containerized (foremanctl) deployment: raising the timeout to 180s reduced failures in the moderate-to-high concurrency range, though it doesn't help at the very top end where failures come from genuine backend saturation rather than this timeout.

Fixes https://projects.theforeman.org/issues/39229

### DIFF
--- a/lib/proxy/request.rb
+++ b/lib/proxy/request.rb
@@ -66,6 +66,9 @@ module Proxy::HttpRequest
       http             = Net::HTTP.new(uri.host, uri.port)
       http.use_ssl     = uri.scheme == 'https'
       http.verify_mode = OpenSSL::SSL::VERIFY_NONE
+      if Proxy::SETTINGS.foreman_request_timeout.to_i > 0
+        http.read_timeout = Proxy::SETTINGS.foreman_request_timeout.to_i
+      end
 
       if http.use_ssl?
         ca_file = Proxy::SETTINGS.foreman_ssl_ca || Proxy::SETTINGS.ssl_ca_file
--- a/config/settings.yml.example
+++ b/config/settings.yml.example
@@ -42,6 +42,12 @@
 #:foreman_ssl_cert: ssl/certs/fqdn.pem
 #:foreman_ssl_key: ssl/private_keys/fqdn.pem
 
+# Timeout in seconds for HTTP requests from smart-proxy to Foreman (e.g. POST /register).
+# Defaults to Ruby's Net::HTTP read timeout (60s) when unset. Consider raising this
+# above your client's timeout to avoid cutting connections the client is still waiting on.
+# subscription-manager's default server_timeout is 180s, so a value above that is recommended.
+#:foreman_request_timeout: 60
+
 # host and ports configuration
 # an array of interfaces to bind ports to (possible values: *, localhost, 0.0.0.0)
 #:bind_host: ['*']
--- a/test/request_test.rb
+++ b/test/request_test.rb
@@ -54,6 +54,18 @@ class RequestTest < Test::Unit::TestCase
     assert_equal("body", result.body)
   end
 
+  def test_read_timeout_applied_when_foreman_request_timeout_configured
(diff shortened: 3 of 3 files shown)


pablomh:
## CI failure: rake 13.4.2 regression (not related to this PR)

All test jobs fail with `rake_test_loader: version unknown` and exit code 1. The test runner never actually runs any tests.

**Root cause:** `gem 'rake'` in `bundler.d/test.rb` has no version constraint and no `Gemfile.lock` is committed, so CI resolves the latest version from rubygems.org. Rake 13.4.2 (released April 16) changed `Rake::TestTask#option_list` to append `-v` to the test command when `verbose = true` — previously this flag was not passed to the test runner. The new `-v` causes `test-unit 3.7.7` to fail silently.

This affects all new PRs on smart-proxy, not just this one. PR #935 passed because its bundle cache still had rake 13.3.1.

**Diff in `lib/rake/testtask.rb` (13.3.1 → 13.4.2):**

```ruby
# 13.3.1 — option_list never adds -v
def option_list
  (ENV["TESTOPTS"] || ENV["TESTOPT"] || ENV["TEST_OPTS"] || ENV["TEST_OPT"] || @options || "")
end

# 13.4.2 — now adds -v when verbose is true
def option_list(verbose: @verbose)
  opts = ENV["TESTOPTS"] || ENV["TESTOPT"] || ENV["TEST_OPTS"] || ENV["TEST_OPT"] || @options || ""
  if verbose && !opts.split.include?("-v")
    opts = opts.empty? ? "-v" : "#{opts} -v"
  end
  opts
end
```

**Fix options:**
- Pin `gem 'rake', '< 13.4'` in `bundler.d/test.rb` (workaround)
- Report the regression upstream to [ruby/rake](https://github.com/ruby/rake)
- Or fix the `test-unit` interaction with `-v` if the issue is on that side

lib/proxy/request.rb
```diff
@@ -66,6 +66,10 @@ def http_init
       http             = Net::HTTP.new(uri.host, uri.port)
       http.use_ssl     = uri.scheme == 'https'
       http.verify_mode = OpenSSL::SSL::VERIFY_NONE
+      if Proxy::SETTINGS.respond_to?(:foreman_request_timeout) &&
+         Proxy::SETTINGS.foreman_request_timeout.to_i > 0
```
ekohl: `respond_to?` is redundant because Ruby's `OpenStruct` responds to `nil` on unknown properties.

```suggestion
      if Proxy::SETTINGS.foreman_request_timeout.to_i > 0
```

ekohl (changes requested):
Makes sense, but 1 minor comment inline.

lib/proxy/request.rb
pablomh: Thanks @ekohl. Done :)

pablomh:
Friendly ping.

pablomh:
Ping.

PR #937, opened 2026-04-19
Implement bootloader-universe script

ekohl:
This script downloads the grub and shim files and places them in the correct locations.

It accepts a repo URL so it can be easily downloaded from Pulp repositories.

It has various TODOs and is incomplete, but helps with providing the structure. Long term this is better to be included in the TFTP module itself and exposed via a REST API call. Sharing this as a draft PR after @stejskalleos and I discussed this.

### DIFF
--- a/extra/bootloader-universe
+++ b/extra/bootloader-universe
@@ -1,0 +1,68 @@
+#!/bin/bash
+
+set -xe
+
+# TODO: read from Smart Proxy config?
+TFTPROOT=/var/lib/tftpboot
+TFTPROOT=$PWD/tftpboot
+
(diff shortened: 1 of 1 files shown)


lzap:
For the record, I implemented a dedicated downloader for RHEL distributions versions 7-10. I made a comment at https://access.redhat.com/solutions/7138202 so Satellite customers can easily find it. The proper solution fully integrated with Foreman is currently available at https://github.com/theforeman/smart-proxy/pull/958

https://gist.github.com/lzap/91257190cd9d4db597e02e418f40572d

PR #939, opened 2026-05-08
Fixes #39300 - Fix Redfish identify LED handling

spesnova717:
## Summary

Fix incorrect `IndicatorLED` value and use ETag-aware PATCH for identify LED operations in the Redfish BMC provider.

### Bug fix: IndicatorLED value
The [Redfish specification (DSP0266_1.23.1.pdf)](https://www.dmtf.org/sites/default/files/standards/documents/DSP0266_1.23.1.pdf) defines `IndicatorLED` accepted values as `Lit`, `Off`, and `Blinking`.
The previous implementation used `'On'` in `identifyon`, which is **not a valid Redfish value** and would be rejected by spec-compliant BMCs.

This PR changes it to the correct value `'Lit'`.

### Improvement: use `patch_if_match` for identify operations
`identifyon` and `identifyoff` previously used `system.patch(payload: ...)`, which does not include ETag concurrency control.
Other state-changing operations like `bootdevice` already use `patch_if_match`.
This PR aligns the identify LED operations to also use `patch_if_match` for consistency and to avoid potential race conditions.

### DIFF
--- a/modules/bmc/redfish.rb
+++ b/modules/bmc/redfish.rb
@@ -59,11 +59,11 @@ module Proxy
       end
 
       def identifyon
-        system.patch(payload: { 'IndicatorLED' => 'On' })
+        system.patch_if_match({ 'IndicatorLED' => 'Lit' })
       end
 
       def identifyoff
-        system.patch(payload: { 'IndicatorLED' => 'Off' })
+        system.patch_if_match({ 'IndicatorLED' => 'Off' })
       end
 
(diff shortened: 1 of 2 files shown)


stejskalleos (approved):
🍏 LGTM

* The scope of change is clear and to the point
* Tests are present
* CI is green

stejskalleos:
Thanks @spesnova717 !

PR #941, opened 2026-05-12
Fix tests

ofedoren:


Rakefile
```diff
@@ -17,7 +17,7 @@ Rake::TestTask.new(:test) do |t|
   t.libs << 'lib'
   t.libs << 'test'
   t.test_files = FileList['test/**/*_test.rb']
-  t.verbose = true
+  t.options = "--verbose"
```
ofedoren, line 20: As an alternative we could pin rake here. I admit, the fix is from Claude:

```
The fix is one of:

 - Pin rake to ~> 13.3.0 in bundler.d/test.rb to avoid the new behavior
 - Remove t.verbose = true from the Rakefile (or set it to false) and use TESTOPTS="--verbose" when needed
 - Add -v handling to test-unit (upstream fix)
 - Set t.verbose = true but add t.options = "--verbose" which test-unit's autorunner does understand, preventing the OptionParser -v → --version mapping
```

ofedoren:
Closing in favor of https://github.com/theforeman/smart-proxy/pull/938

PR #943, opened 2026-05-14
Expose bmc_default_provider setting via v2/features API

arvind4501:
The bmc_default_provider setting is used as a fallback when no bmc_provider param is passed in BMC API requests. Exposing it allows Foreman to discover the configured default provider from the features endpoint.


evgeni (approved):
Tested with foreman-proxy-3.20.0-0.20260514075140831453.pr943.738.g01b066f.el9.noarch
- When BMC feature is off, returns `"settings":{"bmc_default_provider":null}`
- When BMC feature is on, returns `"settings":{"bmc_default_provider":"ipmitool"}` (as the config dictates)

PR #944, opened 2026-05-26
Fixes #39359 - Add shared Faraday transport for Foreman requests

pablomh:
## Summary
- replace the per-request outbound Foreman client path with a shared Faraday transport using faraday-net_http_persistent
- reuse persistent upstream connections on the smart-proxy to Foreman or Satellite leg while keeping the existing request-building API stable
- add layered connect and read timeout support plus a cleaner seam for future TLS and PQC transport changes

## Test plan
- [x] Run ruby -c lib/proxy/foreman_transport.rb
- [x] Run ruby -c lib/proxy/request.rb
- [x] Run ruby -c test/request_test.rb
- [x] Run RuboCop on changed files with Ruby 3.2
- [ ] Run focused request tests in a full project Ruby and Bundler environment
- [ ] Run broader smart-proxy test coverage in CI

### DIFF
--- a/lib/proxy/foreman_transport.rb
+++ b/lib/proxy/foreman_transport.rb
@@ -1,0 +1,127 @@
+require 'concurrent/map'
+require 'faraday'
+require 'faraday/net_http_persistent'
+require 'net/http'
+require 'openssl'
+
+module Proxy::HttpRequest
+  class ForemanTransport
+    TRANSPORT_RETRY_EXCEPTIONS = [
+      Faraday::SSLError,
+      Faraday::ConnectionFailed,
+    ].freeze
+
+    class << self
+      def connection_cache
+        @connection_cache ||= Concurrent::Map.new
+      end
+
+      def reset!
+        @connection_cache = Concurrent::Map.new
+      end
+
+      def invalidate!(uri = nil)
+        return reset! if uri.nil?
+
+        connection_cache.delete(connection_cache_key(uri))
+      end
+
+      def connection_for(uri)
+        connection_cache.compute_if_absent(connection_cache_key(uri)) { build_connection(uri) }
+      end
+
+      def run_request(uri, request)
+        attempt = 0
+        begin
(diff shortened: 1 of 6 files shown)


lib/proxy/foreman_transport.rb
```diff
@@ -0,0 +1,174 @@
...
+          options[:ca_file] = ssl_paths[:ca_file]
+          options[:verify] = true
+        end
+
+        if ssl_paths[:certificate] && ssl_paths[:private_key]
+          options[:client_cert] = OpenSSL::X509::Certificate.new(File.read(ssl_paths[:certificate]))
+          options[:client_key] = OpenSSL::PKey.read(File.read(ssl_paths[:private_key]), nil)
+        end
+
+        apply_future_tls_profile!(options)
```
ekohl: IMHO this shouldn't be included. We've been steering at using system wide crypto-policies and avoid using application config within Foreman. Smart Proxy isn't quite there today, but application code shouldn't be aware of it if possible.

ekohl (changes requested):
What's the advantage of using all of this custom code over plain Faraday? IMHO this makes it complicated to review. Is it an idea to first rewrite it to Faraday and only then add connection pooling?

I see a lot of custom classes and I'd prefer to pass options to Faraday like in https://lostisland.github.io/faraday/#/customization/connection-options.

Then I also know that we have various HTTP "implementations". Perhaps it's also good to do some research on how to unify those. We have those because we didn't include an external dependency, but if you're adding that then it should be considered.

pablomh:
Quick cleanup/update for review:

- I reshaped the branch into two commits so the Faraday introduction and the shared connection pooling are easier to review independently.
- The first commit now covers the Faraday migration plus the explicit `foreman_open_timeout` handling.
- The second commit contains the shared transport/pooling change.

The goal here was to keep the diff easier to reason about and make it simpler to split the pooling part later if that is requested.

PR #946, opened 2026-06-01
Fixes #39320 - Expose bmc_default_provider setting via v2/features API

evgeni:
(cherry picked from commit eaa7d95432f2bb45685aff24f6fa6f7bb543830e)


evgeni:
@ogajduse @arvind4501 do you think we can pick that into 3.19? would make it easier on the foremanctl side as then we don't need to differentiate between 3.19 and newer in the test suite

arvind4501 (approved):
Yes, make sense to cherry-pick

ogajduse (approved):

PR #947, opened 2026-06-08
Fixes #39405 - Add tls_min_version and tls_ciphers settings

adamlazik1:
Made with Cursor.

### DIFF
--- a/lib/launcher.rb
+++ b/lib/launcher.rb
@@ -1,12 +1,17 @@
+require 'openssl'
 require 'proxy/log'
 require 'proxy/settings'
 require 'proxy/signal_handler'
 require 'proxy/log_buffer/trace_decorator'
 require 'sd_notify'
 
-CIPHERS = ['ECDHE-RSA-AES128-GCM-SHA256', 'ECDHE-RSA-AES256-GCM-SHA384',
-           'AES128-GCM-SHA256', 'AES256-GCM-SHA384', 'AES128-SHA256',
-           'AES256-SHA256', 'AES128-SHA', 'AES256-SHA'].freeze
+CRYPTO_POLICIES_CONFIG = '/etc/crypto-policies/back-ends/opensslcnf.config'.freeze
+TLS_MIN_VERSION_MAP = {
+  '1.0' => OpenSSL::SSL::TLS1_VERSION,
+  '1.1' => OpenSSL::SSL::TLS1_1_VERSION,
+  '1.2' => OpenSSL::SSL::TLS1_2_VERSION,
+  '1.3' => OpenSSL::SSL::TLS1_3_VERSION,
+}.freeze
 
 module Proxy
   class Launcher
@@ -66,26 +71,8 @@ module Proxy
         plugins.each { |p| instance_eval(p.https_rackup) }
       end
 
-      ssl_options = OpenSSL::SSL::SSLContext::DEFAULT_PARAMS[:options]
-      ssl_options |= OpenSSL::SSL::OP_CIPHER_SERVER_PREFERENCE if defined?(OpenSSL::SSL::OP_CIPHER_SERVER_PREFERENCE)
-      # This is required to disable SSLv3 on Ruby 1.8.7
-      ssl_options |= OpenSSL::SSL::OP_NO_SSLv2 if defined?(OpenSSL::SSL::OP_NO_SSLv2)
-      ssl_options |= OpenSSL::SSL::OP_NO_SSLv3 if defined?(OpenSSL::SSL::OP_NO_SSLv3)
-      ssl_options |= OpenSSL::SSL::OP_NO_TLSv1 if defined?(OpenSSL::SSL::OP_NO_TLSv1)
-      ssl_options |= OpenSSL::SSL::OP_NO_TLSv1_1 if defined?(OpenSSL::SSL::OP_NO_TLSv1_1)
-      # Disable client initiated renegotiation
-      ssl_options |= OpenSSL::SSL::OP_NO_RENEGOTIATION if defined?(OpenSSL::SSL::OP_NO_RENEGOTIATION)
-
-      Proxy::SETTINGS.tls_disabled_versions&.each do |version|
-        constant = OpenSSL::SSL.const_get("OP_NO_TLSv#{version.to_s.tr('.', '_')}") rescue nil
-
-        if constant
-          logger.info "TLSv#{version} will be disabled."
-          ssl_options |= constant
-        else
-          logger.warn "TLSv#{version} was not found."
-        end
-      end
+      tls_ciphers = resolve_tls_ciphers
+      cipher_list, ciphersuites = validate_tls_ciphers!(tls_ciphers)
 
       https_settings = {
         :app => app,
@@ -96,10 +83,91 @@ module Proxy
         :SSLPrivateKey => load_ssl_private_key(settings.ssl_private_key),
         :SSLCertificate => load_ssl_certificate(settings.ssl_certificate),
         :SSLCACertificateFile => settings.ssl_ca_file,
-        :SSLOptions => ssl_options,
-        :SSLCiphers => CIPHERS - Proxy::SETTINGS.ssl_disabled_ciphers,
+        :SSLOptions => build_ssl_options,
+        :SSLCiphers => cipher_list,
+        :SSLCiphersuites => ciphersuites,
+        :SSLMinVersion => resolve_tls_min_version,
       }
       base_app_settings.merge(https_settings)
+    end
+
+    def build_ssl_options
+      ssl_options = OpenSSL::SSL::SSLContext::DEFAULT_PARAMS[:options]
+      ssl_options |= OpenSSL::SSL::OP_CIPHER_SERVER_PREFERENCE
+      # Disable client initiated renegotiation
+      ssl_options |= OpenSSL::SSL::OP_NO_RENEGOTIATION
+
+      ssl_options
+    end
+
+    def resolve_tls_min_version
+      return nil unless settings.tls_min_version
+
+      min_version = settings.tls_min_version.to_s
+      unless TLS_MIN_VERSION_MAP.key?(min_version)
+        raise "Invalid tls_min_version '#{min_version}'. Valid values: #{TLS_MIN_VERSION_MAP.keys.join(', ')}"
+      end
+
+      logger.info "Setting minimum TLS version to #{min_version}."
+      TLS_MIN_VERSION_MAP[min_version]
+    end
+
+    def resolve_tls_ciphers
+      configured = settings.tls_ciphers
+      raise "Invalid tls_ciphers value '#{configured}': must be a String" if !configured.nil? && !configured.is_a?(String)
+      return nil if configured&.empty?
+      return configured unless configured.nil?
+
+      if File.exist?(CRYPTO_POLICIES_CONFIG)
+        logger.info "Crypto-policies detected, using PROFILE=SYSTEM for TLS ciphers."
+        'PROFILE=SYSTEM'
+      else
+        logger.debug "No crypto-policies detected, using HIGH cipher string as default."
+        'HIGH'
+      end
+    end
+
+    def validate_tls_ciphers!(ciphers)
+      return [nil, nil] if ciphers.nil?
+
+      if ciphers == 'PROFILE=SYSTEM' && !settings.tls_min_version.nil? && settings.tls_min_version != ''
+        logger.warn "tls_min_version is configured together with tls_ciphers 'PROFILE=SYSTEM'. " \
+                    "The system crypto policy minimum TLS version may be overridden by this setting."
+      end
+
+      cipher_list = begin
+        OpenSSL::SSL::SSLContext.new.ciphers = ciphers
+        ciphers
+      rescue OpenSSL::SSL::SSLError
+        nil
+      end
+
+      if OpenSSL::SSL::SSLContext.method_defined?(:ciphersuites=)
+        ciphersuites = begin
+          OpenSSL::SSL::SSLContext.new.ciphersuites = ciphers
+          ciphers
+        rescue OpenSSL::SSL::SSLError
+          nil
+        end
+      elsif settings.tls_ciphers
+        logger.warn "tls_ciphers is configured but this Ruby/OpenSSL build does not support " \
+                    "OpenSSL::SSL::SSLContext#ciphersuites=. TLS 1.3 connections will not be " \
+                    "restricted by tls_ciphers."
+      end
+
+      # If the cipher list and ciphersuites are not valid, return the original string to let the SSL server fail at startup
+      return [ciphers, nil] unless cipher_list || ciphersuites
+
+      if cipher_list.nil? && ciphersuites && settings.tls_min_version != '1.3'
+        logger.warn "tls_ciphers '#{ciphers}' is only valid for TLS 1.3. " \
+                    "Set tls_min_version to '1.3' to prevent TLS 1.2 and lower connections with unrestricted ciphers."
+      end
+
+      if settings.tls_ciphers && settings.tls_min_version == '1.3' && ciphersuites.nil?
+        raise "tls_ciphers '#{ciphers}' is not valid for TLS 1.3 but tls_min_version is 1.3."
+      end
+
+      [cipher_list, ciphersuites]
     end
 
     def load_ssl_private_key(path)
@@ -118,8 +186,24 @@ module Proxy
 
     def webrick_server(app, addresses, port)
       server = ::WEBrick::HTTPServer.new(app)
-      addresses.each { |a| server.listen(a, port) }
+      begin
+        addresses.each { |a| server.listen(a, port) }
+      rescue ::OpenSSL::SSL::SSLError => e
+        raise "Invalid tls_ciphers value '#{app[:SSLCiphers]}': #{e.message}"
+      end
       server.mount "/", Rack::Handler::WEBrick, app[:app]
+
+      # WEBrick 1.9.x does not support :SSLMinVersion in its config hash, so we
+      # apply min_version= directly on the SSL context after WEBrick creates it.
+      # This patch should be removed once WEBrick adds support for :SSLMinVersion.
+      #
+      # :SSLCiphers is passed to SSL_CTX_set_cipher_list(), covering TLS 1.0–1.2.
+      # :SSLCiphersuites is applied here via SSL_CTX_set_ciphersuites() for TLS 1.3.
+      if app[:SSLEnable]
+        server.ssl_context.min_version = app[:SSLMinVersion] if app[:SSLMinVersion]
+        server.ssl_context.ciphersuites = app[:SSLCiphersuites] if app[:SSLCiphersuites]
+      end
+
       server
     end
 
--- a/config/settings.yml.example
+++ b/config/settings.yml.example
@@ -7,16 +7,16 @@
 #:ssl_ca_file: ssl/certs/ca.pem
 #:ssl_private_key: ssl/private_keys/fqdn.key
 
-# Use this option only if you need to disable certain cipher suites.
-# Note: we use the OpenSSL suite name, such as "RC4-MD5".
-# The complete list of cipher suite names can be found at:
-# https://www.openssl.org/docs/manmaster/man1/ciphers.html#CIPHER-SUITE-NAMES
-#:ssl_disabled_ciphers: [CIPHER-SUITE-1, CIPHER-SUITE-2]
+# TLS cipher string passed directly to OpenSSL. Accepts any valid OpenSSL cipher
+# string. When unset, foreman-proxy auto-detects: 'PROFILE=SYSTEM' is used if
+# /etc/crypto-policies/back-ends/opensslcnf.config, otherwise 'HIGH' is used
+# as a secure default.
+#:tls_ciphers: PROFILE=SYSTEM
 
-# Use this option only if you need to strictly specify TLS versions to be
-# disabled. SSLv3 and TLS v1.0 are always disabled and cannot be configured.
-# Specify versions like: '1.1', or '1.2'
-#:tls_disabled_versions: []
+# Minimum TLS protocol version to accept. When unset, the minimum version is
+# determined by the system's OpenSSL configuration (crypto-policies).
+# Valid values: '1.0', '1.1', '1.2', '1.3'.
+#:tls_min_version: '1.2'
 
 # Hosts which the proxy accepts connections from
 # commenting the following lines would mean every verified SSL connection allowed
--- a/lib/proxy/settings/global.rb
+++ b/lib/proxy/settings/global.rb
@@ -14,8 +14,8 @@ module ::Proxy::Settings
       :bind_host => ["*"],
       :log_buffer => 2000,
       :log_buffer_errors => 1000,
-      :ssl_disabled_ciphers => [],
-      :tls_disabled_versions => [],
+      :tls_ciphers => nil,
+      :tls_min_version => nil,
       :dns_resolv_timeouts => [5, 8, 13], # Ruby default is [5, 20, 40] which is a bit too much for us
     }
 
--- a/test/launcher_test.rb
+++ b/test/launcher_test.rb
@@ -22,3 +22,177 @@ class LauncherTest < Test::Unit::TestCase
     @launcher.launched([:app1, :app2])
   end
 end
+
+class LauncherTlsCiphersTest < Test::Unit::TestCase
+  def setup
+    @launcher = Proxy::Launcher.new
+  end
+
+  def launcher_with(settings_hash)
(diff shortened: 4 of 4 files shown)


adamlazik1:
Since tls_disabled_ciphers became dead code because of tls_ciphers, I decided to remove the deprecated opts in the same PR.

lib/launcher.rb
```diff
@@ -76,30 +96,49 @@ def https_app(https_port, plugins = https_plugins)
...
+      TLS_MIN_VERSIONS[0...min_idx].each do |version|
+        constant = OpenSSL::SSL.const_get("OP_NO_TLSv#{version.tr('.', '_')}") rescue nil
         if constant
-          logger.info "TLSv#{version} will be disabled."
+          logger.info "Disabling TLSv#{version} due to tls_min_version setting."
           ssl_options |= constant
-        else
-          logger.warn "TLSv#{version} was not found."
         end
       end
```
ekohl: This part surprises me because OpenSSL gained the `min_version` option: https://docs.ruby-lang.org/en/master/OpenSSL/SSL/SSLContext.html#method-i-min_version-3D. I have some recollection that conflicts with the classic `OP_NO_SSL*` and `OP_NO_TLS*`.

ekohl (commented):
This is a breaking change so perhaps this is good to share in the community. Perhaps an upstream RFC, even if it's just informational? Then in the release note you can refer to it.

lib/launcher.rb
```diff
@@ -66,6 +65,27 @@ def https_app(https_port, plugins = https_plugins)
...
+        :SSLOptions => ssl_options,
+        :SSLCiphers => tls_ciphers,
+      }
+      base_app_settings.merge(https_settings)
+    end
+
+    def build_ssl_options
       ssl_options = OpenSSL::SSL::SSLContext::DEFAULT_PARAMS[:options]
       ssl_options |= OpenSSL::SSL::OP_CIPHER_SERVER_PREFERENCE if defined?(OpenSSL::SSL::OP_CIPHER_SERVER_PREFERENCE)
       # This is required to disable SSLv3 on Ruby 1.8.7
```
adamruzicka: Could we also include a cleanup of those things?

lib/launcher.rb
adamruzicka: [Ruby 3.0 docs](https://ruby-doc.org/stdlib-3.0.0/libdoc/openssl/rdoc/OpenSSL/SSL/SSLContext.html#method-i-min_version-3D) carry this warning

> Be careful that you don't overwrite OpenSSL::SSL::OP_NO_{SSL,TLS}v* options by [options=](https://ruby-doc.org/stdlib-3.0.0/libdoc/openssl/rdoc/OpenSSL/SSL/SSLContext.html#method-i-options-3D) once you have called [min_version=](https://ruby-doc.org/stdlib-3.0.0/libdoc/openssl/rdoc/OpenSSL/SSL/SSLContext.html#method-i-min_version-3D) or [max_version=](https://ruby-doc.org/stdlib-3.0.0/libdoc/openssl/rdoc/OpenSSL/SSL/SSLContext.html#method-i-max_version-3D).

lib/launcher.rb
```diff
@@ -76,30 +96,49 @@ def https_app(https_port, plugins = https_plugins)
...
+    end
+
+    def apply_tls_min_version(ssl_options, min_version)
+      unless TLS_MIN_VERSIONS.include?(min_version)
+        raise "Invalid tls_min_version '#{min_version}'. Valid values: #{TLS_MIN_VERSIONS.join(', ')}"
+      end
 
+      min_idx = TLS_MIN_VERSIONS.index(min_version)
+      TLS_MIN_VERSIONS[0...min_idx].each do |version|
+        constant = OpenSSL::SSL.const_get("OP_NO_TLSv#{version.tr('.', '_')}") rescue nil
```
adamruzicka: I know it was done in a similar way before, but could we spell out the constants somewhere?

lib/launcher.rb
adamlazik1: Can you specify?

adamruzicka: Ruby 1.8 is long gone, yet we still carry a comment saying we're doing something specifically for it. All of the constants we check with `if defined?` should always be defined since ruby 2.6-ish

adamlazik1:
> This is a breaking change so perhaps this is good to share in the community. Perhaps an upstream RFC, even if it's just informational? Then in the release note you can refer to it.

Done. https://community.theforeman.org/t/rfc-new-tls-configuration-options-in-smart-proxy-tls-ciphers-tls-min-version/46777

lib/launcher.rb
adamlazik1: Oh I see what you mean. I will clean it up.

ekohl:
https://github.com/theforeman/puppet-candlepin/blob/033a4f1d8d1e5bdfefdc9f3d01ebb9fe374ab747/spec/acceptance/basic_candlepin_spec.rb#L14-L28 is my tip on how to verify the outcome.

lib/launcher.rb
adamlazik1: Apllied ctx.min_version with a tiny bit of workaround since webrick does not support `:SSLMinVersion` directly.

lib/launcher.rb
adamlazik1: Done.

lib/launcher.rb
adamlazik1: With switching to ctx.min_version the constants were removed. This should be resolved.

lib/launcher.rb
```diff
@@ -96,12 +81,53 @@ def https_app(https_port, plugins = https_plugins)
...
+      ssl_options = OpenSSL::SSL::SSLContext::DEFAULT_PARAMS[:options]
+      ssl_options |= OpenSSL::SSL::OP_CIPHER_SERVER_PREFERENCE
+      # Disable client initiated renegotiation
+      ssl_options |= OpenSSL::SSL::OP_NO_RENEGOTIATION
+
+      ssl_options
+    end
+
+    def resolve_tls_min_version
+      min_version = settings.tls_min_version&.to_s || '1.2'
```
ofedoren: Do we really need `||`? If min version is not present or not supported, it would fail the next check. Also, || implies that `1.2` is default value, shouldn't it be a part of `tls_min_version` setting itself?

```suggestion
      min_version = settings.tls_min_version&.to_s
```

ekohl: If crypto policies are applied, you should not set a default minimum version

lib/launcher.rb
```diff
@@ -96,12 +81,53 @@ def https_app(https_port, plugins = https_plugins)
...
+    def resolve_tls_ciphers
+      configured = settings.tls_ciphers
+      return configured unless configured.nil?
+
+      if File.exist?(CRYPTO_POLICIES_CONFIG)
+        logger.info "Crypto-policies detected, using PROFILE=SYSTEM for TLS ciphers."
+        'PROFILE=SYSTEM'
+      else
+        logger.debug "No crypto-policies detected, using OpenSSL defaults for TLS ciphers."
+        nil
```
ekohl: I think this could use some testing and consider my suggestion of using `HIGH` from https://community.theforeman.org/t/rfc-new-tls-configuration-options-in-smart-proxy-tls-ciphers-tls-min-version/46777/3

lib/launcher.rb
ofedoren: Or at least 
```suggestion
      min_version = settings.tls_min_version&.to_s || TLS_MIN_VERSION_MAP.keys.first
```

lib/launcher.rb
adamlazik1: Should `HIGH` be default in the repository or submitted as a patch in deb packaging?

lib/launcher.rb
adamlazik1: Fixed.

adamlazik1: Removed the default entirely as per https://github.com/theforeman/smart-proxy/pull/947#discussion_r3389874691.

lib/launcher.rb
adamlazik1: For testing: This is a good candidate for automation.

lib/launcher.rb
```diff
@@ -96,12 +83,55 @@ def https_app(https_port, plugins = https_plugins)
...
+      end
+    end
+
+    def validate_tls_ciphers!(ciphers)
+      return if ciphers.nil?
+
+      OpenSSL::SSL::SSLContext.new.ciphers = ciphers
```
adamruzicka: Passing in an empty string in the hopes of getting openssl defaults raises an exception

```
2026-06-11T09:20:46  [W] Error details for Error during startup, terminating: <RuntimeError>: Invalid tls_ciphers value '': SSL_CTX_set_cipher_list: no cipher match
/usr/share/foreman-proxy/lib/launcher.rb:132:in `rescue in validate_tls_ciphers!'
/usr/share/foreman-proxy/lib/launcher.rb:127:in `validate_tls_ciphers!'
/usr/share/foreman-proxy/lib/launcher.rb:75:in `https_app'
/usr/share/foreman-proxy/lib/launcher.rb:170:in `launch'
/usr/share/foreman-proxy/bin/smart-proxy:6:in `<main>'
```

lib/launcher.rb
adamruzicka: If I explicitly set `tls_min_version`, but I leave `tls_ciphers` unset and autodetection kicks in and crypto policies get applied, then I'd expect to get (and do get) an intersection of the ciphers allowed by the min version and by the crypto policy. This expectation (and behaviour) is correct, right?

adamruzicka (commented):
Seems to work well, except for the empty string.

For what I checked see https://gist.github.com/adamruzicka/4b6370339dd36b9f9836aeba2080f40a . There are more combinations that could be made, but the essentials should be covered.

lib/launcher.rb
adamruzicka: I'd vote for an in-repo default

adamlazik1:
> If I explicitly set tls_min_version, but I leave tls_ciphers unset and autodetection kicks in and crypto policies get applied, then I'd expect to get (and do get) an intersection of the ciphers allowed by the min version and by the crypto policy. This expectation (and behaviour) is correct, right?

I'd say yes.

lib/launcher.rb
adamlazik1: Fixed.

lib/launcher.rb
adamlazik1: Applied.

lib/launcher.rb
adamruzicka: Not quite what I expected. Now an empty string means the same thing as an unset value, which leads to autodetection.

adamruzicka:
An array with arbitrary values can be passed in as a value for `tls_ciphers`. It gets passed around, doesn't seem to raise any errors anywhere, but ultimately doesn't seem to have any effect on anything. This is misleading at best

lib/launcher.rb
```diff
@@ -96,12 +83,55 @@ def https_app(https_port, plugins = https_plugins)
...
+        raise "Invalid tls_min_version '#{min_version}'. Valid values: #{TLS_MIN_VERSION_MAP.keys.join(', ')}"
+      end
+
+      logger.debug "Setting minimum TLS version to #{min_version}."
+      TLS_MIN_VERSION_MAP[min_version]
+    end
+
+    def resolve_tls_ciphers
+      configured = settings.tls_ciphers
+      return configured unless configured.nil? || configured.empty?
```
adamruzicka: edge case, setting tls_ciphers to an integer raises an undefined method error
```
2026-06-12T05:54:30  [W] Error details for Error during startup, terminating: <NoMethodError>: undefined method `empty?' for 12:Integer
/usr/share/foreman-proxy/lib/launcher.rb:117:in `resolve_tls_ciphers'
/usr/share/foreman-proxy/lib/launcher.rb:74:in `https_app'
/usr/share/foreman-proxy/lib/launcher.rb:172:in `launch'
/usr/share/foreman-proxy/bin/smart-proxy:6:in `<main>'
```

lib/launcher.rb
```diff
@@ -120,6 +150,14 @@ def webrick_server(app, addresses, port)
       server = ::WEBrick::HTTPServer.new(app)
       addresses.each { |a| server.listen(a, port) }
       server.mount "/", Rack::Handler::WEBrick, app[:app]
+
+      # WEBrick 1.9.x does not support :SSLMinVersion in its config hash, so we
+      # apply min_version= directly on the SSL context after WEBrick creates it.
+      # This patch should be removed once WEBrick adds support for :SSLMinVersion.
+      if app[:SSLEnable] && app[:SSLMinVersion]
+        server.ssl_context.min_version = app[:SSLMinVersion]
+      end
```
adamruzicka, line 205: Could we have some test for this, just to be extra sure?

lib/launcher.rb
```diff
@@ -96,12 +83,55 @@ def https_app(https_port, plugins = https_plugins)
...
+
+    def resolve_tls_min_version
+      return nil unless settings.tls_min_version
+
+      min_version = settings.tls_min_version.to_s
+      unless TLS_MIN_VERSION_MAP.key?(min_version)
+        raise "Invalid tls_min_version '#{min_version}'. Valid values: #{TLS_MIN_VERSION_MAP.keys.join(', ')}"
+      end
+
+      logger.debug "Setting minimum TLS version to #{min_version}."
```
adamruzicka:
```suggestion
      logger.info "Setting minimum TLS version to #{min_version}."
```

lib/launcher.rb
adamlazik1: My bad. Fixed.

lib/launcher.rb
adamlazik1: Fixed.

lib/launcher.rb
adamlazik1: Applied.

lib/launcher.rb
adamlazik1: Added.

ekohl:
> https://github.com/theforeman/puppet-candlepin/blob/033a4f1d8d1e5bdfefdc9f3d01ebb9fe374ab747/spec/acceptance/basic_candlepin_spec.rb#L14-L28 is my tip on how to verify the outcome.

And I also found https://testssl.sh/

lib/launcher.rb
```diff
@@ -96,12 +83,57 @@ def https_app(https_port, plugins = https_plugins)
         :SSLPrivateKey => load_ssl_private_key(settings.ssl_private_key),
         :SSLCertificate => load_ssl_certificate(settings.ssl_certificate),
         :SSLCACertificateFile => settings.ssl_ca_file,
-        :SSLOptions => ssl_options,
-        :SSLCiphers => CIPHERS - Proxy::SETTINGS.ssl_disabled_ciphers,
+        :SSLOptions => build_ssl_options,
+        :SSLCiphers => tls_ciphers,
+        :SSLMinVersion => resolve_tls_min_version,
```
lhellebr, line 89: If we set crypto-policy to DEFAULT:TLS13-ONLY (TLS13-ONLY being a custom subpolicy) but in the config we set tls_min_version to 1.2, this would keep the correct cipher set but would degrade TLS min version to 1.2 (in conflict with crypto-policies).

adamruzicka: Chmm, in theory we could go back to the old approach of disabling individual versions which should interact with crypto policies in a slightly better way

ekohl: I think it's a bit of an edge case. I'd prefer to have 2 use cases:
* crypto-policies (RPM-based installs)
* explicitly configured (where crypto-policies is unavailable, like Debian, FreeBSD and Windows)
The case where you configure 1 option, but not the other is IMHO up to the user who does that.

adamlazik1: We can keep min_tls version in repo and remove it in the foreman-packaging rpm patch so that rpm builds rely only on crypto policies. Would that work @adamruzicka @ekohl ?

adamruzicka: I'd prefer to have as few patches in packaging as possible. How about making `PROFILE=SYSTEM` special and refusing to start if `tls_ciphers == 'PROFILE=SYSTEM'` and `tls_min_version` is set?

adamlazik1: Yeah, I can do that.

lhellebr: Agree, and please add a test for it.

lib/launcher.rb
```diff
@@ -96,12 +83,57 @@ def https_app(https_port, plugins = https_plugins)
         :SSLPrivateKey => load_ssl_private_key(settings.ssl_private_key),
         :SSLCertificate => load_ssl_certificate(settings.ssl_certificate),
         :SSLCACertificateFile => settings.ssl_ca_file,
-        :SSLOptions => ssl_options,
-        :SSLCiphers => CIPHERS - Proxy::SETTINGS.ssl_disabled_ciphers,
+        :SSLOptions => build_ssl_options,
+        :SSLCiphers => tls_ciphers,
```
lhellebr: This works for TLS 1.2 but not for TLS 1.3. Unfurtunately, `SSL_CTX_set_cipher_list()` is TLS 1.2 only and `SSL_CTX_set_ciphersuites()` is TLS 1.3 only and the Webrick/Ruby stack under the hood only seems to do the former.
The same hack as in https://github.com/theforeman/smart-proxy/pull/947/changes#diff-54572e87725df68a63244713ca93aa40c81de79d76cba589c0a1d0329812db04R160 might help.

lhellebr: This may make your work easier: https://github.com/warmcat/libwebsockets/issues/1440

lib/launcher.rb
ekohl: > We can keep min_tls version in repo and remove it in the foreman-packaging rpm patch so that rpm builds rely only on crypto policies.

If we're confident enough in the auto-detection then we wouldn't need any packaging patch. I'd be OK with that. Emitting a log message on startup that can be asserted in automated tests is probably enough.

> I'd prefer to have as few patches in packaging as possible. How about making `PROFILE=SYSTEM` special and refusing to start if `tls_ciphers == 'PROFILE=SYSTEM'` and `tls_min_version` is set?

Is it really a problem? If the user configures it that way, what will break? If `foremanctl` doesn't provide you any option to set that then IMHO it's really up to the user. Perhaps emit a log warning?

lib/launcher.rb
adamlazik1: Currently trying to resolve this, for now I have no idea why the tests fail. I need to investigate further.

lib/launcher.rb
adamlazik1: > I'd prefer to have as few patches in packaging as possible. How about making PROFILE=SYSTEM special and refusing to start if tls_ciphers == 'PROFILE=SYSTEM' and tls_min_version is set?

Applied.

lhellebr:
The CI failure is real, I've just reproduced on Debian with the current PR version. The proxy refuses to start because
```
Errors detected on startup, see log for details. Exiting: undefined method `ciphersuites=' for #<OpenSSL::SSL::SSLContext:0x00007f28a61b8890 @verify_mode=0, @verify_hostname=false>
```
We probably need to check whether `ciphersuites=` is understood by underlying ruby/openssl. If yes, call it, if not... those older versions probably still support TLS 1.3 so just not calling it that would mean the cipher setting gets silently ignored. Maybe just make that not silent and refuse to start if ciphers are explicitly set in the config but we can't set them?

Interestingly, it seems that Red Hat's version of Ruby has this patched in earlier Ruby versions (https://gitlab.com/redhat/centos-stream/rpms/ruby/-/commit/42815870), which is why it works on RHEL with Ruby 3.0.7, but fails in CI on Ruby 3.0 and 3.1 (upstream, it got patched in 3.2 https://github.com/ruby/openssl/pull/493).

adamlazik1:
> Maybe just make that not silent and refuse to start if ciphers are explicitly set in the config but we can't set them?

Applied. Went with this suggestion.

lib/launcher.rb
```diff
@@ -96,12 +83,86 @@ def https_app(https_port, plugins = https_plugins)
...
+      end
+
+      if OpenSSL::SSL::SSLContext.method_defined?(:ciphersuites=)
+        ciphersuites = begin
+          OpenSSL::SSL::SSLContext.new.ciphersuites = ciphers
+          ciphers
+        rescue OpenSSL::SSL::SSLError
+          nil
+        end
+      elsif settings.tls_ciphers && settings.tls_min_version == '1.3'
```
lhellebr: 1) What if the user didn't specify TLS 1.3 explicitly (thus it's still allowed together with 1.2) but has specified some constraints for a cipher that is applicable to TLS 1.3? Wouldn't then those be ignored on a machine without `ciphersuites=` and wouldn't TLS 1.3 still run, but with any cipher available, even those that the user forbid?

2) What if the user allows both TLS 1.2 and 1.3 but specifies a cipher only recognized by `ciphersuites=` (because it's only in TLS 1.3)? `ciphers=` would return nil and allow *any* TLS 1.2 cipher, wouldn't it?

lib/launcher.rb
```diff
@@ -96,12 +83,86 @@ def https_app(https_port, plugins = https_plugins)
...
+        raise "tls_min_version '1.3' is configured but this Ruby/OpenSSL build does not support TLS 1.3. " \
```
lhellebr: Misleading message. It supports TLS 1.3 but doesn't support constraining it by `ciphersuites=`.

lhellebr (changes requested):
More issues found - I haven't really tested yet but this seems sus on reading. These things are getting quite complicated.

lib/launcher.rb
adamlazik1: 1
You are right. And to be honest I have no idea how to solve that. Consulting AI gave me these options: 

a. Widen the elsif guard to settings.tls_ciphers (without the tls_min_version check) — raises whenever any explicit cipher is set and ciphersuites= is absent. This is safe but means PROFILE=SYSTEM without tls_min_version also refuses to start on Ruby 3.0, which is overly strict.

b. Accept the limitation and document it — on Ruby 3.0 without the Red Hat backport, TLS 1.3 cipher restrictions from tls_ciphers are silently not applied. Practically, PROFILE=SYSTEM is only meaningful on RHEL where ciphersuites= is always available anyway, so real-world exposure is low.

To me, b. seems more viable. Let me know your thoughts, or if you know about another solution.

2
Added logger warning for this case, let me know if I should change it to exception.

lib/launcher.rb
adamlazik1: Edited.

adamlazik1:
Removed the packaging patch inclusion.

lhellebr:
**ACK** — Verified on RHEL 9.8 (Satellite, packit build `pr947.753.gb7d2025`) and Debian 12 (upstream Foreman, manually patched).

I'm ACKing this PR because the known issues listed below are **not regressions** — the same behavior exists in the current code without this PR. However, I want to be transparent that this PR introduces a cipher configuration interface that **does not fully deliver on its promise**: an admin setting `tls_ciphers` would reasonably expect it to restrict ciphers for all TLS versions, but it silently leaves the other TLS version unrestricted in several scenarios. These are security concerns that should be addressed in follow-up work.

### What was tested

Tested `tls_ciphers` and `tls_min_version` settings with various values (valid, invalid, edge cases including non-string types), their interaction with RHEL crypto-policies (ciphers, TLS versions, signatures), the auto-detection logic (`PROFILE=SYSTEM` on RHEL, `HIGH` fallback on Debian), auto-detection resilience (broken/missing/unreadable files), and the `ciphersuites=` compatibility fix on Debian where the method is missing.

### Security concerns (not regressions, tracked separately)

1. On platforms where the `ciphersuites=` method is missing (e.g., Debian 12), TLS 1.3 ciphers cannot be controlled at all. When `tls_ciphers` is explicitly set, all TLS 1.3 ciphers are allowed regardless of the configured value.

2. Even on platforms where `ciphersuites=` is available (e.g., RHEL 9), when `tls_ciphers` is set to a value that only matches one TLS version, the other TLS version is left with unrestricted ciphers. This goes both ways: a TLS 1.2-only cipher leaves TLS 1.3 unrestricted, and a TLS 1.3-only cipher leaves TLS 1.2 unrestricted. The existing warning only covers the latter case — there is **no warning** for the TLS 1.2-only cipher case. Note that this does not apply when ciphers are controlled through crypto-policies — crypto-policies correctly restricts both TLS versions.

### Notes

- The packaging patch (foreman-packaging#13625) is no longer needed — auto-detection handles RHEL correctly.
- The deprecated `ssl_disabled_ciphers` and `tls_disabled_versions` options seem to be effectively removed by this PR.
- Deployment tooling (`foreman-installer` / `foremanctl`) needs updating to expose the new options and remove the deprecated ones from templates.

lhellebr (approved):

adamlazik1:
Squashed.

adamruzicka (approved):

adamruzicka:
Thank you @adamlazik1 !

PR #948, opened 2026-06-09
Use smart-proxy mTLS auth for container registry

sbernhard:
Allow smart-proxies to pull/push container images against Katello's registry using their existing Foreman SSL client certificate. Adds a reusable `Proxy::ContainerRegistry::PodmanAuth` module to smart-proxy for podman cert-dir setup and cleanup.

Katello PR: https://github.com/Katello/katello/pull/11776
Smart Proxy PR: https://github.com/theforeman/smart-proxy/pull/948
Smart Proxy Ansible Director: https://github.com/ATIX-AG/smart_proxy_ansible_director/pull/24

Note: I have created this PRs to discuss how to use the katello container registry with smart-proxies.

### DIFF
--- a/lib/proxy/container_registry/podman_auth.rb
+++ b/lib/proxy/container_registry/podman_auth.rb
@@ -1,0 +1,53 @@
+# frozen_string_literal: true
+
+require 'fileutils'
+require 'shellwords'
+require 'tmpdir'
+
+module Proxy
+  module ContainerRegistry
+    # Provides mTLS certificate setup for authenticating podman CLI commands
+    # against a Katello container registry using the smart-proxy's existing
+    # foreman SSL client certificate (foreman_ssl_cert / foreman_ssl_key).
+    #
+    # Usage in an async runner (initialize_command pattern):
+    #
+    #   def start
+    #     @cert_dir = Proxy::ContainerRegistry::PodmanAuth.setup_cert_dir
+    #     cmd = "podman push #{Proxy::ContainerRegistry::PodmanAuth.tls_args(@cert_dir)} ..."
+    #     initialize_command('bash', '-c', cmd)
+    #   end
+    #
+    #   def close
+    #     Proxy::ContainerRegistry::PodmanAuth.cleanup(@cert_dir)
+    #   end
+    module PodmanAuth
+      # Creates a temporary directory with symlinks named as podman expects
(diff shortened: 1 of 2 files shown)


lib/proxy/container_registry/podman_auth.rb
```diff
@@ -0,0 +1,53 @@
...
+      # (client.cert, client.key, ca.crt) pointing at the smart-proxy's
+      # existing SSL certificate files. Symlinks are used rather than copies
+      # to avoid duplicating the private key on disk.
+      # Call cleanup when done.
+      def self.setup_cert_dir
+        cert_file = Proxy::SETTINGS.foreman_ssl_cert || Proxy::SETTINGS.ssl_certificate
+        key_file  = Proxy::SETTINGS.foreman_ssl_key  || Proxy::SETTINGS.ssl_private_key
+        ca_file   = Proxy::SETTINGS.foreman_ssl_ca   || Proxy::SETTINGS.ssl_ca_file
+
+        dir = Dir.mktmpdir('podman_registry_cert')
```
ianballou, line 35: This temporary cert directory works well because we do support setting up container certs on hosts when they register, and users do register smart proxies as hosts. So, the certificate directories will not clash.

ekohl (commented):
What's the benefit of having this code in core? Do you intend to use it in multiple places?

sbernhard:
> What's the benefit of having this code in core? Do you intend to use it in multiple places?

to be able to re-use it and not implement it multiple times. e.g. re-used in https://github.com/ATIX-AG/smart_proxy_ansible_director/pull/24/changes#diff-d10917afb8a06f6552550287738ecd520da52438cdf29983fa2bbd6fdc0d9688R105

ekohl:
But that's still only within a single repository, right?

sbernhard:
> But that's still only within a single repository, right?

Yes, but I think the ability for a Smart Proxy plugin to use Katello’s Docker registry could also be useful for other plugins.

ekohl:
I have my doubts. Why optimize it prematurely and make merging your PR harder?

sbernhard:
> I have my doubts. Why optimize it prematurely and make merging your PR harder?

It was a thought. I an move everything to smart proxy Ansible Director, too.

ekohl:
I'd lean to starting there. Gives you a bit more freedom to backport the plugin to older releases too. We can always add it if it's needed.

PR #949, opened 2026-06-15
Fixes #39429 - Add KEA DHCP provider

archanaserver:
Ref: https://github.com/theforeman/foremanctl/pull/531#issuecomment-4679221932

KEA DHCP integration for  Smart Proxy as we're transitioning away from ISC DHCP (which reached EOL in 2022) to KEA now.

Adds Smart Proxy DHCP provider for ISC KEA DHCP server using the KEA Control Agent API, providing a migration path for users currently using the deprecated dhcp_isc provider.

Note: Based on the `smart_proxy_dhcp_kea_api` gem by Sam McCarthy, to integrate into Smart Proxy core. https://gitlab.surrey.ac.uk/sm0049/smart-proxy-dhcp-kea-api/-/issues/3

Fixes: https://redhat.atlassian.net/browse/SAT-27739


modules/dhcp_kea/dhcp_kea_main.rb
```diff
@@ -0,0 +1,135 @@
...
+
+    private
+
+    def netmask_from_cidr(cidr)
+      network, prefix = cidr.split('/')
+      prefix = prefix.to_i
+
+      mask = (0xffffffff << (32 - prefix)) & 0xffffffff
+
+      [mask].pack('N').unpack('C4').join('.')
```
ekohl: IPAddr has this built in:
```suggestion
      cidr.netmask
```

bundler.d/dhcp_kea.rb
```diff
@@ -0,0 +1,4 @@
+group :dhcp_kea do
```
evgeni: Then we don't need this group (file) at all?

config/settings.d/dhcp.yml.example
```diff
@@ -3,10 +3,14 @@
 :enabled: false
 
 # valid providers:
-#   - dhcp_isc (ISC dhcp server)
+#   - dhcp_kea (ISC KEA dhcp server)
 #   - dhcp_native_ms (Microsoft native implementation)
 #   - dhcp_libvirt
-#:use_provider: dhcp_isc
+#   - dhcp_isc (ISC dhcp server - DEPRECATED, EOL October 2022)
```
evgeni: I wouldn't mention the "EOL October 2022" part, as it's not us who declares the EOL date. Just say it's deprecated.

config/settings.d/dhcp_kea.yml.example
```diff
@@ -0,0 +1,16 @@
+---
+# KEA DHCP provider configuration
+# Requires ISC KEA DHCP server with Control Agent API enabled
+
+# URL of the KEA Control Agent API
+#:dhcp_kea_api_url: http://127.0.0.1:8000/
```
evgeni: I wonder if we should drop the `api_` part from all settings, given that we're calling the provider `dhcp_kea` (and not `dhcp_kea_api`)

evgeni:
Given this is an import of an previously external provider, I'd vote for "make CI green and merge it", with any further cleanup in a separate PR.

config/settings.d/dhcp_kea.yml.example
archanaserver: Yup agree, Updated!

bundler.d/dhcp_kea.rb
archanaserver: Yes. we don't

modules/dhcp_kea/dhcp_kea_main.rb
evgeni: Sadly, this is Ruby 3.1+ only :/

config/settings.d/dhcp.yml.example
```diff
@@ -3,10 +3,14 @@
...
 # valid providers:
-#   - dhcp_isc (ISC dhcp server)
+#   - dhcp_kea (ISC KEA dhcp server)
 #   - dhcp_native_ms (Microsoft native implementation)
 #   - dhcp_libvirt
-#:use_provider: dhcp_isc
+#   - dhcp_isc (ISC dhcp server - DEPRECATED)
+#
+# NOTE: ISC DHCP reached end-of-life in October 2022. ISC recommends
+# migrating to KEA DHCP. Use `dhcp_kea` for new deployments.
```
evgeni:
```suggestion
```

modules/dhcp_kea/dhcp_kea_main.rb
ekohl: Sadly the Ruby's API docs didn't mention that :(

modules/dhcp_kea/dhcp_kea.rb
```diff
@@ -0,0 +1,3 @@
+require 'dhcp_common/dhcp_common'
+require 'dhcp_kea/plugin_configuration'
+require 'dhcp_kea/dhcp_kea_plugin'
```
stejskalleos: Is it a plugin? IMO it's not.

```suggestion
require 'dhcp_kea/configuration'
require 'dhcp_kea/dhcp_kea'
```

config/settings.d/dhcp_kea.yml.example
```diff
@@ -0,0 +1,16 @@
...
+
+# Optional HTTP Basic Authentication credentials
+#:dhcp_kea_username: ~
+#:dhcp_kea_password: ~
+
+# SSL certificate verification (set to false for self-signed certificates)
+#:dhcp_kea_verify_ssl: true
+
+# Timeout for lease operations in seconds
+#:dhcp_kea_lease_timeout: 60
```
stejskalleos, line 16: Discussion: 60 seconds is quite generous, and I don't know how it will work with all those queue steps and JS actions in the create host UI form. 

IMO, we should lower this to <30 seconds.

stejskalleos (dismissed):

config/settings.d/dhcp_kea.yml.example
evgeni: Where is that used anyway?

modules/dhcp_kea/dhcp_kea_plugin.rb
```diff
@@ -0,0 +1,20 @@
...
+                     :dhcp_kea_username => nil,
+                     :dhcp_kea_password => nil,
+                     :dhcp_kea_verify_ssl => true,
+                     :dhcp_kea_lease_timeout => 60
+
+    requires :dhcp, ::Proxy::VERSION
+
+    load_classes ::Proxy::DHCP::Kea::PluginConfiguration
+    load_programmable_settings ::Proxy::DHCP::Kea::PluginConfiguration
+    load_dependency_injection_wirings ::Proxy::DHCP::Kea::PluginConfiguration
```
ekohl: Note that all of these are optional. If they're empty, you can leave them out. You can also use a block. For example:
https://github.com/theforeman/smart-proxy/blob/3e249dd019dd36f41b546fe44c045ab055c1da6c/modules/httpboot/httpboot_plugin.rb#L7-L11

I'm fine either way, but I want to point out the option since it may be easier to understand to have fewer files.

The trade off is that it's harder to test. For `load_classes` that's not a concern since we don't test that anyway.

config/settings.d/dhcp_kea.yml.example
archanaserver: currently it is only passed into the KEA provider and stored as `@lease_timeout`, but it is not actually used in the lease/reservation operation calls yet.

soeither wire this into the KEA API client timeout handling, or remove the config option for now if we don’t think we need it here. that's my thought on this

modules/dhcp_kea/dhcp_kea_plugin.rb
archanaserver: Yeah that make sense, Thanks for pointing this out!

modules/dhcp_kea/dhcp_kea.rb
archanaserver: I kept `dhcp_kea_plugin` because we already have `dhcp_kea` as the entry point. Renaming the plugin file to `dhcp_kea.rb` would create confusion, so let's stay with this?

archanaserver:
One test changes and now it's failing completely, FYI this is in-progress right now.

archanaserver:
@evgeni @stejskalleos can you help how it keep failing here workflow? i'm not sure what i'm missing

test/dhcp_kea/dhcp_kea_main_test.rb
```diff
@@ -0,0 +1,167 @@
...
+
+    assert_nil @subnet_service.find_host_by_mac('192.168.1.0/24', 'aa:bb:cc:dd:ee:ff')
+  end
+
+  def test_del_record_subnet_not_found
+    record = Proxy::DHCP::Reservation.new(
+      'test.example.com',
+      '172.16.0.50',
+      'aa:bb:cc:dd:ee:ff',
+      Proxy::DHCP::Subnet.new('172.16.0.0/24', '255.255.255.0')
```
evgeni:
```suggestion
      Proxy::DHCP::Subnet.new('172.16.0.0', '255.255.255.0')
```

`Subnet.new` wants an IP and a netmask, separately

modules/dhcp_kea/dhcp_kea_main.rb
```diff
@@ -0,0 +1,131 @@
...
+
+      begin
+        kea_client.delete_reservation_by_ip(subnet_id, record.ip)
+      rescue => e
+        logger.error "Failed to delete KEA reservation: #{e.message}"
+        raise Proxy::DHCP::Error, "Failed to delete reservation from KEA: #{e.message}"
+      end
+
+      if record.is_a?(::Proxy::DHCP::Reservation)
+        service.delete_host(subnet.network, record)
```
evgeni: `delete_host` only expects one parameter: a record. you're passing two, and thus it fails.

modules/dhcp_kea/dhcp_kea_main.rb
```diff
@@ -0,0 +1,131 @@
...
+      elsif record.is_a?(::Proxy::DHCP::Lease)
+        service.delete_lease(subnet.network, record)
```
evgeni, line 112: `delete_lease` *also* expects just one parameter, seems there is no test for this case.

evgeni: also, for this to work, we should have issued `lease4-del`, not `reservation-del` as done by `delete_reservation_by_ip`, right?

evgeni:
Which version (commit) of https://gitlab.surrey.ac.uk/sm0049/smart-proxy-dhcp-kea-api/ was this based on? I have a hard time comparing the code and the issues :(

ekohl:
> Which version (commit) of https://gitlab.surrey.ac.uk/sm0049/smart-proxy-dhcp-kea-api/ was this based on? I have a hard time comparing the code and the issues :(

Perhaps you can write 2 commits: 1 doing a verbatim import with proper ownership (`git commit --author '...'`) and then another with the changes that you made to it. I strongly feel that proper attribution of the original owner is very important in open source. There may be some legal reasons (since there still is a license), but IMHO also ethical. Someone took the effort to write it and we should acknowledge that.

You can use `Refs #39429 - ...` on the first commit so make the Redmine checker happy.

config/settings.d/dhcp_kea.yml.example
lzap: More importantly, why this configuration is set up when it is not used at all? If there are some plans for follow-ups, let's document that the setting is currently unused.

Also, please improve the comment so it is immediately obvious whether this is a timeout for HTTP REST management API operation, or time of actual DHCP leases. It is a bit vague.

Additionally, we need the REST API open/read timeout here with some reasonable default value (5 secs). I suggest to have just a single variable like `rest_api_timeout` because for localhost connections this can be both very small value and unlikely to be subject of configuration. Maybe we can even hardcode, but we must not rely on OS default timeouts which are way to long usually.

modules/dhcp_kea/dhcp_kea_main.rb
```diff
@@ -0,0 +1,131 @@
...
+    end
+
+    private
+
+    def netmask_from_cidr(cidr)
+      prefix = cidr.split('/').last.to_i
+
+      mask = (0xffffffff << (32 - prefix)) & 0xffffffff
+      [mask].pack('N').unpack('C4').join('.')
+    end
```
lzap, line 129: Please drop this function and replace it with `IPAddr.new(cidr).netmask.to_s` which does the same thing.

```
irb(main):005> IPAddr.new("192.168.42.13/24").netmask.to_s
=> "255.255.255.0"
```

modules/dhcp_kea/kea_api_client.rb
```diff
@@ -0,0 +1,136 @@
...
+
+    def send_command(service, command, arguments = {})
+      payload = {
+        'command' => command,
+        'service' => [service],
+        'arguments' => arguments,
+      }
+
+      logger.debug "Sending KEA command: #{command} to service: #{service}"
+      logger.debug "Arguments: #{arguments.inspect}"
```
lzap, line 26: Nitpick: could have been one line.

modules/dhcp_kea/kea_api_client.rb
```diff
@@ -0,0 +1,136 @@
...
+
+      response = http_post('/', payload)
+
+      result = response.first
+
+      if result['result'] != 0
+        error_msg = "KEA command '#{command}' failed: #{result['text']}"
+        logger.error error_msg
+        raise error_msg
```
lzap, line 35: Raise `Proxy::DHCP::Error`.

modules/dhcp_kea/kea_api_client.rb
```diff
@@ -0,0 +1,136 @@
...
+                     'subnet-id' => subnet_id.to_i,
+        'ip-address' => ip_address,
+                   })
+    end
+
+    def reservation_by_ip(subnet_id, ip_address)
+      send_command('dhcp4', 'reservation-get', {
+                     'subnet-id' => subnet_id.to_i,
+        'ip-address' => ip_address,
+                   })
```
lzap, line 84: Broken indentation. Same above.

modules/dhcp_kea/kea_api_client.rb
```diff
@@ -0,0 +1,136 @@
...
+    end
+
+    def list_leases
+      send_command('dhcp4', 'lease4-get-all')
+    end
+
+    def lease_by_ip(ip_address)
+      result = send_command('dhcp4', 'lease4-get', {
+                              'ip-address' => ip_address,
+                            })
```
lzap, line 97: Nitpick: Could have been a single line.

README.md
```diff
@@ -9,7 +9,7 @@
 # Supported Modules
 Currently Supported modules:
  * BMC - BMC management of devices supported by freeipmi and ipmitool
- * DHCP - ISC DHCP and MS DHCP Servers
+ * DHCP - ISC KEA, MS DHCP, and ISC DHCP (legacy, EOL 2022) Servers
```
lzap, line 12:
```suggestion
 * DHCP - ISC KEA, MS DHCP, and ISC DHCP (deprecated) Servers
```

lzap (changes requested):
This needs more changes. The current implementation is basically broken, so I am not even trying to test this. Specifically leases are never loader properly into memory and deletion incorrectly handles types (and also is broken code). Finally, `load_subnets` is not idempotent.

We have almost zero system testing in smart proxy, but I think in today's world (containers, AI) we could change that. This would make reviews easier for all of us and set a good standard going forward. But I do not want to block this, only if you want, we can stick to testing manually. Such test would be something like:

* Special test that only runs when `SYSTEM_TEST` variable is set.
* Starts an ephemeral `--rm` container with Kea
* Configures a network.
* Tests reservation CRUD.
* Tests lease CRUD.
* Optional: enable `lease4-add` special hook and create some leases (this could be done via DHCP but that is more elaborate test).
* Optional: test lease loading and reservations against leases.

modules/dhcp_kea/dhcp_kea_main.rb
ekohl: See https://github.com/theforeman/smart-proxy/pull/949#discussion_r3413178198. tl;dr: that is Ruby 3.1 and we need to support Ruby 3.0. It could use `respond_to?(:netmask)` with a comment though.

PR #950, opened 2026-06-25
Fixes #XXXXX - Reuse persistent Foreman connections

pablomh:
## Summary
- reuse a memoized `Net::HTTP::Persistent` client for outbound `smart-proxy -> Foreman` requests instead of creating a fresh `Net::HTTP` client for each request
- keep `ForemanRequest` as the existing entry point so the request-building flow used by registration and template callers stays unchanged
- preserve the existing SSL configuration behavior and update the focused tests for the shared client and the keep-alive headers emitted by the persistent client

## Test plan
- [x] `bundle3.2 exec ruby3.2 -Itest test/request_test.rb`
- [x] `bundle3.2 exec ruby3.2 -Itest test/templates/template_proxy_request_test.rb`
- [x] `bundle3.2 exec ruby3.2 -Itest test/registration/registration_api_test.rb`
- [x] `bundle3.2 exec rubocop lib/proxy/request.rb test/request_test.rb test/templates/template_proxy_request_test.rb test/registration/registration_api_test.rb smart_proxy.gemspec`
- [ ] `bundle3.2 exec rake test` (`rubyipmi` is missing in this environment, so the suite stops in `test/bmc/...` before reaching completion)

### DIFF
--- a/lib/proxy/request.rb
+++ b/lib/proxy/request.rb
@@ -1,4 +1,5 @@
 require 'net/http'
+require 'net/http/persistent'
 require 'net/https'
 require 'uri'
 require 'cgi'
@@ -44,8 +45,36 @@ module Proxy::HttpRequest
   end
 
   class ForemanRequest
+    class << self
+      def http
+        @http ||= begin
+          uri = URI.parse(Proxy::SETTINGS.foreman_url.to_s)
+          http = Net::HTTP::Persistent.new(name: 'smart_proxy_foreman')
+          http.verify_mode = OpenSSL::SSL::VERIFY_NONE
+
+          if uri.scheme == 'https'
+            ca_file = Proxy::SETTINGS.foreman_ssl_ca || Proxy::SETTINGS.ssl_ca_file
+            certificate = Proxy::SETTINGS.foreman_ssl_cert || Proxy::SETTINGS.ssl_certificate
+            private_key = Proxy::SETTINGS.foreman_ssl_key || Proxy::SETTINGS.ssl_private_key
+
+            if ca_file && !ca_file.to_s.empty?
+              http.ca_file     = ca_file
+              http.verify_mode = OpenSSL::SSL::VERIFY_PEER
+            end
+
+            if certificate && !certificate.to_s.empty? && private_key && !private_key.to_s.empty?
+              http.certificate = OpenSSL::X509::Certificate.new(File.read(certificate))
+              http.private_key = OpenSSL::PKey.read(File.read(private_key), nil)
+            end
+          end
+
+          http
+        end
+      end
+    end
+
     def send_request(request)
-      http.request(request)
+      http.request(uri, request)
     end
 
     def request_factory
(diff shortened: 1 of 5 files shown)


pablomh:
This would need https://github.com/theforeman/foreman-packaging/pull/13670.

ekohl:
Alternative smaller implementation: https://github.com/theforeman/smart-proxy/pull/951.

pablomh:
Reduced version of the code.

pablomh:
We deployed PR #950 on 6 capsules under incremental registration load from 76 to 1140 concurrent hosts. Satellite was configured with `--tuning large` (24 Puma workers × 5 threads).

**CLOSE-WAIT accumulation:** All capsules accumulated 80-103 CLOSE-WAIT sockets to Satellite:443 with 0 ESTAB, visible even at rest between batches. Satellite closes connections after `MaxKeepAliveRequests 100` or `KeepAliveTimeout 15`, but the closed sockets sit in CLOSE-WAIT between pool checkouts. WEBrick's thread-per-request model creates higher connection churn than the stable thread pools [`connection_pool`](https://github.com/mperham/connection_pool) was designed for, amplifying the accumulation.

**ConnectionPool::TimeoutError under load:** Starting at ~300 concurrent hosts, requests failed with `ConnectionPool::TimeoutError: Waited 0.5 sec`. [`net-http-persistent`'s Pool](https://github.com/drbrain/net-http-persistent/blob/master/lib/net/http/persistent/pool.rb) checkout inherits a [0.5-second pop timeout](https://github.com/mperham/connection_pool/blob/main/lib/connection_pool/timed_stack.rb#L62) from `TimedStack`. When pooled connections are busy serving Satellite requests that take 10-27 seconds, new threads cannot acquire a connection and fail immediately.

**POST /register failures:** ~2-3% of POSTs failed with `EOFError` or `ECONNRESET`. `Net::HTTP` [does not retry non-idempotent methods](https://bugs.ruby-lang.org/issues/10674), so these propagated as 500s with no recovery.

Connection reuse worked correctly under sequential load (20 requests, 1 socket). The issues appeared under concurrent WEBrick load.

### Why `net-http-persistent` is a poor fit for WEBrick

`net-http-persistent` v4 delegates to [`connection_pool`](https://github.com/mperham/connection_pool), which was designed for servers with stable, long-lived thread pools (Puma, Sidekiq). WEBrick creates short-lived threads per request, producing higher checkout/checkin churn and making pool sizing difficult: too small triggers `TimeoutError` under load; too large (default ~256) causes excessive CLOSE-WAIT accumulation.

The 0.5-second checkout timeout — inherited from [`TimedStack#pop`](https://github.com/mperham/connection_pool/blob/main/lib/connection_pool/timed_stack.rb#L62) and not configurable through `net-http-persistent`'s API — is adequate for fast backends but too aggressive when Satellite response times reach 10+ seconds under registration load.

Similar issues have been reported by users of [Stripe](https://github.com/stripe/stripe-ruby/issues/780), [Elasticsearch](https://github.com/elastic/elasticsearch-rails/issues/416), [Faraday](https://github.com/lostisland/faraday/issues/1174), and [Mechanize](https://github.com/sparklemotion/mechanize/issues/123) with `net-http-persistent` under high concurrency, with workarounds typically being `idle_timeout` tuning or switching to a different HTTP library ([net-http-persistent#37](https://github.com/drbrain/net-http-persistent/issues/37), [#19](https://github.com/drbrain/net-http-persistent/issues/19)).

PR #952, opened 2026-07-01
Fixes #38120 - Use Sinatra 4

credbbl:


smart_proxy.gemspec
```diff
@@ -18,9 +18,10 @@ Gem::Specification.new do |s|
   s.add_dependency 'logging'
   s.add_dependency 'ostruct'
   s.add_dependency 'rack', '>= 1.3'
+  s.add_dependency 'rackup'
```
ekohl, line 20: Note to self: https://github.com/rack/rack/blob/main/UPGRADE-GUIDE.md#binrackup-rackserver-rackhandlerand--racklobster-were-moved-to-a-separate-gem

smart_proxy.gemspec
```diff
@@ -18,9 +18,10 @@ Gem::Specification.new do |s|
   s.add_dependency 'logging'
   s.add_dependency 'ostruct'
   s.add_dependency 'rack', '>= 1.3'
```
ekohl: `sinatra` depends on `rack`. Perhaps this should be dropped?

ekohl (commented):
Not a complete review, but just some things that jumped out to me that helped me understand more.

https://github.com/sinatra/sinatra/blob/main/CHANGELOG.md is helpful and points to https://github.com/rack/rack/blob/main/UPGRADE-GUIDE.md. Now I wonder what else depends on Rack. Probably Rails so we'll need to verify that can also work with Rack 3.

ekohl:
> Probably Rails so we'll need to verify that can also work with Rack 3.

We're currently on Rails 7.0, but Rails 7.1 added Rack 3 support. That's a blocker to upgrading in our packaging.

credbbl:
> We're currently on Rails 7.0, but Rails 7.1 added Rack 3 support. That's a blocker to upgrading in our packaging.

Where does the proxy use Rails?

adamruzicka:
> Where does the proxy use Rails?

It does not, but foreman does. If foreman and proxy were deployed on the same machine (which they often do), then you'd be able to satisfy the dependencies of only one of them.

ekohl:
> Where does the proxy use Rails?

It doesn't, but https://github.com/theforeman/foreman does. Then in our RPM packaging there is just a single package: https://github.com/theforeman/foreman-packaging/tree/rpm/develop/packages/foreman/rubygem-rack.

Now I'd love to upgrade Rails as well because 7.0 is already EOL and even 7.1 is already EOL: https://endoflife.date/rails. I'm trying to say we need some more coordination.

Last I checked going from 7.0 to 7.1 isn't too hard. According to https://www.fastruby.io/blog/ruby/rails/versions/compatibility-table.html we can still run on Ruby 2.7. Upgrading to 7.2 is harder because it requires us to also bump the minimum Ruby version to 3.1, which is its own can of worms for us. We're working on that, but an area we've neglected for too long.

evgeni:
I'd love to see Rack3/Sinatra4 support in Smart Proxy, as that's what Debian 13 and Ubuntu 26.04 ship.

Would it be a viable way forward to extend the patch here to *try* Rackup and fallback to plain Rack2 if not possible? Then we could drop the `rackup` dep for now and users on Rack3 have to install that manually?

evgeni:
Given #955 was merged, is there anything left in here or should we close it?

PR #955, opened 2026-08-18
Fixes #38120 - Support Sinatra 4 and Rack 3

lhellebr:
Updates bundle to new versions.

In Rack 3, headers must be lowercase https://github.com/rack/rack/issues/1592.

We now use Rackup::... instead of Rack::...

test_global_500 and test_host_500 in RegistrationRegisterApiTest have been failing, both raising `NoMethodError: undefined method   'exception' for #<Logger:...>` from `modules/registration/registration_api.rb`'s error-handling blocks. The cause was that `Proxy::Registration::Api` never included helpers ` ::Proxy::Helpers` (unlike every sibling API class), so its logger calls resolved through Sinatra's own built-in logger helper rather than smart-proxy's real logger; under  Sinatra 2 that helper pointed at `Rack::NullLogger` when logging was disabled, and while `Rack::NullLogger` never actually had an #exception method either, the test's Mocha stub on that exact class silently intercepted the call anyway, whereas Sinatra 4 replaced that null-logger path with a plain `stdlib ::Logger`, which the test's stub no longer targets and which genuinely has no `#exception` method, turning a previously-masked bug into a real crash. The fix was adding `helpers ::Proxy::Helpers` to `Proxy::Registration::Api`, which makes logger resolve to `Proxy::LogBuffer::Decorator.instance` (which does implement `#exception`), matching the pattern already used by every other API class in the codebase — after which the full test suite passed.

### DIFF
--- a/lib/launcher.rb
+++ b/lib/launcher.rb
@@ -1,5 +1,13 @@
 require 'openssl'
 require 'proxy/log'
+
+begin
+  require 'rackup/handler/webrick'
+  WEBRICK_HANDLER = Rackup::Handler::WEBrick
+rescue LoadError
+  require 'rack'
+  WEBRICK_HANDLER = Rack::Handler::WEBrick
+end
 require 'proxy/settings'
 require 'proxy/signal_handler'
 require 'proxy/log_buffer/trace_decorator'
@@ -191,7 +199,7 @@ module Proxy
       rescue ::OpenSSL::SSL::SSLError => e
         raise "Invalid tls_ciphers value '#{app[:SSLCiphers]}': #{e.message}"
       end
-      server.mount "/", Rack::Handler::WEBrick, app[:app]
+      server.mount "/", WEBRICK_HANDLER, app[:app]
 
       # WEBrick 1.9.x does not support :SSLMinVersion in its config hash, so we
       # apply min_version= directly on the SSL context after WEBrick creates it.
--- a/lib/proxy/hsts_middleware.rb
+++ b/lib/proxy/hsts_middleware.rb
@@ -4,14 +4,18 @@ module Proxy
   # is needed.
   # https://www.tenable.com/plugins/nessus/142960
   class HstsMiddleware
+    # Lowercase is required by Rack 3 (https://github.com/rack/rack/issues/1592);
+    # Rack 2 doesn't care about header key case, so this works for both.
+    HEADER_KEY = 'strict-transport-security'.freeze
+
     def initialize(app)
       @app = app
     end
 
     def call(env)
       status, headers, body = @app.call(env)
-      if env['HTTPS'] == 'on' && !headers.include?('Strict-Transport-Security')
-        headers['Strict-Transport-Security'] = 'max-age=31536000'
+      if env['HTTPS'] == 'on' && !headers.include?(HEADER_KEY)
+        headers[HEADER_KEY] = 'max-age=31536000'
       end
       [status, headers, body]
     end
--- a/lib/smart_proxy_main.rb
+++ b/lib/smart_proxy_main.rb
@@ -46,7 +46,11 @@ module Proxy
   ::Sinatra::Base.use ::Proxy::RequestIdMiddleware
   ::Sinatra::Base.use ::Proxy::LoggerMiddleware
   ::Sinatra::Base.use ::Proxy::HstsMiddleware
-  ::Sinatra::Base.set :env, :production
+  # Without this, Sinatra defaults to :development, which in Sinatra 4
+  # enables strict Rack::Protection::HostAuthorization that rejects
+  # requests whose Host header isn't localhost — breaking all real
+  # deployments where Foreman reaches smart-proxy by hostname.
+  ::Sinatra::Base.set :environment, :production
   ::Sinatra::Base.register ::Sinatra::Authorization
 
   require 'root/root'
--- a/modules/registration/registration_api.rb
+++ b/modules/registration/registration_api.rb
@@ -1,6 +1,11 @@
 require 'registration/proxy_request'
 
 class Proxy::Registration::Api < ::Sinatra::Base
+  # Needed so `logger` resolves to Proxy::LogBuffer::Decorator (which implements
+  # #exception, used in the rescue blocks below) instead of Sinatra's own null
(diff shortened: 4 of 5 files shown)


lhellebr:
Converting to draft because we need to investigate how this will work on rpm based deployments with Foreman: it will likely require bump to Rails 7.1.

ekohl:
This is largely a duplicate of https://github.com/theforeman/smart-proxy/pull/952. Based on https://github.com/theforeman/smart-proxy/pull/952#issuecomment-5103577842 it would be great to separate the Rack 3 compatibility into its own PR to be cherry picked into Foreman 5.0.

modules/registration/registration_api.rb
```diff
@@ -1,6 +1,11 @@
 require 'registration/proxy_request'
 
 class Proxy::Registration::Api < ::Sinatra::Base
+  # Needed so `logger` resolves to Proxy::LogBuffer::Decorator (which implements
+  # #exception, used in the rescue blocks below) instead of Sinatra's own null
+  # logger, which is a plain ::Logger with no #exception method as of Sinatra 4.
+  helpers ::Proxy::Helpers
```
ekohl, line 7: Looks like this is a bug we've had that went unnoticed for a long time

lhellebr:
Updated so this supports EITHER Sinatra 2 + Rack 2 OR Sinatra 4 + Rack 3.
This way, this PR shouldn't depend on Foreman's Rails 7.1 PR.

lib/proxy/hsts_middleware.rb
```diff
@@ -6,12 +8,18 @@ module Proxy
   class HstsMiddleware
     def initialize(app)
       @app = app
+      # Rack 3 requires lowercase header keys; Rack 2 uses uppercase
```
ekohl: I think Rack 2 didn't really care. I wonder if it works to just unconditionally use lowercase.

lib/launcher.rb
```diff
@@ -1,6 +1,11 @@
 require 'openssl'
-require 'rackup/handler/webrick'
+require 'rack'
 require 'proxy/log'
+
+# Support both Rack 2 (Rack::Handler) and Rack 3 (Rackup::Handler)
+if Gem::Requirement.new('>= 3.0').satisfied_by?(Gem::Version.new(Rack.release))
+  require 'rackup/handler/webrick'
+end
```
ekohl: I'm not sure how well that gem stuff works in production. That's often weird. Perhaps avoid it and use something like:
```ruby
begin
  require 'rackup/handler/webrick'
  HANDLER_CLASS = Rackup::Handler::WEBrick
rescue LoadError
  require 'rack'
  HANDLER_CLASS = Rack::Handler::WEBrick
end
```

lib/proxy/hsts_middleware.rb
lhellebr: I wanted to keep the changes minimal but since you proposed it, I'll try if lowercase works.

lib/launcher.rb
lhellebr: Will do. Though it would be interesting to see how it works on rpm based gems.

ibuclaw:
Note: you'll also need to set `APP_ENV` or `RACK_ENV` to something other than `development`, otherwise the proxy will refuse all connections even if `:trusted_hosts:` is unset.

See https://github.com/sinatra/sinatra/pull/2053 which defaults to permitting localhost and test domains only.

Alternatively forwarding foreman-proxy trusted hosts setting to Sinatra permitted hosts I guess could also work. To be added in all classes that inherit from `Sinatra::Base` via `configure do ... end`?

lib/launcher.rb
ekohl: There is a packit build - I'd expect that works to test it out

lib/smart_proxy_main.rb
```diff
@@ -46,7 +46,11 @@ module Proxy
   ::Sinatra::Base.use ::Proxy::RequestIdMiddleware
   ::Sinatra::Base.use ::Proxy::LoggerMiddleware
   ::Sinatra::Base.use ::Proxy::HstsMiddleware
-  ::Sinatra::Base.set :env, :production
+  # Without this, Sinatra defaults to :development, which in Sinatra 4
+  # enables strict Rack::Protection::HostAuthorization that rejects
+  # requests whose Host header isn't localhost — breaking all real
+  # deployments where Foreman reaches smart-proxy by hostname.
+  ::Sinatra::Base.set :environment, :production
```
lhellebr, line 53: This is intended to be a fix for an issue that @ibuclaw raised.
The other option (using trusted_hosts in permitted_hosts context) wouldn't IMO make sense because one is about request origin, the other is about request target.

This line has been there for a long time and I think it was a bug - the correct symbol is `:environment` https://sinatrarb.com/configuration.html . It seems like it has been unnoticed for a long time. Note that changing environment can have consequences beyond just this fix - different security rules or logging content and others - I'm not certain what exactly. So this change needs to be weighed carefully.
In either case, the test suite is still green and the proxy still seems to work after this change.

michalgritzbach (approved):
Looks good to me, I think that the hardcoded :production env is a sane default. From what I've found, it could potentially hide traces in failed tests - if this would be an issue, adding `::Sinatra::Base.set :raise_errors, true` to `lib/smart_proxy_for_testing.rb` should fix that. But I wouldn't consider it a blocker.

lib/smart_proxy_main.rb
ibuclaw: Makes sense, thanks.

That wrong name explains why I couldn't find `environment` anywhere.

ekohl:
@lhellebr could you please fix up the Redmine issue reference in the commit message so that test passes?

lhellebr:
@ekohl done

lib/smart_proxy_main.rb
lhellebr: Thanks for warning me about this, @ibuclaw !

ofedoren (approved):
LGTM as well. Do we consider configuring `permitted_hosts` in a follow up?

lhellebr:
@ofedoren yes: https://redhat.atlassian.net/browse/SAT-49512

ekohl (approved):
Thanks!

PR #958, opened 2026-09-15
Fixes #39765 - TFTP archive extraction feature

lzap:
Adds new TFTP capabilities with extracting tarballs and ISO files. Use the test script to perform a live download test on CentOS or Debian to see it in action. See https://community.theforeman.org/t/managed-bootloader-universe/47609 for more context. The script follows the universe contract: https://docs.theforeman.org/nightly/Provisioning_Hosts/index-katello.html#grub2-uefi-boot-loader-and-the-bootloader-universe-structure

The patch extends the API with three new fields:

* `source` is just an alias for `path` (which remains unchanged)
* `destination` can be used instead of `prefix` in case a file must be downloaded under a different filename, previously this was not possible
* `extract` is a new field that allows downloading and extraction of specific files from archives including making symlinks

The API is idempotent - when called with same URL, it will only check if the file on server is newer, if not, download (and extraction) is skipped. If some files were deleted for any reason, extraction will be performed. The source archives are always kept in the TFTP directory.

The legacy API call was asynchronous, this is done by the downloader utility class. The new extraction code retain this behavior and spawns a new thread, then download the archive synchronously and performs the extraction logging everything to the log.

Additionally, an extra check is performed before download of any file via the utility downloader - HTTP HEAD is performed first, if the file does not exist it fails immediately. This will improve experience by emitting errors right away for incorrect URLs or misconfigured setups.

Writing of any files is now much more strict, paths are normalized, symlinked directories are not allowed, and symlinks are never extracted. But this should not affect existing contract.

It is a little bit more code that I expected, most of it are extra checks against path escape and various other attacks.

Example result when you run `extra/download_test.sh`:

```
bootloader-universe/pxegrub2/ubuntu/26.04/amd64/grubx64.efi
bootloader-universe/pxegrub2/ubuntu/26.04/amd64/shimx64.efi
bootloader-universe/pxegrub2/ubuntu/26.04/amd64/linux
bootloader-universe/pxegrub2/ubuntu/26.04/amd64/initrd.gz
bootloader-universe/pxegrub2/ubuntu/26.04/amd64/netboot.tar.gz
bootloader-universe/pxegrub2/ubuntu/26.04/amd64/boot.iso
bootloader-universe/pxegrub2/debian/bookworm/amd64/linux
bootloader-universe/pxegrub2/debian/bookworm/amd64/initrd.gz
bootloader-universe/pxegrub2/debian/bookworm/amd64/grubx64.efi
bootloader-universe/pxegrub2/debian/bookworm/amd64/netboot.tar.gz
bootloader-universe/pxegrub2/centos/10/x86_64/shimx64.efi
bootloader-universe/pxegrub2/centos/10/x86_64/grubx64.efi
bootloader-universe/pxegrub2/centos/10/x86_64/boot.iso
```

The following symlinks are also created:

```
bootloader-universe/pxegrub2/ubuntu/26.04/amd64/boot-sb.efi
bootloader-universe/pxegrub2/ubuntu/26.04/amd64/boot.efi
bootloader-universe/pxegrub2/debian/bookworm/amd64/boot.efi
bootloader-universe/pxegrub2/centos/10/x86_64/boot-sb.efi
bootloader-universe/pxegrub2/centos/10/x86_64/boot.efi
```


modules/tftp/extractor_iso.rb
```diff
@@ -0,0 +1,71 @@
...
+      end
+      [status, error]
+    end
+
+    def run_file_extraction_command(command, temporary)
+      _output, error, status = Open3.capture3(*command)
+      [status, error]
+    end
+
+    def find_command(command)
```
ekohl, line 63: Shouldn't this use [Proxy::Util.which](https://github.com/theforeman/smart-proxy/blob/c2af3d35497058fd7dc8146dcbca3adf60334b9e/lib/proxy/util.rb#L59-L73)?

ekohl (commented):
Not a real review. I should admit I neglected to review https://github.com/theforeman/smart-proxy/pull/837 for a long time. It has very related functionality. Mind taking a look?

modules/tftp/extractor_iso.rb
```diff
@@ -0,0 +1,71 @@
...
+        Proxy::TFTP.create_relative_symlink(link, target)
+      end
+    end
+
+    def extract_member(archive, member, destination)
+      temporary = destination.dirname.join(".#{destination.basename}.#{$$}.tmp")
+      Proxy::TFTP.ensure_tftp_file_path(destination)
+
+      commands = [
+        [[find_command('bsdtar'), '-xOf', archive.to_s, member], false],
```
adamruzicka, line 23: Why bsdtar over gnu tar which is more likely to be available?

modules/tftp/extractor_iso.rb
```diff
@@ -0,0 +1,71 @@
...
+      return run_file_extraction_command(command, temporary) if writes_file
+
+      error = nil
+      status = nil
+      Open3.popen3(*command) do |stdin, stdout, stderr, wait_thread|
+        stdin.close
+        File.open(temporary, 'wb') { |file| IO.copy_stream(stdout, file) }
+        error = stderr.read
+        status = wait_thread.value
+      end
```
adamruzicka, line 54: Could we spawn a shell and do the redirection there rather than pulling all the data through ruby?

modules/tftp/server.rb
```diff
@@ -14,13 +17,11 @@ def set(mac, config)
       end
 
       host_pxe_dir = File.join(path, 'host-config', dashed_mac(mac).downcase, host_pxe_dir_name)
-      FileUtils.mkdir_p(host_pxe_dir)
+      Proxy::TFTP.ensure_tftp_path(host_pxe_dir, create_parents: true, reject_final_symlink: true)
```
lzap, line 20: This new method was introduced to increase security of TFTP manage files, but it is a good idea to also apply it to configuration files. There are several similar changes through this patch. Since we need to carefully test this change, I think this is worth it.

modules/tftp/extractor_iso.rb
```diff
@@ -0,0 +1,71 @@
+require 'fileutils'
+require 'open3'
+
+module Proxy::TFTP
+  module ExtractorIso
+    module_function
+
+    def extract(archive, files, symlinks)
+      files.each do |member, destination|
+        extract_member(archive, member, destination)
```
adamruzicka, line 10: I'm pretty sure both bsdtar and xorriso support extracting multiple files in one pass. Any particular reason for extracting them one by one, re-reading the archive over and over?

modules/tftp/server.rb
```diff
@@ -264,16 +265,232 @@ def host_pxe_dir_name
...
 
-  def self.fetch_boot_file(dst, src)
-    filename    = boot_filename(dst, src)
-    destination = Pathname.new(File.expand_path(filename, Proxy::TFTP::Plugin.settings.tftproot)).cleanpath
-    tftproot    = Pathname.new(Proxy::TFTP::Plugin.settings.tftproot).cleanpath
-    raise "TFTP destination outside of tftproot" unless destination.to_s.start_with?(tftproot.to_s)
+  EXTRACTORS = {
+    'tgz' => Proxy::TFTP::ExtractorTgz,
+    'iso' => Proxy::TFTP::ExtractorIso,
+  }.freeze
```
lzap, line 271: These extractors will cover Fedora, Red Hat, CentOS and clone, Debian and Ubuntu and clones. It is a good start, it is easy to add more if needed.

adamruzicka (commented):
just some tidbits after a quick read-through

modules/tftp/server.rb
```diff
@@ -284,7 +501,7 @@ def self.choose_protocol_and_fetch(src, destination)
                                 verify_server_cert: Proxy::TFTP::Plugin.settings.verify_server_cert).start
 
     when 'nfs'
-      logger.debug "NFS as a protocol for installation medium detected."
+      logger.debug "TFTP: NFS as a protocol for installation medium detected."
```
lzap, line 504: All the other log statements had this prefix...

lzap:
I did not expect such a quick reviews, I pushed an ugly draft so I can take this with me offline (not working today). Well, you guys are fast, this is a very rough code (it works but needs more love).

> #837 for a long time. It has very related functionality. Mind taking a look?

Yeah, I was not aware. This implementation in comparsion:

* bootloader-universe aware
* flexible API with 
* modular extraction
* supports Fedora/RH/Debian/Ubuntu out of box

The other implementation:

* just plainly extracts (not create files for bootloader universe)
* simpler API
* only focuses on Ubuntu

There are some ideas which I like and I could inspire, e.g. using the command task for running the extraction subcommands.

modules/tftp/extractor_iso.rb
lzap: I only found two utilities in RHEL9 which can extract ISOs without mounting: bsdtar and xorriso.

modules/tftp/extractor_iso.rb
lzap: Any specific reason to do so? My concern is security, if I can avoid shell I will. Need to think about it, if your suggestion will lead to less code while retaining security, then why not.

I will say that I want to look into how the util CommandTask class could be reused here, so I think this can be improved.

modules/tftp/server.rb
```diff
@@ -264,16 +265,232 @@ def host_pxe_dir_name
...
+  def self.legacy_tftp_path(path)
+    pathname = Pathname.new(path.to_s)
+    return tftp_path(pathname) unless pathname.absolute?
+
+    configured_root = configured_tftp_root
+    pathname = pathname.cleanpath
+    raise "TFTP destination outside of tftproot" unless pathname == configured_root || pathname.to_s.start_with?("#{configured_root}/")
+
+    tftp_path(pathname.relative_path_from(configured_root))
+  end
```
lzap, line 386: I am going to drop the `legacy_tftp_path` and only use `tftp_path` going forward. It kept one silly edge case (when `prefix` was absolute path) which is never actually called from Foreman.

modules/tftp/server.rb
```diff
@@ -264,16 +265,232 @@ def host_pxe_dir_name
...
+      begin
+        stat = File.lstat(current)
+      rescue Errno::ENOENT
+        break
+      end
+      final = index == parts.length - 1
+      raise "TFTP path contains a symlink: #{current}" if stat.symlink? && (!final || reject_final_symlink)
+      raise "TFTP path component is not a directory: #{current}" unless current == pathname || stat.directory?
+    end
+  end
```
lzap, line 475: Both those `ensure*` methods are a bit silly, they do slightly different thing. My instruction was to make sure that extraction or download is not subject of path attacks. I need to take a closer look on how this could be simplified.

Could not figure out how to simplify this, I strongly believe that security is key when it comes to extracting unknown content. Here is the breakdown:

Method `ensure_tftp_path` walks down the tree and ensures directories are present and are NOT symlinks. It also can create directories, like `mkdir_p` and is used instead when path must be created.

Method `ensure_tftp_existing_path` is used for incoming HTTP request data ensuring the path does not escape the TFTP root. It uses similar techniques like the other method, but it serves a different purpose.

modules/tftp/extractor_iso.rb
lzap: Confirmed, taking a look.

modules/tftp/extractor_iso.rb
lzap: Actually, this was nicely aligned with your other command, the code now extracts all files through temp directory directly.

lzap:
Ok rebased, thanks for initial reviews. This is now ready for full reviews. Once I get approvals, I will work on the Foreman part and test this end to end.

modules/tftp/extractor_iso.rb
```diff
@@ -0,0 +1,79 @@
+require 'fileutils'
+require 'open3'
+require 'tmpdir'
+
+module Proxy::TFTP
+  module ExtractorIso
```
ekohl: Perhaps a tiny nit pick, but would it make sense to have an `Extractor` module` so you end up with `Proxy::TFTP::Extractor::Iso`?

modules/tftp/extractor_iso.rb
```diff
@@ -0,0 +1,79 @@
...
+
+        staged_files = files.map do |member, destination|
+          staged = Pathname.new(staging).join(member)
+          [staged, destination]
+        end
+        return 'requested ISO member was not extracted' unless staged_files.all? { |staged, _destination| regular_file?(staged) }
+
+        staged_files.each do |staged, destination|
+          File.chmod(0o644, staged)
+          FileUtils.mv staged.to_s, destination.to_s, force: true
```
adamruzicka: For the sake of consistency.
```suggestion
          FileUtils.mv(staged.to_s, destination.to_s, force: true)
```

This also applies to L60 and maybe some other places.

modules/tftp/extractor_tgz.rb
```diff
@@ -0,0 +1,45 @@
+require 'fileutils'
+require 'English'
+require 'rubygems/package'
+require 'rubygems/package/tar_reader'
+require 'zlib'
+
+module Proxy::TFTP
+  module ExtractorTgz
+    module_function
```
adamruzicka: This is mostly personal preference, but I'd still prefer a class with class methods over a module with this

adamruzicka (commented):
Haven't tested, but it fits together nicely in my head. It still feels a bit go-esque though in some places

lib/proxy/http_download.rb
```diff
@@ -44,6 +49,53 @@ def initialize(src, dst, connect_timeout: DEFAULT_CONNECT_TIMEOUT, verify_server
...
+      proxy = uri.find_proxy
+      http = if proxy
+               Net::HTTP.new(uri.host, uri.port, proxy.host, proxy.port, proxy.user, proxy.password)
+             else
+               Net::HTTP.new(uri.host, uri.port, nil)
+             end
+      http.use_ssl = uri.scheme == 'https'
+      http.open_timeout = connect_timeout
+      http.read_timeout = connect_timeout
+      http.verify_mode = OpenSSL::SSL::VERIFY_NONE unless verify_server_cert
```
ekohl, line 93: These days I'm leaning to always verifying unless explicitly disabled. Any thoughts?

ekohl (commented):
Not a complete review.

lib/proxy/http_download.rb
lzap: The default value for the HTTP downloader util method is false, however, for all the TFTP downloads this is overwritten by a setting which has a different default value: true. So we are safe for TFTP.

Good point, irrelevant for this patch tho. We can flip it in a separate PR, this might break things.

lzap:
Thanks amended your comments, I actually squashed the two commits into one because I had to fix how "destination" alias works. Previously, it was only prefix so it worked like a prefix when file path was provided and for directory (ending with slash) it created the very same filename. But after I tested this also with Ubuntu, we need to be also able to download ISO alone. This is because last 3 releases of Ubuntu provides both netboot tarball and installation ISO, however, for Ubuntu installer the ISO is always required. Thus, it will make sense to download BOTH netboot tarball and the ISO:

```bash

curl --fail --silent --show-error \
  --header 'Content-Type: application/json' \
  --data '{
  "extract": {
    "source": "https://releases.ubuntu.com/26.04/ubuntu-26.04-netboot-amd64.tar.gz",
    "destination": "bootloader-universe/pxegrub2/ubuntu/26.04/amd64/netboot.tar.gz",
    "type": "tgz",
    "files": {
      "bootloader-universe/pxegrub2/ubuntu/26.04/amd64/linux": "amd64/linux",
      "bootloader-universe/pxegrub2/ubuntu/26.04/amd64/initrd.gz": "amd64/initrd",
      "bootloader-universe/pxegrub2/ubuntu/26.04/amd64/grubx64.efi": "amd64/grubx64.efi",
      "bootloader-universe/pxegrub2/ubuntu/26.04/amd64/shimx64.efi": "amd64/bootx64.efi"
    },
    "symlinks": {
      "bootloader-universe/pxegrub2/ubuntu/26.04/amd64/boot.efi": "bootloader-universe/pxegrub2/ubuntu/26.04/amd64/grubx64.efi",
      "bootloader-universe/pxegrub2/ubuntu/26.04/amd64/boot-sb.efi": "bootloader-universe/pxegrub2/ubuntu/26.04/amd64/shimx64.efi"
    }
  }
}' \
  "${PROXY_URL}/tftp/fetch_boot_file"

# Ubuntu ISO is also needed
curl --fail --silent --show-error \
  --header 'Content-Type: application/json' \
  --data '{
  "source": "https://releases.ubuntu.com/26.04/ubuntu-26.04-live-server-amd64.iso",
  "destination": "bootloader-universe/pxegrub2/ubuntu/26.04/amd64/boot.iso"
}' \
  "${PROXY_URL}/tftp/fetch_boot_file"
```

The old flag called `prefix` still works and it has unchanged behavior. Only `source` is an alias, the old `path` also still works so the API compatibility is remained. I tested this works fine.

ekohl:
> ```shell
>     "destination": "bootloader-universe/pxegrub2/ubuntu/26.04/amd64/netboot.tar.gz",
> ```

I don't think Foreman should send the exact path. In other places we send the `os`, `release`, and `arch` and I'd like to remain consistent. Neither should it send which files and symlinks because that's already done server side:
https://github.com/theforeman/smart-proxy/blob/c2af3d35497058fd7dc8146dcbca3adf60334b9e/modules/tftp/server.rb#L161-L186

lzap:
> I don't think Foreman should send the exact path. In other places we send the `os`, `release`, and `arch` and I'd like to remain consistent. Neither should it send which files and symlinks because that's already done server side:

That is exactly what the changed API point does, `prefix` is a relative path within the TFTP root. I would like to stay consistent here.

It also aligns how this works in Foreman core - the code there fully works with full (relative) TFTP paths when constructing the `filename` option and orchestrating the downloads. And this has a reason - we centralize all OS-related know how in Foreman code, specifically in Operatingsystem class (and descendants). Designing the API the generic way, which is generally preferred and it is hard to doubt that, would mean we also need to copy some of the logic (of constructing the paths) in here. And because this change will be gradual, I only plan to implement RedHat, Ubuntu, Debian this would mean this is scattered between Foreman core and Foreman Proxy.

For this reason I think it is better to keep the "download this file in here" contract also for extraction.

